PR #1, opened 2018-10-09
Data storage for registry API

ecordell:
- Adds a `DirectoryLoader` that takes a directory of manifests and loads them into a database
- Adds a `Load` interface and an implementation for SQLlite to add data into the sqlite store (called by the `DirectoryLoader`
- Adds a `Query` interface that implements the queries we will need for the http API
- Adds a dockerfile that loads a set of manifests into a sqlite.db

The next step is defining the http interface, and then talking to it from OLM.


aravindhp:
I see that the CRDs are not packaged into the sqlite store. What is the plan for that? Otherwise I tried it out locally and it LGTM.

ecordell:
@aravindhp the CRDs are all stored in the `operatorbundle.bundle` field in the DB. Basically the bundle is "everything you need", the CSV + the CRDs + potentially extra objects like an SCC, etc.

aravindhp:
Sounds good, @ecordell. The `time="2018-10-10T13:50:28Z" level=info msg="skipping non-csv file" dir=manifests file=etcdcluster.crd.yaml load=bundles` messages misled me.

PR #3, opened 2018-10-18
feat(configmap): add configmap loader

ecordell:
This is to support the older catalog format, which stores data in a configmap.

OLM will use this to interact with configmaps and catalog images using the same interface.


pkg/sqlite/configmap.go
```diff
@@ -0,0 +1,162 @@
...
+// entries under "packages" will be parsed as Packages
+type ConfigMapLoader struct {
+	store     registry.Load
+	configMap v1.ConfigMap
+	crds      map[registry.APIKey]*unstructured.Unstructured
+}
+
+var _ SQLPopulator = &ConfigMapLoader{}
+
+func NewSQLLoaderForConfigMap(store registry.Load, configMap v1.ConfigMap) *ConfigMapLoader {
```
alecmerdler: Wouldn't it be better to remove the intermediate `ConfigMap` and just consume a directory of manifests? Since the `ConfigMap` will probably be too large to `kubectl create` anyway?

ecordell: That's a decision we can make on the OLM side (what to do with the configmap after we have them converted here). I was thinking that it would make sense to continue to support the configmaps as a low barrier to entry; not really convinced they're that much easier though.

PR #7, opened 2018-11-26
fix(bundle): fix CRD <plural>.<group> parsing

ecordell:


stevekuznetsov:
Y'all forgot to give the robots access to this repo

ecordell:
@stevekuznetsov we aren't yet shipping this repo, and haven't yet configured ci for it

pkg/registry/bundle.go
```diff
@@ -120,7 +120,10 @@ func (b *Bundle) RequiredAPIs() (map[APIKey]struct{}, error) {
 		return nil, err
 	}
 	for _, api := range csv.Spec.CustomResourceDefinitions.Required {
-		parts := strings.SplitAfterN(api.Name, ".", 2)
+		parts := strings.SplitN(api.Name, ".", 2)
+		if len(parts) < 2 {
```
jpeeler: The paranoid in me wants to write it like:
```
substrings := 2
parts := strings.SplitN(api.Name ",", substrings)
delimiterCount := strings.Count(api.Name, ",")
if len(parts) != substrings || delimiterCount != substrings-1 {
	  fmt.Println("ERROR")
}
```

ecordell: The format of the CRD name is something like `<pluralname>.<grouppart>.<grouppart>.<grouppart>` - we want to parse it into `<pluralname>` and `<grouppart>.<grouppart>.<grouppart>`. For example: `etcdclusters.etcd.database.coreos.com` into `etcdclusters` and `etcd.database.coreos.com`.

`delimiterCount` will often  `!= substrings-1` in this case. 

The docs for `SplitN`:
```
n > 0: at most n substrings; the last substring will be the unsplit remainder.
n == 0: the result is nil (zero substrings)
n < 0: all substrings
```
so for n = 2, we will get 0, 1, or 2 substrings. It will never be more than 2, so we just need to check the invalid cases of 0 and 1, hence the `len(parts) < 2` condition

jpeeler: Ah, I was not thinking of the right format for the name. But thanks for the detailed explanation.

PR #8, opened 2018-12-11
Write panic and fatal log levels to a termination log file

njhale:
### Description
- Adds `termination-log` flag to `registry-server` and `configmap-server` binaries to specify termination log path
- Writes panic and fatal log levels to a termination log
- Adds a WriteHook type for splitting log levels between Writers (for logrus)
- Fixes checksum issue with client-go go module

### Motivation
When an `operator-registry` container is being orchestrated by a Kubernetes Pod resource, exit information (possibly due to failed catalog parsing) needs to be surfaced to make decisions in orchestrating controllers. The canonical way to do this is by [writing a termination message](https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/#writing-and-reading-a-termination-message) to a termination log file in the container.


cmd/configmap-server/main.go
```diff
@@ -39,16 +41,49 @@ func main() {
...
 	}
 }
 
 func runCmdFunc(cmd *cobra.Command, args []string) error {
+	// Immediately set up termination log
+	terminationLogPath, err := cmd.Flags().GetString("termination-log")
+	if err != nil {
+		return err
+	}
+	terminationLogFile, err := os.OpenFile(terminationLogPath, os.O_WRONLY|os.O_CREATE, 0755)
```
ecordell: only comment: we probably want to do this across all of our binaries (open the termination log file and set up the logging hooks) so it could be factored out into its own thing.

njhale: I was thinking about making a parent cobra command with shared flags and behavior. Although, that may constrain us to a single binary.

pkg/lib/log/writerhook.go
```diff
@@ -0,0 +1,77 @@
...
+func AddHooks(hooks ...*WriterHook) {
+	// Send all logs to nowhere by default
+	logrus.SetOutput(ioutil.Discard)
+
+	for _, hook := range hooks {
+		logrus.AddHook(hook)
+	}
+}
+
+func AddDefaultWriterHooks(terminationLogPath string) error {
```
njhale: Moved the shared setup here.

PR #10, opened 2018-12-14
fix(health): move health to its own package with the right package name

ecordell:
this is required for the health check protocol to work properly


Dockerfile
```diff
@@ -8,7 +8,6 @@ COPY cmd cmd
 COPY pkg pkg
 COPY Makefile Makefile
 COPY go.mod go.mod
-COPY go.sum go.sum
```
njhale: probably for the best right now.

PR #19, opened 2019-01-16
fix(bundle): fix CRD conversion with minimum validation field

njhale:
- Marshals to JSON then Unmarshals to CustomResourceDefinition when
updating bundle cache to avoid a FromUnstructured call.
- Runs `make vendor`.

### DIFF
--- a/pkg/registry/bundle.go
+++ b/pkg/registry/bundle.go
@@ -3,6 +3,7 @@ package registry
 import (
 	"fmt"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"encoding/json"
 	"strings"
 
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/registry/bundle.go
```diff
@@ -205,7 +206,13 @@ func (b *Bundle) cache() error {
 	for _, o := range b.Objects {
 		if o.GetObjectKind().GroupVersionKind().Kind == "CustomResourceDefinition" {
 			crd := &apiextensions.CustomResourceDefinition{}
-			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.UnstructuredContent(), crd); err != nil {
+			// Marshal Unstructured and Unmarshal as CustomResourceDefinition. FromUnstructured has issues
```
alecmerdler: Can we file a bug upstream to document this for other devs?

njhale: @alecmerdler yeah, this fix works for now but I need to nail down the exact wording. Maybe there's something additional we're doing wrong that's causing `FromUnstructured` to not work.

PR #20, opened 2019-01-17
feat(bundles): add GetBundle request

njhale:
Allows bundles to be queried by CSV name.


pkg/server/server.go
```diff
@@ -38,6 +38,18 @@ func (s *RegistryServer) GetPackage(ctx context.Context, req *api.GetPackageRequ
 	return api.PackageManifestToAPIPackage(packageManifest), nil
 }
 
+func (s *RegistryServer) GetBundle(ctx context.Context, req *api.GetBundleRequest) (*api.Bundle, error) {
```
ecordell: I think I misunderstood what you were asking before - I think we should take `package` and `channel` as arguments to `GetBundle`. In theory multiple channelentries can point to the same bundle (e.g. if we had a "promotion" mechanism to push bundles through channels) but (channel, package, bundlename) should be unique.

njhale: SGTM 😎

PR #27, opened 2019-02-12
Add support to download from a remote registry 

tkashem:
Add a new binary 'appregistry-server' that can be used to download operator manifest(s) from a remote app registry and generate a sqlite database to be served using grpc.

https://jira.coreos.com/browse/MKTPLC-108
https://jira.coreos.com/browse/MKTPLC-240

TBD:
* e2e test(s) for the binary
* I am running into error while building upstream docker files for both configmap-server and the appregistry-server binaries. This needs to be fixed in a separate PR later.


pkg/apprclient/adapter.go
```diff
@@ -0,0 +1,77 @@
+package apprclient
+
+import (
+	"bytes"
+
+	appr "github.com/operator-framework/go-appr/appregistry"
+	appr_blobs "github.com/operator-framework/go-appr/appregistry/blobs"
```
njhale: nit: [underscores in package names?](https://golang.org/doc/effective_go.html#package-names)

pkg/apprclient/adapter.go
```diff
@@ -0,0 +1,77 @@
...
+	appr_package "github.com/operator-framework/go-appr/appregistry/package_appr"
```
njhale: nit: [underscores in package names?](https://golang.org/doc/effective_go.html#package-names)

pkg/apprclient/adapter.go
```diff
@@ -0,0 +1,77 @@
...
+	appr_models "github.com/operator-framework/go-appr/models"
+)
+
+const (
+	mediaType string = "helm"
```
njhale: nit: no need for explicit typing here

pkg/apprclient/adapter.go
tkashem: Removed underscores

pkg/apprclient/adapter.go
tkashem: Removed explicit typing

tkashem:
@njhale Addressed the comments you posted, let me know if you have more questions.

pkg/apprclient/apprclient.go
```diff
@@ -0,0 +1,38 @@
+package apprclient
+
+import (
+	"net/url"
+
+	"github.com/go-openapi/runtime"
+	httptransport "github.com/go-openapi/runtime/client"
+	"github.com/go-openapi/strfmt"
+	apprclient "github.com/operator-framework/go-appr/appregistry"
```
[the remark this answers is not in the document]
tkashem: you are right, didn't realize this at all. I will fix this in a separate PR next sprint.

pkg/apprclient/decoder.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"io"
+)
+
+type blobDecoder interface {
+	// Decode decodes package blob into plain unencrypted byte array
+	Decode(encoded []byte) ([]byte, error)
+}
+
+type blobDecoderImpl struct {
```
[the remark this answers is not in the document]
tkashem: yeah, I like it, makes it concise.

pkg/appregistry/input.go
```diff
@@ -0,0 +1,103 @@
...
+type Input struct {
+	// Sources is the set of namespaced name(s) of OperatorSource objects from
+	// which we need to pull packages.
+	Sources []*types.NamespacedName
+
+	// Packages is the set of package name(s) specified.
+	Packages []string
+}
+
+func (i *Input) PackagesToMap() map[string]bool {
```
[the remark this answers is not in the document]
tkashem: I was thinking of using `truct{}`, but then came across this in effective go -
`A set can be implemented as a map with value type bool. Set the map entry to true to put the value in the set, and then test it by simple indexing`
```
attended := map[string]bool{
    "Ann": true,
    "Joe": true,
    ...
}

if attended[person] { // will be false if person is not in the map
    fmt.Println(person, "was at the meeting")
}
```

PR #28, opened 2019-02-15
Use auth from opsrc

kevinrizza:
-Query opsrc for secret name
-Update downloader to get auth token secret
-Use auth token in apprclient request header


kevinrizza:
Put this wip pr up for now to review, still need to merge the auth pr on operator-marketplace and then update vendored dependencies.

pkg/appregistry/downloader.go
```diff
@@ -178,3 +198,24 @@ func (d *downloader) QuerySource(key *types.NamespacedName) (spec *v1alpha1.Oper
...
+	}
+
+	auth := spec.AuthorizationToken
+	if auth.SecretName != "" {
+		secret, err := d.secretClient.CoreV1().Secrets(namespace).Get(auth.SecretName, metav1.GetOptions{})
+		if err != nil {
+			return options, err
+		}
+
+		options.AuthToken = string(secret.Data["token"])
```
tkashem: I thought we needed to base64 decode it.

pkg/appregistry/appregistry.go
```diff
@@ -100,3 +107,23 @@ func NewClient(kubeconfig string, logger *logrus.Entry) (clientset marketplace.I
 	clientset, err = marketplace.NewForConfig(config)
 	return
 }
+
+func NewSecretClient(kubeconfig string, logger *logrus.Entry) (clientset *kubernetes.Clientset, err error) {
```
tkashem: nit: maybe `kubeclient` is a more appropriate name?

pkg/appregistry/downloader.go
kevinrizza: @tkashem Nope, the api just obfuscates it as base64, but when pulling from the client directly it is returned as a string.

pkg/appregistry/appregistry.go
kevinrizza: Sure, updated in latest commit. I also updated the field name in the downloader to match

PR #30, opened 2019-02-22
WIP - adding ability to dynamically change the container images in CSV.

shawn-hurley:
I am submitting the PR to describe what I would like to see. 

I think it is useful to change the container image the CSV will use dynamically. This will be useful in local development and CI.

This is pretty hacky but is a decent enough POC to see how this could work. 

Example say the CSV's deployment has the image: docker.io/shurley/image1 setting this env IMAGE_REGISTRY=quay.io will update the container section to the image: quay.io/shurley/image1.

### DIFF
--- a/pkg/api/conversion.go
+++ b/pkg/api/conversion.go
@@ -3,9 +3,12 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/operator-framework/operator-registry/pkg/registry"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
@@ -71,7 +74,13 @@ func BundleStringToAPIBundle(bundleString string, entry *registry.ChannelEntry) 
 		}
 		if unst.GetKind() == "ClusterServiceVersion" {
 			out.CsvName = unst.GetName()
-			out.CsvJson = o
+			// Here if the environment var REGISTRY_BASE and/or IMAGE_TAG is set, we will update the
+			// json string to change the image that will be exposed
+			csvJSON, err := updateCSVJSON(unst)
+			if err != nil {
+				return nil, err
+			}
+			out.CsvJson = csvJSON
 			break
 		}
 	}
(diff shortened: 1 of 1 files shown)


pmorie:
The context for this change is that @shawn-hurley and I have been working to build CI/CD pipelines for operators as part of an effort to define our story around this for operators shipping in the marketplace. There are a couple specific needs we want to address that Shawn is sharing his POC for:

- During development, allow developers to deploy an operator and operator registry using their own quay or image registry account
- During CI/CD, allow tests to deploy operators using images built in the pipeline for the operators

### Examples

- As a developer, I want to deploy a catalog source that using an operator registry from an image in my image registry account, so that I can develop the operator without my image registry coordinates sneaking into a config that gets checked in
- As a person delivering an operator, I want to have a test that deploys my operator with the images built in my pipeline, without having to manually munge configs and rebuild image for registry

ecordell:
I understand the motivation, and this type of tooling/workflow is absolutely missing from OLM at the moment.

I've been thinking about a slightly different approach to this problem by enabling OLM to "resolve" a CSV alone (in the absence of a subscription). Creating a CSV will create an InstallPlan for that CSV's embedded resources (serviceaccount, etc). This is similar to OLM's old behavior, except that we would have a place to gate creation of the CSV's resources.

The workflow for local dev / CI would then be:

1. Have a directory for your operator bundle (csv + any extras, but not your SA/Roles/Bindings/etc)
2. `kubectl apply -f ./bundle`
3. InstallPlan created, applies your CSV + rbac, anything else is created directly from your bundle as if it came from a catalog.

If you want to test an update to an existing published operator:

1. Add a `replaces` field to your CSV that points the last release
2. Place a catalogsource in the cluster containing the last release
3. Create a subscription for the the last release and wait for the operator to install
4. `kubectl apply -f ./bundle` with your updated CSV
5. InstallPlan created for your new CSV
6. Subscription detects that you've updated manually and notes the newer version in its status

Am I thinking about these UX problems in the same way that you are? Do you like this proposed solution or is there an advantage to this PR's implementation that I'm missing?

shawn-hurley:
Installing the CSV this way would make a solution achievable. But making a user constantly changing the CSV feels like the wrong thing to do especially when there is an obvious mechanism to make this easier. IMO something like this would enable users building a CI pipeline and/or attempting to test new changes in a cluster better than forcing them to have to change the CSV for each type of change they want. I just think that an automated solution via a tool that already exists > having users use SED and manage multiple files.

pmorie:
@ecordell I don't think your comment addresses the use-case for serving modified CSVs without forcing the user to manually modify them. That's pretty central to the CI story for deploying the operator image built in the pipeline without having to manually munge the manifests, rebuild operator registry, etc. Did I miss it?

ecordell:
cf. https://github.com/openshift/federation-v2-operator/blob/master/olm-testing/Dockerfile

shawn-hurley:
Closing this PR. I think this use case this is attempting to solve will have to be solved in some other way.

PR #31, opened 2019-02-26
pkg/sqlite: Add error checking and improve looping through rows

LiliC:
Ran into this problem while testing adding an operator locally, the for loop never exited because it reached the end of rows but there was no break in place, so it just kept looping.

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -172,35 +172,39 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error 
 
 		channelEntryCSVName := c.CurrentCSVName
 		depth := 1
-		for {
-			rows, err := getReplaces.Query(channelEntryCSVName)
-			if err != nil {
+		rows, err := getReplaces.Query(channelEntryCSVName)
+		if err != nil {
+			return err
+		}
+		defer rows.Close()
+
+		for rows.Next() {
+			var replaced sql.NullString
+			if err := rows.Scan(&replaced); err != nil {
 				return err
 			}
 
-			if rows.Next() {
-				var replaced sql.NullString
-				if err := rows.Scan(&replaced); err != nil {
-					return err
-				}
-
-				if !replaced.Valid || replaced.String == "" {
-					break
-				}
-
-				replacedChannelEntry, err := addChannelEntry.Exec(c.Name, manifest.PackageName, replaced.String, depth)
-				if err != nil {
-					return err
-				}
-				replacedID, err := replacedChannelEntry.LastInsertId()
-				if err != nil {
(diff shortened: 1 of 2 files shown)


LiliC:
I tested this locally, by building the registry with the changes and then adding the operator packages and it worked on two different operator packages for me. Not sure why the unit tests is failing. 🤔

pkg/sqlite/load.go
```diff
@@ -172,35 +172,39 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
 
 		channelEntryCSVName := c.CurrentCSVName
 		depth := 1
-		for {
-			rows, err := getReplaces.Query(channelEntryCSVName)
-			if err != nil {
+		rows, err := getReplaces.Query(channelEntryCSVName)
```
ecordell: Removing the outer for loop here means that we only every build update chains of depth 1

pkg/sqlite/load.go
```diff
@@ -172,35 +172,39 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
-				}
-				replacedID, err := replacedChannelEntry.LastInsertId()
-				if err != nil {
-					return err
-				}
-				addReplaces.Exec(replacedID, currentID)
-				currentID = replacedID
-				channelEntryCSVName = replaced.String
-				depth += 1
+			if !replaced.Valid || replaced.String == "" {
```
ecordell: This is the part that's supposed to break out of the loop when we hit the end of the upgrade chain (the oldest CSV in a channel)

If there's an infinite loop it's probably because this isn't being hit for some reason - if you can provide some data to reproduce we could hopefully provide good error messages for those cases.

LiliC: As mentioned on slack:
When I tested the prometheus operator change from https://github.com/operator-framework/community-operators/pull/65, the build went into an endless loop without this patch. I expected it to either work or throw the appropriate error. I'm not too familiar with this code base to know whether this is 100% correct semantically but it made it work and I could successfully test the new Prometheus Operator CSV introduced in that PR.

ecordell: The reason this works is because it sidesteps the validation of the "replacement" chain by only going one layer deep. The root cause is the replaces field pointing to a CSV that doesn't exist. In #32 I added an error message to detect this case. I checked that the prometheus csv returned that error when I tried to load it, and I checked that changing it to the right value (0.22.2) removed the error and loaded everything as expected.

ecordell:
Closing in favor of #32 - 

Thank you for tracking down this issue and helping us work through it! It's really appreciated and helps make the registry better to use.

PR #34, opened 2019-03-06
add instructions how to use grpc catalog

dmesser:


README.md
```diff
@@ -80,15 +80,61 @@ manifests
...
+metadata:
+  name: example-manifests
+spec:
+  sourceType: grpc
+  image: example-registry:latest
+```
+
+When you create the `CatalogSource`, watch the `catalog-operator` as part of your OLM deployment (assuming OLM is deloyed in a namespace called `olm`):
+
+```sh
```
ecordell: This highlights some pretty bad UX around catalogs... you should be able to watch a CatalogSource's status to know when it's ready.

What do you think about suggesting:

```sh
$ watch kubectl get packagemanifests
```
until we have improved the status of the CatalogSource?

(also just an implementation note - the package server and the catalog operator are separate binaries with separate connections to the registry pods. The catalog operator getting a connection doesn't necessarily mean the package server will also have a healthy connection, though they're usually correlated)

PR #40, opened 2019-03-21
Remove OperatorSource dependency

tkashem:
Currently the way we specify an operator source is by providing a namespaced name of an OperatorSource CR. This introduces a tight coupling between operator registry and marketplace operator.

Make the following changes to remove OperatorSource dependency:
- Specify a single operator source using the following format
{base url with cnr prefix}|{quay registry namespace}|{secret namespace/secret name}
 -- Use '|' as a delimiter to separate operator source elements.
 -- k8s secret object is optional and if specified should have the
    format - {k8s namespace}/{secret name}.
 -- Add a new flag 'registry' to specify operator source using the
    above format.

- Convert the existing flag 'sources' from string to string slice so that we
  can accept multiple operator source specifier(s).

We can maintain backwards compatibility by having both flags for now.

Example:
--registry="https://quay.io/cnr|community-operators" --registry="https://quay.io/cnr|custom-operators|mynamespace/mysecret"

### DIFF
--- a/cmd/appregistry-server/main.go
+++ b/cmd/appregistry-server/main.go
@@ -33,10 +33,12 @@ func main() {
 	rootCmd.Flags().Bool("debug", false, "enable debug logging")
 	rootCmd.Flags().StringP("kubeconfig", "k", "", "absolute path to kubeconfig file")
 	rootCmd.Flags().StringP("database", "d", "bundles.db", "name of db to output")
-	rootCmd.Flags().StringP("sources", "s", "", "comma separated list of OperatorSource object(s) {namespace}/{name}")
+	rootCmd.Flags().StringSliceP("sources", "s", []string{}, "comma separated list of OperatorSource object(s) {namespace}/{name}")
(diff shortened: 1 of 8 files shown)


aravindhp:
@ecordell yeah that might make things a bit easier. @tkashem is this easily doable?

tkashem:
If we use a different name for the the new arg, it means that I have to keep both old and new parsing active which is not the case now.

njhale:
To second @ecordell, this looks really good! Thank you for doing this.

That being said, it does have me wondering if using JSON to represent this configuration would be more easily maintainable. We could pull it in from stdin or a file, and we would benefit from standard marshaling/unmarshaling and adding new fields would be relatively trivial.

tkashem:
`e2e-aws` tests failed because marketplace registry pods we crashing due to the changes. So I layered the two different ways of specifying operator source into individual units. This allowed me to isolate the `operator source CR dependent` logic. So now we are backwards compatible.  

And the future PR to remove OperatorSource CR dependency should be simple, I believe.

@aravindhp @kevinrizza @njhale @ecordell

aravindhp:
@ecordell @njhale please review when you get the chance. You can also remove the hold when you lgtm the PR.

PR #41, opened 2019-03-26
Permit skipping operators during updates 

ecordell:
Use a new "skips" field to specify additional operators that
the current one should also replace. This is to allow building upgrade
graphs that omit previously released versions of operators. For example,
if version 2 of an operator has such a serious bug that it should never
be installed if it hasn't already, then version 3 can specify:

```yaml
replaces: operator.v1
skips:
- operator.v2
```

and this will tell the registry to synthesize data for the release 2.

The goal is for clients that have never installed 2 to skip straight
to 3, and for clusters that have installed 2 to be directed to 3 as well


njhale:
Looks like [this](https://github.com/operator-framework/operator-registry/pull/41/files?file-filters%5B%5D=.go#diff-3e4bbabd322c31b037c64a71d50b272bR121) is deferred twice.

PR #48, opened 2019-05-15
fix(registry): Validate alm-examples in operator registry

dinhxuanvu:
Validate alm-examples for CSV from registry to ensure it is parsable
and match provided APIs

### DIFF
--- a/pkg/schema/schema.go
+++ b/pkg/schema/schema.go
@@ -241,6 +241,10 @@ func validateKind(kind string, fileBytes []byte) error {
 		if err != nil {
 			return err
 		}
+		err = validateExamplesAnnotations(&csv)
+		if err != nil {
+			return err
+		}
 		return err
 	case "CatalogSource":
 		cs := v1alpha1.CatalogSource{}
@@ -305,7 +309,8 @@ func validateResources(directory string) error {
 		}
 
 		fmt.Printf("validate %s\n", path)
-		if validateResource(path, f, err) != nil {
+		err = validateResource(path, f, err)
+		if err != nil {
 			return err
 		}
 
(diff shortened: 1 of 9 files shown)


pkg/schema/schema.go
```diff
@@ -317,3 +337,80 @@ func validateResources(directory string) error {
...
+
+func getProvidedAPIs(csv *v1alpha1.ClusterServiceVersion) (map[opregistry.APIKey]struct{}, error) {
+	provided := map[opregistry.APIKey]struct{}{}
+
+	for _, owned := range csv.Spec.CustomResourceDefinitions.Owned {
+		parts := strings.SplitN(owned.Name, ".", 2)
+		if len(parts) < 2 {
+			return nil, fmt.Errorf("couldn't parse plural.group from crd name: %s", owned.Name)
+		}
+		provided[opregistry.APIKey{Group: parts[1], Version: owned.Name, Kind: owned.Kind, Plural: ""}] = struct{}{}
```
jpeeler: I'd expect this line to look more like this:
https://github.com/operator-framework/operator-lifecycle-manager/blob/b2d1cd21368bc8cc10e4ca11a231f09077630c33/pkg/controller/registry/resolver/operators.go#L277

pkg/schema/schema.go
```diff
@@ -317,3 +328,80 @@ func validateResources(directory string) error {
...
+
+func validateExamplesAnnotations(csv *v1alpha1.ClusterServiceVersion) error {
+	var examples []CRDTemplate
+	annotations := csv.ObjectMeta.GetAnnotations()
+	// Return right away if no alm-examples are found.
+	if annotations == nil {
+		return fmt.Errorf("Unable to get annotations")
+	}
+
+	almExamples, ok := annotations["alm-examples"]
```
ecordell: It might be nice to go ahead and support both `alm-examples` and `olm.examples` or something, so that we can start to phase out the ALM name.

pkg/schema/schema.go
```diff
@@ -241,6 +249,9 @@ func validateKind(kind string, fileBytes []byte) error {
 		if err != nil {
 			return err
 		}
+		if validateExamplesAnnotations(&csv) != nil {
```
ecordell: I don't think this is actually running - I tried to mess up an example and it still passed validation when loading ☹️

pkg/schema/schema.go
```diff
@@ -317,3 +328,90 @@ func validateResources(directory string) error {
...
+	var examples []CRDTemplate
+	var annotationsNames = []string{"alm-examples", "olm.examples"}
+	var found = false
+	annotations := csv.ObjectMeta.GetAnnotations()
+	// Return right away if no alm-examples are found.
+	if annotations == nil {
+		return fmt.Errorf("Unable to get annotations")
+	}
+
+	for _, name := range annotationsNames {
```
ecordell: I think you can clean this up a little bit:

```go
var annotationsExamples string
for _, name := range annotationsNames {
	annotationsExamples, ok := annotations[name]
	if !ok {
		continue
	}
}
if annotationsExamples == "" {
	return fmt.Errorf("Unable to get examples annotations")
}

// the rest
if json.Unmarshal([]byte(annotationsExamples), &examples) == nil {
//... 
}
```

it's a minor change but I think it makes it a little easier to read by removing the need for the `found` variable and letting you de-indent the rest of the function.

ecordell:
/lgtm

we should also follow up with validating the examples against their validation blocks, but this is a good start.

PR #49, opened 2019-05-20
feat(client): add health check to high-level client interface

ecordell:
Without this, client consumers need to access the underlying Conn object to perform a health check.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -2,8 +2,10 @@ package client
 
 import (
 	"context"
+	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 
 	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/operator-framework/operator-registry/pkg/api/grpc_health_v1"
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -61,6 +65,30 @@ func (c *Client) GetBundleThatProvides(ctx context.Context, group, version, kind
...
+		return nil
+	}
+	return c.Conn.Close()
+}
+
+func (c *Client) HealthCheck(ctx context.Context, reconnectTimeout time.Duration) (bool, error) {
+	res, err := c.Health.Check(ctx, &grpc_health_v1.HealthCheckRequest{Service: "Registry"})
+	if err != nil {
+		if c.Conn.GetState() == connectivity.TransientFailure {
+			ctx, _ := context.WithTimeout(ctx, reconnectTimeout)
```
jpeeler: Why do you discard the cancel function instead of calling it with defer?

ecordell: `WaitForStateChange` blocks, so the only time we'd hit the deferred cancel would be after we either successfully changed states or after the timeout defined here. 

I can call cancel here just for consistency though

PR #52, opened 2019-06-25
[bug] Remove duplicate packages from list

awgreene:
This change fixes a bug in the sanatizePackageList function that
allows duplicate packages to be included in the list.

### DIFF
--- a/pkg/appregistry/input_test.go
+++ b/pkg/appregistry/input_test.go
@@ -1,0 +1,16 @@
+package appregistry
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDuplicatePackage(t *testing.T) {
+	parser := inputParser{sourceSpecifier: &registrySpecifier{}}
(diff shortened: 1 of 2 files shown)


pkg/appregistry/input_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDuplicatePackage(t *testing.T) {
+	input := []string{"Jaeger", "Jaeger"}
+
+	expected := []string{"Jaeger"}
+	actual := sanitizePackageList(input)
```
tkashem: nit: Since we don't have any unit test for this layer, can we add a test for `Parse` instead of the private function `sanitizePackageList`? This way we can verify `Parse` works as expected.

awgreene: Good idea - addressed in the latest commit.

awgreene:
PR has been updated as requested @tkashem

PR #53, opened 2019-06-25
Allow users to specify a release for a package

awgreene:
Currently, the AppRegistry Server will download the latest release of
a package from quay. This update allows users to specify a specific
release.

### DIFF
--- a/pkg/appregistry/input_test.go
+++ b/pkg/appregistry/input_test.go
@@ -6,11 +6,29 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestDuplicatePackage(t *testing.T) {
+func TestDuplicatePackageResultHasNoRelease(t *testing.T) {
 	parser := inputParser{sourceSpecifier: &registrySpecifier{}}
-	expectedPackages := []string{"Jaeger"}
+	expectedPackages := []*Package{&Package{Name: "Jaeger", Release: ""}}
 
-	actual, err := parser.Parse([]string{"https://quay.io/cnr|community-operator|"}, "Jaeger,Jaeger")
+	actual, err := parser.Parse([]string{"https://quay.io/cnr|community-operator|"}, "Jaeger,Jaeger,Jaeger:0.0.1,Jaeger:0.0.2")
+	assert.NoError(t, err)
+	assert.Equal(t, expectedPackages, actual.Packages)
+}
+
+func TestDuplicatePackageResultHasRelease(t *testing.T) {
+	parser := inputParser{sourceSpecifier: &registrySpecifier{}}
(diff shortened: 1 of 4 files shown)


awgreene:
Updated based on feedback from @tkashem's [comment](https://github.com/operator-framework/operator-registry/pull/52#discussion_r297270873) from PR #52

ecordell:
Sorry @awgreene, this slipped by! Would you mind rebasing?

awgreene:
@ecordell will do!

awgreene:
I have finished rebasing @ecordell

pkg/appregistry/package.go
```diff
@@ -0,0 +1,28 @@
+package appregistry
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Package struct {
+	Name    string
```
ecordell: readability nit: we have established in registry / olm that "package" is specifically _not_ referring to a particular release, so I would rename this (and the other references to it) to indicate that it represents a particular release and not a package. 

Elsewhere in registry we refer to these as "ChannelEntries", but that may not fit this case since channel info is not used in appregistry.

pkg/appregistry/input_test.go
```diff
@@ -6,11 +6,29 @@ import (
...
 )
 
-func TestDuplicatePackage(t *testing.T) {
+func TestDuplicatePackageResultHasNoRelease(t *testing.T) {
 	parser := inputParser{sourceSpecifier: &registrySpecifier{}}
-	expectedPackages := []string{"Jaeger"}
+	expectedPackages := []*Package{&Package{Name: "Jaeger", Release: ""}}
 
-	actual, err := parser.Parse([]string{"https://quay.io/cnr|community-operator|"}, "Jaeger,Jaeger")
+	actual, err := parser.Parse([]string{"https://quay.io/cnr|community-operator|"}, "Jaeger,Jaeger,Jaeger:0.0.1,Jaeger:0.0.2")
```
ecordell: Can you explain these two test cases? Why is the order important?

awgreene: These tests ensure that existing behavior persists. When a packageList provided to `sanatizePackageList` contains multiple packages with the same name, only the first instance of the package is added. These test together ensure that behavior persists with the change to add a `Package struct` that allows users to specify a `release`.

pkg/appregistry/package.go
awgreene: Sure! This PR isn't a huge priority for our team so I will try and get around to it soon!

rthallisey:
how would a user specify the package? Can you document an example?

awgreene:
> how would a user specify the package? Can you document an example?

If you want to specify version 2.0.0 of the Jaeger operator you would provide "Jaeger:0.0.2". I will add documentation.

PR #54, opened 2019-07-02
fix(query): return all channels when querying

ecordell:
Previously this was returning a maximum of 2 channels

### DIFF
--- a/pkg/sqlite/query.go
+++ b/pkg/sqlite/query.go
@@ -94,7 +94,7 @@ func (s *SQLQuerier) GetPackage(ctx context.Context, name string) (*registry.Pac
 		},
 	}
 
-	if rows.Next() {
+	for rows.Next() {
 		if err := rows.Scan(&pkgName, &defaultChannel, &channelName, &bundleName); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


pkg/sqlite/query.go
```diff
@@ -94,7 +94,7 @@ func (s *SQLQuerier) GetPackage(ctx context.Context, name string) (*registry.Pac
 		},
 	}
 
-	if rows.Next() {
+	for rows.Next() {
```
njhale: 👍

tkashem: nit: we can collapse the `rows.Scan` for the first item into the loop, prepare the `channels` array first and then populate the `&registry.PackageManifest{`.

davidfestal:
@ecordell is there a way to benefit from this fix on an OS 4.1 cluster ?

ecordell:
@davidfestal We need to backport it if you need it in a 4.1 cluster, which is subject to a cherry-picking/approval process. Do you need it backported or can you use a 4.2 cluster?

davidfestal:
Would be great if it could be backported. We would use additional channels to test PRs of our operator. And the place to run the tests is on  4.1 cluster as for now.

PR #59, opened 2019-07-19
Refactor load/query to remove sqlite json requirement and dependency on OLM

ecordell:
This makes it significantly easier to re-use the load/query code in other projects.

Removing the json dependency means that projects that import the `sqlite` package don't have to specify additional build tags. Instead of loading the bundles into the db and then querying with the json extensions, we just query for the bundle directly and grab the parts we need from the whole CSV.

Removing the OLM dependency means that importing the `sqlite` package won't pull in a bunch of dependencies to satisfy the kube deps (for `runtime` from our API codegen). Instead of parsing bundles as real CSVs, we just treat them as unstructured blobs and pull the information we need from them. This will also help in the future when we make api changes to CSVs / migrate their data to another object.


pkg/registry/bundle.go
```diff
@@ -120,15 +116,24 @@ func (b *Bundle) RequiredAPIs() (map[APIKey]struct{}, error) {
 	if err != nil {
 		return nil, err
 	}
-	for _, api := range csv.Spec.CustomResourceDefinitions.Required {
+
+	_, requiredCRDs, err := csv.GetCustomResourceDefintions()
```
njhale: No difference between required and provided?

pkg/registry/bundle.go
```diff
@@ -143,8 +148,12 @@ func (b *Bundle) AllProvidedAPIsInBundle() error {
 	if err != nil {
 		return err
 	}
-	shouldExist := make(map[APIKey]struct{}, len(csv.Spec.CustomResourceDefinitions.Owned))
-	for _, crdDef := range csv.Spec.CustomResourceDefinitions.Owned {
+	ownedCRDs, _, err := csv.GetCustomResourceDefintions()
```
njhale: Is this function missing a check for APIServices?

pkg/registry/csv.go
```diff
@@ -0,0 +1,164 @@
...
+		return
+	}
+
+	rawValue, ok := objmap[customResourceDefinitions]
+	if !ok || rawValue == nil {
+		return
+	}
+
+	var definitions struct {
+		Owned    []*DefinitionKey `json:"owned"`
```
njhale: Any reason to use pointers in a slice here? Useful zero value?

ecordell: No particular reason

PR #62, opened 2019-07-23
[WIP] feat(load,query): when loading bundles, extract a list of images

ecordell:
/hold 
(this may or may not be included in 4.2 depending on the exception status)

This parses bundles to collect their images and stores them in the metadb so that we can query for them later. For now, it only collects operator images; before this merges it should also collect operand images.

Question: is it useful to be able to query separately for operator vs. operand images? (should we add labels to the image list to distinguish?)

TODO:

- [ ] load "related" images so that Operand images can be pulled as well


ecordell:
closed in favor of #101

PR #63, opened 2019-07-26
Bug 1783030: Detect replacement cycles while adding package channels

tmckayus:
If a cycle is created via "replaces" attributes in CSVs, the appregistry-server will infinite loop in the sqlite loader and never serve. This is a simple fix to detect the cycle and fail.

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -190,8 +190,10 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error 
 
 		channelEntryCSVName := c.CurrentCSVName
 		depth := 1
-		for {
 
+		// Since this loop depends on following 'replaces', keep track of where it's been
+		replaceCycle := map[string]bool{channelEntryCSVName: true}
+		for {
 			// Get CSV for current entry
 			channelEntryCSV, err := s.getCSV(tx, channelEntryCSVName)
 			if err != nil {
@@ -265,6 +267,14 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error 
 				errs = append(errs, err)
 				break
 			}
+
+			// If we find 'replaces' in the circuit list then we've seen it already, break out
+			if _, ok := replaceCycle[replaces]; ok {
+			        errs = append(errs, fmt.Errorf("Cycle detected, %s replaces %s", channelEntryCSVName, replaces))
+				break
+			}
+			replaceCycle[replaces] = true
+
 			replacedID, err := replacedChannelEntry.LastInsertId()
 			if err != nil {
 				errs = append(errs, err)
--- a/pkg/sqlite/configmap_test.go
+++ b/pkg/sqlite/configmap_test.go
@@ -1,11 +1,14 @@
 package sqlite
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"fmt"
+	"io/ioutil"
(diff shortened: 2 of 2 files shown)


tmckayus:
I ran into this accidentally while editing CSVs :)  On a local OCP instance running in a VM, this actually caused qemu to suck up near 100% CPU.   This change, in a pod, should cause the pod to fail and enter a crash backoff loop with a log message which should be more clear to a user if they happen to do what I did.

It might be nice to detect this in bundle validation in operator-courier as well.

tmckayus:
Attempted fix for https://github.com/operator-framework/operator-registry/issues/64

pkg/sqlite/load.go
```diff
@@ -254,6 +256,12 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
 				return fmt.Errorf("%s specifies replacement that couldn't be found", c.CurrentCSVName)
 			}
 
+			// If we find 'replaces' in the circuit list then we've seen it already, break out
+			if _, ok := replaceCircuit[replaces]; ok {
+			   return fmt.Errorf("Circuit detected, %s replaces %s", channelEntryCSVName, replaces)
+			}
+			replaceCircuit[channelEntryCSVName] = true
```
tmckayus: this should actually be "replaces", it still works but it will cause an extra time around the loop.

tmckayus: done

pkg/sqlite/load.go
```diff
@@ -268,7 +270,14 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
 			if err != nil {
 				errs = append(errs, err)
 				break
+			}	
+
+			// If we find 'replaces' in the circuit list then we've seen it already, break out
+			if _, ok := replaceCycle[replaces]; ok {
+			   return fmt.Errorf("Cycle detected, %s replaces %s", channelEntryCSVName, replaces)
```
njhale: Instead of returning immediately, I think we should append the error to the errors list and break from the for loop. This way, we can continue loading other channels and leave the decision to fail up to the client.

tmckayus: @njhale okay thanks for the feedback! I'll give it a try

tmckayus: okay, done! apologies for the delay

tmckayus:
rebased on master

jwforres:
@njhale is this good to merge now?

njhale:
@jwforres will review today. Currently, we also need to associate a BZ for this to merge, so I'll take care of that today as well.

tmckayus:
Thanks for the BZ @njhale .  TIL :)

> @jwforres will review today. Currently, we also need to associate a BZ for this to merge, so I'll take care of that today as well.

tmckayus:
@ecordell @njhale I'm not familiar enough with the CI to know if the failing test is the fault of this PR, or a larger issue.  Any insight?  Perhaps a rebase maybe?

njhale:
> Any insight? Perhaps a rebase maybe?
Prow merges master onto the branch before testing. Since it seems there were no conflicts there, I don't think rebasing will help in this case. I retested and everything seemed to work. 

After this merges I will manually check master, but I'm fairly certain that this patch doesn't cause any flakes on its own.

ecordell:
/retest

we're just hitting issues with test infra, doesn't look related to the project itself.

PR #68, opened 2019-08-05
Bug 1737586: Ignore files if we are unable to decode into kubernetes objects

shawn-hurley:
* Add log statement to make it clear that the file is being skipped
* Appears that go-format is making some white space changes.

### DIFF
--- a/pkg/sqlite/directory.go
+++ b/pkg/sqlite/directory.go
@@ -110,7 +110,7 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
 		return fmt.Errorf("error loading objs in dir: %s", err.Error())
 	}
 
-	if bundle.Size() == 0 {
+	if bundle == nil || bundle.Size() == 0 {
 		log.Warnf("no bundle objects found")
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/sqlite/directory.go
```diff
@@ -155,8 +155,10 @@ func (d *DirectoryLoader) LoadBundle(dir string) (*registry.Bundle, error) {
 
 		log.Info("found csv, loading bundle")
 		if err = decoder.Decode(obj); err != nil {
-			log.Infof("could not decode contents of file %s into file: %v", path, err)
-			return nil, nil
+			log.Infof("could not decode contents of file %s into file: %v - skipping", path, err)
+			// Explicitly set obj to nil, to make sure that nothing
+			// is added to the bundle
+			obj = nil
```
ecordell: I think `continue` would be clearer here

ecordell: This is also `continue` in master

knobunc:
@shawn-hurley does this reference the correct bug?  The bug has no link to this PR and is ON_QA... how can that be, if this hasn't merged?

bparees:
@shawn-hurley bump.  same issue @knobunc noted except now that BZ this PR references is already verified.....  (also now needs a rebase)

eparis:
Month later, wrong BZ, is this a PR we are actually working on, so should it just be closed?

njhale:
Looks like this [was fixed and verified](https://bugzilla.redhat.com/show_bug.cgi?id=1737586#c2) -- closing out.

PR #69, opened 2019-08-06
Bug 1732579: Best-effort loading

njhale:
This PR represents a more concise attempt to make appregistry-server more permissive of errors by allowing more operations to be best-effort; allowing loading to progress while aggregating errors (instead of stopping immediately). In lieu of simple logging, error aggregation is necessary to preserve the strict behavior of the configmap-server and initializer commands since appregistry-server uses both underlying load methods (configmap and directory) in various scenarios.


cmd/appregistry-server/main.go
```diff
@@ -57,7 +57,7 @@ func runCmdFunc(cmd *cobra.Command, args []string) error {
 	}
 	err = log.AddDefaultWriterHooks(terminationLogPath)
 	if err != nil {
-		return err
+		logrus.WithError(err).Warn("unable to set termination log path")
```
ecordell: should this actually error out at all? won't it prevent you from running locally in a lot of cases?

pkg/sqlite/directory.go
```diff
@@ -136,13 +154,14 @@ func (d *DirectoryLoader) LoadBundle(csvName string, dir string) (*registry.Bund
 		path := filepath.Join(dir, f.Name())
 		fileReader, err := os.Open(path)
 		if err != nil {
-			return nil, fmt.Errorf("unable to load file %s: %v", path, err)
+			errs = append(errs, fmt.Errorf("unable to load file %s: %s", path, err))
+			continue
```
ecordell: 👍

pkg/sqlite/load.go
```diff
@@ -192,86 +199,103 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
 			if err != nil {
-				return err
+				errs = append(errs, err)
+				break
 			}
 
 			if err := s.addProvidedAPIs(tx, channelEntryCSV, currentID); err != nil {
-				return err
+				errs = append(errs, err)
+				break
```
ecordell: We should be able to continue processing even if the provided apis can't be added. It just means the registry won't be able to answer dependency questions about this particular bundle.

pkg/sqlite/load.go
```diff
@@ -192,86 +199,103 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
 			}
 
 			skips, err := channelEntryCSV.GetSkips()
 			if err != nil {
-				return err
+				errs = append(errs, err)
+				break
```
ecordell: We should be able to continue processing even if we can't find the skips

cmd/appregistry-server/main.go
njhale: With these changes, this doesn't error out, it just prints a warning log on error.

pkg/sqlite/load.go
```diff
@@ -192,86 +199,101 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
 			}
 
 			if replaces == "" {
 				// we've walked the channel until there was no replacement
 				break
 			}
 
-			replaced, err := s.getCSV(tx, replaces)
+			replacedChannelEntry, err := addChannelEntry.Exec(c.Name, manifest.PackageName, replaces, depth)
```
shawn-hurley: I would like to understand why the order of this changed?

cmd/appregistry-server/main.go
```diff
@@ -89,17 +89,20 @@ func runCmdFunc(cmd *cobra.Command, args []string) error {
...
+		logger.Fatalf("error initializing: %s", err)
 	}
 
 	store, err := loader.Load(sources, packages)
 	if err != nil {
-		logger.Fatalf("error loading manifest from remote registry - %v", err)
+		logger.WithError(err).Warn("error loading app registry manifests")
+	}
+	if store == nil {
+		logger.Fatal("store failed to initialize")
```
shawn-hurley: If there are no bundles that are valid this will crash loop correct? Wondering if it makes sense to expose an empty catalog instead?

njhale: If there are no bundles the store will not be `nil`. My previous PR exposes an "Empty" `Query` implementation, but  this will only be nil when something is *really* wrong. If you think it's best I can pull in that empty implementation for this PR too.

pkg/sqlite/load.go
njhale: The change in ordering lets us add a channel entry for a "missing" replaced CSV.

pkg/sqlite/load.go
```diff
@@ -192,86 +199,101 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+
 			currentID = replacedID
 			channelEntryCSVName = replaces
-			depth += 1
+			depth++
 		}
 	}
-	return tx.Commit()
+
+	if err := tx.Commit(); err != nil {
```
ecordell: We should revisit this, and probably have more granular transactions. This one failing would nullify a lot of the work above to load as much as possible.

cmd/appregistry-server/main.go
njhale: I added the empty querier just to be safe.

pkg/sqlite/directory_test.go
```diff
@@ -2,11 +2,15 @@ package sqlite
...
 	"context"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/otiai10/copy"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
njhale: At a cursory glance, there was no encoder to match the decoder in the k8s package.

I didn't look into the ghodss package. Will follow-up soon.

PR #74, opened 2019-08-15
Permissive/Strict Options

njhale:
- Adds a permissive option to the intializer and configmap-server commands that, when given, permits load errors with out failing.
- Adds a strict option to the appregistry-server command that, when given, fails when load errors are encountered.

### DIFF
--- a/cmd/configmap-server/main.go
+++ b/cmd/configmap-server/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


cmd/configmap-server/main.go
```diff
@@ -96,26 +103,34 @@ func runCmdFunc(cmd *cobra.Command, args []string) error {
...
+	}
+	if store == nil {
+		store = registry.NewEmptyQuerier()
 	}
 
 	// sanity check that the db is available
 	tables, err := store.ListTables(context.TODO())
 	if err != nil {
-		logger.Fatalf("couldn't list tables in db, incorrect config: %v", err)
+		logger.WithError(err).Warnf("couldn't list tables in db")
```
jpeeler: Is there a chance of success at this point?

njhale: At this point, the store is effectively empty. I added this to keep in line with our theme of being permissive where we can. I figured an empty store should be non-fatal. Do you think it makes sense for the `permissive` option to gate this branch point too?

jpeeler: I was going back and forth on it.

PR #76, opened 2019-08-16
Skip non yaml or json files in initializer

jharrington22:
This PR attempts to fix an error where the initializer attempts to load files that are not yaml or json. The change from non permissive to permissive errors in #74 broke our builds because we have a README.md in all our repos at the package.yaml level.

### DIFF
--- a/pkg/sqlite/directory.go
+++ b/pkg/sqlite/directory.go
@@ -201,6 +201,11 @@ func (d *DirectoryLoader) LoadPackagesWalkFunc(path string, f os.FileInfo, err e
 		return nil
 	}
 
+	if !strings.HasSuffix(f.Name(), ".yaml") && !strings.HasSuffix(f.Name(), ".json") {
+		log.Info("skipping non yaml or json file")
+		return nil
+	}
+
 	fileReader, err := os.Open(path)
 	if err != nil {
 		return fmt.Errorf("unable to load package from file %s: %s", path, err)
--- a/pkg/sqlite/directory_test.go
+++ b/pkg/sqlite/directory_test.go
@@ -30,6 +30,39 @@ func TestDirectoryLoader(t *testing.T) {
 	require.NoError(t, loader.Populate())
 }
 
+func TestLoadBundleWalkFuncBadSuffix(t *testing.T) {
+	logrus.SetLevel(logrus.DebugLevel)
+
+	store, err := NewSQLLiteLoader("test.db")
+	require.NoError(t, err)
+	defer func() {
+		if err := os.Remove("test.db"); err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	// Copy golden manifests to a temp dir
+	dir, err := ioutil.TempDir("testdata", "manifests-")
+	require.NoError(t, err)
+	defer func() {
+		if err := os.RemoveAll(dir); err != nil {
+			t.Fatal(err)
+		}
+	}()
+	require.NoError(t, copy.Copy("../../manifests", dir))
+
+	// Create a non json or yaml file to trip parser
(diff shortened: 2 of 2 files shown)


pkg/sqlite/directory.go
```diff
@@ -201,6 +201,11 @@ func (d *DirectoryLoader) LoadPackagesWalkFunc(path string, f os.FileInfo, err e
 		return nil
 	}
 
+	if !strings.HasSuffix(f.Name(), ".yaml") && !strings.HasSuffix(f.Name(), ".json") {
```
cblecker: May wish to also add `.yml`

jharrington22: Yup this was a conscious decision though as I was following along with another example from this project so I wasn't sure if it was a choice to not look include `.yml` https://github.com/operator-framework/operator-registry/blob/master/pkg/schema/schema.go#L301-L320 - happy to add it though.

njhale:
@jharrington22 Thanks for taking the time to write a fix PR! The only additional thing I would like to see is a unit test confirming that this fixes the issue.

As an side:

> The change from non permissive to permissive errors in #74 broke our builds because we have a README.md in all our repos at the package.yaml level.

#74 just adds a `--permissive` option to the `initializer` and `configmap-server` commands and a `--strict` option to the `appregistry-server` command. It does not change the default behavior of these commands; `initializer` is still **not** permissive by default, so anything that failed before that PR would still fail after it was merged. That being said, I think that adding the `--permissive` option to your catalog build would fix the issue until this PR is merged.

jharrington22:
@njhale test added :)

jmelis:
@njhale it seems like the job failed for some random reason, but now it passed. Can you review and approve?

jharrington22:
@njhale anything I can do here?

jharrington22:
@ecordell maybe you could take a look at this for us? Thanks :slightly_smiling_face:

pbergene:
@njhale Can you please have a look at this, it's a functional regression that is currently affecting the delivery pipelines for hive and openshift dedicated.

pbergene:
Might be a bit blunt calling it a regression, but would love to get this loop completed. :)

pbergene:
Thanks Nick! (@njhale) Owe you a beverage of your choice.

njhale:
/hold

@pbergene Sorry lgtm was premature, need to get confirmation on something first. I just remembered some restrictions regarding merges to master right now. 

@ecordell does this need a BZ to merge?

jharrington22:
> @jharrington22 Thanks for taking the time to write a fix PR! The only additional thing I would like to see is a unit test confirming that this fixes the issue.
> 
> As an side:
> 
> > The change from non permissive to permissive errors in #74 broke our builds because we have a README.md in all our repos at the package.yaml level.
> 
> #74 just adds a `--permissive` option to the `initializer` and `configmap-server` commands and a `--strict` option to the `appregistry-server` command. It does not change the default behavior of these commands; `initializer` is still **not** permissive by default, so anything that failed before that PR would still fail after it was merged. That being said, I think that adding the `--permissive` option to your catalog build would fix the issue until this PR is merged.

I believe the issue is here https://github.com/operator-framework/operator-registry/pull/74/files#diff-ee55d2909ac16fc9f7cc6a6de6ac73c6R65-R68 as before we didn't return an err and now we do which is what is breaking our builds.

I'll add `--permissive` in the meantime though thanks :)

njhale:
I apologize for this sitting so long. This fell through the cracks during the release process. The initializer itself has been replaced with `opm index` commands, so I'm not sure this is as pressing anymore.

I'm going to close this out. Please feel free to reopen if it's is still a serious concern. Sorry again!

PR #78, opened 2019-08-19
chore(make): add rules to build/run upstream-example

njhale:


Makefile
```diff
@@ -1,7 +1,7 @@
 MOD_FLAGS := $(shell (go version | grep -q -E "1\.(11|12)") && echo -mod=vendor)
 CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | xargs -I{} basename {}))
 
-.PHONY: build test vendor clean
+.PHONY: build test vendor clean run-example image-upstream clean-example
```
jpeeler: I think you want example-image instead of image-upstream.

njhale:
This has been sitting around long enough -- I'm also not sure if these rules will be useful in the future.

PR #81, opened 2019-08-21
[release-4.1] Bug 1732580: Best-effort loading

njhale:
This PR reimplements best-effort loading and default permissiveness for the `appregistry-server` command for `release-4.1`.

/hold

### DIFF
--- a/pkg/sqlite/configmap.go
+++ b/pkg/sqlite/configmap.go
@@ -7,12 +7,14 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
-	"github.com/operator-framework/operator-registry/pkg/registry"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


ecordell:
/hold cancel

Couldn't reproduce the issue referenced above

pkg/sqlite/configmap.go
```diff
@@ -100,76 +104,84 @@ func (c *ConfigMapLoader) Populate() error {
...
+		errs = append(errs, fmt.Errorf("error loading CSV list: %s", err))
+		return utilerrors.NewAggregate(errs)
 	}
 
 	var parsedCSVList []v1alpha1.ClusterServiceVersion
 	err = json.Unmarshal([]byte(csvListJson), &parsedCSVList)
 	if err != nil {
-		c.log.WithError(err).Debug("error parsing CSV list")
-		return err
+		errs = append(errs, fmt.Errorf("error parsing CSV list: %s", err))
```
ecordell: This is fine given the goals of the PR are really geared towards appregistry, but -

if we wanted to be permissive here, wouldn't we treat this as a generic list, and attempt to parse each CSV one by one? As it stands now, if one doesn't parse, the whole thing will bail.

eparis:
labeling based on request from @mfojtik

PR #86, opened 2019-09-11
pkg/registry: use v1beta1.CustomResourceDefinition, as the apiextensions type is internal

estroz:
The type [`apiextensions.CustomResourceDefinition`](https://godoc.org/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions#CustomResourceDefinition) is an [internal type](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/register.go#L27) that should not be used for unmarshaling. Instead [`v1beta1.CustomResourceDefinition`](https://godoc.org/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1#CustomResourceDefinition) should be used. When unmarshaling the former type, fields are not properly populated, ex. `objectmeta.name` is missing. Using a `runtime.Decoder` with `Scheme` will not cause JSON number issues (at least with one manual test).

/cc @ecordell @njhale

### DIFF
--- a/pkg/registry/bundle.go
+++ b/pkg/registry/bundle.go
@@ -1,11 +1,9 @@
 package registry
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -35,7 +33,7 @@ type Bundle struct {
 	Package    string
 	Channel    string
 	csv        *ClusterServiceVersion
-	crds       []*apiextensions.CustomResourceDefinition
+	crds       []*v1beta1.CustomResourceDefinition
 	cacheStale bool
 }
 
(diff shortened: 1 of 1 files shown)


pkg/registry/bundle.go
```diff
@@ -210,19 +211,21 @@ func (b *Bundle) cache() error {
...
 			// converting JSON numbers to float64 for CRD minimum/maximum validation.
-			bytes, err := o.MarshalJSON()
+			cb, err := o.MarshalJSON()
 			if err != nil {
 				return err
 			}
-			if err := json.Unmarshal(bytes, &crd); err != nil {
-				return err
+			crd := &v1beta1.CustomResourceDefinition{}
+			dec := serializer.NewCodecFactory(Scheme).UniversalDeserializer()
```
estroz: This will throw an error if a CRD version created in the future wasn't registered with `Scheme`. This is could be considered a feature ("we don't explicitly support that CRD version yet, throw an error"); simply using `json.Unmarshal` may cause silent marshaling errors like the one I describe in the PR comments. Thoughts?

pkg/registry/bundle.go
```diff
@@ -210,19 +211,21 @@ func (b *Bundle) cache() error {
...
+			_, _, err = dec.Decode(cb, nil, crd)
```
estroz: This will throw an error if a CRD version created in the future wasn't registered with `Scheme`. This is could be considered a feature ("we don't explicitly support that CRD version yet, throw an error"); simply using `json.Unmarshal` may cause silent marshaling errors like the one I describe in the PR comments. Thoughts?

ecordell: That was the reasoning behind using the internal types in the first place, but it's reasonable to only support certain types.

PR #87, opened 2019-09-25
Bump to kube 1.16

ecordell:

This bumps the kube libraries to the versions used in kube 1.16.

It also adds one supporting method to the bundle `csv` abstraction so that OLM can get the version data without fully parsing the csv from the bundle.

**Reviewer Checklist**


pkg/registry/csv.go
```diff
@@ -66,6 +69,28 @@ func (csv *ClusterServiceVersion) GetReplaces() (string, error) {
 	return replaces, nil
 }
 
+// GetVersion returns the version of the CSV
+//
+// If not defined, the function returns an empty string.
+func (csv *ClusterServiceVersion) GetVersion() (string, error) {
```
ecordell: This is the only non-dependency change

jpeeler:
I'm curious how to properly solve this:

vendor/github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient/serviceaccount.go:38:2: not enough arguments to return
vendor/github.com/operator-framework/operator-lifecycle-manager/pkg/lib/operatorclient/serviceaccount.go:38:10: c.Core undefined (type *Client has no field or method Core)

The first OLM release to not use c.Core is 4.2. Perhaps the rule should be to vendor the latest OLM that contains the most recent rebase?

dinhxuanvu:
@ecordell I assume you remove the schema because we have a validation library now and will vendor that library at a later time?

ecordell:
@dinhxuanvu Yep! I mentioned that in the commit message as well.

ecordell:
/hold 

Just noticed the version of olm being vendored is wrong.

ecordell:
/hold cancel

operator-registry no longer depends on OLM at all 🙃

PR #88, opened 2019-09-26
Add migration proposal

kevinrizza:
### Introduce database migrations to the operator-registry database

This proposal intends to introduce a method of migrating database schema to new versions on additive operations to existing registry databases.

### DIFF
--- a/docs/contributors/design-proposals/db-migrations.md
+++ b/docs/contributors/design-proposals/db-migrations.md
@@ -1,0 +1,61 @@
+# Add database versioning and migrations to output database
+
+Status: Pending
+
+Version: alpha
+
+Implementation owner: TBD
+
+## Abstract
+
+Proposal to add database versioning and migration capability to the output of operator-registry commands
+
+## Motivation
+
+In an attempt at optimizing the registry database usage in OLM to allow the registry database to be stored in a container image, a method is needed to add to the database over time. Since these database files will be long living, it becomes a requirement that the database can migrate to a new schema over time. This proposal attempts to solve that problem by implementing a method with which database versions can be migrated with existing db files over time.
+
+## Proposal
+
+### Migration Method
+
+At a high level, the migration method is relatively straightforward. The database will contain a current migration version and when any additive operation is called with an existing database as input, the first step will be to run the migration to upgrade the database schema to the latest version.
+
+Rather than reinventing the database migration process, operator-registry will use an existing database migration tool that has already defined a set of semantics and conventions. This proposal suggests using the [golang-migrate](https://github.com/golang-migrate/migrate) tool for this purpose. Golang-migrate uses a set of conventions to define migrations.
+
+Firstly, it defines migrations as a set of sql files that live in a flat database folder. Migrations are individually defined as a pair of up and down migration scripts. These up and down scripts each define a method of going to and back from a particular database version and should be opposites semantically. Each script pair has a unique 64 bit unsigned integer identifier followed by an underscore, as well as a title, a direction, and ends in the `.sql` extension: `${version}_${title}.${direction(down/up)}.sql`. The versions are ordered, with the lowest integer value coming first and the latest version defining the current database migration version in the db. For more details on the migration format, please see https://github.com/golang-migrate/migrate/blob/master/MIGRATIONS.md.
+
+```
+ # example migrations folder
+ db_migrations
+ ├── 201909251522_add_users_table.up.sql
+ ├── 201909251522_add_users_table.down.sql
+ ├── 201909251510_first_migration.up.sql
+ └── 201909251510_first_migration.down.sql
+```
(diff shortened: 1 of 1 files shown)


doc/contributors/design-proposals/db-migrations.md
```diff
@@ -0,0 +1,40 @@
...
+ ├── 201909251510_first_migration.up.sql
+ └── 201909251510_first_migration.down.sql
+```
+
+Once that migration schema is defined, whenever any additive operation is run against the database we will first use the migrate API to upgrade the schema to the latest version. See https://godoc.org/github.com/golang-migrate/migrate#Migrate.Up for more details on that API.
+
+### Versioning fresh databases
+
+One other consideration of note here is that the operator-registry database is not a long living database in the traditional sense, since the database is often created from scratch. All common migration tools do not generally account for such an edge case, so this proposal also defines a method of initializing the database at a particular migration version. Since the migration version is a matter of convention on file names, we can infer the migration version by parsing the migration folder and finding the latest migration script version. Then, once we initialize the database on startup we can use `golang-migrate` API to force the initial migration version. See https://godoc.org/github.com/golang-migrate/migrate#Migrate.Force for more details on that API.
```
ecordell: This seems to imply that we maintain two things:

- a set of migrations
- a sql script that can initialize the database to whatever the latest schema is

Do we have a plan for keeping them in sync?

kevinrizza: I'm not suggesting that at all. All of the migrations themselves will be run in go source code, so the thing that initializes the database looks something like:

```
initVersion, err := getLatestMigrationVersion()
if err != nil {
	return err
}

err = migrator.Force(initVersion)
if err != nil {
	return err
}
```

We would just implement the `getLatestMigrationVersion()` to parse the migration folder and pull the latest version every time. Then we just call something that wraps this whenever the `initialize` command is called.

kevinrizza: Also, sorry if this isn't clear, the `migrator.Force(initVersion)` API is something provided by `golang-migrate`, so we wouldn't have to be aware of the implementation of how the version is stored in the database.

ecordell: Migrations typically just contain the "diffs" to the schema via sql commands.

If you start from the schema we have today ("initial schema") you need to:

- initial schema
- `+` migration 1 up
- `+` migration 2 up
... 
- `+` migration latest up

in order to get to the "latest schema". 

An alternative is to keep the full "latest schema" as well, and use that when initializing from scratch (and then force the migration version like you mentioned). But then you need to ensure that "latest schema" = "initial schema" + "migrations" always. 

Since you mentioned wanting to force-set the schema version to latest, I assumed that means we also need the full "latest schema" somewhere, so I was asking about how we plan to keep that invariant true ("lates schema" = "initial schema" + "migrations"). 

Does my concern make sense?

kevinrizza: Do we not have the "latest schema" already defined? https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/load.go#L29

The developer workflow then becomes:

Update that file to the new schema however you want to.
Create a migration with the diff to migrate the existing tables from that version to the one you just defined by updating `load.go`.

Or is there some other schema modification that I'm not aware of? I'm not sure that moving it into a script is much better since we will still have to call the force method, otherwise we will need to be aware of changes to the way that `golang-migrate` defines its schema version. I could see the argument that we might want to move that thing into a separate file and serialize it, though I'm not sure that really changes anything.

This is an attempt to try to avoid having to keep the "original schema" around, since that seems a lot more difficult to reason about than having to think about the diff every time.

ecordell: It's these two things I'm concerned about keeping in sync:

- Update that file to the new schema however you want to.
- Create a migration with the diff to migrate the existing tables from that version to the one you just defined by updating load.go.

There's nothing in this proposal that ensures "existing tables + diff" == "new schema defined in load.go".

kevinrizza: Yep, my bad, this wasn't clear enough. I added a section explicitly calling out the fact that the latest schema should continue to be defined in `load.go`/wherever if it moves and also explicitly calls out testing the diff between from scratch dbs and migrated dbs.

doc/contributors/design-proposals/db-migrations.md
```diff
@@ -0,0 +1,61 @@
+# Add database versioning and migrations to output database
```
ecordell: Super nit: I think the convention used everywhere else is `docs/...` not `doc/...` and we made a mistake in OLM repo.

kevinrizza: great point. updated.

PR #89, opened 2019-09-26
Adding OPM cli with registry add and rm commands

gallettilance:

Adding a new cli to operator-registry called `opm`. It can add bundles to a database, delete packages from that db and serve its contents at a specified host and port.

Having built the binaries, a user can run the old binaries

```
./bin/initializer ...
./bin/appregistry-server ...
./bin/registry-server ...
./bin/configmap-server ...
```

or can use the new cli that supports (and will support) our new operator bundle format. Here is an example flow:

Build the database (or add to an existing one) using

```
./bin/opm registry add -d=bundles.db -b="quay.io/username/imagename,quay.io/username/otherimagename"
```

Remove operators from an existing db using

```
./bin/opm registry rm -d=bundles.db -o="operator-package-name"
```

Serve the db using (can optionally specify a host and port)

```
./bin/opm registry serve -t=path/to/file.log -d=bundles.db
```

This is a first step toward creating new commands to interface with operator-registry as part of its corresponding epic.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/registry/add.go
+++ b/cmd/opm/registry/add.go
@@ -1,0 +1,81 @@
+package registry
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/pkg/sqlite"
+)
+
+func newRegistryAddCmd() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "add",
+		Short: "add operator bundle to operator registry DB",
+		Long:  `add operator bundle to operator registry DB`,
+
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if debug, _ := cmd.Flags().GetBool("debug"); debug {
+				logrus.SetLevel(logrus.DebugLevel)
+			}
+			return nil
+		},
+
+		RunE: addFunc,
+	}
+
+	rootCmd.Flags().Bool("debug", false, "enable debug logging")
+	rootCmd.Flags().StringP("database", "d", "bundles.db", "relative path to database file")
+	rootCmd.Flags().StringSliceP("bundle-images", "b", []string{}, "comma separated list of links to bundle image")
+	rootCmd.Flags().Bool("permissive", false, "allow registry load errors")
+
+	return rootCmd
+}
+
+func addFunc(cmd *cobra.Command, args []string) error {
+	bundleImages, err := cmd.Flags().GetStringSlice("bundle-images")
+	if err != nil {
+		return err
+	}
+
+	fromFilename, err := cmd.Flags().GetString("database")
+	if err != nil {
+		return err
+	}
+	permissive, err := cmd.Flags().GetBool("permissive")
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+
+	db, err := sql.Open("sqlite3", fromFilename)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	dbLoader, err := sqlite.NewSQLLiteLoader(db)
+	if err != nil {
+		return err
+	}
+	if err := dbLoader.Migrate(context.TODO()); err != nil {
+		return err
+	}
+
+	for _, bundleImage := range bundleImages {
+		loader := sqlite.NewSQLLoaderForImage(dbLoader, bundleImage)
+		if err := loader.Populate(); err != nil {
+			err = fmt.Errorf("error loading bundle from image: %s", err)
+			if !permissive {
+				logrus.WithError(err).Fatal("permissive mode disabled")
+				errs = append(errs, err)
+			}
+			logrus.WithError(err).Warn("permissive mode enabled")
+		}
+	}
+	return nil
+}
--- a/pkg/containertools/command.go
+++ b/pkg/containertools/command.go
@@ -1,0 +1,66 @@
+package containertools
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+const (
+	// Podman cli tool
+	Podman = "podman"
+	// Docker cli tool
+	Docker = "docker"
+)
+
+// CommandRunner is configured to select a container cli tool and execute commands with that
+// tooling.
+type CommandRunner struct {
+	containerTool string
+}
+
+// NewCommandRunner takes the containerTool as an input string and returns a CommandRunner to
+// run commands with that cli tool
+func NewCommandRunner(containerTool string) *CommandRunner {
+	r := &CommandRunner{}
+
+	switch containerTool {
+	case Podman:
+		r.containerTool = Podman
+	case Docker:
+		r.containerTool = Docker
+	default:
+		r.containerTool = Podman
+	}
+
+	return r
+}
+
+// Pull takes a container image path hosted on a container registry and runs the pull command to
+// download it onto the local environment
+func (r *CommandRunner) Pull(image string) error {
+	args := []string{"pull", image}
+
+	command := exec.Command(r.containerTool, args...)
+
+	out, err := command.Output()
+	if err != nil {
+		return fmt.Errorf("error with %s %s: %s", command.Args, string(out), err)
+	}
+
+	return nil
+}
+
+// Save takes a local container image and runs the save commmand to convert the image into a specified
+// tarball and push it to the local directory
+func (r *CommandRunner) Save(image, tarFile string) error {
+	args := []string{"save", image, "-o", tarFile}
+
+	command := exec.Command(r.containerTool, args...)
+
+	out, err := command.Output()
+	if err != nil {
+		return fmt.Errorf("error with %s %s: %s", command.Args, string(out), err)
+	}
+
+	return nil
+}
--- a/pkg/registry/interface.go
+++ b/pkg/registry/interface.go
@@ -8,7 +8,10 @@ import (
 
 type Load interface {
 	AddOperatorBundle(bundle *Bundle) error
+	AddBundlePackageChannels(manifest PackageManifest, bundle Bundle) error
 	AddPackageChannels(manifest PackageManifest) error
+	RmPackageName(packageName string) error
+	ClearNonDefaultBundles(packageName string) error
 }
 
 type Query interface {
--- a/pkg/sqlite/directory.go
+++ b/pkg/sqlite/directory.go
@@ -107,7 +107,7 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
 	log.Info("found csv, loading bundle")
 
 	var errs []error
-	bundle, err := d.loadBundle(csv.GetName(), filepath.Dir(path))
+	bundle, err := loadBundle(csv.GetName(), filepath.Dir(path))
 	if err != nil {
 		errs = append(errs, fmt.Errorf("error loading objs in directory: %s", err))
 	}
@@ -128,10 +128,56 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
 	return utilerrors.NewAggregate(errs)
 }
 
-// LoadBundle takes the directory that a CSV is in and assumes the rest of the objects in that directory
+// LoadPackagesWalkFunc attempts to unmarshal the file at the given path into a PackageManifest resource.
+// If unmarshaling is successful, the PackageManifest is added to the loader's store.
+func (d *DirectoryLoader) LoadPackagesWalkFunc(path string, f os.FileInfo, err error) error {
+	if f == nil {
+		return fmt.Errorf("invalid file: %v", f)
+	}
+
+	log := logrus.WithFields(logrus.Fields{"dir": d.directory, "file": f.Name(), "load": "package"})
+	if f.IsDir() {
+		if strings.HasPrefix(f.Name(), ".") {
+			log.Info("skipping hidden directory")
+			return filepath.SkipDir
+		}
+		log.Info("directory")
+		return nil
+	}
+
+	if strings.HasPrefix(f.Name(), ".") {
+		log.Info("skipping hidden file")
+		return nil
+	}
+
+	fileReader, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("unable to load package from file %s: %s", path, err)
+	}
+
+	decoder := yaml.NewYAMLOrJSONDecoder(fileReader, 30)
+	manifest := registry.PackageManifest{}
+	if err = decoder.Decode(&manifest); err != nil {
+		if err != nil {
+			return fmt.Errorf("could not decode contents of file %s into package: %s", path, err)
+		}
+
+	}
+	if manifest.PackageName == "" {
+		return nil
+	}
+
+	if err := d.store.AddPackageChannels(manifest); err != nil {
+		return fmt.Errorf("error loading package into db: %s", err)
+	}
+
+	return nil
+}
+
+// loadBundle takes the directory that a CSV is in and assumes the rest of the objects in that directory
 // are part of the bundle.
-func (d *DirectoryLoader) loadBundle(csvName string, dir string) (*registry.Bundle, error) {
-	log := logrus.WithFields(logrus.Fields{"dir": d.directory, "load": "bundle"})
+func loadBundle(csvName string, dir string) (*registry.Bundle, error) {
+	log := logrus.WithFields(logrus.Fields{"dir": dir, "load": "bundle"})
 	files, err := ioutil.ReadDir(dir)
 	if err != nil {
 		return nil, err
@@ -177,47 +223,4 @@ func (d *DirectoryLoader) loadBundle(csvName string, dir string) (*registry.Bund
 	}
 
 	return bundle, utilerrors.NewAggregate(errs)
-}
-
-// LoadPackagesWalkFunc attempts to unmarshal the file at the given path into a PackageManifest resource.
-// If unmarshaling is successful, the PackageManifest is added to the loader's store.
-func (d *DirectoryLoader) LoadPackagesWalkFunc(path string, f os.FileInfo, err error) error {
-	if f == nil {
-		return fmt.Errorf("invalid file: %v", f)
-	}
-
-	log := logrus.WithFields(logrus.Fields{"dir": d.directory, "file": f.Name(), "load": "package"})
-	if f.IsDir() {
-		if strings.HasPrefix(f.Name(), ".") {
-			log.Info("skipping hidden directory")
-			return filepath.SkipDir
-		}
-		log.Info("directory")
-		return nil
-	}
-
-	if strings.HasPrefix(f.Name(), ".") {
-		log.Info("skipping hidden file")
-		return nil
-	}
-
-	fileReader, err := os.Open(path)
-	if err != nil {
-		return fmt.Errorf("unable to load package from file %s: %s", path, err)
-	}
-
-	decoder := yaml.NewYAMLOrJSONDecoder(fileReader, 30)
-	manifest := registry.PackageManifest{}
-	if err = decoder.Decode(&manifest); err != nil {
-		return fmt.Errorf("could not decode contents of file %s into package: %s", path, err)
-	}
-	if manifest.PackageName == "" {
-		return nil
-	}
-
-	if err := d.store.AddPackageChannels(manifest); err != nil {
-		return fmt.Errorf("error loading package into db: %s", err)
-	}
-
-	return nil
 }
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -1,0 +1,239 @@
+package sqlite
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/yaml"
+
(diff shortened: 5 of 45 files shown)


pkg/cli/appregistryserver.go
```diff
@@ -0,0 +1,127 @@
+package cli
```
kevinrizza: I think this code should still live in the ./cmd folder. Why are we moving it to the ./pkg folder?

kevinrizza: And this comment goes for all the different sub commands

gallettilance: So for example have a folder `cmd/commands` where the code that is now in `pkg/cli` lives? I don't see why that wouldn't work. We would just have to change the `Makefile` a bit I think. I don't have a good argument why it's in `pkg` :)

kevinrizza: I'm actually suggesting that you can just leave 95% of this code exactly where it is. You can create your operator-registry command as a file in the root of the existing `./cmd/` folder called `main.go` that just imports those sub packages. Then just refactor the existing `main()` functions in each command slightly so that they do exactly what you have done to update them. Right now this PR has a ton of noise that makes it hard to read and I'm just suggesting that we can make this more readable without creating any extraneous files in the `./pkg/` folder.

kevinrizza:
looks good to me

Makefile
```diff
@@ -1,5 +1,5 @@
 MOD_FLAGS := $(shell (go version | grep -q -E "1\.(11|12)") && echo -mod=vendor)
-CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | xargs -I{} basename {}))
+CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | xargs -I{} basename {} | grep -Fv "cli"))
```
jpeeler: Nit: I think it'd be better to at least use "cmd/cli" here, but it's probably fine.

jpeeler: Oops, I didn't think about the xargs here. Ignore me.

cmd/operator-registry/main.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	cli "github.com/operator-framework/operator-registry/cmd/cli"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "operator-registry",
```
ecordell: let's start the bikeshed: 

`opreg`?

gallettilance: I liked what @njhale said at scrum: `opm`

kevinrizza: +1

njhale: `krab` 🦀

jpeeler: > I liked what @njhale said at scrum: `opm`

What does opm stand for?

tkashem: `op`erator `m`atters :)

ecordell: I think it was `operator package manager`

What about  `opreg` for the server stuff (configmap, appregistry, registry-server today), `opm` for bundle / index stuff (initializer today)?

gallettilance: It makes sense to me but didn't we want to combine all these into a single binary?

ecordell:
@gallettilance This is looking good!

I think we just need:

- the result of the bikeshedding discussion recorded (`opm serve`, `opm build`, etc)
- an update to the readme to acknowledge these new commands
- a scaffold of a man page / doc that we can update as we add functionality to the root command

cmd/opm/registry/add.go
```diff
@@ -0,0 +1,85 @@
...
+	fromFilename, err := cmd.Flags().GetString("database")
+	if err != nil {
+		return err
+	}
+	permissive, err := cmd.Flags().GetBool("permissive")
+	if err != nil {
+		return err
+	}
+
+	bundleList := strings.Split(bundleImages, ",")
```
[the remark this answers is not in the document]
gallettilance: Thanks - I was copying the pattern from the registry-server where the parsing happens outside of cobra command

pkg/containertools/bundlereader_test.go
```diff
@@ -0,0 +1,41 @@
...
+	temp, err := ioutil.TempDir("./", "bundlereader_test_")
+	require.NoError(t,err)
+	defer func() {
+		require.NoError(t, os.RemoveAll(temp))
+	}()
+
+	logger := logrus.NewEntry(logrus.New())
+	mockCmd := mock.NewContainerToolsCommandRunner(controller)
+
+	mockCmd.EXPECT().Pull(image).Return(nil)
```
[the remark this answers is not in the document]
gallettilance: Oh that's right my bad - Kevin mentioned this to me yesterday and I haven't refactored that yet.

pkg/sqlite/directory.go
```diff
@@ -20,14 +20,68 @@ const ClusterServiceVersionKind = "ClusterServiceVersion"
...
+// PackageRemover loads a directory of resources into the database
+type PackageRemover struct {
+	store    registry.Load
+	packages string
+}
+
 var _ SQLPopulator = &DirectoryLoader{}
+var _ SQLRemover = &PackageRemover{}
+
+func NewSQLRemoverForPackages(store registry.Load, packages string) *PackageRemover {
```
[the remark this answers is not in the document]
gallettilance: makes sense to me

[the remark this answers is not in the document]
gallettilance: will do

pkg/sqlite/migrations/000_init.go
```diff
@@ -12,22 +12,22 @@ var initMigration = &Migration{
...
-			csv TEXT UNIQUE,
+			name TEXT PRIMARY KEY,  
+			csv TEXT UNIQUE, 
 			bundle TEXT
 		);
 		CREATE TABLE IF NOT EXISTS package (
 			name TEXT PRIMARY KEY,
 			default_channel TEXT,
-			FOREIGN KEY(name, default_channel) REFERENCES channel(package_name,name)
+			FOREIGN KEY(name, default_channel) REFERENCES channel(package_name,name) ON DELETE CASCADE
```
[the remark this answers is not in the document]
gallettilance: Something needs to clean up that package table when the channel is deleted from the channel table

cmd/opm/registry/add.go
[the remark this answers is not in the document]
kevinrizza: `thing.Flags().StringSliceP()` is what you're looking for

pkg/containertools/bundlereader_test.go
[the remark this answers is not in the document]
kevinrizza: if it's easier you can just leave this as is and I will update it to use counterfeiter in the index pr commit where i'm introducing it.

pkg/containertools/command.go
```diff
@@ -0,0 +1,66 @@
...
+
+// CommandRunner is configured to select a container cli tool and execute commands with that
+// tooling.
+type CommandRunner struct {
+	containerTool string
+}
+
+// NewCommandRunner takes the containerTool as an input string and returns a CommandRunner to
+// run commands with that cli tool
+func NewCommandRunner(containerTool string) *CommandRunner {
```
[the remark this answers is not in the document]
kevinrizza: i'll just rebase on top of this to include all the tests I added and remove this file.

pkg/sqlite/migrations/000_init.go
```diff
@@ -12,22 +12,22 @@ var initMigration = &Migration{
...
-			name TEXT,
-			package_name TEXT,
+			name TEXT, 
+			package_name TEXT, 
 			head_operatorbundle_name TEXT,
 			PRIMARY KEY(name, package_name),
-			FOREIGN KEY(package_name) REFERENCES package(name),
-			FOREIGN KEY(head_operatorbundle_name) REFERENCES operatorbundle(name)
+			FOREIGN KEY(package_name) REFERENCES package(name) DEFERRABLE INITIALLY DEFERRED,
+			FOREIGN KEY(head_operatorbundle_name) REFERENCES operatorbundle(name) ON DELETE CASCADE
```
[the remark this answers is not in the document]
kevinrizza: @ecordell for later we should probably introduce a test that guards against this

pkg/sqlite/directory.go
```diff
@@ -20,14 +20,68 @@ const ClusterServiceVersionKind = "ClusterServiceVersion"
...
+
+	return utilerrors.NewAggregate(errs)
+}
+
+// sanitizePackageList sanitizes the set of package(s) specified. It removes
+// duplicates and ignores empty string.
+func sanitizePackageList(in []string) []string {
+	out := make([]string, 0)
+
+	inMap := map[string]bool{}
```
kevinrizza: nit:

you can just use `map[string]struct{}` for a "hash list" implementation
instead of
`inMap[item] = true`

just use
`inMap[item] = struct{}{}`

In the general use case for that data structure it's nice because you can iterate over the list and get a sort of `is in map` syntax a little more easily and it's slightly less data to pass around.

second nit:
Also what is the purpose of checking `item == ""` in the map check function? Filtering empty string from the list seems like a totally unrelated check. Can we break them apart? It's a little easier to read that way.
```
if item == "" {
	continue
}

if _, ok := inMap[item]; ok {
	continue
}
```

pkg/sqlite/directory.go
```diff
@@ -128,10 +182,56 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
...
+// If unmarshaling is successful, the PackageManifest is added to the loader's store.
+func (d *DirectoryLoader) LoadPackagesWalkFunc(path string, f os.FileInfo, err error) error {
+	if f == nil {
+		return fmt.Errorf("invalid file: %v", f)
+	}
+
+	log := logrus.WithFields(logrus.Fields{"dir": d.directory, "file": f.Name(), "load": "package"})
+	if f.IsDir() {
+		if strings.HasPrefix(f.Name(), ".") {
+			log.Info("skipping hidden directory")
```
kevinrizza: Should this log message have slightly more context?

pkg/sqlite/directory.go
```diff
@@ -128,10 +182,56 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
...
+			return filepath.SkipDir
+		}
+		log.Info("directory")
+		return nil
+	}
+
+	if strings.HasPrefix(f.Name(), ".") {
+		log.Info("skipping hidden file")
```
kevinrizza: Same comment about log context as above

pkg/sqlite/directory.go
```diff
@@ -128,10 +182,56 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
...
+		return fmt.Errorf("invalid file: %v", f)
+	}
+
+	log := logrus.WithFields(logrus.Fields{"dir": d.directory, "file": f.Name(), "load": "package"})
+	if f.IsDir() {
+		if strings.HasPrefix(f.Name(), ".") {
+			log.Info("skipping hidden directory")
+			return filepath.SkipDir
+		}
+		log.Info("directory")
```
kevinrizza: Should this be a debug statement? It doesn't seem to mean anything from a user perspective.

pkg/sqlite/directory.go
```diff
@@ -128,10 +182,56 @@ func (d *DirectoryLoader) LoadBundleWalkFunc(path string, f os.FileInfo, err err
 	return utilerrors.NewAggregate(errs)
 }
 
-// LoadBundle takes the directory that a CSV is in and assumes the rest of the objects in that directory
+// LoadPackagesWalkFunc attempts to unmarshal the file at the given path into a PackageManifest resource.
+// If unmarshaling is successful, the PackageManifest is added to the loader's store.
+func (d *DirectoryLoader) LoadPackagesWalkFunc(path string, f os.FileInfo, err error) error {
```
kevinrizza: Sorry just realized, was this whole function copied and pasted from a different part of the file? Why?

gallettilance: yeah I'm not sure why git thinks I re-wrote it...

pkg/sqlite/directory.go
gallettilance: this is code I stole from somewhere else in the code base... But yes I agree and happy to fix it

pkg/sqlite/directory.go
ecordell: The log will have all of the fields added to it, so the output shows us visiting each node.

pkg/sqlite/directory.go
ecordell: It's in the fields on the log

pkg/sqlite/image.go
```diff
@@ -0,0 +1,239 @@
...
+	log.Infof("loading Bundle %s", i.image)
+	errs := make([]error, 0)
+	if err := i.LoadBundleFunc(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+// LoadBundleFunc walks the directory. When it sees a `.clusterserviceversion.yaml` file, it
```
ecordell: nit: description is wrong

pkg/sqlite/image.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return fmt.Errorf("Could not translate annotations file into packageManifest %s", err)
+	}
+
+	if err := i.loadOperatorBundle(packageManifest, *bundle); err != nil {
+		return fmt.Errorf("Error adding package %s", err)
+	}
+
+	// Finally let's delete all the old bundles
+	if err = i.store.ClearNonDefaultBundles(packageManifest.PackageName); err != nil {
```
ecordell: this should be behind a flag at least - it limits our options for transitioning (i.e. if I add an image bundle to an existing database, I lose all of my operator history...)

pkg/sqlite/load.go
```diff
@@ -386,3 +487,459 @@ func (s *SQLLoader) addAPIs(tx *sql.Tx, csv *registry.ClusterServiceVersion, cha
...
+	}
+	// TODO: bulk insert
+	for img := range imgs {
+		if _, err := addImage.Exec(img, csvName); err != nil {
+			return err
+		}
+	}
+
+	if err := tx.Commit(); err != nil {
+		return err
```
ecordell: if a commit fails, you should call `tx.Rollback()`

pkg/sqlite/image.go
kevinrizza: Are you imagining that we want to support a workflow like that? imo if you are using `opm` it means you are just getting the new stuff that version of opm provides. otherwise it will start getting really hard about reasoning about migrations, for example.

if they want an old catalog, they are still able to use `initialize`

gallettilance: Is anything in the csv that isn't in the db (as far as function is concerned)?

cmd/opm/registry/add.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+			return nil
+		},
+
+		RunE: addFunc,
+	}
+
+	rootCmd.Flags().Bool("debug", false, "enable debug logging")
+	rootCmd.Flags().StringP("database", "d", "bundles.db", "relative path to database file")
+	rootCmd.Flags().StringSliceP("bundle-images", "b", []string{}, "comma separated list of links to bundle image")
```
ecordell: non-blocking: did we think about making these positional? 

```
opm add -d bundles.db quay.io/my/manifests:1 quay.io/my/manifests:2 
```

pkg/registry/interface.go
```diff
@@ -8,7 +8,10 @@ import (
 
 type Load interface {
 	AddOperatorBundle(bundle *Bundle) error
+	AddBundlePackageChannels(manifest PackageManifest, bundle Bundle) error
 	AddPackageChannels(manifest PackageManifest) error
+	RmPackageName(packageName string) error
+	ClearNonDefaultBundles(packageName string) error
```
ecordell: non-blocking: I think these should probably be separate interfaces. we can refactor later.

cmd/opm/registry/add.go
gallettilance: That looks cool

pkg/sqlite/load.go
```diff
@@ -135,8 +135,11 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
 	defer addReplaces.Close()
 
 	if _, err := addPackage.Exec(manifest.PackageName); err != nil {
```
ecordell: instead of this, can we just always "update" package channels, so that it creates if it's not there? (rather than trying to write data and falling back to update on failure?)

non-blocking, we can follow up.

gallettilance: In the old format, it was not allowed to have packagename collisions and that's exactly the case in the new format where we are trying to update the db to contain a new version of some operator. I tried to take advantage of that such that when there are no previous versions of an operator in the db it defaults to the old format behavior

pkg/sqlite/migrations/004_cascade_delete_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-registry/pkg/sqlite/migrations"
+)
+
+func TestCascadeDeleteUp(t *testing.T) {
+	// migrate up to, but not including, this migration
+	db, migrator, cleanup := CreateTestDbAt(t, migrations.CascadeDeleteMigrationKey-1)
+	defer cleanup()
+
```
ecordell: you should validate that the table doesn't have the cascades here, before migrating up

pkg/sqlite/migrations/004_cascade_delete_test.go
```diff
@@ -0,0 +1,272 @@
...
+	require.NoError(t, err)
+	require.Equal(t, createNewRelatedImageTable, createStatement)
+	err = table.Close()
+	require.NoError(t, err)
+}
+
+func TestCascadeDeleteDown(t *testing.T) {
+	db, migrator, cleanup := CreateTestDbAt(t, migrations.CascadeDeleteMigrationKey)
+	defer cleanup()
+
```
ecordell: again - you should validate the correct state here

pkg/sqlite/load.go
```diff
@@ -135,8 +135,11 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
-		// This should be terminal
-		return err
+		err = s.updatePackageChannels(tx, manifest)
+		if err != nil {
+			return err
```
ecordell: should add to the errlist and return the errorlist, right?

pkg/sqlite/load.go
```diff
@@ -135,8 +135,11 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+		}
+		return tx.Commit()
```
ecordell: same here

pkg/sqlite/load.go
```diff
@@ -274,6 +277,100 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+	if err != nil {
+		return err
+	}
+	defer getChanHead.Close()
+
+	chanHeadRows, err := getChanHead.Query()
+	if err != nil {
+		return err
+	}
+	if !chanHeadRows.Next() {
```
ecordell: you need to call `chanHeadRows.Close()`

pkg/sqlite/load.go
```diff
@@ -274,6 +277,100 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+	bundles := make(map[string]struct{}, 0)
+	for chanBundleRows.Next() {
+		var bundleToUpdate sql.NullString
+		if err := chanBundleRows.Scan(&bundleToUpdate); err != nil {
+			return err
+		}
+		bundles[bundleToUpdate.String] = struct{}{}
+	}
+
+	if len(bundles) > 0 {
```
ecordell: this is fine, but I think we can do this loop with one sql query

pkg/sqlite/load.go
```diff
@@ -380,9 +487,473 @@ func (s *SQLLoader) addAPIs(tx *sql.Tx, csv *registry.ClusterServiceVersion, cha
...
+	var csvName string
+	csvNames := []string{}
+	for rows.Next() {
+		err := rows.Scan(&csvName)
+		if err != nil {
+			return nil, err
+		}
+		csvNames = append(csvNames, csvName)
+	}
+
```
ecordell: close `rows`

pkg/sqlite/load.go
```diff
@@ -274,6 +277,100 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+	defaultChannelRows, err := getDefChan.Query()
+	if err != nil {
+		return err
+	}
+	if !defaultChannelRows.Next() {
+		return fmt.Errorf("no default channel found for package %s", packageName)
+	}
+	var defaultChannel sql.NullString
+	if err := defaultChannelRows.Scan(&defaultChannel); err != nil {
```
ecordell: close defaultChannelRows

pkg/sqlite/load.go
```diff
@@ -274,6 +277,100 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+		return err
+	}
+	defer getChanHead.Close()
+
+	chanBundleRows, err := getChannelBundles.Query()
+	if err != nil {
+		return err
+	}
+	bundles := make(map[string]struct{}, 0)
+	for chanBundleRows.Next() {
```
ecordell: close rows

PR #90, opened 2019-09-27
Database migration initialization

kevinrizza:
This commit introduces a day 0 set of building blocks to implement a
database migration framework into the sqlite database. Adding the
following changes:

- Introduce the `golang-migrate` project as a dependency and vendor it
- Introduce a `db_migrations` folder with an initial no-op migration to
bootstrap the initial database version for newly created databases
- Update the initialization of the database when creating the initial
schema to hard set the database migration version on db creation
before loading any data

This commit does not introduce a method of actually running future
migrations against existing databases -- that is left for future
implementation based on this groundwork when a set of commands that
updates the database in place exists and requires the migration.

The intention of this commit is also to ensure that future initialized
registry databases have an initial version that can be used for
migrations in future releases.


pkg/sqlite/migrator.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// This function does not take any input and relies on the db defined in the SQLMigrator object
+// to connect to the database.
+func (m *SQLMigrator) InitMigrationVersion() error {
+	instance, err := sqlite3.WithInstance(m.db, &sqlite3.Config{})
+	if err != nil {
+		return err
+	}
+
+	migrator, err := migrate.NewWithDatabaseInstance(fmt.Sprintf("file://%s", m.migrationsPath), "registrydb", instance)
```
ecordell: Do we need the database name for anything anywhere else?

pkg/sqlite/migrator.go
```diff
@@ -0,0 +1,122 @@
...
+	"strings"
+
+	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database/sqlite3"
+	_ "github.com/golang-migrate/migrate/v4/source/file" // indirect import required by golang-migrate package
+)
+
+const (
+	// Hardcoded path where the db_migrations live
+	defaultMigrationsPath = "./pkg/sqlite/db_migrations"
```
ecordell: doesn't this mean that migrations only work if you have the `operator-registry` repo check out locally?

pkg/sqlite/migrator.go
kevinrizza: That name exists purely for logging purposes per the API documentation:

https://godoc.org/github.com/golang-migrate/migrate#NewWithDatabaseInstance

It gets attached to the migration object when performing updates internally since the only other identifier is the name of the db file and gets thrown away afterwards.

pkg/sqlite/migrator.go
kevinrizza: Yep, that is true. Regardless you at the very least need the migrations folder. For the initializer, this is currently only run today from inside a container image that has the relevant files copied into it. We could also add a flag to the initializer that takes the migration folder as input, but in either case those files will need to be pulled down.

We could also, in theory, save the files themselves as source code so they could be compiled as part of the binary and then written to a directory before being read. That would make this more portable if we found that significantly valuable.

ecordell: A quick google led me to this: https://github.com/gobuffalo/packr - if this looks reasonable to add to this PR quickly (say <1hr?) can you go ahead and do that? I think the UX is much better if we don't have to ship migration files around.

If it looks like it would take longer than that, we don't have to bother with it now, and we can revisit the idea in the future.

kevinrizza: @ecordell I don't really like this tool because it makes a bunch of assumptions about the underlying filesystem of your files and also requires that you wrap `go build` inside their `packr` command. I just messed around with https://github.com/go-bindata/go-bindata and it seems like it's a little bit better for our specific use case -- it just generates a `.go` file with compressed binary versions of the files. We still need to use a binary tool to generate them, but we should be able to just add something to the makefile to download and run it when we need to generate the files.

I'm going to update this PR in a little bit with a second commit with an implementation that includes that.

ecordell: Awesome, thanks for finding that!

cmd/configmap-server/main.go
```diff
@@ -98,7 +98,7 @@ func runCmdFunc(cmd *cobra.Command, args []string) error {
 		logger.Fatalf("error getting configmap: %s", err)
 	}
 
-	sqlLoader, err := sqlite.NewSQLLiteLoader(dbName)
+	sqlLoader, err := sqlite.NewSQLLiteLoader(dbName, "")
```
ecordell: We use some different patterns in other places that I think could clean this up. 

Either a builder with good defaults

```go
// uses the compiled migrations
sqlite.NewSQLLiteLoader(dbName)

// for tests
sqlite.NewSQLLiteLoader(dbName).WithMigrator(*testMigrator)
```
or the variadic option pattern
```go
sqlite.NewSQLLiteLoader(options.WithDBName(dbName), options.WithMigrator(default))
```

pkg/sqlite/load_test.go
```diff
@@ -131,7 +131,7 @@ func TestAddPackageChannels(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.description, func(t *testing.T) {
 			db := fmt.Sprintf("%d.db", rand.Int())
-			store, err := NewSQLLiteLoader(db)
+			store, err := NewSQLLiteLoader(db, "./db_migrations")
```
ecordell: We should probably be testing the compiled migrations, right?

kevinrizza: We do further down in the file:

```
migrator, err := NewSQLLiteMigrator(store.db, "")
```

But actually testing the values of the default migrations would involve either keeping the latest version in a configuration file or actually parsing the version. At that point, we would be maintaining test code that provides no functionality except to execute tests: IMO that is unnecessary overhead and technical debt.

That's why I implemented all the value checking using a set of test migrations that can be statically defined and their values referenced directly in the tests, and the actual check of the real data is just "does this throw an error when parsed".

Do you feel like there is significant value to be gained from ensuring something I'm missing?

cmd/configmap-server/main.go
kevinrizza: Yep, this is a good point. I went with the trivial implementation instead of using a builder because the constructor currently has actual parsing logic that requires the migrator to exist since there is currently no distinction between calling `NewSQLLiteLoader` and actually initializing the database -- I was attempting to retain that step as an atomic function call instead of splitting it into several steps.

But using the option pattern is a really good idea, I'll try to make this more digestible for the default case that way.

pkg/sqlite/load_test.go
kevinrizza: Ah, just realized this is for `load_test.go`, not the new test file I added. Yeah in this case you are right this should be updated to just use the correct default path.

kevinrizza:
@ecordell Updated. ptal

Makefile
```diff
@@ -3,12 +3,12 @@ CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | xargs -I{} b
...
 .PHONY: build test vendor clean
 
-all: clean test build
+all: clean install-go-bindata test build
 
 $(CMDS):
 	go build $(MOD_FLAGS) $(extra_flags) -o $@ ./cmd/$(shell basename $@)
 
-build: clean $(CMDS)
+build: clean generate-migration-bundle $(CMDS)
```
ecordell: I don't think we want this as part of the `build` command so that CI will catch when we don't correctly commit the generated migration bundle to git.

Makefile
```diff
@@ -37,4 +43,4 @@ container-codegen:
 
 clean:
 	@rm -rf ./bin
-
+	@rm -f ./pkg/sqlite/migrations.go
```
ecordell: like the rest of our codegen, we probably want to go ahead and commit these artifacts.

go.mod
```diff
@@ -3,25 +3,28 @@ module github.com/operator-framework/operator-registry
...
 	github.com/ghodss/yaml v1.0.0
+	github.com/go-bindata/go-bindata v3.1.2+incompatible // indirect
+	github.com/golang-migrate/migrate/v4 v4.6.2
 	github.com/golang/mock v1.2.0
 	github.com/golang/protobuf v1.3.1
 	github.com/googleapis/gnostic v0.2.0 // indirect
 	github.com/grpc-ecosystem/grpc-health-probe v0.2.1-0.20181220223928-2bf0a5b182db
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/mattn/go-sqlite3 v1.10.0
+	github.com/operator-framework/operator-lifecycle-manager v3.11.0+incompatible // indirect
```
ecordell: Where did this come from? We recently did some work to remove the dependency on OLM. 

(also this is picking up the 3.11 version which is like 0.7.0 I think)

pkg/sqlite/db_options.go
```diff
@@ -0,0 +1,23 @@
+package sqlite
+
+type DbOptions struct {
```
ecordell: 👍

ecordell:
Discussed offline:

We'll stick with the current implementation and revisit in the future if/when we need migrations that require more than SQL.

Just need those dependencies fixed up and we can merge @kevinrizza !

pkg/sqlite/load.go
```diff
@@ -20,7 +20,7 @@ type SQLLoader struct {
 
 var _ registry.Load = &SQLLoader{}
 
-func NewSQLLiteLoader(outFilename string) (*SQLLoader, error) {
+func NewSQLLiteLoader(outFilename, migrationsPath string) (*SQLLoader, error) {
```
ecordell: Did I hallucinate a change to this API? I thought this became variadic with config options?

kevinrizza: you did not, i rebased poorly :) latest should contain the right bits.

PR #92, opened 2019-10-01
(feat) Load operator registry bundle from ConfigMap

tkashem:
- Iterate through the Data section of a ConfigMap, process each
  resource and construct a registry.Bundle object for use.
- The ConfigMap may have a resource that contains a `PackageManifest`
  resource. Process this resource if present and deserialize.


jpeeler:
/hold cancel
Removing the hold as the added commit made by me verifies that this code will work correctly and no further changes are known at present. (Though it might be nice if #94 is merged first just in case.)

ecordell:
this will need to change slightly when this merges: https://github.com/openshift/enhancements/pull/70

jpeeler:
I forgot this depends on constants in https://github.com/operator-framework/operator-registry/pull/94.

jpeeler:
Removed the dependency on #94. Also, decided to keep the latest commit separate since these commits evolved over time with different authors instead of squashing it all together.

PR #94, opened 2019-10-01
Expose bundle data from bundle image

jpeeler:
This provides a command to expose data in a bundle image and parses both /manifests and /metadata as proposed in https://github.com/operator-framework/operator-lifecycle-manager/pull/1054.

~This might need some shuffling around as I'm not sure the sqlite package is the best place for this code to live. Also, currently it includes another WIP PR which is likely to change.~

This includes functionality for launching the container to execute the new serve command, to be called from OLM once this and other PRs are pulled in.

OLM-1275

### DIFF
--- a/pkg/client/kubeclient.go
+++ b/pkg/client/kubeclient.go
@@ -1,0 +1,35 @@
+package client
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func NewKubeClient(kubeconfig string, logger *logrus.Logger) (clientset *kubernetes.Clientset, err error) {
+	var config *rest.Config
+
+	if overrideConfig := os.Getenv(clientcmd.RecommendedConfigPathEnvVar); overrideConfig != "" {
+		kubeconfig = overrideConfig
+	}
+
+	if kubeconfig != "" {
+		logger.Infof("Loading kube client config from path %q", kubeconfig)
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
+	} else {
+		logger.Infof("Using in-cluster kube client config")
+		config, err = rest.InClusterConfig()
+	}
+
+	if err != nil {
+		err = fmt.Errorf("Cannot load config for REST client: %v", err)
+		return
+	}
+
+	clientset, err = kubernetes.NewForConfig(config)
+	return
+}
--- a/pkg/configmap/configmap_writer.go
+++ b/pkg/configmap/configmap_writer.go
@@ -1,0 +1,222 @@
+package configmap
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"regexp"
+
+	"github.com/ghodss/yaml"
+	_ "github.com/mattn/go-sqlite3"
+	"github.com/sirupsen/logrus"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/operator-framework/operator-registry/pkg/client"
+	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
+)
+
+// configmap keys can contain underscores, but configmap names can not
+var unallowedKeyChars = regexp.MustCompile("[^-A-Za-z0-9_.]")
+
+const (
+	EnvContainerImage           = "CONTAINER_IMAGE"
+	ConfigMapImageAnnotationKey = "olm.sourceImage"
+)
+
+type AnnotationsFile struct {
+	Annotations struct {
+		Resources      string `json:"operators.operatorframework.io.bundle.manifests.v1"`
+		MediaType      string `json:"operators.operatorframework.io.bundle.mediatype.v1"`
+		Metadata       string `json:"operators.operatorframework.io.bundle.metadata.v1"`
+		Package        string `json:"operators.operatorframework.io.bundle.package.v1"`
+		Channels       string `json:"operators.operatorframework.io.bundle.channels.v1"`
+		ChannelDefault string `json:"operators.operatorframework.io.bundle.channel.default.v1"`
+	} `json:"annotations"`
+}
+
+type ConfigMapWriter struct {
+	manifestsDir  string
+	configMapName string
+	namespace     string
+	clientset     *kubernetes.Clientset
+}
+
+func NewConfigMapLoaderForDirectory(configMapName, namespace, manifestsDir, kubeconfig string) *ConfigMapWriter {
+	clientset, err := client.NewKubeClient(kubeconfig, logrus.StandardLogger())
+	if err != nil {
+		logrus.Fatalf("cluster config failed: %v", err)
+	}
+
+	return &ConfigMapWriter{
+		manifestsDir:  manifestsDir,
+		configMapName: configMapName,
+		namespace:     namespace,
+		clientset:     clientset,
+	}
+}
+
+func TranslateInvalidChars(input string) string {
+	validConfigMapKey := unallowedKeyChars.ReplaceAllString(input, "~")
+	return validConfigMapKey
+}
+
+func (c *ConfigMapWriter) Populate(maxDataSizeLimit uint64) error {
+	subDirs := []string{"manifests/", "metadata/"}
+
+	configMapPopulate, err := c.clientset.CoreV1().ConfigMaps(c.namespace).Get(c.configMapName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	configMapPopulate.Data = map[string]string{}
+
+	var totalSize uint64
+	for _, dir := range subDirs {
+		completePath := c.manifestsDir + dir
+		files, err := ioutil.ReadDir(completePath)
+		if err != nil {
+			logrus.Errorf("read dir failed: %v", err)
+			return err
+		}
+
+		for _, file := range files {
+			log := logrus.WithField("file", completePath+file.Name())
+			log.Info("Reading file")
+			content, err := ioutil.ReadFile(completePath + file.Name())
+			if err != nil {
+				log.Errorf("read failed: %v", err)
+				return err
+			}
+			totalSize += uint64(len(content))
+			if totalSize > maxDataSizeLimit {
+				log.Errorf("File with size %v exceeded %v limit, aboring", len(content), maxDataSizeLimit)
+				return fmt.Errorf("file %v bigger than total allowed limit", file.Name())
+			}
+
+			validConfigMapKey := TranslateInvalidChars(file.Name())
+			if validConfigMapKey != file.Name() {
+				logrus.WithFields(logrus.Fields{
+					"file.Name":         file.Name(),
+					"validConfigMapKey": validConfigMapKey,
+				}).Info("translated filename for configmap comptability")
+			}
+			if file.Name() == bundle.AnnotationsFile {
+				var annotationsFile AnnotationsFile
+				err := yaml.Unmarshal(content, &annotationsFile)
+				if err != nil {
+					return err
+				}
+				configMapPopulate.SetAnnotations(map[string]string{
+					bundle.ManifestsLabel:      annotationsFile.Annotations.Resources,
+					bundle.MediatypeLabel:      annotationsFile.Annotations.MediaType,
+					bundle.MetadataLabel:       annotationsFile.Annotations.Metadata,
+					bundle.PackageLabel:        annotationsFile.Annotations.Package,
+					bundle.ChannelsLabel:       annotationsFile.Annotations.Channels,
+					bundle.ChannelDefaultLabel: annotationsFile.Annotations.ChannelDefault,
+				})
+			} else {
+				configMapPopulate.Data[validConfigMapKey] = string(content)
+			}
+		}
+	}
+
+	if sourceImage := os.Getenv(EnvContainerImage); sourceImage != "" {
+		annotations := configMapPopulate.GetAnnotations()
+		annotations[ConfigMapImageAnnotationKey] = sourceImage
+	}
+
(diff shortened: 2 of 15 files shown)


jpeeler:
This update includes the code for launching a container to execute the new serve command and is feature complete, minus the tests.

jpeeler:
I just need to plumb the e2e test (potentially) into the makefile and CI.

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/operator-framework/operator-registry/pkg/client"
+)
+
+// configmap keys can contain underscores, but configmap names can not
+var unallowedKeyChars = regexp.MustCompile("[^-A-Za-z0-9_.]")
```
awgreene: Rather than whitelisting valid characters wouldn't it be better to check for valid k8s object names?

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+	return &ConfigMapWriter{
+		manifestsDir:  manifestsDir,
+		configMapName: configMapName,
+		namespace:     namespace,
+		clientset:     clientset,
+	}
+}
+
+func TranslateInvalidChars(input string) string {
+	validConfigMapKey := unallowedKeyChars.ReplaceAllString(input, "~")
```
awgreene: It is not apparent to me why you are providing `~` as an input (even after googling). Could you explain?

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+
+	configMapPopulate, err := c.clientset.CoreV1().ConfigMaps(c.namespace).Get(c.configMapName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	configMapPopulate.Data = map[string]string{}
+
+	var totalSize uint64
+	for _, dir := range subDirs {
+		completePath := c.manifestsDir + dir
```
awgreene: Should you be using [path.Join](https://golang.org/pkg/path/#Join) here?

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+		files, err := ioutil.ReadDir(completePath)
+		if err != nil {
+			return err
```
awgreene: Is there a reason you elected not to log an error here? A similar read was done on line 85  which included a log:
```
if err != nil {
	log.Errorf("read failed: %v", err)
	return err
}
```

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+			content, err := ioutil.ReadFile(completePath + file.Name())
+			if err != nil {
+				log.Errorf("read failed: %v", err)
+				return err
+			}
+			totalSize += uint64(len(content))
+			if totalSize > (1 << 20) {
+				// over 1 MiB limit, bail out
+				log.Errorf("File with size %v exceeded 1 MiB limit, aboring", len(content))
+				return fmt.Errorf("file %v bigger than total allowed limit", file.Name())
```
awgreene: :+1:

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+
+			validConfigMapKey := TranslateInvalidChars(file.Name())
+			if validConfigMapKey != file.Name() {
+				logrus.Warningf("translated filename from '%v' to '%v' for configmap compatability", file.Name(), validConfigMapKey)
```
awgreene: Did you mean to use `logrus` here? Perhaps you wanted to avoid printing the field you set on `log` earlier?

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			if file.Name() == "annotations.yaml" {
```
awgreene: Should `annotations.yaml` be a `const`?

pkg/sqlite/configmap_writer.go
awgreene: For example, `a--` would get through this. I understand that this is being used to fix invalid configmap names, but I'm not sure if that's something the tool should be responsible for.

pkg/sqlite/configmap_writer.go
awgreene: Do we want to explain to users why there is a limit?

pkg/sqlite/configmap_writer.go
jpeeler: There was some discussion as to whether or not the code here should be responsible for validating "valid" file names ("valid" because technically it's a configmap implementation detail leaking into what's considered valid from a user perspective). However, given how short this code is, I personally like ensuring that things don't explode. And since the resource names aren't in play here, I believe the chosen regex is best.

pkg/sqlite/configmap_writer.go
jpeeler: This ties into the above comment. If you want to argue that failing instead of translating to valid is best, then this code is up for potential removal. But I randomly chose the tilda character as a character that is typically rarely used to replace any invalid characters. In practice, I hope the string is not modified, but did make sure to include a warning message to point out the potential issue.

pkg/sqlite/configmap_writer.go
jpeeler: Definitely.

pkg/sqlite/configmap_writer.go
jpeeler: Not really, probably good to have consistency.

pkg/sqlite/configmap_writer.go
jpeeler: I thought about that, but then I saw that I wasn't 100% sure of the current limit. It seems in the past the etcd storage limit was 1 MB, but now it might be 1.5 MB? https://github.com/etcd-io/etcd/blob/master/Documentation/dev-guide/limit.md#request-size-limit. Personally this was an addition made at the last minute to handle what I consider to be a very unlikely scenario. If we're worried about file size limit I'm not sure the current design is good with (currently just one) configmap backing.

pkg/sqlite/configmap_writer.go
jpeeler: Definitely.

pkg/client/kubeclient.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func NewKubeClient(kubeconfig string, logger *logrus.Logger) (clientset *kubernetes.Clientset, err error) {
+	var config *rest.Config
+
+	if overrideConfig := os.Getenv("KUBECONFIG"); overrideConfig != "" {
```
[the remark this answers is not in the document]
jpeeler: Good idea. And amazingly the constant is already defined from an already imported package.

pkg/sqlite/configmap_writer.go
jpeeler: Yeah, it seemed to be a little redundant.

pkg/sqlite/configmap_writer.go
ecordell: Maybe this should be an input on the tool, that defaults to 1mb? Then ask kube changes we can configure it differently? 

(this is also technically a per-cluster limit and can be configured in etcd, and things like k3s don't use etcd could potentially have a different limit - but most clusters probably use the same defaults)

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,214 @@
...
+			totalSize += uint64(len(content))
+			if totalSize > (1 << 20) {
+				// over 1 MiB limit, bail out
+				log.Errorf("File with size %v exceeded 1 MiB limit, aboring", len(content))
+				return fmt.Errorf("file %v bigger than total allowed limit", file.Name())
+			}
+
+			validConfigMapKey := TranslateInvalidChars(file.Name())
+			if validConfigMapKey != file.Name() {
+				logrus.Warningf("translated filename from '%v' to '%v' for configmap compatability", file.Name(), validConfigMapKey)
```
ecordell: Personal preference here: I think it's easier to read this type of log with `Fields`

cmd/bundle-extractor/main.go
```diff
@@ -0,0 +1,70 @@
...
+	configmapName, err := cmd.Flags().GetString("configmapname")
+	if err != nil {
+		return err
+	}
+	namespace, err := cmd.Flags().GetString("namespace")
+	if err != nil {
+		return err
+	}
+
+	loader := sqlite.NewConfigMapLoaderForDirectory(configmapName, namespace, manifestsDir, kubeconfig)
```
ecordell: could we find a better place to put this than in the `sqlite` package?

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type ConfigMapWriter struct {
+	manifestsDir  string
+	configMapName string
+	namespace     string
+	clientset     *kubernetes.Clientset
+}
+
+var _ SQLPopulator = &ConfigMapWriter{}
```
ecordell: It's a little weird to have this object that populates configmaps conform to the interface that populates sql databases?

pkg/sqlite/configmap_writer.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// LaunchBundleImage will launch a bundle image and also create a configmap for
+// storing the data that will be updated to contain the bundle image data. It is
+// the responsibility of the caller to delete both the job and configmap when done.
+func LaunchBundleImage(kubeclient kubernetes.Interface, bundleImage, initImage, namespace string) (*corev1.ConfigMap, *batchv1.Job, error) {
```
ecordell: The intention is that this is called from OLM, correct?

cmd/bundle-extractor/main.go
jpeeler: I figured this would be brought up. But I didn't know where else to put it. Suggestions?

pkg/sqlite/configmap_writer.go
jpeeler: Yeah - would it be better to create another interface or just drop it altogether?

pkg/sqlite/configmap_writer.go
jpeeler: Yes!

jpeeler: Changed the comment to be more clear about the intention.

pkg/sqlite/configmap_writer.go
jpeeler: Modified as suggested, new parameter added with a default of 1 MB.

cmd/bundle-extractor/main.go
jpeeler: Moved to new configmap package.

cmd/bundle-extractor/main.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/pkg/configmap"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
```
ecordell: now that the `opm` command has merged, should this be `opm bundle extract`?

test/e2e/image-bundle/metadata/annotations.yaml
```diff
@@ -0,0 +1,3 @@
+annotations:
+  operators.coreos.com.bundle.resources: "manifests+metadata"
```
ecordell: This is the only change I think we need. Please see my comment on Vu's PR: https://github.com/operator-framework/operator-registry/pull/97#discussion_r336463215

This should be a minor change, otherwise everything looks great!

Before I lgtm, do you want to hold this for further testing on the OLM side? Or would you rather merge and make new PRs if something comes up?

cmd/bundle-extractor/main.go
jpeeler: Updated!

test/e2e/image-bundle/metadata/annotations.yaml
jpeeler: I think it's easier to go ahead and get in. Hopefully any follow ups are pretty minor.

cmd/bundle-extractor/main.go
jpeeler: This has been re-updated to be `opm alpha bundle extract`.

pkg/configmap/configmap_writer.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+type ConfigMapWriter struct {
+	manifestsDir  string
+	configMapName string
+	namespace     string
+	clientset     *kubernetes.Clientset
+}
+
+func NewConfigMapLoaderForDirectory(configMapName, namespace, manifestsDir, kubeconfig string) *ConfigMapWriter {
```
ecordell: nit: `NewConfigMapLoaderForDirectory`

nit: might want to use the functional option pattern for arguments we've been using elsewhere

pkg/configmap/configmap_writer.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return nil
+}
+
+// LaunchBundleImage will launch a bundle image and also create a configmap for
+// storing the data that will be updated to contain the bundle image data. It is
+// the responsibility of the caller to delete the job, the pod, and the configmap
+// when done. This function is intended to be called from OLM, but is put here
+// for locality.
+func LaunchBundleImage(kubeclient kubernetes.Interface, bundleImage, initImage, namespace string) (*corev1.ConfigMap, *batchv1.Job, error) {
```
ecordell: nit: should be able to set the path to find `opm` in `initImage`?

test/e2e/bundle_image_test.go
```diff
@@ -0,0 +1,773 @@
+package e2e_test
```
ecordell: 👍

PR #96, opened 2019-10-10
Adding opm CLI

gallettilance:

**Reviewer Checklist**


Makefile
```diff
@@ -1,5 +1,6 @@
+
 MOD_FLAGS := $(shell (go version | grep -q -E "1\.(11|12)") && echo -mod=vendor)
-CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | xargs -I{} basename {}))
+CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | grep -Fv "github.com/operator-framework/operator-registry/cmd/cli" |xargs -I{} basename {}))
```
njhale: If we keep the command declarations in separate top level packages, we don't need to do this right?

cmd/opm/main.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	cli "github.com/operator-framework/operator-registry/cmd/cli"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "opm",
```
njhale: 😄

pkg/sqlite/load.go
```diff
@@ -77,6 +77,9 @@ func NewSQLLiteLoader(outFilename string) (*SQLLoader, error) {
 	if _, err = db.Exec(createTable); err != nil {
 		return nil, err
 	}
+	if _, err := db.Exec("PRAGMA foreign_keys = ON", nil); err != nil {
```
njhale: So the foreign keys in the definition before weren't active? I thought we've been relying on foreign keys to help validate relationships on insert?

ecordell:
@gallettilance I made a separate PR to address the foreign key issues; can you remove it from this one so that tests are passing?

cmd/cli/opmregistry.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+	}
+
+	rootCmd.AddCommand(newOpmRegistryServeCmd())
+
+	return rootCmd
+}
+
+func newOpmRegistryServeCmd() *cobra.Command {
+	rootCmd := &cobra.Command{
```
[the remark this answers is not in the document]
gallettilance: Right now it is a duplicate but it may not be in the future and if the goal is to eventually move to only using opm, I would prefer to keep these separate. It also provides an example for those trying to add new sub-commands

Makefile
gallettilance: You mean put this in `pkg` instead of `cmd`?

cmd/cli/opmregistry.go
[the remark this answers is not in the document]
njhale: That's fair.

Makefile
njhale: I mean `cmd/initializer`, `cmd/registry`, `cmd/opm`, and forgo extracting the commands out to `cmd/cli` -- but it seems like everyone has a different opinion on this, so feel free to ignore.

Makefile
```diff
@@ -1,5 +1,5 @@
 MOD_FLAGS := $(shell (go version | grep -q -E "1\.(11|12)") && echo -mod=vendor)
-CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | xargs -I{} basename {}))
+CMDS  := $(addprefix bin/, $(shell go list $(MOD_FLAGS) ./cmd/... | grep -v opm/... | xargs -I{} basename {}))
```
kevinrizza: Do we want to continue to use `go list` like this? Maybe we should just explicitly list the things we build rather than rely on using this command and ignoring part of the result? Or at least should we have a heuristic for determining it (like only look at the packages in the root of `./cmd/`?

@dinhxuanvu Any thoughts on this? I know you most likely have a similar problem with the bundle build command you are adding.

cmd/opm/main.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/spf13/cobra"
+
+	cli "github.com/operator-framework/operator-registry/cmd/opm/registry"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "opm",
+		Short: "operator package manager",
+		Long:  "Top level CLI for operator-registry",
```
kevinrizza: Is this long description accurate anymore?

cmd/opm/registry/cmd.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+// NewOpmRegistryCmd returns the appregistry-server command
+func NewOpmRegistryCmd() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "registry",
+		Short: "registry",
```
kevinrizza: I think you want `Short` to have a slightly longer description that defines what it actually does. Otherwise `--help` won't really tell you anything.

cmd/opm/registry/serve.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/operator-framework/operator-registry/pkg/lib/log"
+	"github.com/operator-framework/operator-registry/pkg/registry"
+	"github.com/operator-framework/operator-registry/pkg/server"
+	"github.com/operator-framework/operator-registry/pkg/sqlite"
+)
+
+func newRegistryServeCmd() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "serve",
+		Short: "serve",
```
kevinrizza: Same comment as above about `Short`

gallettilance: addressed

cmd/opm/registry/cmd.go
gallettilance: addressed

cmd/opm/main.go
gallettilance: addressed

Makefile
dinhxuanvu: So the plan for me is to separate the functional code of my bundle PR to have it as a package living under `/pkg` and the cobra command code (the rest) will live under `/opm/bundle/...` and it will consume the bundle package.

ecordell: Why do we want to omit `opm` from the build?

kevinrizza: @dinhxuanvu I thought we wanted to expose the bundle build as a command that wasn't built as part of any of our existing utilities? and then we create another pr against operator-sdk that consumed that cobra thing? We can create our own test utility that can also use that cobra thing but we won't publish it, just use it to test. but I was expecting it to NOT be part of `opm`.

@ecordell I think he just meant that he doesn't want the subcommand packages to be part of the list of things that should be called by `go build`. Those packages don't have `main.go` files so the command that's in the makefile now will throw an error.

ecordell: I think this will do what we want:

```
CMDS  := $(addprefix bin/, $(shell ls ./cmd))
```

dinhxuanvu: @kevinrizza You are correct. The ultimate usage for bundle library would be on SDK side. The users won't use bundle commands as a part of `opm` binary. I'm planning to add bundle commands with `hidden` flag on cobra. So we know they are there and available for internal use but they won't show up in the list of available commands.

PR #97, opened 2019-10-11
Add bundle library to build bundle manifest image

dinhxuanvu:
1. Add top level CLI command `opm`
2. Add bundle command to generate annotations.yaml and Dockerfile
3. Add bundle command to build Bundle image
* The build command will generate annotations.yaml and Dockerfile
as well

**Reviewer Checklist**


kevinrizza:
@dinhxuanvu won't this break the makefile because you are nesting an additional package into the `cmd` folder that doesn't compile directly? I think `make build` relies on `go list` to get its list to compile.

dinhxuanvu:
@kevinrizza This is just a cherry-pick PR from original PR in OLM repo. I'm working on separating the functional code into an actual package under `pkg` and cli/cobra code remaining in `/cmd` for internal use. Stay tuned.

docs/design/operator-bundle.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Bundle Dockerfile
+
+This is an example of a `Dockerfile` for operator bundle:
+```
+FROM scratch
+
+# We are pushing an operator-registry bundle
+# that has both metadata and manifests.
+LABEL operators.operatorframework.io.bundle.resources=manifests+metadata
```
ecordell: We had some discussion but merged before resolving on the bundle enhancement:

https://github.com/openshift/enhancements/pull/27#discussion_r330779064

Basically, can we make the values here be the locations in the bundle (and probably versioned as well)?

```
LABEL operators.operatorframework.io.bundle.resources.manifests.v1=/manifests
LABEL operators.operatorframework.io.bundle.resources.metadata.v1=/metadata
```
This indicates that the `metadata` resources can be found in `/manifests` in the image, etc.

ecordell:
Ah, one issue to resolve for the tests:
```
pkg/lib/bundle/build_test.go:55:27: cmd.String undefined (type *exec.Cmd has no field or method String) 
```

cmd/opm/bundle/generate.go
```diff
@@ -0,0 +1,60 @@
+package bundle
+
+import (
+	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+type AnnotationMetadata struct {
```
kevinrizza: @gallettilance

kevinrizza: @dinhxuanvu Instead of putting this type in the cmd package, can we expose it in one of the packages in `pkg/` so that it can be imported in multiple places? The database function that parses these bundles will need to use this type to read the annotations in the other direction.

pkg/lib/bundle/build.go
```diff
@@ -0,0 +1,62 @@
...
+	log.Debugf("Running %#v", cmd.Args)
+
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("Failed to exec %#v: %v", cmd.Args, err)
+	}
+
+	return nil
+}
+
+func BuildFunc(directory, imageTag, imageBuilder, packageName, channels, channelDefault string, overwrite bool) error {
```
kevinrizza: I assume this is the function that the SDK will reference when it imports this package? If so, can we add a comment to this function that defines each parameter and its purpose? I know they are defined in our hidden opm command, but it would be useful to reiterate them here so that IDEs can see the doc text

cmd/opm/bundle/build.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	bundleBuildCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "", "The list of channels that bundle image belongs to")
+	if err := bundleBuildCmd.MarkFlagRequired("channels"); err != nil {
+		log.Fatalf("Failed to mark `channels` flag for `build` subcommand as required")
+	}
+
+	bundleBuildCmd.Flags().StringVarP(&imageBuilderArgs, "image-builder", "b", "docker", "Tool to build container images. One of: [docker, podman, buildah]")
+
+	bundleBuildCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "", "The default channel for the bundle image")
```
kevinrizza: maybe `defaultChannel` is a better name here?

cmd/opm/bundle/build.go
```diff
@@ -0,0 +1,81 @@
...
+        `,
+		RunE: buildFunc,
+	}
+
+	bundleBuildCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "", "The directory where bundle manifests are located")
+	if err := bundleBuildCmd.MarkFlagRequired("directory"); err != nil {
+		log.Fatalf("Failed to mark `directory` flag for `build` subcommand as required")
+	}
+
+	bundleBuildCmd.Flags().StringVarP(&tagBuildArgs, "tag", "t", "", "The name of the bundle image will be built")
```
kevinrizza: nit: help message should be more like: `The image tag applied to the bundle image`

cmd/opm/bundle/build.go
```diff
@@ -0,0 +1,81 @@
...
+	bundleBuildCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "", "The list of channels that bundle image belongs to")
+	if err := bundleBuildCmd.MarkFlagRequired("channels"); err != nil {
+		log.Fatalf("Failed to mark `channels` flag for `build` subcommand as required")
+	}
+
+	bundleBuildCmd.Flags().StringVarP(&imageBuilderArgs, "image-builder", "b", "docker", "Tool to build container images. One of: [docker, podman, buildah]")
+
+	bundleBuildCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "", "The default channel for the bundle image")
+
+	bundleBuildCmd.Flags().BoolVarP(&overwriteArgs, "overwrite", "o", false, "To overwrite annotations.yaml if existing")
```
kevinrizza: Can we define what happens if this flag is not set but you already have an `annotations.yaml` file in this helper text? If we are explicit it may help users to understand that we are writing that file to their local disk as well.

pkg/lib/bundle/build.go
kevinrizza: Or are they actually just importing our cobra function?

kevinrizza:
Sorry one more request @dinhxuanvu , can you squash your commits? There's a lot of old history that isn't really relevant anymore.

cmd/opm/bundle/generate.go
dinhxuanvu: @kevinrizza The type has been moved to bundle package already but I forgot to remove it here.

pkg/lib/bundle/build.go
dinhxuanvu: SDK will not import our cobra code as both SDK and `opm` will use bundle library with cobra command code wrapped around it. For that reason, the cobra code for SDK and `opm` will be basically identical.

cmd/opm/bundle/build.go
dinhxuanvu: The `defaultChannel` is a fine name but the reason I use `default` here is to keep the command flag to be as short as possible. Long name only makes the command even longer and difficult to read. I try to keep all flags to be single word with the exception of `image-builder` because SDK has the a flag with that name with the same functionality. So I decided to keep it the same as SDK to keep the consistency given this command will be added to SDK.

dinhxuanvu:
Ready for another round of review.

pkg/lib/bundle/build.go
kevinrizza: Cool. I think this is a good idea, and I think we should start using this `lib` pattern everywhere in `opm`

kevinrizza:
lgtm

cmd/opm/alpha/bundle/build.go
```diff
@@ -0,0 +1,81 @@
...
+
+        For example: The command will generate annotations.yaml metadata plus
+        Dockerfile for bundle image and then build a container image from
+        provided operator bundle manifests generated metadata
+        e.g. "quay.io/example/operator:v0.0.1".
+
+        After the build process is completed, a container image would be built
+        locally in docker and available to push to a container registry.
+
+        $ opm bundle build --directory /test/ --tag quay.io/example/operator:v0.1.0 \
```
gallettilance: I guess this should be `opm alpha bundle ...` now?

docs/design/operator-bundle.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+*Notes:*
+* In case of a mismatch, the `annotations.yaml` file is authoritative because the on-cluster operator-registry that relies on these annotations has access to the yaml file only.
+* The potential use case for the `LABELS` is - an external off-cluster tool can inspect the image to check the type of a given bundle image without downloading the content.
+
+This example uses [Operator Registry Manifests](https://github.com/operator-framework/operator-registry#manifest-format) format to build an operator bundle image. The source directory of an operator registry bundle has the following layout.
+```
+$ tree test
+test
```
gallettilance: @kevinrizza can correct me if I'm wrong but isn't the format:
```
test
└── manifests
    ├── testbackup.crd.yaml
    ├── testcluster.crd.yaml
    ├── testoperator.v0.1.0.clusterserviceversion.yaml
    ├── testrestore.crd.yaml
└── metadata
    └── annotations.yaml
```

docs/design/operator-bundle.md
```diff
@@ -0,0 +1,199 @@
...
+# that has both metadata and manifests.
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=test-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=beta,stable
+LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
+
+ADD test/*.yaml /manifests
+ADD test/metadata/annotations.yaml /metadata/annotations.yaml
```
gallettilance: if above comment holds, it's more convenient if we don't have the manifests and metadata in different layers (which is what the separate ADD commands will do afaik). So we could change the above to

```
ADD test /
```

docs/design/operator-bundle.md
dinhxuanvu: This section is for the directory format where users have their manifests *locally*. Below in line 79, you will see the directory format in the bundle image filesystem.

docs/design/operator-bundle.md
dinhxuanvu: Give this is a CLI tool that is controlled by us, we want a strict build process that only adds yaml manifests + metadata (`annotations.yaml`) that we either generate or validate. We want to avoid adding everything in the directory which may include unwanted garbage such as non-consumable files or binaries by OLM. One of the side goals of this enhancement is to keep the image to be as small as possible. If adding everything in the directory, we potentially add a ton of stuff that users may leave there and inflate the image size.

gallettilance: I agree with building the manifests and metadata to avoid unwanted garbage. Can we add these folders to the same layer? Is there a reason we want separate layers for each folder?

dinhxuanvu: Give the ADD can't be chained, not sure how I can add both in the same layer. They are going in two separate directories as well.

gallettilance: That's a good point about chaining not being supported. Ok I guess we have to do it this way.

gallettilance:
lgtm

pkg/lib/bundle/generate.go
```diff
@@ -0,0 +1,267 @@
...
+)
+
+const (
+	defaultPermission   = 0644
+	registryV1Type      = "registry+v1"
+	plainType           = "plain"
+	helmType            = "helm"
+	annotationsFile     = "annotations.yaml"
+	dockerFile          = "Dockerfile"
+	manifestsDir        = "manifests/"
```
jpeeler: Can we make this consts here and below upper case so they can be used in other packages?

docs/design/operator-bundle.md
jpeeler: Really should be using COPY here instead, but that can be changed later if anybody else agrees. Is there anything aside from efficiency that is a concern here for how many layers are used?

PR #98, opened 2019-10-12
fix(sqlite): enable foreign keys for sqlite

ecordell:
Foreign key constraints were not being enforced - thanks @gallettilance for finding this issue!

There were a couple of problems with the foreign keys that are fixed in this PR

We want to rely on sqlite's enforcement of foreign keys so that we don't need to enforce them ourselves.

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -35,7 +35,7 @@ func NewSQLLiteLoader(outFilename string) (*SQLLoader, error) {
 	CREATE TABLE IF NOT EXISTS package (
 		name TEXT PRIMARY KEY,
 		default_channel TEXT,
-		FOREIGN KEY(default_channel) REFERENCES channel(name)
+		FOREIGN KEY(name, default_channel) REFERENCES channel(package_name,name)
 	);
 	CREATE TABLE IF NOT EXISTS channel (
 		name TEXT, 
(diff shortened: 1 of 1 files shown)


pkg/sqlite/load.go
```diff
@@ -35,7 +35,7 @@ func NewSQLLiteLoader(outFilename string) (*SQLLoader, error) {
 	CREATE TABLE IF NOT EXISTS package (
 		name TEXT PRIMARY KEY,
 		default_channel TEXT,
-		FOREIGN KEY(default_channel) REFERENCES channel(name)
+		FOREIGN KEY(name, default_channel) REFERENCES channel(package_name,name)
```
ecordell: referencing a field that's part of a compound primary key requires referencing the full primary key - this was the main issue.

pkg/sqlite/load.go
```diff
@@ -53,9 +53,7 @@ func NewSQLLiteLoader(outFilename string) (*SQLLoader, error) {
 		replaces INTEGER,
 		depth INTEGER,
 		FOREIGN KEY(replaces) REFERENCES channel_entry(entry_id)  DEFERRABLE INITIALLY DEFERRED, 
-		FOREIGN KEY(channel_name) REFERENCES channel(name),
-		FOREIGN KEY(package_name) REFERENCES channel(package_name),
-		FOREIGN KEY(operatorbundle_name) REFERENCES operatorbundle(name)
```
ecordell: Had to remove the foreign key to operator_bundle, since synthesized channel entries for `skip`ed releases may not have a corresponding bundle.

PR #99, opened 2019-10-15
Image digests for referencing bundles

anik120:

- Change schema of OperatorBundle table to include field 'bundlepath'.
- Extend API to allow querying of the new `bundlepath` field.
- Modify API to return empty struct instead of an error while querying empty bundle 

**Reviewer Checklist**

### DIFF
--- a/pkg/api/conversion.go
+++ b/pkg/api/conversion.go
@@ -82,3 +82,10 @@ func BundleStringToAPIBundle(bundleString string, entry *registry.ChannelEntry) 
 	out.PackageName = entry.PackageName
 	return out, nil
 }
+
+func BundlePathStringToAPIBundlePath(bundlePath string) *BundlePath {
+	out := &BundlePath{
+		Path: bundlePath,
+	}
+	return out
+}
--- a/pkg/api/registry.pb.go
+++ b/pkg/api/registry.pb.go
@@ -3,13 +3,14 @@
 
 package api
 
-import proto "github.com/golang/protobuf/proto"
-import fmt "fmt"
-import math "math"
-
 import (
-	context "golang.org/x/net/context"
+	context "context"
+	fmt "fmt"
+	proto "github.com/golang/protobuf/proto"
 	grpc "google.golang.org/grpc"
+	codes "google.golang.org/grpc/codes"
+	status "google.golang.org/grpc/status"
+	math "math"
 )
 
 // Reference imports to suppress errors if they are not otherwise used.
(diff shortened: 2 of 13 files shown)


ecordell:
@anik120 I like this approach and think it will give us a good transition path on the client side.

Couple of things to think about:
 
- Do we want to use this schema change to test our new migration process? If so, we should wait for #90 to merge, and add a migration here
- Tests 😄

anik120:
@ecordell thank you for your feedback.  

- I've looked at #90 and looks like we still don't need to migrate our database as we're at ground zero with storing image-digests. Once we have long lived clusters with image digests stored in the registry and there needs to be a change in the schema, that is when we'll need to migrate the db.  
- I've included a API test to test querying a bundle path(currently it just checks for an empty path as all `bundlePath` values are set to `null`). Also, seems like at some point we should refactor the API tests to mock a database instead of using an actual database which seems more like integration tests?

ecordell:
> Also, seems like at some point we should refactor the API tests to mock a database instead of using an actual database which seems more like integration tests?

It would be good to have two layers, but `sqlite` is often used as a mock sql database anyway, so IMO it's low on the list of priorities.

ecordell:
/hold 

I missed some things on my initial review.

There are multiple API calls in OLM that will need to have `Path`s returned to them. See the querier: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/resolver/querier.go

(the `source` in that package is the grpc client to the registry server)

Because there are multiple places that we need to return this, could we extend the `Bundle` struct itself? We could add a new field `path`, and then clients can prefer the path field if it exists?

cmd/opm/registry/serve.go
```diff
@@ -57,6 +58,23 @@ func runRegistryServeCmdFunc(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Migrate database to latest version before serving
+	migrator, err := sqlite.NewSQLLiteMigrator(db, "")
+	if err != nil {
+		return err
+	}
+	defer migrator.CleanUpMigrator()
```
ecordell: probably shouldn't defer this, it won't get run until the server exits (which could be never)

anik120:
> There are multiple API calls in OLM that will need to have `Path`s returned to them. See the querier: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/resolver/querier.go
> 
> (the `source` in that package is the grpc client to the registry server)
> 
> Because there are multiple places that we need to return this, could we extend the `Bundle` struct itself? We could add a new field `path`, and then clients can prefer the path field if it exists?

@ecordell that is actually a very good point. I've opened another [PR](#103) where I've extended the `Bundle` struct and `GetBundle` API instead of adding a new API. I'll go ahead and close this PR if that looks good.

anik120:
Closing in favor of #103

PR #100, opened 2019-10-22
Migrate database up before serving using opm

anik120:

- Add wrapper around Up migration
- Call wrapper before serving database using opm cli

Merge before #99 
 
**Reviewer Checklist**

### DIFF
--- a/cmd/opm/registry/serve.go
+++ b/cmd/opm/registry/serve.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"context"
+	"database/sql"
 	"net"
 
 	"github.com/sirupsen/logrus"
@@ -64,8 +65,19 @@ func runRegistryServeCmdFunc(cmd *cobra.Command, args []string) error {
 
 	logger := logrus.WithFields(logrus.Fields{"database": dbName, "port": port})
 
+	db, err := sql.Open("sqlite3", dbName)
+	if err != nil {
+		return err
+	}
+
+	// Migrate database to latest version before serving
+	err = migrateToLatest(db, dbName)
+	if err != nil {
+		return err
+	}
+
 	var store registry.Query
-	store, err = sqlite.NewSQLLiteQuerier(dbName)
+	store = sqlite.NewSQLLiteQuerierFromDb(db)
 	if err != nil {
 		logger.WithError(err).Warnf("failed to load db")
 	}
@@ -96,5 +108,19 @@ func runRegistryServeCmdFunc(cmd *cobra.Command, args []string) error {
 		logger.Fatalf("failed to serve: %s", err)
 	}
 
+	return nil
+}
+
+func migrateToLatest(db *sql.DB, dbName string) error {
+	migrator, err := sqlite.NewSQLLiteMigrator(db, "")
+	if err != nil {
+		return err
+	}
+	defer migrator.CleanUpMigrator()
+
+	err = migrator.MigrateUp(dbName)
+	if err != nil {
+		return err
+	}
 	return nil
 }
--- a/pkg/sqlite/migrator.go
+++ b/pkg/sqlite/migrator.go
@@ -18,7 +18,7 @@ import (
 type SQLMigrator struct {
 	db             *sql.DB
 	migrationsPath string
-	generated bool
+	generated      bool
 }
 
 // NewSQLLiteMigrator returns a SQLMigrator. The SQLMigrator takes a sql database and directory for migrations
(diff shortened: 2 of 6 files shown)


cmd/opm/registry/serve.go
```diff
@@ -98,3 +111,17 @@ func runRegistryServeCmdFunc(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+func migrateToLatest(db *sql.DB) error {
```
kevinrizza: I don't think this is the right place for this function. If we want to reuse it in other places (like when we add other commands to `opm registry`, maybe we should move it into the `sqlite` package? Otherwise we should just leave it in line.

pkg/sqlite/testdata/test_db_migrations/up_migration/194708150000_test_up_migration.down.sql
```diff
@@ -0,0 +1 @@
+/* Not required */
```
kevinrizza: For completions sake, maybe this should be an actual down migration? I see that the test isn't going to test this since it's probably hard to simulate an error here for right now, but I think we may want to revisit this later if we decided to add a `migrate` command.

pkg/sqlite/query.go
```diff
@@ -25,6 +25,10 @@ func NewSQLLiteQuerier(dbFilename string) (*SQLQuerier, error) {
 	return &SQLQuerier{db}, nil
 }
 
+func NewSQLLiteQuerierFromDb(db *sql.DB) (*SQLQuerier, error) {
```
kevinrizza: Why return an error here if it will always be `nil`?

pkg/sqlite/migrator.go
```diff
@@ -111,6 +111,28 @@ func (m *SQLMigrator) InitMigrationVersion() error {
 	return nil
 }
 
+// MigrateUp is a wrapper around golang-migrate's Up. Up
+// looks at the currently active migration version and will
+// migrate all the way up (applying all up migrations).
+func (m *SQLMigrator) MigrateUp() error {
+	instance, err := sqlite3.WithInstance(m.db, &sqlite3.Config{})
```
ecordell: This will make the migrator sqlite-only. 

If we change 
```go
type SQLMigrator struct {
	db             *sql.DB
	migrationsPath string
	generated bool
}
```

to 

```go
type SQLMigrator struct {
	db             *database.Driver
	migrationsPath string
	generated bool
}
```
then we can use the migrator for any db?

kevinrizza: We can make this change, but the golang-migrate needs that instance object to understand which database and explicitly support it. IMO we should probably revisit this later if we want to abstract the migrator to understand multiple database types.

pkg/sqlite/query.go
anik120: I honestly don't know how it got there. Changing it.

cmd/opm/registry/serve.go
anik120: @kevinrizza this is a stop gap measure till we have the `migrate` command available with the opm cli. Once we have that this function can be removed and the corresponding function from the `registry` package should be used [here](https://github.com/operator-framework/operator-registry/blob/9d0489171c30c887bae1341760584bb967ceef3a/cmd/opm/registry/serve.go#L75)

pkg/sqlite/testdata/test_db_migrations/up_migration/194708150000_test_up_migration.down.sql
anik120: @kevinrizza there's no harm in adding a down migration. And in that note, I'm changing the folder name from `up_migration` to `migrations` to test both up and down migrations later.

pkg/sqlite/migrator.go
anik120: +1 with Kevin.

ecordell: > golang-migrate needs that instance object to understand which database and explicitly support it.

I don't think this is the case, but we can revisit it in the future.

I do think we should be careful to pass in a `DatabaseName` to `sqlite3.Config`, right?

anik120: > I do think we should be careful to pass in a DatabaseName to sqlite3.Config, right?

@ecordell I've updated the PR to pass in a `DatabaseName`. Why do you say `careful` though? I didn't seem to find any "best practice guide" around passing the database name in the docs so was curious.

anik120:
Closing in favor or #101

PR #101, opened 2019-10-22
Add support for storing and retrieving related images

ecordell:

This is based on #100, which should merge first.

This adds two new `Query` methods:

```go
type Query interface { 
  ListImages(ctx context.Context) ([]string, error)
  GetImagesForBundle(ctx context.Context, bundleName string) ([]string, error)
  // ...
}
```

The bundle abstraction now also understands the images in the `containers` and `initContainers` fields of a CSV, and the `relatedImages` field:

```yaml
kind: ClusterServiceVersion 
spec:
  relatedImages:
  - name: default
    image: quay.io/coreos/etcd@sha256:12345 
  - name: etcd-2.1.5
    image: quay.io/coreos/etcd@sha256:12345 
  - name: etcd-3.1.1
    image: quay.io/coreos/etcd@sha256:12345 
```

See the [proposal](https://github.com/operator-framework/enhancements/blob/master/enhancements/opm-mirroring.md#specify-related-images), this is primarily to support disconnected mirroring of a catalog.

**Reviewer Checklist**

Replaces #62


pkg/sqlite/migrator.go
```diff
@@ -1,167 +1,175 @@
...
-	if err != nil {
+	if err := m.ensureMigrationTable(ctx, tx); err != nil {
 		return err
 	}
 
-	initVersion, err := getLatestMigrationVersion(m.migrationsPath)
-	if err != nil {
-		return err
-	}
+	for _, migration := range migrations {
```
ecordell: The bulk of the migrator is here

pkg/sqlite/migrations/migrations.go
```diff
@@ -0,0 +1,56 @@
...
+
+type Migration struct {
+	Id int
+	Up func(context.Context, *sql.Tx) error
+	Down func(context.Context, *sql.Tx) error
+}
+
+type MigrationSet map[int]*Migration
+
+var migrations MigrationSet = make(map[int]*Migration)
```
ecordell: this map is built up in `init` methods on the migration fields

pkg/sqlite/migrations/000_init.go
```diff
@@ -0,0 +1,78 @@
...
+			FOREIGN KEY(channel_entry_id) REFERENCES channel_entry(entry_id),
+			FOREIGN KEY(group_name, version, kind) REFERENCES api(group_name, version, kind)
+		);
+		`
+		_, err := tx.ExecContext(ctx, sql)
+		return err
+	},
+	Down: func(ctx context.Context, tx *sql.Tx) error {
+		sql := `
+			DROP TABLE operatorbundle;
```
anik120: Won't trying to drop tables without switching off foreign key constraints fail?

pkg/sqlite/migrations/000_init.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+	"database/sql"
+)
+
+var InitMigrationKey = 0
+
+var initMigration = &Migration{
+	Id: InitMigrationKey,
+	Up: func(ctx context.Context, tx *sql.Tx) error {
```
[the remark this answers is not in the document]
kevinrizza: If we do this we lose the ability to run migrations on registry databases that were built before this migrator was added. I think that's fine, though.

pkg/sqlite/migrations/000_init.go
ecordell: Great note - this will succeed because it will also drop all of the tables with the foreign key refs.

I'll update to enable foreign keys on these tests so that we catch future issues like this.

pkg/registry/csv.go
```diff
@@ -187,3 +191,74 @@ func (csv *ClusterServiceVersion) GetApiServiceDefinitions() (owned []*Definitio
...
+		imageSet[img.Ref] = struct{}{}
+	}
+
+	return
+}
+
+// GetOperatorImages returns a list of any images used to run the operator.
+// Currently this pulls any images in the pod specs of operator deployments.
+func (csv *ClusterServiceVersion) GetOperatorImages() (map[string]struct{}, error) {
+	type dep struct {
```
kevinrizza: Can we import these types from somewhere rather than redefine them here? If/When the spec changes this may be hard to find.

ecordell:
@anik120 I enabled foreign keys for the tests and added a doc on adding migrations.

pkg/sqlite/migrations/000_init.go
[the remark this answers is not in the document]
ecordell: Thanks for catching this - I had intended to go back and make another migrator method that migrates from the current version up to the latest version. Added and updated.

pkg/registry/csv.go
ecordell: This is specifically to avoid importing OLM

PR #103, opened 2019-10-23
Image digests for referencing bundles

anik120:

- Change schema of OperatorBundle to include field 'bundlepath'
- Extend GetBundle API to return bundle image path
- Modify GetBundle API to return struct with avaiable fields
  instead of error while querrying empty bundles

**Reviewer Checklist**


anik120:
@ecordell @kevinrizza I've rebased it on top of #101. PTAL

pkg/sqlite/migrations.go
```diff
@@ -0,0 +1,310 @@
+// Code generated by go-bindata.
```
ecordell: I think you can delete this

anik120:
> can you add positive tests for the grpc apis (i.e. if bundle = nil and bundlepath = path in the db, the grpc apis return the right thing)?

@ecordell looks like there's two ways we can add this test with the way the tests are laid out now:
1) Wait for #89 to merge which will then allow us to `add`  an operator bundle from a `bundle path` 
2) Merge this now and add the test after #89 merges.

There is also a third option: Write a test that creates a database and populates it with the information that we need to query, but I feel that'll just be dead code soon.

ecordell:
/lgtm

That's fine, we can come back with tests after #89 merges

PR #105, opened 2019-10-25
opm registry build / mirror (library only)

ecordell:
This adds supporting libraries to be used in the `oc` commands for disconnected. 

Eventually, these can be used to provide similar features to `opm`, but there is a lot of legwork to do first (`opm` does not currently  have ways to push, pull, append, or extract images like `oc` does tody).

Enable disconnected mirroring for appregistry catalogs via `oc`.

**Reviewer Checklist**


cmd/opm/registry/build.go
```diff
@@ -0,0 +1,48 @@
+package registry
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/pkg/appregistry"
+)
+
+func BuildCmd() *cobra.Command {
```
ecordell: These are not actually registered with the parent, they are inert (not just "hidden")

pkg/appregistry/builder.go
```diff
@@ -0,0 +1,222 @@
...
+	DatabasePath        string
+	CacheDir            string
+
+	// derived
+	CleanOutput bool
+	ManifestDir string
+	DatabaseDir string
+}
+
+func NewAppregistryImageBuilder(config *AppregistryBuildOptions, options ...AppregistryBuildOption) (*AppregistryImageBuilder, error) {
```
njhale: Taking the config be a formal parameter defeats at least one reason for using the variadic options pattern -- that is, always requiring the config argument (i.e. `NewAppregistryImageBuilder(nil)` vs. `NewAppregistryImageBuilder()`). 

Also, a common configuration can always be stored as a slice of pre-baked option closures.

pkg/appregistry/builder.go
```diff
@@ -0,0 +1,222 @@
...
+	Append(from, to, layer string) error
+}
+
+type ImageAppendFunc func(from, to, layer string) error
+
+func (f ImageAppendFunc) Append(from, to, layer string) error {
+	return f(from, to, layer)
+}
+
+type AppregistryImageBuilder struct {
```
njhale: nit: do these need to be public?

pkg/appregistry/builder.go
njhale: Would it be more flexible to have a return type of `RegistryImageBuilder` (an interface)?

pkg/appregistry/manifest_downloader.go
```diff
@@ -0,0 +1,55 @@
...
+	packages, err := d.client.ListPackages(namespace)
+	if err != nil {
+		return err
+	}
+
+	for _, pkg := range packages {
+		klog.V(4).Infof("Downloading %s", pkg)
+		manifest, err := d.client.RetrieveOne(namespace+"/"+pkg.Name, pkg.Release)
+		if err != nil {
+			return err
```
njhale: We should aggregate errors instead of bailing out immediately so we can maintain our best-effort behavior.

pkg/mirror/mirror.go
```diff
@@ -0,0 +1,116 @@
...
+	ImageMirrorer       ImageMirrorer
+	DatabaseExtractor   DatabaseExtractor
+
+	// options
+	Source, Dest        string
+}
+
+var _ Mirrorer = &IndexImageMirrorer{}
+
+func NewIndexImageMirror(config *IndexImageMirrorerOptions, options ...ImageIndexMirrorOption) (*IndexImageMirrorer, error) {
```
njhale: see [comment](https://github.com/operator-framework/operator-registry/pull/https://github.com/operator-framework/operator-registry/blob/cfc214de5a0ee31a76f32060a97d3175d606fdfe/pkg/appregistry/builder.go#r339886519) for this

pkg/mirror/mirror.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer db.Close()
+
+	migrator, err := sqlite.NewSQLLiteMigrator(db)
+	if err != nil {
+		return nil, err
+	}
+	if err := migrator.Migrate(context.TODO()); err != nil {
```
njhale: > Is there a clever way to plumb an external context to this?

pkg/mirror/mirror.go
```diff
@@ -0,0 +1,116 @@
...
+
+	querier := sqlite.NewSQLLiteQuerierFromDb(db)
+	images, err := querier.ListImages(context.TODO())
+	if err != nil {
+		return nil, err
+	}
+
+	mapping := map[string]string{}
+
+	errs := []error{}
```
njhale: nit: `var err []error` would work nicely as a zero value too.

pkg/mirror/mirror_test.go
```diff
@@ -0,0 +1,108 @@
...
+				"quay.io/coreos/etcd@sha256:3816b6daf9b66d6ced6f0f966314e2d4f894982c6b1493061502f8c2bf86ac84":"localhost/coreos/etcd@sha256:3816b6daf9b66d6ced6f0f966314e2d4f894982c6b1493061502f8c2bf86ac84",
+				"quay.io/coreos/etcd@sha256:49d3d4a81e0d030d3f689e7167f23e120abf955f7d08dbedf3ea246485acee9f":"localhost/coreos/etcd@sha256:49d3d4a81e0d030d3f689e7167f23e120abf955f7d08dbedf3ea246485acee9f",
+				"quay.io/coreos/prometheus-operator@sha256:0e92dd9b5789c4b13d53e1319d0a6375bcca4caaf0d698af61198061222a576d":"localhost/coreos/prometheus-operator@sha256:0e92dd9b5789c4b13d53e1319d0a6375bcca4caaf0d698af61198061222a576d",
+				"quay.io/coreos/prometheus-operator@sha256:3daa69a8c6c2f1d35dcf1fe48a7cd8b230e55f5229a1ded438f687debade5bcf":"localhost/coreos/prometheus-operator@sha256:3daa69a8c6c2f1d35dcf1fe48a7cd8b230e55f5229a1ded438f687debade5bcf",
+				"quay.io/coreos/prometheus-operator@sha256:5037b4e90dbb03ebdefaa547ddf6a1f748c8eeebeedf6b9d9f0913ad662b5731":"localhost/coreos/prometheus-operator@sha256:5037b4e90dbb03ebdefaa547ddf6a1f748c8eeebeedf6b9d9f0913ad662b5731",
+			},
+			// strimzi has invalid images in its manifests, so we both return a list of maps and an error
+			wantErr: true,
```
njhale: nit: Why not just check the value of the error returned?

pkg/appregistry/builder.go
ecordell: I did this so that I can directly use the options built by cobra flags (which is the style `oc` uses).

I'll update with a compromise: a method on `AppregistryBuildOptions`, `ToOption` that builds an `AppregistryBuildOption` from an existing option struct. Then the initializer can only have variadic options.

pkg/appregistry/builder.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := db.Close(); err != nil {
+			klog.Warningf(err.Error())
+		}
+	}()
+
+	if err := dbLoader.Migrate(context.TODO()); err != nil {
```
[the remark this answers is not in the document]
ecordell: This is being called by oc which doesn't offer a context and which does kill the process with signals, so I think this is a valid but low-value suggestion to implement. When we implement the opm versions (for real) we should do that.

pkg/appregistry/manifest_downloader.go
```diff
@@ -0,0 +1,59 @@
...
+			errs = append(errs, err)
+			continue
+		}
+		if _, err = decoder.Decode([]*apprclient.OperatorMetadata{manifest}, directory); err != nil {
+			errs = append(errs, err)
+			continue
+		}
+	}
+
+	return nil
```
njhale: This still needs to return the aggregate error, no?

pkg/appregistry/manifest_downloader.go
```diff
@@ -0,0 +1,59 @@
...
+	klog.V(4).Infof("Downloading manifests at namespace %s to %s", namespace, directory)
+
+	log := logrus.New().WithField("ns", namespace)
+
+	packages, err := d.client.ListPackages(namespace)
+	if err != nil {
+		return err
+	}
+
+	errs := []error{}
```
njhale: nit: If you need to correct the error aggregation anyway, I would be consistent with `mirror.go` by using `var errs []error`.

PR #106, opened 2019-10-28
Add index build commands

kevinrizza:
- Added `opm index add` which generates indexes (either as a
dockerfile or shelling out to container tools) with additional bundles
- Added `opm index rm` which generates an index based on a previous
index minus a specified operator

In order to write tests for these, added mock files for the new
`containertools` package via `gomocks`. Added a readme in the
`doc/contributors/` folder that defines the process of updating these
mocks.

Note: this needs to be rebased on top of https://github.com/operator-framework/operator-registry/pull/89/ once that merges. For now, marking this pr as WIP

**Reviewer Checklist**


pkg/containertools/command.go
```diff
@@ -0,0 +1,133 @@
...
+	GetToolName() string
+	Pull(image string) error
+	Build(dockerfile, tag string) error
+	Save(image, tarFile string) error
+	Inspect(image string) ([]byte, error)
+}
+
+// CommandRunnerImpl is configured to select a container cli tool and execute commands with that
+// tooling.
+type CommandRunnerImpl struct{
```
ecordell: I kind of expected a `DockerCommandRunner` and a `PodmanCommandRunner` implementation - I know their cli's are supposed to be compatible, but are they the only two we may support? Are there no differences other than the root command?

(also not blocking; more a question for future changes)

pkg/containertools/dockerfilegenerator.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	baseImage = "scratch"
+	defaultBinarySourceImage = "quay.io/kevinrizza/upstream-opm-builder"
```
ecordell: probably should be in operator-framework 😄

pkg/containertools/dockerfilegenerator.go
```diff
@@ -0,0 +1,98 @@
...
+func NewDockerfileGenerator(containerTool string, logger *logrus.Entry) DockerfileGenerator {
+	return &DockerfileGeneratorImpl{
+		logger: logger,
+		labelReader: NewLabelReader(containerTool, logger),
+	}
+}
+
+// GenerateIndexDockerfile builds a string representation of a dockerfile to use when building
+// an operator-registry index image
+func (g *DockerfileGeneratorImpl) GenerateIndexDockerfile(binarySourceImage, previousIndexImage string, bundlesToAdd, operatorsToDelete []string) (string, error) {
```
ecordell: super-nit: my brain keeps reading "binary source image" as "source code written in binary" and not "the source image containing the binaries"

pkg/containertools/dockerfilegenerator.go
```diff
@@ -0,0 +1,98 @@
...
+		// copy the old database file in the scratch image
+		dockerfile += fmt.Sprintf("COPY --from=previous-index %s /previous-index.db\n", previousDbLocation)
+	} else { // Otherwise just define the builder as a place to get the binary and generate the database
+		dockerfile += fmt.Sprintf("FROM %s AS builder\n", binarySourceImage)
+	}
+
+	// Generate the registry database
+	if len(bundlesToAdd) > 0 {
+		add := strings.Join(bundlesToAdd, ",")
+		dockerfile += fmt.Sprintf("RUN /build/bin/opm registry add \"%s\" -o \"index.db\"\n", add)
```
ecordell: nit: seems weird to intentionally place our binaries in `/build`?

pkg/lib/index/indexer.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+const (
+	defaultDockerfileName = "index.Dockerfile"
+	defaultImageTag = "operator-registry-index:latest"
+)
+
+// Indexer allows the creation of index container images from scratch or
+// based on previous index images
+type Indexer interface {
```
ecordell: This feels like it should be two interfaces, one for adding and one for removing?

pkg/lib/index/indexer.go
```diff
@@ -0,0 +1,129 @@
...
+	AddToIndex(AddToIndexRequest) error
+	DeleteFromIndex(DeleteFromIndexRequest) error
+}
+
+// IndexerImpl is a struct implementation of the Indexer interface
+type IndexerImpl struct {
```
ecordell: This is mostly a style thing, but I don't love `<interface> interface` and `<interface>Impl struct`, if you can think if something more descriptive.

For this one, I could imagine a a non-image-based implementation of this that operates on database files only. Then I would have `IndexerImpl` and `DatabaseFileIndexerImpl`, at which point why not have `ImageIndexer` and `DatabaseFileIndexer` so I can tell their intent apart?

cmd/opm/index/add.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+
+	logger := logrus.WithFields(logrus.Fields{"bundles": bundles})
+
+	logger.Info("building the index")
+
+	indexer := index.NewIndexer(containerTool, logger)
+
+	request := index.AddToIndexRequest{
```
ecordell: This is fine for now, for some other commands I started doing something similar to what `oc` does for building cobra commands:

https://github.com/operator-framework/operator-registry/blob/master/pkg/mirror/options.go#L7

It's not a huge problem the way it's written now, but it makes it difficult to use the defaulting/validation that's here if I wanted to use an indexer as a library (because that is done solely though the cobra flag setting).

docs/contributors/mocking.md
```diff
@@ -0,0 +1,15 @@
+# Mocking code in operator-registry
```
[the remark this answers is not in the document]
kevinrizza: ah, I was going off this:

https://github.com/operator-framework/operator-registry/tree/master/pkg/apprclient/mock

but now I realize that that's probably there because we never cleaned it up? I can look into this and see if I can do it quickly. We probably want to have all of the mocks in the project on the same framework at some point

pkg/lib/index/indexer.go
```diff
@@ -0,0 +1,129 @@
+package index
```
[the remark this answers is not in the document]
kevinrizza: IMO the right way to do all of these is to stop putting functional logic into the `cmd` package and instead define that as just the steps to collect parameters from the terminal. Then have the first order function that actually implements those in the `lib` package. If we do this, it allows any arbitrary project to import our calls without actually having to import the Cobra stuff. Also, it makes testing the higher order library functions significantly easier.This is a first pass attempt to do that.

pkg/containertools/command.go
kevinrizza: That was my first thought too, but it turned out that so far the implementation is identical so it seemed like a bit of a waste to create two implementations of the interface. I was imagining that if/when the actual implementation differed we could split it out at that point.

pkg/containertools/dockerfilegenerator.go
kevinrizza: yeah, this is one of the things that needs to change once this pr is rebased. :)

pkg/containertools/dockerfilegenerator.go
kevinrizza: Yeah I had a lot of trouble coming up with these names (there are actually 3 different images involved). I was hoping someone could tell me this is bad and have a better idea. Any suggestions?

pkg/containertools/dockerfilegenerator.go
kevinrizza: Yeah this is an artifact of the fact that the builder image (https://github.com/operator-framework/operator-registry/blob/master/upstream-builder.Dockerfile) just runs `make` on the source. If we wanted to change that we may have to create a new dockerfile for that build since I assume other things depend on that one.

pkg/lib/index/indexer.go
kevinrizza: good point

cmd/opm/index/add.go
kevinrizza: interesting, i'll take a look

kevinrizza:
@ecordell Updated based on your feedback, ptal

pkg/containertools/dockerfilegenerator.go
ecordell: Ah, I see that in the final image it's in a better location. Thanks!

pkg/containertools/dockerfilegenerator_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+func TestGenerateDockerfile(t *testing.T) {
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	binarySourceImage := "quay.io/operator-framework/builder"
+	previousIndexImage := ""
+	exampleBundles := []string{"quay.io/operator-framework/examplebundle"}
+	expectedDockerfile := `FROM quay.io/operator-framework/builder AS builder
+RUN /build/bin/opm registry add "quay.io/operator-framework/examplebundle" -o "index.db"
```
gallettilance: nit: can we tab this?

pkg/lib/indexer/indexer.go
```diff
@@ -0,0 +1,287 @@
...
+	BinarySourceImage string
+	FromIndex string
+	OutDockerfile string
+	Bundles []string
+	Tag string
+}
+
+// AddToIndex is an aggregate API used to generate a registry index image with additional bundles
+func (i ImageIndexer) AddToIndex(request AddToIndexRequest) error {
+	databaseFile := defaultDatabaseFile
```
gallettilance: nit: spacing looks off in this function

pkg/lib/indexer/indexer.go
```diff
@@ -0,0 +1,287 @@
...
+	BinarySourceImage string
+	FromIndex string
+	OutDockerfile string
+	Tag string
+	Operators []string
+}
+
+// DeleteFromIndex is an aggregate API used to generate a registry index image 
+// without specific operators
+func (i ImageIndexer) DeleteFromIndex(request DeleteFromIndexRequest) error {
```
gallettilance: nit: spacing looks off in this function too

pkg/containertools/dockerfilegenerator_test.go
```diff
@@ -0,0 +1,66 @@
...
+	dockerfileGenerator := containertools.IndexDockerfileGenerator{
+		Logger: logger,
+	}
+
+	dockerfile := dockerfileGenerator.GenerateIndexDockerfile(binarySourceImage, databaseFolder)
+	require.Equal(t, dockerfile, expectedDockerfile)
+}
+
+func TestGenerateDockerfile_EmptyBaseImage(t *testing.T) {
+	controller := gomock.NewController(t)
```
ecordell: nit: I don't think this is used anymore?

pkg/containertools/labelreader.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// parse output of inspect to get labels
+	switch containerTool := r.Cmd.GetToolName(); containerTool {
+	case "docker":
+		var data []DockerImageData
+		err := json.Unmarshal(imageData, &data)
+		if err != nil {
+			return nil, err
+		}
+		return data[0].Config.Labels, nil
```
ecordell: nit/followup: should check length of data to be safe, right?

pkg/containertools/labelreader_test.go
```diff
@@ -0,0 +1,566 @@
+package containertools_test
+
+import (
+    "fmt"
```
ecordell: nit: gofmt

PR #112, opened 2019-10-31
GRPC client wrapper returns api objects

ecordell:
This changes the client wrapper to return the grpc api objects directly, instead of abstracting them with the `registry.Bundle` object.

The registry.Bundle object is not a particularly useful abstraction now that the api is returning data directly (instead of returning a CSV that is then parsed for the data). On the OLM side, we just want to use the fields on the object directly instead of calculating them, when possible.

**Reviewer Checklist**


dinhxuanvu:
lgtm

jpeeler:
/lgtm

I didn't see anywhere in OLM that used crds (the only field on the registry bundle that doesn't have a compliment), but it seems like that'll come up eventually.

PR #121, opened 2019-11-06
Make imagereader work for all layer paths

kevinrizza:
Problem:
Docker manifests can define the location of the tarball outside the
root directory of the tarball. For those manifests, the image reader
fails to pull all the required data

Solution:
If the manifest.json defines the location outside that root dir, we need
to traverse the reader past the root files to read all files

**Reviewer Checklist**


pkg/containertools/test_files/expected_unpack/manifests/prometheusoperator.0.14.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,240 @@
+#! parse-kind: ClusterServiceVersion
```
njhale: nit: in go, the idiomatic location for "ancillary test data" is the `testdata` directory; see https://golang.org/pkg/cmd/go/internal/test/

pkg/containertools/imagereader_test.go
```diff
@@ -0,0 +1,145 @@
...
+const (
+	expectedFilePath = "test_files/expected_unpack"
+)
+
+func TestReadImageLayersDocker(t *testing.T) {
+	image := "quay.io/operator-framework/example"
+	testWorkingDir := "test_files/docker"
+	testOutputDir := "test_files/output"
+
+	expectedFiles, err := helperGetExpectedFiles()
```
njhale: nit: the indentation is a little funky here.

pkg/containertools/imagereader_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+    logger := logrus.NewEntry(logrus.New())
```
ecordell: nit: gofmt

pkg/containertools/imagereader.go
```diff
@@ -42,7 +42,12 @@ func NewImageReader(containerTool string, logger *logrus.Entry) ImageReader {
 	}
 }
 
-func (b ImageLayerReader) GetImageData(image, outputDir string) error {
+func (b ImageLayerReader) GetImageData(image, outputDir string, opts ...GetImageDataOption) error {
```
ecordell: I don't see anything (non-test) calling this with the new option - did I miss that somewhere?

pkg/containertools/getimagedata_options.go
```diff
@@ -0,0 +1,14 @@
+package containertools
+
+type GetImageDataOptions struct {
+	WorkingDir string
```
njhale: Nice. You could probably make `OutputDir` an option too (and remove the second parameter of `GetImageData`).

Aside from that, I can't help but think of taking a `Writer` and `Reader` instead of directory strings -- thought that's probably premature optimization.

pkg/containertools/imagereader.go
kevinrizza: it's just to inject the temp dir for tests so that we can inject the result of the `docker save`

pkg/containertools/getimagedata_options.go
kevinrizza: My attempt here was to add just the optional parameters and `OutputDir` is required.

kevinrizza:
Addressed comments, ptal

PR #122, opened 2019-11-08
Bug 1771522: add version and skiprange to operatorbundle table

gallettilance:

**Reviewer Checklist**


dinhxuanvu:
lgtm

pkg/sqlite/load.go
```diff
@@ -639,7 +655,21 @@ func (s *SQLLoader) AddBundlePackageChannels(manifest registry.PackageManifest,
...
-	if _, err := stmt.Exec(csvName, csvBytes, bundleBytes, bundleImage); err != nil {
+	csv, err := bundle.ClusterServiceVersion()
+	if err != nil {
+		return err
+	}
+
+	skiprange, ok := csv.Annotations[SkipRangeAnnotationKey]
+	if !ok {
+		skiprange = ""
+	}
```
njhale: What do you think about having a `SkipRange()` method on the `ClusterServiceVersion` struct and handling this there?

pkg/sqlite/load.go
```diff
@@ -77,7 +79,21 @@ func (s *SQLLoader) AddOperatorBundle(bundle *registry.Bundle) error {
...
+	version, err := csv.GetVersion()
+	if err != nil {
+		version = ""
+	}
```
njhale: It looks like we're doing this in multiple places. Can we surface these as methods of `registry.Bundle()` and memoize this work?

PR #124, opened 2019-11-13
Bug 1771273: Fix bundle build command with Docker and update documentation

dinhxuanvu:
The bundle build command using Docker fails in some cases due
to Docker can't access some files which in turn due to issue
with generated Docker command doesn't account for manifests
directory. The `ADD` command in Dockerfile is now changed to
`COPY` given we don't use any features from ADD.

Update documentation + unit test cases for those above changes.

**Reviewer Checklist**


docs/design/operator-bundle.md
```diff
@@ -159,6 +159,12 @@ The `--package` or `-p` is the name of package fo the operator such as `etcd` wh
 
 All information in `annotations.yaml` is also existed in `LABEL` section of `Dockerfile`.
 
+The `Dockerfile` can used to manually build bundle image using container image tools such as Docker, Podman or Buildah. For example, the Docker build command would be:
```
jpeeler: manually to build the bundle image using an image build tool such as ...

pkg/lib/bundle/generate.go
```diff
@@ -43,10 +43,14 @@ type AnnotationMetadata struct {
 // @channelDefault: The default channel for the bundle image
 // @overwrite: Boolean flag to enable overwriting annotations.yaml locally if existed
 func GenerateFunc(directory, packageName, channels, channelDefault string, overwrite bool) error {
-	var mediaType string
+	_, err := os.Stat(directory)
+	if os.IsNotExist(err) {
+		return err
+	}
 
+	var mediaType string
```
jpeeler: You can get rid of this and change below back to using short assignment.

pkg/lib/bundle/generate.go
```diff
@@ -77,7 +81,7 @@ func GenerateFunc(directory, packageName, channels, channelDefault string, overw
 	log.Info("Building Dockerfile")
 
 	// Generate Dockerfile
-	content, err = GenerateDockerfile(directory, mediaType, ManifestsDir, MetadataDir, packageName, channels, channelDefault)
+	content, err = GenerateDockerfile(mediaType, ManifestsDir, MetadataDir, packageName, channels, channelDefault)
```
jpeeler: :+1:

pkg/lib/bundle/generate.go
```diff
@@ -246,8 +250,8 @@ func GenerateDockerfile(directory, mediaType, manifests, metadata, packageName,
 	fileContent += fmt.Sprintf("LABEL %s=%s\n\n", ChannelDefaultLabel, chanDefault)
 
 	// CONTENT
-	fileContent += fmt.Sprintf("ADD %s %s\n", filepath.Join(directory, "*.yaml"), "/manifests/")
-	fileContent += fmt.Sprintf("ADD %s %s%s\n", filepath.Join(directory, metadata, AnnotationsFile), "/metadata/", AnnotationsFile)
+	fileContent += fmt.Sprintf("COPY %s %s\n", "/*.yaml", "/manifests/")
```
jpeeler: Yeah, I think COPY is better too.

PR #125, opened 2019-11-14
Bug 1771850: Error logging for opm alpha bundle extract command

anik120:

The `opm alpha bundle extract` command does not log any error
if the input annotations yaml has incorrect keys. This PR
introduces error logging for the command when the keys are
incorrect. The command will also error out before building the
configmap.

**Reviewer Checklist**

### DIFF
--- a/pkg/configmap/configmap_writer.go
+++ b/pkg/configmap/configmap_writer.go
@@ -1,6 +1,7 @@
 package configmap
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -8,10 +9,12 @@ import (
 
 	"github.com/ghodss/yaml"
 	_ "github.com/mattn/go-sqlite3"
+	errorwrap "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/operator-framework/operator-registry/pkg/client"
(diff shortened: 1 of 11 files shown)


pkg/configmap/configmap_writer.go
```diff
@@ -63,6 +64,32 @@ func TranslateInvalidChars(input string) string {
 	return validConfigMapKey
 }
 
+func validateAnnotationKeys(file AnnotationsFile) error {
+	var errors []string
+	if file.Annotations.Resources == "" {
+		errors = append(errors, fmt.Errorf("No value set for operators.operatorframework.io.bundle.manifests.v1").Error())
```
jpeeler: all of these can use the constant value in pkg/lib/bundle

anik120: @jpeeler thank you for pointing the constants out. I've updated the PR to use the constants.

pkg/configmap/configmap_writer.go
```diff
@@ -108,6 +135,10 @@ func (c *ConfigMapWriter) Populate(maxDataSizeLimit uint64) error {
 				if err != nil {
 					return err
 				}
+				err = validateAnnotationKeys(annotationsFile)
```
jpeeler: Sorry for not pointing this out earlier. The idea for validation is to check the configmap data. While it's not much different from the bundle data, it'll check a little more code.

pkg/configmap/configmap_writer.go
```diff
@@ -63,6 +64,32 @@ func TranslateInvalidChars(input string) string {
 	return validConfigMapKey
 }
 
+func validateAnnotationKeys(file AnnotationsFile) error {
+	errors := []error{}
+	if file.Annotations.Resources == "" {
+		errors = append(errors, fmt.Errorf("No value set for %s", bundle.ManifestsLabel))
```
jpeeler: You might consider just doing `errors.New(bundle.ManifestsLabel)` and wrapping the error with an appropriate error message if this function returns with an error.

pkg/configmap/configmap_writer.go
anik120: @jpeeler In the bug report, the invalid `annotation` yaml contains invalid keys, as a result of which the configmap is not populated the way it was expected. 
This PR checks that the valid keys are present in the `annotation` yaml, and throws an error if any expected key is missing. It does not however attempt to validate the values of the keys. 
I was under the assumption that the values are validated elsewhere, do you think we should validate that the keys are present, and that the values of those keys are of the right format here?

jpeeler: Yes, I think the configmap values should be verified. It doesn't change the code all that much aside from moving this call to below the configmap call and validating the annotations instead of the file data. Put another way - when generating the bundle there will be validation there checking the format, so on this side checking the configmap is best in my opinion.

pkg/configmap/configmap_writer.go
anik120: @jpeeler is there any specific reason we'd want to use `errors.New()` instead of `fmt.Errorf()`? `fmt.Errorf()` allows us to have the words `No value set for ___` for each error, and then `strings.Join()` allows us to display all the errors in a comprehensible way: 
```
No value set for operators.operatorframework.io.bundle.manifests.v1
No value set for operators.operatorframework.io.bundle.package.v1
```
Is there a better practice for error aggregating with `errors,New()` that allows us to display the output like that?

jpeeler: This is just a way to reduce code duplication and make the reported errors a little less verbose. If you return just the annotation with a problem, you can do something like:
```
if err = validateAnnotationKeys(configMapPopulate.GetAnnotations()); err != nil {
    return errorwrap.Wrap(err, "annotation validation failed, missing or empty values")
}
```
I don't know if it's really a best practice though.

pkg/configmap/configmap_writer.go
anik120: I see your point. Updated the PR.

pkg/configmap/configmap_writer.go
anik120: @jpeeler Updated the PR to wrap the aggregated errors. Had to vendor in the `errors` package though. PTAL

jpeeler:
I rebased this locally and tested it with the happy path case and a missing annotation, error output looks like this:

Error: error loading manifests from directory: annotation validation failed, missing or empty values: operators.operatorframework.io.bundle.mediatype.v1.

However, as a sanity check I did a `go mod vendor` and that ended up removing a bunch of files in the vendor directory.

anik120:
> However, as a sanity check I did a `go mod vendor` and that ended up removing a bunch of files in the vendor directory.

@jpeeler I rebased locally, updated go to 1.13.4 and did `go mod vendor` and nothing got deleted for me. Not sure why your installation of go deleted all those files. In any case, I just vendored in the `pkg/errors` package and did not remove any package in my PR.

jpeeler:
/lgtm
I've run this through the e2e locally and it works fine.

PR #126, opened 2019-11-14
fix(e2e): update ginkgo configmap test

jpeeler:
Since this test isn't running in the gate (yet), there were a few things that needed updating.

(Eventually this test will be modified to use an image builder that doesn't require a daemon.)


jpeeler:
/hold
Don't really want to change the image pull policy, so just leaving this for now.

PR #128, opened 2019-11-20
Bug 1774630: Fix panic caused by incorrect bundle data

anik120:

Incorrect data in CSV causes a panic due to nil pointer
deference. This PR avoids the panic by checking for an error
after trying to retrive a CSV, and then logging the error if
there is one.

### DIFF
--- a/incorrectbundle/3scale-community-operator/3scale-community-operator.package.yaml
+++ b/incorrectbundle/3scale-community-operator/3scale-community-operator.package.yaml
@@ -1,0 +1,5 @@
+packageName: 3scale-community-operator
+channels:
+- currentCSV: 3scale-community-operator.v0.3.0
+  name: alpha
+
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -806,7 +806,11 @@ func (s *SQLLoader) updatePackageChannels(tx *sql.Tx, manifest registry.PackageM
 		// because this is caught by primary key of operatorbundle table
 
 		channelEntryCSV, err := s.getCSV(tx, c.CurrentCSVName)
-
+		if err != nil {
+			errs = append(errs, err)
+			break
+		}
+		
 		// check replaces
 		replaces, err := channelEntryCSV.GetReplaces()
 		if err != nil {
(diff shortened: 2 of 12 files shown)


anik120:
>  I'm wondering if we can think of an easy way to write a test for serializing invalid yaml to make sure we aren't missing this for any other files in the operator package?

That is a good point @kevinrizza. I'm looking into it.

pkg/sqlite/load.go
```diff
@@ -806,7 +806,11 @@ func (s *SQLLoader) updatePackageChannels(tx *sql.Tx, manifest registry.PackageM
 		// because this is caught by primary key of operatorbundle table
 
 		channelEntryCSV, err := s.getCSV(tx, c.CurrentCSVName)
-
```
gallettilance: `updatePackageChannels()` should only be called in the `opm registry add` flow. I'm curious as to why this function is being called in the `initializer` flow in this instance - if you look at `AddPackageChannels` it only calls this function when adding a packageName to the database fails. Could you investigate the root of that failure?

anik120: @gallettilance looks like when if there's a problem with unmarshaling the CSV, the CSV is not added to the [`operatorBundle` table](https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/migrations/000_init.go#L18) and the [`channel` table](https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/migrations/000_init.go#L28), i.e it is skipped. Therefore when we try to add a new [package](https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/migrations/000_init.go#L23) which references the `channel` table, the Foreign Key constraint fails, as a result adding a package fails, and therefore, it tries to update the channel.

anik120:
> > I'm wondering if we can think of an easy way to write a test for serializing invalid yaml to make sure we aren't missing this for any other files in the operator package?
> 
> That is a good point @kevinrizza. I'm looking into it.

@kevinrizza it looks like the easiest way to write a test for this is to add some incorrect bundle test data. Note that reusing the `manifests` we already have for testing would require some refactoring to [enable parsing](https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/csv.go#L50-L53) of the `spec` block of the CSV first.
I've added a test for incorrect operator bundle. PTAL when you get a chance. Thank you!

kevinrizza:
nice work @anik120 . As long as lance is fine with your responses this lgtm

incorrectbundle/3scale-community-operator/3scale-community-operator.package.yaml
```diff
@@ -0,0 +1,5 @@
+packageName: 3scale-community-operator
```
njhale: Can we put this in a `testdata` directory under the package being tested (`sqlite`)? I don't think we want to put bad test manifests at the root of the repo.

PR #134, opened 2019-12-09
Allow a single empty document at the beginning of a yaml file

### LINKED ISSUE #133: appregistry interprets whitespace followed by directives end mark differently than operator-courier
Consider a yaml file that begins like this (whitespace followed a directives end mark:
"

(dash dash dash ---)
"

(TL;DR although this is an edge case ,it really happened to a real company who was stymied  trying to run an operator on OpenShift :) ) 

The YAML spec is not quite clear on how this should be interpreted. In the Python3 standard yaml library, this will be interpreted as a single document. However, in k8s.io/apimachinery/pkg/util/yaml/decoder.go this will be treated as an empty document followed by a second document beginning at the directives end mark (---).

This leads to a scenario where a bundle can be validated by operator-courier and successfully upload but will fail when downloaded by the appregistry because the empty document will occlude the second document, resulting in a missing CRD in the bundle.

This in turn leads to a ClusterServiceVersion stuck in pending, with no clear error to the end user unless they are knowledgeable enough to look at pod logs in the openshift-marketplace and figure out what went wrong.

The solution is to reconcile the handling of the initial whitespace by operator-courier and appregistry. Either make operator-courier flag this as an error, even though standard Python3 treats it unambiguously, or make the appregistry handle the empty document without failing the bundle.

tmckayus:

This change allows the appregistry to handle a single empty document
at the beginning of a yaml file.

It's possible for operator-courier to validate a bundle with initial whitespace in yaml files
(accepted by Python3) and then have that bundle fail in appregistry. This change
reconciles that case, which is ambiguous in the yaml spec and handled differently
by Python and k8s.

Closes #133

**Reviewer Checklist**

### DIFF
--- a/bundles/prometheus.0.23.0_whitespace/manifests/alertmanager.crd.yaml
+++ b/bundles/prometheus.0.23.0_whitespace/manifests/alertmanager.crd.yaml
@@ -1,0 +1,2400 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: alertmanagers.monitoring.coreos.com
+spec:
+  group: monitoring.coreos.com
+  names:
+    kind: Alertmanager
+    plural: alertmanagers
+  scope: Namespaced
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          description: 'Specification of the desired behavior of the Alertmanager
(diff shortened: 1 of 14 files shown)


njhale:
Closing this out since it has be pending rebase for quite some time.

@tmckayus please feel free to reopen with rebased changes if you feel this is still important.

PR #138, opened 2019-12-18
Fix appr upstream dockerfile

gbenhaim:

- Update to golang 12, like we have in the other images.

- Copy grpc-health-probe binary from github, like we in the u/s builder
  image.

- Copy the go.mod and go.sum to the build container, since they are
  required for the compilation.

- Don't run the container as root, since it's not needed.

- Since the container doesn't run with root, a directory where the 1001
  user has write permissions should be created (this is where the db is
  created and the temporary files are downloaded).

  For that, change the base image to be alpine 3.10 - It's not possible
  to create directories and change their permissions in scratch
  containers, and copying the directory from the build container doesn't
  work.

The upstream Dockerfile is useful for testing during development.
The downstream Dockerfile can't be used outside of a build system, since it expects that repositories 
will be injected to it.

**Reviewer Checklist**

### DIFF
--- a/appr-registry.Dockerfile
+++ b/appr-registry.Dockerfile
@@ -1,28 +1,30 @@
-FROM golang:1.10-alpine as builder
+FROM golang:1.12-alpine AS builder
 
 RUN apk update && apk add sqlite build-base git mercurial
-WORKDIR /go/src/github.com/operator-framework/operator-registry
+WORKDIR /build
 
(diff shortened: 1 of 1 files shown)


njhale:
Closing out as this has been pending rebase for some time.

@gbenhaim please feel free to rebase and reopen if you're still interested.

PR #143, opened 2019-12-20
Bundle validate lib functions

kevinrizza:
This pull request adds bundle validation lib functions to validate the format of the image itself
It also includes an example implementation of these validation functions in `opm alpha bundle validate`

This change is intended to provide a common functional library for the operator-sdk and any other callers to validate that bundle images are in the prescribed format. Additionally, this pr provides a convenient function that understands how to shell out to container tooling and pull the contents of bundle images into a specified local directory (via `podman/docker save` and `pull`)

**Reviewer Checklist**


pkg/lib/bundle/generate.go
```diff
@@ -93,19 +98,29 @@ func GenerateFunc(directory, packageName, channels, channelDefault string, overw
...
+		}
+
+		filewithpath := filepath.Join(directory, item.Name())
+		fileblob, err := ioutil.ReadFile(filewithpath)
+		if err != nil {
+			return "", fmt.Errorf("Unable to read file %s in bundle", filewithpath)
+		}
+
+		var yamlfile map[string]interface{}
+		err = yaml.Unmarshal(fileblob, &yamlfile)
```
ecordell: 👍

pkg/lib/bundle/generate.go
```diff
@@ -114,13 +129,19 @@ func GetMediaType(directory string) (string, error) {
 	}
 
 	// Validate the file names to determine media type
-	for _, file := range files {
-		if file == "Chart.yaml" {
+	for fileName, fileBlob := range files {
+		// TODO: be more robust here, we should validate the format of helm charts
```
ecordell: I think the `Chart.yaml` filename is the only defining feature of a helm chart. I'll see if I can find something else to look for and update here if I do.

cmd/opm/alpha/bundle/validate.go
```diff
@@ -0,0 +1,64 @@
...
+
+func newBundleValidateCmd() *cobra.Command {
+	bundleValidateCmd := &cobra.Command{
+		Use:   "validate",
+		Short: "Validate bundle image",
+		Long: `The "opm alpha bundle validate" command will validate bundle image
+from a remote source to determine if its format and content information are
+accurate.
+
+		$ opm alpha bundle validate --tag quay.io/test/test-operator:latest \
```
njhale: nit: there is also an `Examples` field for this

pkg/lib/bundle/generate.go
ecordell: There's also a "yamlOrJson" decoder somewhere that we can use - don't see any reason to exclude json

pkg/lib/bundle/generate.go
joelanford: There's a helm library to validate a chart. Here's an example from SDK:

https://github.com/operator-framework/operator-sdk/blob/ba68281353e529aefadb0c86c74979dd2cf69f12/pkg/helm/watches/watches.go#L83-L85

kevinrizza:
@njhale @ecordell @joelanford updated based on latest feedback. Although I couldn't find that yamlorjson unmarshaler you were referring to Evan. Maybe you could point me to one?

pkg/lib/bundle/generate.go
```diff
@@ -33,6 +33,11 @@ type AnnotationMetadata struct {
 	Annotations map[string]string `yaml:"annotations"`
 }
 
+type ClusterServiceVersion struct {
```
joelanford: Unexport this since it is only used for unmarshalling?
```suggestion
type clusterServiceVersion struct {
```

pkg/lib/bundle/generate.go
```diff
@@ -33,6 +33,11 @@ type AnnotationMetadata struct {
...
+	// Type metadata.
+	Kind string `json:"Kind"`
```
joelanford: Kube resources use lowercase kind metadata keys right?

```suggestion
	Kind string `json:"kind"`
```

ecordell:
@kevinrizza https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/resolver/steps.go#L119-L127

pkg/lib/bundle/interfaces.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+
+	"github.com/sirupsen/logrus"
+)
+
+// BundleImageValidator provides a toolset for pulling and then validating bundle container images
+type BundleImageValidator interface {
+	PullBundleImage(string, string) error
+	ValidateBundle(string) error
```
joelanford: It might be helpful to users of the interface to document the usages of the method parameters either in go doc or by using named parameters.

pkg/lib/bundle/interfaces.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// NewImageValidator is a constructor that returns an ImageValidator
+func NewImageValidator(containerTool string, logger *logrus.Entry) BundleImageValidator {
+	return ImageValidator{
```
joelanford: Since this function returns the `BundleImageValidator` interface, does `ImageValidator` need to be exported? If it does need to be exported, would it make sense to return the concrete type and do a separate compile time interface assignment/check. E.g.
```go
var _ BundleImageValidator = &ImageValidator{} 
```

pkg/lib/bundle/interfaces.go
```diff
@@ -0,0 +1,21 @@
...
+		ImageReader: containertools.NewImageReader(containerTool, logger),
+		Logger:      logger,
```
joelanford: If `logger == nil`, would it make sense to set it to a null logger?

```go
logger := logrus.New()
logger.Out = ioutil.Discard
```

pkg/lib/bundle/validate.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+
+	log "github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
+)
+
+// ImageValidator is a struct implementation of the Indexer interface
+type ImageValidator struct {
+	ImageReader containertools.ImageReader
+	Logger      *log.Entry
```
joelanford: Since there's a constructor for this struct, do users need access to the struct fields? Can they be unexported to make the API surface smaller?

pkg/lib/bundle/validate.go
```diff
@@ -0,0 +1,147 @@
...
+// to a specified directory for further validation
+func (i ImageValidator) PullBundleImage(tag, dir string) error {
+	i.Logger.Debug("Pulling and unpacking container image")
+
+	err := i.ImageReader.GetImageData(tag, dir)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
joelanford: Nit: this can be simplified to
```go
return i.ImageReader.GetImageData(tag, dir)
```

pkg/lib/bundle/validate.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		formatErrors = append(formatErrors, fmt.Errorf("Unable to parse annotations.yaml file"))
+	}
+
+	for label, item := range annotations {
+		val, ok := fileAnnotations.Annotations[label]
+		if ok {
+			i.Logger.Debugf(`Found annotation "%s" with value "%s"`, label, val)
+		} else {
+			aErr := fmt.Errorf(`Missing annotation "%s"`, label)
```
joelanford: Nit: You can use `%q` to quote strings in format strings:
```suggestion
			aErr := fmt.Errorf("Missing annotation %q", label)
```

test/e2e/bundle_image_test.go
```diff
@@ -678,8 +678,11 @@ spec:
 }
 
 func buildContainer(tag, dockerfilePath, context string) {
+	err:= bundle.ValidateFunc(tag,builderCmd)
```
ecordell: nit: gofmt

pkg/lib/bundle/generate.go
```diff
@@ -93,34 +95,53 @@ func GenerateFunc(directory, packageName, channels, channelDefault string, overw
...
 	}
 
-	// Validate the file names to determine media type
-	for _, file := range files {
-		if file == "Chart.yaml" {
+	// Validate the files to determine media type
+	for _, fileName := range files {
+		// TODO: be more robust here, we should validate the format of helm charts
+		// instead of file name
+		fmt.Println(fileName)
```
estroz: Leftover from testing?

pkg/lib/bundle/validate.go
```diff
@@ -0,0 +1,142 @@
...
+
+// ValidateBundle takes a directory containing the contents of a bundle and validates
+// the format of that bundle for correctness
+func (i imageValidator) ValidateBundle(directory string) error {
+	var manifestsFound, metadataFound bool
+	var annotationsDir, manifestsDir string
+	var annotationErrors []error
+	var formatErrors []error
+
+	items, _ := ioutil.ReadDir(directory)
```
estroz: Any reason not to return this error?

pkg/lib/bundle/validate.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, item := range items {
+		if item.IsDir() {
+			switch s := item.Name(); s {
```
estroz: Direct matching seems brittle here since the manifests or metadata dirs can be children, unless I'm missing something. `strings.HasSuffix(s, strings.TrimSuffix(ManifestsDir, "/"))` would be a better comparison.

pkg/lib/bundle/validate.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		formatErrors = append(formatErrors, err)
+	}
+
+	// Validate annotations.yaml
+	annotationsFile, err := ioutil.ReadFile(filepath.Join(annotationsDir, AnnotationsFile))
+	if err != nil {
+		fmtErr := fmt.Errorf("Unable to read annotations.yaml file: %s", err.Error())
+		formatErrors = append(formatErrors, fmtErr)
+	}
```
estroz: This method should return here since `annotationsFile` is used below, unless the intent is to run through the method body with a potentially empty `[]byte` and append resulting errors below.

pkg/lib/bundle/validate.go
dinhxuanvu: The only potential error here is the directory doesn't exist. The logic below (between 39 to 59) will handle this case as it will notify users if we are unable to locate the directory for manifests and/or metadata.

dinhxuanvu:
@ecordell @joelanford @estroz Ready for final review

pkg/containertools/imagereader.go
```diff
@@ -58,16 +58,17 @@ func (b ImageLayerReader) GetImageData(image, outputDir string, opts ...GetImage
...
-	workingDir := options.WorkingDir
-	if workingDir == "" {
+	rootTarfile := filepath.Join(options.WorkingDir, "bundle.tar")
+
+	if options.WorkingDir == "" {
 		workingDir, err := ioutil.TempDir("./", "bundle_staging_")
 		if err != nil {
 			return err
 		}
 		defer os.RemoveAll(workingDir)
```
Bowenislandsong: should we be consistent and catch this removeall error too?

PR #146, opened 2020-01-06
Add bundle contents validation

exdx:
Add validation of bundle contents as part of the registry epic. Makes use of the https://github.com/operator-framework/api library for core validation logic. Meant to run along with the bundle image validation. Currently only CSV and CRD manifests are fully validated, other types are simply vetted as valid kube objects. 

**Reviewer Checklist**


exdx:
One thing I've noticed is the api validation library throws a fatal error if the CSV `installMode` is not explicitly set. I think this is the desired behavior, but all our example manifests including the CSVs in this repo do not have an install mode set. The previous PR on bundle image validation included a valid prometheus CSV that was considered invalid by the api library.

exdx:
I think we need to add `installMode` to the example bundles in this repo (and probably elsewhere) to ensure that our validation passes for our own examples per @ecordell

exdx:
After meeting with the sdk team we want to decouple the image validation and contents validation so that the SDK can import the operator-framework/API directly. 
We should

- Make a validate bundle command that does not validate image contents (for SDK)
- Make a validate bundle command that validates image and image contents (for opm alpha)
cc @kevinrizza

pkg/lib/bundle/validate.go
```diff
@@ -132,11 +170,159 @@ func (i imageValidator) ValidateBundle(directory string) error {
...
+	supportedTypes := map[string]string{
+		csvKind:                "",
+		crdKind:                "",
+		secretKind:             "",
+		clusterRoleKind:        "",
+		clusterRoleBindingKind: "",
+		serviceKind:            "",
+		serviceAccountKind:     "",
+		roleKind:               "",
+		roleBindingKind:        "",
```
ecordell: This will also need to include the prometheus types, see: https://github.com/operator-framework/operator-lifecycle-manager/pull/1227

pkg/lib/bundle/validate_test.go
```diff
@@ -73,10 +73,74 @@ func TestValidateBundle_InvalidRegistryVersion(t *testing.T) {
...
+	}
+
+	var table = []struct {
+		description string
+		mediaType   string
+		directory   string
+		numErrors   int
+		errString   string
+	}{
+		{
```
ecordell: a test case for "no csv"? It might change in the future, but for now that's a requirement.

pkg/lib/bundle/validate_test.go
```diff
@@ -73,10 +73,74 @@ func TestValidateBundle_InvalidRegistryVersion(t *testing.T) {
...
+		var validationError ValidationError
+		isValidationErr := errors.As(err, &validationError)
+		require.True(t, isValidationErr)
+		require.Len(t, validationError.Errors, tt.numErrors, table[i].description)
+		if len(validationError.Errors) > 0 {
+			e := validationError.Errors[0]
+			require.Contains(t, e.Error(), tt.errString)
+		}
+	}
 }
```
ecordell: I can't comment on the files - but we should `.gitignore` the ` pkg/sqlite/migrations/1048603298946505393.db-journal` files

pkg/lib/bundle/validate.go
kevinrizza: can we export these somehow rather than hardcoding them?

kevinrizza: even if that is in the API package and not in the registry? the more things like this that we start adding everywhere the harder it is going to be to change them.

pkg/lib/bundle/validate_test.go
dinhxuanvu: I understand this case but not sure if it is helpful in this specific situation. If there is no CSV, it is considered `PlainType` which we don't validate the contents. So if it is not `RegistryV1` format, the `ValidateBundleContent` func will just return `nil`.

pkg/lib/bundle/validate.go
dinhxuanvu: At the moment, there is nothing from OLM or registry to export. For OLM, we use a switch statement to handle all supported types. What OLM supports is a selected set of kube objects so there is no standard or logic for how we pick and choose.
I can create some struct with supported types and a function that validates if the type is supported. However, it is still some sort of "hardcoded".

kevinrizza: sure yeah, I undestand that. i mostly just mean now we have to keep track of this in two places and when the list updates remember to update them in both. what happens when the API project wants its own version of this validation for static analysis outside the scope of pulling down the image? Then the list is in 3 places.

not sure if we can solve that here right now, but i think this is a problem that we are facing sooner rather than later for a lot of things that are going to go across all of these projects. what is our import relationship between all of these repositories? or are we trying to explicitly avoid that completely by reimplementing the apis everywhere?

dinhxuanvu: I agree. There is a need to consolidate some agreements between OLM and registry. I would expect this change to begin with OLM. OLM needs to define a set of kube objects that it is supporting in a way that registry can import and then registry will enforce it.

pkg/lib/bundle/validate.go
```diff
@@ -28,12 +58,20 @@ func (i imageValidator) PullBundleImage(imageTag, directory string) error {
...
+// Outputs:
+// error: ValidattionError which contains a list of errors
+func (i imageValidator) ValidateBundleFormat(directory string) error {
 	var manifestsFound, metadataFound bool
 	var annotationsDir, manifestsDir string
-	var annotationErrors []error
-	var formatErrors []error
+	var validationErrors []error
 
 	items, _ := ioutil.ReadDir(directory)
```
exdx: should we really drop the error here? there could be a permissions issue or something else going on when accessing the directory, and none of the subsequent validation would make any sense. Same for L225 with the content validation.

pkg/lib/bundle/validate.go
```diff
@@ -132,11 +170,159 @@ func (i imageValidator) ValidateBundle(directory string) error {
...
+			validationErrors = append(validationErrors, fmt.Errorf("Unable to read file %s in supported types", fileWithPath))
+			continue
+		}
+
+		dec := k8syaml.NewYAMLOrJSONDecoder(strings.NewReader(string(data)), 30)
+		k8sFile := &unstructured.Unstructured{}
+		if err := dec.Decode(k8sFile); err == nil {
+			unstObjs = append(unstObjs, k8sFile)
+			kind := k8sFile.GetObjectKind().GroupVersionKind().Kind
+			i.logger.Debugf(`Validating file "%s" with type "%s"`, item.Name(), kind)
```
exdx: Maybe it makes sense here to log the GVK in its entirety along with the filename for clarity, something like `Validating %s from file %s, gvk.String(), item.Name`?

pkg/lib/bundle/validate.go
dinhxuanvu: We can certainly acknowledge the error here. The reason I keep I ignore the error is because we inspect the directory that is actually created by this problem. Users supply the image tag and the program will pull the image down and unpack the contents into directories that it creates. So there shouldn't be a permission issue here. This function is meant to be use in conjunction with other functions to create a validation flow. However, I understand from standalone API standpoint, maybe we should return every error that occurs to avoid confusions.

pkg/lib/bundle/validate.go
```diff
@@ -132,11 +170,159 @@ func (i imageValidator) ValidateBundle(directory string) error {
...
+		fileWithPath := filepath.Join(manifestDir, item.Name())
+		data, err := ioutil.ReadFile(fileWithPath)
+		if err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("Unable to read file %s in supported types", fileWithPath))
+			continue
+		}
+
+		dec := k8syaml.NewYAMLOrJSONDecoder(strings.NewReader(string(data)), 30)
+		k8sFile := &unstructured.Unstructured{}
+		if err := dec.Decode(k8sFile); err == nil {
```
exdx: why don't we invert this to the typical `if err != nil` and have 
```
err := dec.Decode(k8sFile)
if err != nil {
  continue
}
```
since we are inside the for loop. that avoids some nesting

pkg/lib/bundle/validate.go
```diff
@@ -132,11 +170,159 @@ func (i imageValidator) ValidateBundle(directory string) error {
...
+		&parsedMeta.ObjectMeta,
+		false,
+		func(s string, prefix bool) []string {
+			return nil
+		},
+		field.NewPath("metadata"),
+	)
+
+	if len(errs) > 0 {
+		return fmt.Errorf("error validating object metadata: %s. %v", errs, parsedMeta)
```
exdx: should unpack the errs type `fields.ErrorList` with something like `errs.ToAggregate()` so it looks cleaner to the end user?

cmd/opm/alpha/bundle/validate.go
```diff
@@ -54,9 +55,14 @@ func validateFunc(cmd *cobra.Command, args []string) error {
...
-	logger.Info("Unpacked image layers, validating bundle image contents")
+	logger.Info("Unpacked image layers, validating bundle image format & contents")
 
-	err = imageValidator.ValidateBundle(dir)
+	err = imageValidator.ValidateBundleFormat(dir)
+	if err != nil {
+		return err
+	}
+
+	err = imageValidator.ValidateBundleFormat(filepath.Join(dir, bundle.ManifestsDir))
```
estroz: Should this be `ValidateBundleContent`?

dinhxuanvu:
@kevinrizza @ecordell Would like to have the final approval from you :). Thanks

pkg/lib/bundle/validate.go
```diff
@@ -132,11 +162,155 @@ func (i imageValidator) ValidateBundle(directory string) error {
...
+			validationErrors = append(validationErrors, fmt.Errorf("%s is not supported type for registryV1 bundle: %s", gvk.Kind, fileWithPath))
+			continue
+		}
+
+		if gvk.Kind == CSVKind {
+			csv := &v1.ClusterServiceVersion{}
+			err := runtime.DefaultUnstructuredConverter.FromUnstructured(k8sFile.Object, csv)
+			if err != nil {
+				validationErrors = append(validationErrors, err)
+			}
```
estroz: A `continue` here will avoid duplicating errors from the validator below, since the unstructured converter will not populate `csv` if an error occurs.
```suggestion
				continue
			}
```

pkg/lib/bundle/validate.go
```diff
@@ -132,11 +162,155 @@ func (i imageValidator) ValidateBundle(directory string) error {
...
+				for _, err := range results[0].Errors {
+					validationErrors = append(validationErrors, err)
+				}
+			}
+		} else if gvk.Kind == CRDKind {
+			crd := &v1beta1.CustomResourceDefinition{}
+			err := runtime.DefaultUnstructuredConverter.FromUnstructured(k8sFile.Object, crd)
+			if err != nil {
+				validationErrors = append(validationErrors, err)
+			}
```
estroz: Same as above.
```suggestion
				continue
			}
```

pkg/lib/bundle/validate.go
dinhxuanvu: Ack

pkg/lib/bundle/validate.go
dinhxuanvu: Ack

dinhxuanvu: Resolved.

pkg/lib/bundle/validate.go
dinhxuanvu: Resolved.

PR #148, opened 2020-01-14
Bug 1790785: feat(appr): exit build early if no manifests were downloaded

ecordell:
Fail out if we don't find any manifests.

This is to improve UX in oc adm catalog build.

**Reviewer Checklist**

### DIFF
--- a/pkg/appregistry/builder.go
+++ b/pkg/appregistry/builder.go
@@ -6,6 +6,7 @@ import (
 	"compress/gzip"
 	"context"
 	"database/sql"
+	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


pkg/appregistry/builder.go
```diff
@@ -212,3 +218,11 @@ func BuildLayer(directory string) (string, error) {
 
 	return archive.Name(), nil
 }
+
+func hasManifests(path string) bool {
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return false
+	}
+	return len(files) > 1
```
awgreene: Should this be `>=1`?

PR #151, opened 2020-01-20
[WIP] Add custom labels to bundle

gallettilance:
Users can add custom labels through the opm CLI while generating or
building their bundle.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/bundle/build.go
+++ b/cmd/opm/alpha/bundle/build.go
@@ -14,6 +14,7 @@ var (
 	channelsArgs       string
 	channelDefaultArgs string
 	overwriteArgs      bool
+	labelsArgs         []string
 )
 
 // newBundleBuildCmd returns a command that will build operator bundle image.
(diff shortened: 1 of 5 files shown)


dinhxuanvu:
@estroz FYI

kevinrizza:
I think we shouldn't do this. If we open the gates to "we can generate whatever kind of dockefile you want" into the cli itself it is going to become a nightmare to maintain. If users want to add custom labels or extra files or use special things in `build` then that is the purpose of `--generate`.

estroz:
Initially I was a bit torn on this, because I agree with @kevinrizza's opinion, but I also think a user modifying metadata or Dockerfile after generating them on-disk may not know (how) to modify either correctly, ex. they may modify one but forget to modify the other.

However, the whole point of `bundle generate` is to allow users to interact with metadata/Dockerfile in a non-standard way to which we say "you can do this, but at your own risk". I think as long as a `bundle validate` command exists that will tell a user what they screwed up, we can rely on `bundle generate` for all non-standard things and avoid feature creep.

gallettilance:
After some discussion, we've decided not to move forward with adding this flag in favor of letting users add labels by editing the Dockerfile. We will revisit designs for the abstractions we want to provide via the CLI.

gallettilance:
After some discussion, we've decided not to move forward with adding this flag in favor of letting users add labels by editing the Dockerfile. We will revisit designs for the abstractions we want to provide via the CLI.

PR #152, opened 2020-01-21
[wip] switch to ql

ecordell:
Switches to QL from Sqlite for evaluation purposes. See https://godoc.org/modernc.org/ql

Current status:

- [x] Load and Query working, tests passing
- [x] Fix tests with `-race` (currently race detector detects a race in ql)
- [ ] Refactoring 
- [ ] Add indexes
- [ ] Make sure there are alternative checks for all of the previous foreign keys

There are currently two issues in ql making some of these queries pretty unweildy right now:

 - You can't do multiple joins without a subquery _and_ renaming each field at every join layer. This makes a couple of the queries unreasonably large.
 - You can't use `group by` in a `min()` query. `min()` may also not do the right thing in some cases, but I need to investigate a bit more to be definitive.

Both of these are things we could potentially fix upstream if they become an issue. We could also fix them by denormalizing some more.

Also note that this is just one option - the other is based on bbolt, which will be a more substantial architectural change than this one.

We need to produce builds for macOS and can't use CGO for mac targets in our build environment.

**Reviewer Checklist**


ecordell:
after another spike on bbolt, we will not switch to ql now.

PR #154, opened 2020-01-22
Index maintainers can export a package from an index

gallettilance:

**Reviewer Checklist**


pkg/lib/indexer/interfaces.go
```diff
@@ -47,3 +47,20 @@ func NewIndexDeleter(containerTool string, logger *logrus.Entry) IndexDeleter {
 		Logger:              logger,
 	}
 }
+
+//counterfeiter:generate . IndexDeleter
+type IndexExporter interface {
+	ExportFromIndex(ExportFromIndexRequest) error
+}
+
+// NewIndexDeleter is a constructor that returns an IndexDeleter
```
exdx: nit: returns an IndexExporter

pkg/lib/indexer/interfaces.go
```diff
@@ -47,3 +47,20 @@ func NewIndexDeleter(containerTool string, logger *logrus.Entry) IndexDeleter {
 		Logger:              logger,
 	}
 }
+
+//counterfeiter:generate . IndexDeleter
```
exdx: nit: I think this codegen should point to an IndexExporter as well

exdx:
I’m a little concerned about tying more code to sqlite just as we are trying to actively move to another backend -- but we have no choice, so we can just update the code later on no big deal.

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +291,119 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+
+	bundles, err := dbQuerier.GetImagesForPackage(context.TODO(), request.Operator)
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+
+	for _, bundleImage := range bundles {
+		loader := sqlite.NewSQLExporterForImage(dbQuerier, bundleImage, request.DownloadPath, request.ContainerTool)
```
ecordell: It's surprising to me that this is in the sqlite package - based on how things are layed out today, I would expect any new queries needed (if any) to be added in the Query interface, and then an `export` package in `lib` that implements the exporting

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +291,119 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+	if err != nil {
+		return err
+	}
+
+	channelList, err := dbQuerier.ListChannels(context.TODO(), request.Operator)
+	if err != nil {
+		return err
+	}
+
+	channels := []pregistry.PackageChannel{}
```
ecordell: typo: `pregistry`?

pkg/sqlite/image.go
```diff
@@ -239,3 +240,45 @@ func translateAnnotationsIntoPackage(annotations *registry.AnnotationsFile, csv
...
+		directory:     directory,
+		containerTool: containerTool,
+	}
+}
+
+func (i *ImageExporter) Export() error {
+
+	log := logrus.WithField("img", i.image)
+
+	version, err := i.store.GetBundleVersion(context.TODO(), i.image)
```
ecordell: What if there is no version? Fallback to the bundle name?

Wouldn't it be easier to just use i.image (appropriately munged to be a filename) as the folder? The loader won't care what the folders are named.

pkg/sqlite/image.go
```diff
@@ -239,3 +240,45 @@ func translateAnnotationsIntoPackage(annotations *registry.AnnotationsFile, csv
 
 	return manifest, nil
 }
+
+// ImageLoader loads a bundle image of resources into the database
```
ecordell: nit: comment is wrong

pkg/sqlite/query.go
```diff
@@ -541,3 +540,104 @@ func (s *SQLQuerier) GetApisForEntry(ctx context.Context, entryId int64) (provid
...
+			return "", err
+		}
+	}
+	if version.Valid {
+		return version.String, nil
+	}
+	return "", fmt.Errorf("bundle %s not found", image)
+}
+
+func (s *SQLQuerier) GetImagesForPackage(ctx context.Context, pkgName string) ([]string, error) {
```
ecordell: Where is this used?

Isn't the field called `bundlepath`, not `image`?

pkg/sqlite/query.go
```diff
@@ -541,3 +540,104 @@ func (s *SQLQuerier) GetApisForEntry(ctx context.Context, entryId int64) (provid
...
+		}
+		if imgName.Valid {
+			images = append(images, imgName.String)
+		}
+	}
+	return images, nil
+}
+
+func (s *SQLQuerier) GetDefaultChannelForPackage(ctx context.Context, pkgName string) (string, error) {
+	query := `SELECT DISTINCT default_channel FROM package WHERE name=?`
```
ecordell: LIMIT 1

pkg/sqlite/query.go
```diff
@@ -541,3 +540,104 @@ func (s *SQLQuerier) GetApisForEntry(ctx context.Context, entryId int64) (provid
...
+			return nil, err
+		}
+		if chName.Valid {
+			channels = append(channels, chName.String)
+		}
+	}
+	return channels, nil
+}
+
+func (s *SQLQuerier) GetCurrentCSVForChannel(ctx context.Context, pkgName, channel string) (string, error) {
```
ecordell: This just gets the name. The existing `GetBundle` will give you the full bundle contents (if it exists)

pkg/lib/indexer/indexer.go
exdx: No I think this is intentionally imported as `pregistry` to avoid collision with the overall `registry` package and points to `"github.com/operator-framework/operator-registry/pkg/registry"`

exdx:
Added a long description of the export command as a form of documentation. Please let me know if anything looks amiss. 

```
Export an operator from an index image into the appregistry format. 

This command will take an index image (specified by the --index option), parse it for the given operator (set by the --operator option) and export the operator metadata into an appregistry compliant format (a package.yaml file). 
This command requires access to docker or podman to complete successfully.

Note: the appregistry format is being deprecated in favor of the new index image and image bundle format. 
```

pkg/sqlite/query.go
gallettilance: I only need the name to generate the package.yaml

pkg/sqlite/query.go
gallettilance: Good point - I'll change the name

pkg/lib/bundle/exporter.go
```diff
@@ -0,0 +1,62 @@
...
+	if folderName == "" {
+		// operator-registry does not care about the folder name
+		folderName = i.image
+	}
+	workingDir := filepath.Join(i.directory, folderName)
+	err = os.Mkdir(workingDir, 0777)
+	if err != nil {
+		return err
+	}
+	tmpDir, err := ioutil.TempDir("./", "bundle_tmp")
```
dinhxuanvu: Missing error handling.

cmd/opm/index/export.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"k8s.io/kubectl/pkg/util/templates"
+
+	"github.com/operator-framework/operator-registry/pkg/lib/indexer"
+)
+
+var exportLong = templates.LongDesc(`
```
dinhxuanvu: This is new command so I think we need a doc for this.

exdx: where could we add a doc for this?

exdx: looks like in here would be good https://github.com/operator-framework/operator-registry/pull/149/files#diff-04c6e90faac2675aa89e2176d2eec7d8R65

dinhxuanvu: Yes I thought Kevin's PR has been merged by now but it is not. We should merge it given it is just a doc PR. Then we can add the information from this PR into that doc. PR: https://github.com/operator-framework/operator-registry/pull/149

dinhxuanvu: The #149 PR has been approved and hopefully merged soon. Then, the doc for this command can be added to that doc.

pkg/lib/bundle/exporter.go
exdx: should we `defer os.RemoveAll()` the tmpDir like we did before?

pkg/lib/bundle/utils.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+func copyDirectory(scrDir, dest string) error {
```
ecordell: should we just import https://github.com/gotestyourself/gotest.tools/blob/master/fs/ops.go ?

pkg/lib/indexer/indexer.go
```diff
@@ -1,45 +1,54 @@
...
 	"io"
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 
 	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
 	"github.com/operator-framework/operator-registry/pkg/lib/registry"
+	pregistry "github.com/operator-framework/operator-registry/pkg/registry"
```
ecordell: it's not for this PR to fix, but this module layout is confusing

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +292,182 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+	}
+
+	err = generatePackageYaml(databaseFile, request.Operator, request.DownloadPath)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func getBundlesToExport(databaseFile, packageName string) ([]string, error) {
```
ecordell: is there a reason that these helpers are opening/closing the file? can we pass in the db reference or the querier instead of the filename?

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +292,182 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+
+	// set a temp directory
+	workingDir, err := ioutil.TempDir("./", "index_tmp")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(workingDir)
+
+	// Pull the fromIndex
+	if request.Index != "" {
```
ecordell: I think `ExportFromIndex` would be more readable if this block were factored out

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +292,182 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+		if err := exporter.Export(); err != nil {
+			err = fmt.Errorf("error exporting bundle from image: %s", err)
+			errs = append(errs, err)
+		}
+	}
+	err = db.Close()
+	if err != nil {
+		errs = append(errs, err)
+		return utilerrors.NewAggregate(errs)
+	}
```
ecordell: Does this need to handle bundles (non-image) as well, or are we sure packages will be migrated to all-image?

exdx:
I added two tests for the refactored functions and also pulled some logic out of `ExportFromIndex` per @ecordell's suggestion. Tests pass so we have some coverage. One test queries the DB directly and we didn't have time to stub it out, so I had to check in an index sqlite DB into the code. Obviously this is not ideal but in the interest of test coverage I thought it would be ok - sqlite is going away in the near future regardless. Feel free to refactor further or revert the commit as needed.

pkg/lib/indexer/indexer.go
gallettilance: IMO it doesn't make sense to support non-image bundles since these cannot create an index (as the opm cli is currently defined). In the future, if we support inserting non-image bundles into an index then we will probably need to add non-image bundle support for exporting.

pkg/lib/bundle/utils.go
gallettilance: I didn't really want to add a dependency for copying files around and couldn't find a safe/stable repo to rely on. If you think this one works I'm happy to add it

pkg/lib/indexer/indexer.go
gallettilance: agreed

pkg/lib/indexer/indexer.go
gallettilance: That's what I did at first but changed it to make testing easier

pkg/lib/bundle/utils.go
Bowenislandsong: Uhh, can we just use code from https://stackoverflow.com/questions/51779243/copy-a-folder-in-go ?

pkg/lib/indexer/indexer.go
ecordell: It depends on how they will be bootstrapped, correct?

I can take a database built with `initializer` and then pass it to `opm` to add another entry to it.

pkg/lib/indexer/indexer.go
ecordell: Can you explain how this simplifies testing?

pkg/lib/bundle/utils.go
```diff
@@ -0,0 +1,110 @@
...
+			if err := copySymLink(sourcePath, destPath); err != nil {
+				return err
+			}
+		default:
+			if err := copy(sourcePath, destPath); err != nil {
+				return err
+			}
+		}
+
+		if err := os.Lchown(destPath, int(stat.Uid), int(stat.Gid)); err != nil {
```
ecordell: uid/gid won't work for windows, which is failing the windows build now.

`https://github.com/otiai10/copy` is already a dependency of operator-registry, can you see if that works instead of this package?

gallettilance: oh good catch! I've seen the build pass with `https://github.com/otiai10/copy` so I'll change that now

cmd/opm/index/export.go
gallettilance: I updated the docs but I'd like to keep this long desc to increase visibility on the fact that we are deprecating appregistry.

pkg/lib/bundle/exporter.go
```diff
@@ -0,0 +1,55 @@
+package bundle
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/operator-framework/operator-registry/pkg/containertools"
```
ecordell: nit: local imports in a third block

pkg/sqlite/query.go
```diff
@@ -541,3 +540,106 @@ func (s *SQLQuerier) GetApisForEntry(ctx context.Context, entryId int64) (provid
 
 	return
 }
+
+func (s *SQLQuerier) GetBundleVersion(ctx context.Context, image string) (string, error) {
```
ecordell: These need tests

pkg/lib/indexer/indexer_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/ghodss/yaml"
+	"io/ioutil"
+	"os"
+	"reflect"
+	"sort"
+	"testing"
+
+	pregistry "github.com/operator-framework/operator-registry/pkg/registry"
+)
+
```
ecordell: Can we have a test that actually extracts bundles to disk?

cmd/opm/index/export.go
dinhxuanvu: I think that is fine.

pkg/lib/indexer/indexer.go
gallettilance: No doubt that this is possible but this is not how we want to or will bootstrap them

pkg/lib/bundle/exporter.go
```diff
@@ -0,0 +1,56 @@
...
+	defer os.RemoveAll(tmpDir)
+
+	// Pull the image and get the manifests
+	reader := containertools.NewImageReader(i.containerTool, log)
+
+	err = reader.GetImageData(i.image, tmpDir)
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(i.directory); err != nil {
```
benluddy: The check-and-create can be replaced with https://golang.org/pkg/os/#MkdirAll, which is analogous to `mkdir -p`.

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +292,190 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+	bundles, err := getBundlesToExport(databaseFile, request.Package)
+	if err != nil {
+		return err
+	}
+	i.Logger.Infof("Preparing to pull bundles %+q", bundles)
+
+	// Creating downloadPath dir
+	if _, err := os.Stat(request.DownloadPath); err != nil {
+		if os.IsNotExist(err) {
+			err = os.Mkdir(request.DownloadPath, 0777)
```
benluddy: Same here, can use `MkdirAll` for this.

pkg/lib/indexer/indexer.go
```diff
@@ -283,3 +292,190 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
...
+	}
+
+	err = generatePackageYaml(databaseFile, request.Package, request.DownloadPath)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func (i ImageIndexer) WriteIndexDBFile(index, workingDir, databaseFile string) (string, error) {
```
benluddy: It's a little confusing to have two modes of operation based on the value of `index`, one of which doesn't write a file, and the other doesn't use the value of `databaseFile`. The behavior of this function might be clearer if it were narrowed down to lines 381-398 and only accepted `index` and `workingDir`. Then the caller can decide what to do if it doesn't have a value for `index`.

docs/design/opm-tooling.md
```diff
@@ -93,6 +93,41 @@ Like `opm registry rm`, this command will remove all versions an entire operator
 
 This will result in the tagged container image `quay.io/operator-framework/monitoring-index:1.0.2` with a registry that no longer contains the `prometheus` operator at all.
 
+#### export
+
+`opm index export` will export a package from an index into a directory. The format of this directory will match the appregistry manifest format: containing all versions of the package in the index along with a `package.yaml` file. This command takes an `--index` flag that points to an index image, a `--package` flag that states a package name, an optional `--download-folder` as the export location (default is `./downloaded`), and just as the other index commands it takes a `--container-tool` flag. Ex:
```
dinhxuanvu: You mean "an index image". We have been unclear about "index" and "index image". I feel in this case you meant "index image". Is that right?

pkg/lib/indexer/indexer_test.go
```diff
@@ -0,0 +1,61 @@
...
+func TestGeneratePackageYaml(t *testing.T) {
+	err := generatePackageYaml("./test/bundles.db", "etcd", ".")
+	if err != nil {
+		t.Fatalf("writing package.yaml: %s", err)
+	}
+
+	var expected pregistry.PackageManifest
+	expectedResults, _  := ioutil.ReadFile("./test/package.yaml")
+	_ = yaml.Unmarshal(expectedResults, expected)
```
benluddy: I'm not sure it's safe to ignore errors from `Unmarshal`, even in a test. What if the decoder encounters an error after partially populating `expected`?

pkg/sqlite/load.go
```diff
@@ -820,7 +820,7 @@ func (s *SQLLoader) updatePackageChannels(tx *sql.Tx, manifest registry.PackageM
 			errs = append(errs, err)
 			break
 		}
-		
+
 		// check replaces
 		replaces, err := channelEntryCSV.GetReplaces()
 		if err != nil {
```
dinhxuanvu: I think the `.db-journal` files shouldn't be there because I updated the gitignore to ignore those. You should rebase against master.

pkg/lib/indexer/indexer.go
exdx: I agree this function is confusing, I just pulled the code out of the the main `ExportFromIndex()` function into this to make it more clear. I started to refactor it but would like to write a test for it first to ensure behavior, it's sort of tricky.

opm-test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
exdx: I think we should `set -euxo pipefail` here
https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

pkg/sqlite/directory_test.go
```diff
@@ -159,21 +185,16 @@ func TestQuerierForDirectory(t *testing.T) {
 
 	etcdChannelEntriesThatProvide, err := store.GetChannelEntriesThatProvide(context.TODO(), "etcd.database.coreos.com", "v1beta2", "EtcdCluster")
 	require.ElementsMatch(t, []*registry.ChannelEntry{
-		{"etcd", "alpha", "etcdoperator.v0.6.1", ""},
-		{"etcd", "alpha", "etcdoperator.v0.9.0", "etcdoperator.v0.6.1"},
+		{"etcd", "alpha", "etcdoperator.v0.9.0", ""},
```
ecordell: What happened here? where's 0.6.1?

pkg/sqlite/directory_test.go
```diff
@@ -89,9 +89,45 @@ func TestQuerierForDirectory(t *testing.T) {
...
 
-	loader := NewSQLLoaderForDirectory(load, "../../manifests")
+	loader := NewSQLLoaderForDirectory(load, "../../manifests/strimzi-kafka-operator/")
 	require.NoError(t, loader.Populate())
 
+	image := "quay.io/test/"
+	etcdFirstVersion := &ImageLoader{
+		store:     loader.store,
+		image:     image + "etcd.0.9.0",
+		directory: "../../bundles/etcd.0.9.0",
```
gallettilance: For some reason these csvs are different than the ones in the `manifest` folder

PR #159, opened 2020-01-23
Add semver graph update option to `opm`

kevinrizza:

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -18,18 +18,20 @@ import (
 
 // ImageLoader loads a bundle image of resources into the database
 type ImageLoader struct {
-	store     registry.Load
-	image     string
-	directory string
+	store         registry.Load
+	image         string
+	directory     string
 	containerTool string
+	mode          registry.Mode
 }
 
-func NewSQLLoaderForImage(store registry.Load, image, containerTool string) *ImageLoader {
+func NewSQLLoaderForImage(store registry.Load, image, containerTool string, mode registry.Mode) *ImageLoader {
 	return &ImageLoader{
-		store:     store,
-		image:     image,
-		directory: "",
+		store:         store,
+		image:         image,
+		directory:     "",
 		containerTool: containerTool,
+		mode:          mode,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


pkg/sqlite/image.go
```diff
@@ -141,6 +143,7 @@ func (i *ImageLoader) loadManifests(manifests string, annotationsFile *registry.
 		return fmt.Errorf("error getting csv from bundle %s: %s", bundle.Name, err)
 	}
 
+	// This isn't necessarily right anymore, because the head of the channels may not be what we are adding
 	packageManifest, err := translateAnnotationsIntoPackage(annotationsFile, bcsv)
```
ecordell: this just shouldn't be called here, right? it should only be called after we know we're adding the head of a channel

pkg/sqlite/image.go
```diff
@@ -211,7 +214,7 @@ func (i *ImageLoader) loadOperatorBundle(manifest registry.PackageManifest, bund
 		return nil
 	}
 
-	if err := i.store.AddBundlePackageChannels(manifest, bundle); err != nil {
+	if err := i.store.AddBundlePackageChannels(manifest, bundle, i.mode); err != nil {
```
ecordell: I sort of expected to see different function calls here based on the mode - seems like there will be a ton of conditional logic based on the mode in this otherwise.

pkg/sqlite/image.go
kevinrizza: I suppose we could make that condition and then build a different object to pass into a totally different function, but I was attempting to make as few changes as possible here. Right now the concept of the synthetic manifest is used to define the set of channel metadata anyway and not really a 1:1 translation of "this manifest is what we are judiciously applying to the cluster".

I think if we wanted to do that it would require a slightly more extensive refactor to what we do for adding bundle images now

pkg/sqlite/image.go
kevinrizza: Theres very little mixed logic, I just rely on the bundle add as is and then branch to a totally different logic path when processing channel data based on mode. I think like above it we wanted to do clean totally separate implementations per mode it would require some extensive refactoring of the calls to the data layer -- which I was explicitly attempting to avoid based on feedback from earlier this week. I do agree it would be cleaner though

PR #161, opened 2020-01-24
Bug 1795658: Remove dependency on sqlite from pkg/configmap.

benluddy:
The package github.com/mattn/go-sqlite3 needs CGO enabled to
build. Since pkg/configmap has both direct and transitive dependencies
on go-sqlite3, anything importing pkg/configmap will also need to
enable CGO.


njhale:
nice!

PR #162, opened 2020-01-27
Bug 1795237: Revert "fix(ci): build opm statically in CI"

ecordell:
Reverts operator-framework/operator-registry#160


kevinrizza:
/hold

I think we actually need to create a bugzilla for this.

PR #164, opened 2020-01-28
Update operator-registry OWNERS file

dinhxuanvu:
Update the OWNERS to update approvers and reviewers list.

**Reviewer Checklist**

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,11 +2,17 @@
 approvers:
   - ecordell
   - njhale
-  - jpeeler
-  - dinhxuanvu 
+  - dinhxuanvu
+  - kevinrizza
 # review == this code is good /lgtm
 reviewers:
   - ecordell
   - njhale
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -2,11 +2,17 @@
 approvers:
   - ecordell
   - njhale
-  - jpeeler
-  - dinhxuanvu 
+  - dinhxuanvu
```
njhale: I wonder if we should just add everyone to `approvers` -- @ecordell thoughts?

dinhxuanvu: My methodology behind this is to pick 4 top code contributors as approvers. I thought it would be beneficial to cap number of approvers so not every PR gets approved right away and one lgtm will allow the PR to get merged too quickly without at least one more person has a chance to review and lgtm it. Just a guarding measure. However, if we want to add everyone, I'm fine with it.

PR #170, opened 2020-01-31
Adding opm e2e test

gallettilance:

**Reviewer Checklist**


gallettilance:
This works with the current images I'm using. What images should we use instead and how can I pass the push credentials so that this test can push to a registry? Or do we want to just be able to run this locally with our own creds?

njhale:
Merging this to non-master branch so travis has access to necessary config.

PR #172, opened 2020-02-04
Update operator-bundle to clarify the bundle format directories

dinhxuanvu:
1. Remove the bundle manifest format reference to README to avoid
duplicate information.
2. Add a new section for Bundle Manifest Format to include format
information
3. Add notes regarding the manifests and metadata constants
4. Clean up the bundle generate section to make directory info
clearer and more intuitive.

**Reviewer Checklist**


docs/design/operator-bundle.md
```diff
@@ -44,18 +62,7 @@ annotations:
...
-$ tree test
-test
-├── testbackup.crd.yaml
-├── testcluster.crd.yaml
-├── testoperator.v0.1.0.clusterserviceversion.yaml
-├── testrestore.crd.yaml
-└── metadata
-    └── annotations.yaml
-```
+* Currently, the annotations for bundle manifests and metadata are `manifests/` and `metadata/` and those values are immutable as they are constants.
```
benluddy: Can you clarify this statement? As a user, I can change the metadata/manifests paths in the `annotations.yaml` that I put into my bundle image, so why shouldn't I?

My understanding is that the registry server currently won't respect the value of either annotation (instead using the hardcoded paths you mentioned). If that's right, let's call it out explicitly.

dinhxuanvu: If users want to customize annotations via their own Dockerfiler and annotations, they are free to do so. This statement is meant for folks who use the bundle library to generate, build and validate the bundle image. I'm happy to adjust the sentence to make it a bit clearer. Do you have a suggestion in mind?

PR #173, opened 2020-02-05
fix(index): copy correct opm bin builder path

njhale:

Use the correct path when copying `opm` from the builder image for generated index Dockerfiles.


jdockter:
@ecordell @jpeeler any chance this could get reviewed and merged?

ecordell:
/lgtm

We need to backport this to 4.4, correct? @njhale

PR #186, opened 2020-02-21
chore(deps): remove helm as a dependency

ecordell:
Removes helm as a dependency

This is purely to unblock vendoring in `oc` - helm brings in many dependencies that we do not currently use and which cause conflicts with oc.

Eventually we may want to pull helm in, but for now I have vendored just the bits for identifying a chart.

**Reviewer Checklist**


pkg/lib/bundle/chartutil.go
```diff
@@ -1,10 +1,11 @@
 /*
 Copyright The Helm Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
```
dinhxuanvu: nit: this is not important but would be nice to have it look prettier. 😄

pkg/lib/bundle/chartutil.go
```diff
@@ -64,31 +108,38 @@ type Metadata struct {
...
-	if md.Version == "" {
-		return ValidationError("chart.metadata.version is required")
+
+	chartContent := new(Metadata)
+	if err := yaml.Unmarshal(chartYamlContent, &chartContent); err != nil {
+		return false, err
 	}
-	if !isValidChartType(md.Type) {
-		return ValidationError("chart.metadata.type must be application or library")
+	if chartContent == nil {
```
benluddy: Does this ever evaluate to true?

PR #196, opened 2020-02-25
Opm test

gallettilance:

**Reviewer Checklist**


opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
+$containerTool push $bundleImage:$bundleTag3
+
 echo "building index image with prometheus"
-./bin/opm index add -b="$bundleImage:0.14.0,$bundleImage:0.15.0,$bundleImage:0.22.2" -t "$indexImage" -c="$containerTool"
+./bin/opm index add -b="$bundleImage:$bundleTag1,$bundleImage:$bundleTag2,$bundleImage:$bundleTag3" -t "$indexImage" -c="$containerTool"
+
 echo "pushing index image"
 $containerTool push $indexImage
+
 echo "sleep for 30s before pulling image"
```
ecordell: why do we need to sleep?

opm-test.sh
```diff
@@ -1,26 +1,58 @@
 #!/bin/bash
```
ecordell: could we make this a go file that `exec`s opm (or just exercises the commands)?

ecordell: that will also let you use build tags to selectively build for os

opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
-set -euxo pipefail
 
-echo "Setting Variables"
-read -p "Bunlde image: " bundleImage
-read -p "Index image tag: " indexImage
-read -p "docker or podman: " containerTool
+username=$2
+token=$3
+echo "logging in to quay"
```
awgreene: nit: `into`

opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
+docker login -u "$username" -p "$token" quay.io
+
+echo "Generating Random Tag"
+bundleTag1=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 6 | head -n 1)
```
awgreene: Could this be a function?

opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
-set -euxo pipefail
 
-echo "Setting Variables"
-read -p "Bunlde image: " bundleImage
-read -p "Index image tag: " indexImage
-read -p "docker or podman: " containerTool
+username=$2
+token=$3
+echo "logging in to quay"
+docker login -u "$username" -p "$token" quay.io
```
awgreene: Shouldn't this support docker or podman?

opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
+    exit 1
+fi
+    
+containerTool="$1"
+echo "Running with container tool $containerTool"
 
 echo "building opm"
 go build -mod=vendor  -o bin/opm ./cmd/opm
+echo "building initializer"
+go build -mod=vendor  -o bin/initializer ./cmd/initializer
```
awgreene: Nit: Double space between `mod` and `o`.

opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
+bundleTag1=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 6 | head -n 1)
+bundleTag2=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 6 | head -n 1)
+bundleTag3=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 6 | head -n 1)
+indexTag=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 6 | head -n 1)
+
+bundleImage="quay.io/olmtest/e2e-bundle"
+indexImage="quay.io/olmtest/e2e-index:$indexTag"
+
+if [ $# -eq 0 ]; then
+    echo "No container tool argument supplied"
```
awgreene: Nit: Technically this means no arguments were supplied right?

opm-test.sh
gallettilance: Oh good catch. I had this for when I was pulling from the same tag locally on every run

awgreene: If there is a reason, please add a comment.

opm-test.sh
gallettilance: sure!

opm-test.sh
```diff
@@ -1,26 +1,58 @@
...
+    exit 1
+fi
+
+if [ "$1" != "docker" && "$1" != "podman" ]; then
+    echo "container tool argument must be podman or docker"
+    exit 1
+fi
+    
+containerTool="$1"
+echo "Running with container tool $containerTool"
```
awgreene: Most of the `echos` are lowercase, please make this consistent.

opm-test.sh
gallettilance: podman shares that auth

.travis.yml
```diff
@@ -10,11 +10,29 @@ os:
 
 install:
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install make; fi
-- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install make; fi
+- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+    sudo apt-get install make;
```
ecordell: I think make is already present?

.travis.yml
```diff
@@ -10,11 +10,29 @@ os:
...
+    sudo add-apt-repository -y ppa:projectatomic/ppa;
+    sudo apt-get update -qq;
+    sudo apt-get -qq -y install podman;
+  fi
 
-script:
-- if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then mingw32-make build; fi
-- if [[ "$TRAVIS_OS_NAME" != "windows" ]]; then make unit build; fi
+script: |
+  if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
```
kevinrizza: Since this seems like technical debt to actually enable these tests in multiple environments, can we just run the tests in only linux to start and create issues to get them enabled everywhere else at a later time? Even if we don't remove these steps for now, I think it would be valuable to track it somewhere.

opm-test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if err := dbLoader.Migrate(context.TODO()); err != nil {
+		return err
+	}
+
+	loader := sqlite.NewSQLLoaderForDirectory(dbLoader, "downloaded")
+	return loader.Populate()
+}
+
+func main() {
```
ecordell: Why not use the testing package and run through all of the options (that exist for the target os?)

.travis.yml
```diff
@@ -10,11 +10,29 @@ os:
...
+    mingw32-make build
+  fi
+  if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    make unit build
+  fi
+  if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+    make unit build
+    go build -mod=vendor -o bin/opm ./cmd/opm
+    go build -mod=vendor -o bin/initializer ./cmd/initializer
+    go run opm-test.go docker
```
kevinrizza: Any reason to use `go run` instead of `go test` here? I think at the very least with `go test` we get some nice configuration and output stuff for free.

opm-test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	packageName    = "prometheus"
+	channels       = "preview"
+	defaultChannel = "preview"
+
+	bundlePath1 = "manifests/prometheus/0.14.0"
+	bundlePath2 = "manifests/prometheus/0.15.0"
+	bundlePath3 = "manifests/prometheus/0.22.2"
+
+	bundleTag1 = StringWithCharset(6)
```
kevinrizza: Are we generating random tags here just to account for concurrently running tests? Is this sufficiently random for that case?

.travis.yml
```diff
@@ -10,11 +10,29 @@ os:
 
 install:
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install make; fi
-- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install make; fi
+- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
```
benluddy: Any reason this can't be condensed into:
```
add-apt-repository
apt-get update
apt-get install
```

opm-test.go
gallettilance: I mean the individual probability of collision is like 1 in 36^6. Even if there are 5000 PRs up (each computing 3 random strings) the probability of collision is about 6%.

.travis.yml
gallettilance: maybe - I assumed not since this was part of the code

.travis.yml
gallettilance: for sure

benluddy:
@kevinrizza @ecordell @awgreene 

I updated this change to run the new tests as ginkgo specs. They'll be skipped when running against PRs from forks, which won't have access to registry credentials (https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions). PTAL!

https://github.com/operator-framework/operator-registry/pull/196/checks

Makefile
```diff
@@ -39,10 +37,11 @@ container-codegen:
...
 generate-fakes:
-	go generate ./...
+	$(GO) generate ./...
 
 clean:
 	@rm -rf ./bin
 
-opm-test:
-	 $(shell ./opm-test.sh || echo "opm-test FAIL")
+.PHONY: e2e
```
ecordell: super-nit: can we either put all phony directives together, or make them all separate like this?

pkg/lib/registry/registry.go
```diff
@@ -5,9 +5,9 @@ import (
 	"database/sql"
 	"fmt"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/operator-framework/operator-registry/pkg/sqlite"
```
ecordell: nit: we have been using  `default - library - current module` order for imports

pkg/lib/registry/registry.go
```diff
@@ -43,14 +43,15 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		if err := loader.Populate(); err != nil {
 			err = fmt.Errorf("error loading bundle from image: %s", err)
 			if !request.Permissive {
-				r.Logger.WithError(err).Fatal("permissive mode disabled")
+				r.Logger.WithError(err).Error("permissive mode disabled")
```
ecordell: since this isn't exiting anymore, is where do we actually check the error and fail if we're not `permissive`?

benluddy: [Here](https://github.com/operator-framework/operator-registry/blob/727f39a7bb54721a11bd026193618c31547a8c27/cmd/opm/registry/add.go#L68-L71) and [here](https://github.com/operator-framework/operator-registry/blob/727f39a7bb54721a11bd026193618c31547a8c27/pkg/lib/indexer/indexer.go#L96-L99)->[here](https://github.com/operator-framework/operator-registry/blob/727f39a7bb54721a11bd026193618c31547a8c27/cmd/opm/index/add.go#L126-L129).

PR #206, opened 2020-03-04
Match bundle image directory structure

kevinrizza:
This commit introduces a change to the way that the way the API function
bundle.GenerateFunc manages directory structure. Currently, the output
of that function creates a metadata folder with generated annotations
inside the given manifest directory. However, this is not how bundle
images are actually formatted, and the Dockerfile does some indirection
in order to copy the files into the correct directory.

However, we should consider the default behvaior of `generate` to be
"run once" (especially given that by default the regeneration does not
happen in the build function by default). The expectation is that
generate is a scaffolding method. Given that, we should scaffold the
metadata directory in the same manner as it would actually appear in the
union filesystem on the resulting container image.

In order to accomplish this, this PR makes a few functional changes:

- By default, create the metadata directory in the same parent directory
as the given manifest directory. This, at a high level, mirrors the directory
tree that will exist on the resulting bundle.

- Generates the dockerfile in the current working directory that the
function is called from. Modifies the dockerfile to point to the
relative path of the generated files

- Adds an option output directory parameter to the function (as well
as the shadowed `opm alpha bundle build/generate` commands) to
allow the user to generate a *new* bundle directory.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/bundle/build.go
+++ b/cmd/opm/alpha/bundle/build.go
@@ -13,6 +13,7 @@ var (
 	packageNameArgs    string
 	channelsArgs       string
 	channelDefaultArgs string
+	outputDirArgs      string
 	overwriteArgs      bool
 )
 
@@ -69,11 +70,13 @@ func newBundleBuildCmd() *cobra.Command {
 
 	bundleBuildCmd.Flags().BoolVarP(&overwriteArgs, "overwrite", "o", false, "To overwrite annotations.yaml locally if existed. By default, overwrite is set to `false`.")
 
+	bundleBuildCmd.Flags().StringVarP(&outputDirArgs, "output-dir", "u", "", "Optional output directory for operator manifests")
+
 	return bundleBuildCmd
 }
 
 func buildFunc(cmd *cobra.Command, args []string) error {
-	err := bundle.BuildFunc(dirBuildArgs, tagBuildArgs, imageBuilderArgs,
+	err := bundle.BuildFunc(dirBuildArgs, outputDirArgs, tagBuildArgs, imageBuilderArgs,
 		packageNameArgs, channelsArgs, channelDefaultArgs, overwriteArgs)
 	if err != nil {
 		return err
--- a/pkg/lib/bundle/build.go
+++ b/pkg/lib/bundle/build.go
@@ -4,22 +4,19 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path"
 
 	log "github.com/sirupsen/logrus"
 )
 
 // Create build command to build bundle manifests image
-func BuildBundleImage(directory, imageTag, imageBuilder string) (*exec.Cmd, error) {
+func BuildBundleImage(imageTag, imageBuilder string) (*exec.Cmd, error) {
 	var args []string
-
-	dockerfilePath := path.Join(directory, DockerFile)
 
 	switch imageBuilder {
 	case "docker", "podman":
-		args = append(args, "build", "-f", dockerfilePath, "-t", imageTag, directory)
+		args = append(args, "build", "-f", DockerFile, "-t", imageTag, ".")
 	case "buildah":
-		args = append(args, "bud", "--format=docker", "-f", dockerfilePath, "-t", imageTag, directory)
+		args = append(args, "bud", "--format=docker", "-f", DockerFile, "-t", imageTag, ".")
 	default:
 		return nil, fmt.Errorf("%s is not supported image builder", imageBuilder)
 	}
@@ -53,21 +50,21 @@ func ExecuteCommand(cmd *exec.Cmd) error {
 // @channels: The list of channels that bundle image belongs to
 // @channelDefault: The default channel for the bundle image
 // @overwrite: Boolean flag to enable overwriting annotations.yaml locally if existed
-func BuildFunc(directory, imageTag, imageBuilder, packageName, channels, channelDefault string, overwrite bool) error {
+func BuildFunc(directory, outputDir, imageTag, imageBuilder, packageName, channels, channelDefault string, overwrite bool) error {
 	_, err := os.Stat(directory)
 	if os.IsNotExist(err) {
 		return err
 	}
 
 	// Generate annotations.yaml and Dockerfile
-	err = GenerateFunc(directory, packageName, channels, channelDefault, overwrite)
+	err = GenerateFunc(directory, outputDir, packageName, channels, channelDefault, overwrite)
 	if err != nil {
 		return err
 	}
 
 	// Build bundle image
 	log.Info("Building bundle image")
-	buildCmd, err := BuildBundleImage(path.Clean(directory), imageTag, imageBuilder)
+	buildCmd, err := BuildBundleImage(imageTag, imageBuilder)
 	if err != nil {
 		return err
 	}
--- a/pkg/lib/bundle/build_test.go
+++ b/pkg/lib/bundle/build_test.go
@@ -12,35 +12,30 @@ func TestBuildBundleImage(t *testing.T) {
 	defer cleanup()
 
 	tests := []struct {
-		directory    string
 		imageTag     string
 		imageBuilder string
 		commandStr   string
 		errorMsg     string
 	}{
 		{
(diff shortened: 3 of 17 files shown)


pkg/lib/bundle/generate.go
```diff
@@ -63,37 +82,83 @@ func GenerateFunc(directory, packageName, channels, channelDefault string, overw
...
+	if outputDir == "" {
+		copyDir = filepath.Dir(manifestDir)
+		resultManifests = manifestDir
+	} else { // otherwise copy the manifests into $outputDir/manifests and create the annotations file in $outputDir/metadata
+		copyDir = outputDir
+
+		log.Info("Generating output manifests directory")
+
+		resultManifests = filepath.Join(copyDir, "/manifests/")
+		// copy the manifest directory into $pwd/manifests/
```
njhale: What do we gain by copying the manifests? Could we create a symlink instead?

Additionally, the bundle specification allows the location of the manifests directory to be configurable, so `manifests` may not match that configuration.

pkg/lib/bundle/generate.go
```diff
@@ -63,37 +82,83 @@ func GenerateFunc(directory, packageName, channels, channelDefault string, overw
...
+		err := copyManifestDir(manifestDir, resultManifests, overwrite)
+		if err != nil {
+			return "", "", err
+		}
+	}
+
+	// Now, generate the `metadata/` dir and write the annotations
+	file, err := ioutil.ReadFile(filepath.Join(copyDir, MetadataDir, AnnotationsFile))
```
njhale: Instead of making the metadata and manifests directory match up, have we thought about inlining the annotations file and hiding it from the end-user completely?

dinhxuanvu: This is a valid point that Eric/SDK suggested a while back about inline annotations/Dockerfile during docker build command instead writing them in disk. However, it is still a valid use case to write those files to disk if someone wants to see them.

ecordell: You mean:
```
/manifests
annotations.json
```
?

pkg/lib/bundle/build.go
```diff
@@ -4,22 +4,19 @@ import (
...
-func BuildBundleImage(directory, imageTag, imageBuilder string) (*exec.Cmd, error) {
+func BuildBundleImage(imageTag, imageBuilder string) (*exec.Cmd, error) {
 	var args []string
 
-	dockerfilePath := path.Join(directory, DockerFile)
-
 	switch imageBuilder {
 	case "docker", "podman":
-		args = append(args, "build", "-f", dockerfilePath, "-t", imageTag, directory)
+		args = append(args, "build", "-f", DockerFile, "-t", imageTag, ".")
```
dinhxuanvu: This needs to be tested out because I'm pretty sure that this won't work if the opm binary is not in the same directory with the manifests. I recall that I changed this from `.` to `diretory` to avoid this issue.

cmd/opm/alpha/bundle/build.go
```diff
@@ -69,11 +70,13 @@ func newBundleBuildCmd() *cobra.Command {
...
 	bundleBuildCmd.Flags().BoolVarP(&overwriteArgs, "overwrite", "o", false, "To overwrite annotations.yaml locally if existed. By default, overwrite is set to `false`.")
 
+	bundleBuildCmd.Flags().StringVarP(&outputDirArgs, "output-dir", "u", "", "Optional output directory for operator manifests")
+
 	return bundleBuildCmd
 }
 
 func buildFunc(cmd *cobra.Command, args []string) error {
-	err := bundle.BuildFunc(dirBuildArgs, tagBuildArgs, imageBuilderArgs,
+	err := bundle.BuildFunc(dirBuildArgs, outputDirArgs, tagBuildArgs, imageBuilderArgs,
```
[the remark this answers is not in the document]
kevinrizza: I avoided making any meaningful changes to this interface because it's currently implemented by the SDK team. Right now these functions are attempting to match the SDK's form and style as much as possible.

Personally, I would prefer to refactor this to pass a single struct that contains all parameters and a constructor that initiated a few defaults to allow for optional parameters given the complexity of the input. It just seems like we don't want to rock the boat too much with sdk adoption.

pkg/lib/bundle/generate.go
```diff
@@ -286,3 +364,52 @@ func WriteFile(fileName, directory string, content []byte) error {
...
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(to); os.IsNotExist(err) {
+		os.MkdirAll(to, os.ModePerm)
+	}
+
+	for _, fromFile := range fromFiles {
+		// ignore nested directories
```
ecordell: this might cause us problems if we want to support helm charts, kustomize, etc

pkg/lib/bundle/generate.go
kevinrizza: The goal of this function is just to scaffold. imo `manifests/` is just a nice sane default if you want to generate a bundle directory. We could create a symlink, but I think that would just add to confusion -- especially given that we have already received some feedback around complexity and different usecases for what and how people want to use this.

pkg/lib/bundle/generate.go
kevinrizza: I still think we want this thing to be "Generate/scaffold this once, then you can do what you want with it". Today it's true that everything in `metadata/` is probably not super important to end users, but if we want to move lots of other metadata in the future that users DO want to configure, I think it's important to make it visible now.

pkg/lib/bundle/build_test.go
```diff
@@ -50,7 +50,7 @@ func TestBuildBundleImage(t *testing.T) {
 
 	for _, item := range tests {
 		var cmd *exec.Cmd
-		cmd, err := BuildBundleImage(item.directory, item.imageTag, item.imageBuilder)
```
harishsurf: Could the directory field be removed from the struct if unused?

pkg/lib/bundle/generate.go
```diff
@@ -43,8 +44,20 @@ type AnnotationMetadata struct {
 // @channels: The list of channels that bundle image belongs to
 // @channelDefault: The default channel for the bundle image
 // @overwrite: Boolean flag to enable overwriting annotations.yaml locally if existed
```
harishsurf: nit: could you add a comment for the outputDir flag?

cmd/opm/alpha/bundle/build.go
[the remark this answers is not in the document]
dinhxuanvu: I don't find variadic options is a good pattern for this case when those variables are used for different purposes. Some as dir, some for image tag and builder options. They should remain separate to avoid confusion. However, variadic options pattern does have the benefit that it may avoid future breaking API changes if we decide to add another string parameter into the func. Just need to handle the order.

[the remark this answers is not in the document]
kevinrizza: I actually like passing in a struct in cases like this, because it lets you just initialize the stuff you care about and check for zero values that you don't care about ex: https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/indexer/indexer.go#L44

It lets you be hands off about requiring or ignoring certain parameters and lets the implementer make choices about it, which works pretty well when the indirection in front of the function is a CLI or library implementation

cmd/opm/alpha/bundle/build.go
```diff
@@ -13,6 +13,7 @@ var (
 	packageNameArgs    string
 	channelsArgs       string
 	channelDefaultArgs string
+	outputDirArgs      string
```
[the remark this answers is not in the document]
kevinrizza: +1, i'll update the pr to include it.

pkg/lib/bundle/build_test.go
kevinrizza: +1

pkg/lib/bundle/generate.go
kevinrizza: +1

pkg/lib/bundle/generate.go
kevinrizza: That's a fair point, I'll update this to copy recursively

PR #209, opened 2020-03-06
Generalize manifest copy

kevinrizza:
Allow arbitrary files to be copied from the manifest directory, not just
those that have the .yaml file extension.

**Reviewer Checklist**


pkg/lib/bundle/generate_test.go
```diff
@@ -208,7 +208,7 @@ func TestGenerateDockerfileFunc(t *testing.T) {
 		"LABEL operators.operatorframework.io.bundle.package.v1=test4\n"+
 		"LABEL operators.operatorframework.io.bundle.channels.v1=test5\n"+
 		"LABEL operators.operatorframework.io.bundle.channel.default.v1=test5\n\n"+
-		"COPY test2/*.yaml /manifests/\n"+
+		"COPY test2 /manifests/\n"+
 		"COPY metadata/annotations.yaml /metadata/annotations.yaml\n", MetadataDir)
```
njhale: nit: I think this should be formatted so that the metadata directory always matches the label.

PR #213, opened 2020-03-12
feat(opm): add unprivileged registry add

njhale:
Add the ability to pull and unpack bundle images without using an external tool or daemon. Make this the only supported mode by removing the `--container-tool` option from the `registry add` subcommand.

Ex.

```sh
$ opm registry add --bundle-images "quay.io/olmtest/kiali:1.2.4"
INFO[0000] adding to the registry                        bundles="[quay.io/olmtest/kiali:1.2.4]"
INFO[0000] resolved name: quay.io/olmtest/kiali:1.2.4
INFO[0000] fetched                                       digest="sha256:a1bec450c104ceddbb25b252275eb59f1f1e6ca68e0ced76462042f72f7057d8"
INFO[0000] fetched                                       digest="sha256:8b19d5ea2c56e663fb886d9ca273dd1f8e20b107c9b6817dedd35aac90e16494"
INFO[0000] fetched                                       digest="sha256:1c019331644e2e826b6b8f11328b31a1c2f4aa2ae6ee225f0ab19a14af086d2b"
INFO[0000] fetched                                       digest="sha256:d9c98a446dd77b5e7cb1394748c2279d83668a519b2badda1259cd64cb3beb3f"
INFO[0000] unpacking layer: {application/vnd.docker.image.rootfs.diff.tar.gzip sha256:1c019331644e2e826b6b8f11328b31a1c2f4aa2ae6ee225f0ab19a14af086d2b 5916 [] map[] <nil>}
INFO[0000] unpacking layer: {application/vnd.docker.image.rootfs.diff.tar.gzip sha256:8b19d5ea2c56e663fb886d9ca273dd1f8e20b107c9b6817dedd35aac90e16494 296 [] map[] <nil>}
INFO[0000] found annotations file searching for csv      dir=bundle_tmp146087939 file=bundle_tmp146087939/metadata load=annotations
INFO[0000] found csv, loading bundle                     dir=bundle_tmp146087939 file=bundle_tmp146087939/manifests load=bundle
INFO[0000] loading bundle file                           dir=bundle_tmp146087939/manifests file=kiali.crd.yaml load=bundle
INFO[0000] loading bundle file                           dir=bundle_tmp146087939/manifests file=kiali.monitoringdashboards.crd.yaml load=bundle
INFO[0000] loading bundle file                           dir=bundle_tmp146087939/manifests file=kiali.package.yaml load=bundle
INFO[0000] loading bundle file                           dir=bundle_tmp146087939/manifests file=kiali.v1.4.2.clusterserviceversion.yaml load=bundle
```

Also, introduce a new bundle oriented image interface to decouple actions (pull, push, pack, unpack) from their underlying implementation.


pkg/image/registry.go
```diff
@@ -0,0 +1,33 @@
...
+
+// At the high-level, we need to be able to do the following w/ images
+// pull(ref)
+// push(ref)
+// unpack(ref, writer)
+// build(ref, reader) => new ref
+
+type Registry interface {
+	// Pull fetches and stores an image by reference.
+	Pull(ctx context.Context, ref string) error
```
exdx: One thing I like is using custom types for specificity such a `type ImageReference string` and then in the interface it would take in a `ref ImageReference` type as the arg. I'm not sure how appropriate it is here but something worth noting as it makes the code more readable IMO.

pkg/image/shellout/registry.go
```diff
@@ -0,0 +1,38 @@
+package shellout
```
anik120: @njhale while reading through the code I realized that I understand what `unprivileged` means in the context on this PR due to the commit message, but I'm not sure what `shellout` is for. What are your thoughts on adding comments that summaries the intent of a package/it's name? 
Even with `unprivileged` I guess I'd have to go back and open up the PR to understand the motive behind it.

njhale: @anik120 I totally understand -- I actually spent a long time thinking about the naming here without coming up with anything better.

FYI: "shellout" is for shelling out to an external tool.

cmd/opm/registry/add.go
```diff
@@ -27,7 +27,6 @@ func newRegistryAddCmd() *cobra.Command {
 	rootCmd.Flags().StringP("database", "d", "bundles.db", "relative path to database file")
 	rootCmd.Flags().StringSliceP("bundle-images", "b", []string{}, "comma separated list of links to bundle image")
 	rootCmd.Flags().Bool("permissive", false, "allow registry load errors")
-	rootCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
```
ecordell: we should consider making this a hidden flag in case anyone was relying on it existing (instead of throwing an error about not knowing the flag).

If we're confident no one is relying on it yet, we can remove.

go.mod
```diff
@@ -31,3 +39,5 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/kubectl v0.17.3
 )
+
+replace github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
```
ecordell: is there any way to resolve this? we just got these all cleaned up...

pkg/image/unprivileged/resolver.go
```diff
@@ -0,0 +1,51 @@
...
+	"os"
+
+	"github.com/containerd/containerd/remotes"
+	"github.com/containerd/containerd/remotes/docker"
+	auth "github.com/deislabs/oras/pkg/auth/docker"
+)
+
+func NewResolver(username, password string, configs ...string) (remotes.Resolver, error) {
+	// TODO(njhale): make plainHTTP and insecure optional
+	plainHTTP := true
```
ecordell: We should figure out these options before we merge - not supporting tls verification will be an issue.

pkg/image/unprivileged/resolver.go
```diff
@@ -0,0 +1,51 @@
...
+	opts.Client = client
+
+	if username != "" || password != "" {
+		opts.Credentials = func(hostName string) (string, string, error) {
+			return username, password, nil
+		}
+		return docker.NewResolver(opts), nil
+	}
+
+	cli, err := auth.NewClient(configs...)
```
ecordell: does this support containerd's keychain integration (so you don't have to have your password in plaintext?)

cmd/opm/registry/add.go
```diff
@@ -27,7 +27,6 @@ func newRegistryAddCmd() *cobra.Command {
 	rootCmd.Flags().StringP("database", "d", "bundles.db", "relative path to database file")
```
ecordell: Don't forget we need a couple of updates / notes in the docs for `opm`!

cmd/opm/registry/add.go
njhale: @benluddy and I were discussing this in [an earlier review](). Within Red Hat, I know of a handful unreleased automation making use of the default value (not specifying `--container-tool`). However, I do think it's safer to use a hidden flag. Can we meet in the middle with a hidden flag that logs a warning when specified and continues with the daemonless implementation?

pkg/image/unprivileged/options.go
```diff
@@ -0,0 +1,131 @@
...
+	if err := config.complete(); err != nil {
+		return nil, err
+	}
+
+	cs, err := contentlocal.NewStore(config.CacheDir)
+	if err != nil {
+		return nil, err
+	}
+
+	bdb, err := bolt.Open(config.DBPath, 0644, nil)
```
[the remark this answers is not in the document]
njhale: Yup, it's always generated at runtime unless you **really** want to use an existing file.

pkg/image/registry.go
```diff
@@ -0,0 +1,33 @@
...
+	// If the referenced image does not exist in the registry, an error is returned.
+	Unpack(ctx context.Context, ref, dir string) error
+
+	// Pack creates and stores an image based on the given reference and returns a reference to the new image.
+	// If the referenced image does not exist in the registry, a new image is created from scratch.
+	// If it exists, it's used as the base image.
+	Pack(ctx context.Context, ref string, from io.Reader) (next string, err error)
+}
+
+// We also need to generate Dockerfiles
```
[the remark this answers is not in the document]
njhale: It should be cleaned up now.

go.mod
```diff
@@ -14,20 +19,23 @@ require (
...
-	google.golang.org/grpc v1.24.0
+	google.golang.org/grpc v1.27.0
 	gopkg.in/yaml.v2 v2.2.8
 	k8s.io/api v0.17.3
 	k8s.io/apiextensions-apiserver v0.17.3
 	k8s.io/apimachinery v0.17.3
 	k8s.io/client-go v0.17.3
 	k8s.io/klog v1.0.0
 	k8s.io/kubectl v0.17.3
+	rsc.io/letsencrypt v0.0.3 // indirect
```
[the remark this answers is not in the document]
njhale: this should be gone now

go.mod
njhale: We could just not run an in-process docker registry, which I expect would make the tests more complex.

pkg/image/unprivileged/resolver.go
njhale: [looks like it does](https://github.com/deislabs/oras/blob/19a0d4482aed33ec4431dc8fa126a484c475f2db/pkg/auth/docker/client.go#L38)

cmd/opm/registry/add.go
njhale: I made it a deprecated flag, which is both hidden and gives a warning when used.

go.mod
njhale: This is also a test dependency, which I think means it's not required by dependents to use the operator-registry module.

pkg/image/unprivileged/options.go
```diff
@@ -0,0 +1,141 @@
...
+		once   sync.Once
+		closed bool
+	)
+	close := func() error {
+		defer func() {
+			once.Do(func() {
+				closed = true
+			})
+		}()
+		if closed {
```
benluddy: Isn't this read unsafe? If `closed` is set to true in the deferred func, then it's going to be set to true regardless of when `close` returns, so you may as well scrap the defer and replace it with a synchronized get-then-set or an atomic swap at the beginning of `close`.

pkg/registry/populator.go
```diff
@@ -0,0 +1,256 @@
...
+	files, err := ioutil.ReadDir(metadata)
+	if err != nil {
+		return fmt.Errorf("unable to read directory %s: %s", metadata, err)
+	}
+
+	// Look for the metadata and manifests sub-directories to find the annotations.yaml file that will inform how the
+	// manifests of the bundle should be loaded into the database.
+	annotationsFile := &AnnotationsFile{}
+	for _, f := range files {
+		fileReader, err := os.Open(filepath.Join(metadata, f.Name()))
```
benluddy: This file descriptor doesn't appear to be closed?

pkg/registry/populator.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+
+		if strings.HasPrefix(f.Name(), ".") {
+			log.Info("skipping hidden file")
+			continue
+		}
+
+		log.Info("loading bundle file")
+		path := filepath.Join(dir, f.Name())
+		fileReader, err := os.Open(path)
```
benluddy: Same here. I don't know what we would consider a large number of files in a bundle, but it could be worthwhile to either call Close outside of a defer or wrap the loop body and defer Close to limit peak open file descriptors. I ran into a problem with FD exhaustion [here](https://github.com/operator-framework/operator-registry/commit/626af900865653e4318855776c84868b038a10ee#diff-9521fe6a08ae105c7acc7037e3369db5), granted, the number of records in one of those archives is probably significantly higher than any sane bundle directory.

pkg/registry/populator.go
```diff
@@ -0,0 +1,256 @@
...
+			continue
+		}
+
+		if strings.HasPrefix(f.Name(), ".") {
+			log.Info("skipping hidden file")
+			continue
+		}
+
+		path := filepath.Join(manifests, f.Name())
+		fileReader, err := os.Open(path)
```
benluddy: Here too. Maybe it's worth splitting out a tiny `func(string, *unstructured.Unstructured)`?

pkg/image/registry.go
benluddy: I wouldn't insist on it, but I also appreciate the extra compile-time safety of derived types for this kind of thing.

pkg/image/unprivileged/registry.go
```diff
@@ -0,0 +1,142 @@
...
+
+func ensureNamespace(ctx context.Context) context.Context {
+	if _, namespaced := namespaces.Namespace(ctx); !namespaced {
+		return namespaces.WithNamespace(ctx, namespaces.Default)
+	}
+	return ctx
+}
+
+func adjustPerms(h *tar.Header) (bool, error) {
+	h.Uid = os.Getuid()
```
benluddy: `os.Getuid` and `os.Getgid` return -1 on Windows, are there any portability implications?

pkg/image/unprivileged/registry_test.go
```diff
@@ -0,0 +1,113 @@
...
+	config := &configuration.Configuration{}
+	config.HTTP.Addr = fmt.Sprintf(":%d", dockerPort)
+	if rootDir != "" {
+		config.Storage = map[string]configuration.Parameters{"filesystem": map[string]interface{}{
+			"rootdirectory": rootDir,
+		}}
+	} else {
+		config.Storage = map[string]configuration.Parameters{"inmemory": map[string]interface{}{}}
+	}
+	config.HTTP.DrainTimeout = time.Duration(2) * time.Second
```
benluddy: Nit: I feel like I usually see the integer literal directly multiplied by the time unit base, but maybe I'm mistaken and that isn't conventional style... the only example in the godoc is written this way. :man_shrugging:

pkg/image/unprivileged/registry_test.go
```diff
@@ -0,0 +1,113 @@
...
+	go func() {
+		require.NoError(t, dockerRegistry.ListenAndServe())
+	}()
+
+	// Return the registry host string
+	return fmt.Sprintf("127.0.0.1:%d", dockerPort)
+}
+
+func dirChecksum(t *testing.T, dir string) string {
+	sum, err := dirhash.HashDir(dir, "", dirhash.DefaultHash)
```
benluddy: That's handy!

pkg/image/unprivileged/registry_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+			// Copy golden manifests to a temp dir
+			dir := "kiali-unpacked"
+			require.NoError(t, r.Unpack(ctx, ref, dir))
+
+			checksum := dirChecksum(t, dir)
+			require.Equal(t, tt.expected.checksum, checksum)
+
+			require.NoError(t, r.Close())
+			require.NoError(t, os.RemoveAll(dir))
```
benluddy: Might want to defer this to increase the odds it happens on test failures, unless this was an intentional decision to help debug failures?

njhale:
I received some internal feedback to implement this with opencontainers instead of containerd if possible.

/hold

pkg/image/registry.go
njhale: Okay -- I'll change this

pkg/image/unprivileged/options.go
njhale: I didn't intend for this to be threadsafe, but you make a good point -- do you think it would suffice to wrap the entire block with `once.Do`?

pkg/image/unprivileged/registry.go
njhale: Nice catch. I think there could be and I'm fairly certain `opm` doesn't work on windows as it stands. There are already multiple failures in the travis windows tests (our config just isn't picking them up).

pkg/image/unprivileged/registry_test.go
njhale: I also thought this was strange -- I used oras as an example for this.

pkg/image/unprivileged/registry_test.go
njhale: This was intentional, but I can put it in a defer.

njhale:
/hold cancel

Now I'm ready :)

PR #217, opened 2020-03-16
[WIP] feat: add bundle commutativity checks to opm registry add command

exdx:
Add support for adding multiple packages in different order to the `opm registry add` command. 

Acceptance Criteria:
The opm commands in operator registry that add bundle images to existing registry databases are order independent.

This comes for free with semver bundles, but these bundles are not treated any differently for the purposes of this PR so the hot codepath stays the same. The idea was to keep the existing code (that generally assumes one bundle is loaded independently of another) as in-tact as possible while extending it (having the command aware of other bundles being loaded simultaneously) 

More opm features around index upgrade paths 

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/image.go
+++ b/pkg/sqlite/image.go
@@ -34,7 +34,6 @@ func NewSQLLoaderForImage(store registry.Load, image, containerTool string) *Ima
 }
 
 func (i *ImageLoader) Populate() error {
-
 	log := logrus.WithField("img", i.image)
 
 	workingDir, err := ioutil.TempDir("./", "bundle_tmp")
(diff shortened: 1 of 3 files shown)


pkg/sqlite/image.go
```diff
@@ -141,12 +150,60 @@ func (i *ImageLoader) loadManifests(manifests string, annotationsFile *registry.
...
+		return nil
+	} else {
+		// there are no bundles left provided to the add arg
+
+		// check to see if bundle types are related
+		// TODO do we have code that does this already somewhere?
+		sortedBundles = checkRelatedBundles(addedBundles)
+
+		// clear array
+		addedBundles = nil
```
exdx: bug: this needs to be `deferred` to ensure it happens on every error exit in the function context

exdx:
closing in favor of #220

PR #219, opened 2020-03-16
Minor text updates to bundle registry doc

rwsu:
Signed-off-by: Richard Su <rwsu@redhat.com>

Readability improvements.
Fixed bundle validate example.

Improve documentation.

**Reviewer Checklist**

### DIFF
--- a/docs/design/operator-bundle.md
+++ b/docs/design/operator-bundle.md
@@ -60,8 +60,8 @@ annotations:
 ```
 
 *Notes:*
-* In case of a mismatch, the `annotations.yaml` file is authoritative because the on-cluster operator-registry that relies on these annotations has access to the yaml file only.
-* The potential use case for the `LABELS` is - an external off-cluster tool can inspect the image to check the type of a given bundle image without downloading the content.
+* In case of a mismatch, the `annotations.yaml` file is authoritative because the on-cluster operator-registry that relies on these annotations have access to the yaml file only.
+* The potential use case for the `LABELS` is - an external off-cluster tool that can inspect the image to check the type of a given bundle image without downloading the content.
 * The annotations for bundle manifests and metadata are reserved for future use. They are set to be `manifests/` and `metadata/` for the time being.
 
 ### Bundle Dockerfile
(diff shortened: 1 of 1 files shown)


docs/design/operator-bundle.md
```diff
@@ -60,8 +60,8 @@ annotations:
 ```
 
 *Notes:*
-* In case of a mismatch, the `annotations.yaml` file is authoritative because the on-cluster operator-registry that relies on these annotations has access to the yaml file only.
-* The potential use case for the `LABELS` is - an external off-cluster tool can inspect the image to check the type of a given bundle image without downloading the content.
+* In case of a mismatch, the `annotations.yaml` file is authoritative because the on-cluster operator-registry that relies on these annotations have access to the yaml file only.
```
ecordell, line 63: The subject for the verb is the singular `on-cluster operator-registry` so it should still be `has`

```suggestion
* In case of a mismatch, the `annotations.yaml` file is authoritative because the on-cluster operator-registry that relies on these annotations has access to the yaml file only.
```

njhale:
/retest

@rwsu If you address the current feedback, we can get this merged. Thanks again for the contribution!

docs/design/operator-bundle.md
```diff
@@ -126,27 +126,27 @@ Flags:
...
 $ tree test
 test
 ├── etcdcluster.crd.yaml
 └── etcdoperator.clusterserviceversion.yaml
 ```
 
-The `--package` or `-p` is the name of package fo the operator such as `etcd` which which map `channels` to a particular application definition. `channels` allow package authors to write different upgrade paths for different users (e.g. `beta` vs. `stable`). The `channels` list is provided via `--channels` or `-c` flag. Multiple `channels` are separated by a comma (`,`). The default channel is provided optionally via `--default` or `-e` flag. If the default channel is not provided, the first channel in channel list is selected as default.
+The `--package` or `-p` is the name of package for the operator such as `etcd` which which map `channels` to a particular application definition. `channels` allow package authors to write different upgrade paths for different users (e.g. `beta` vs. `stable`). The `channels` list is provided via `--channels` or `-c` flag. Multiple `channels` are separated by a comma (`,`). The default channel is provided optionally via `--default` or `-e` flag. If the default channel is not provided, the first channel in the channel list is selected as the default.
```
AndrienkoAleksandr, line 145: `which which`  -> `which`

njhale:
Looks like this is stale.

I'm going to close this out. Please feel free to reopen.

PR #224, opened 2020-03-24
Graph Loader initial implementation

kevinrizza:
*Create new graph type that can describe channel graphs of packages
*Create initial implementation of loading it from existing sqlite db

This graph implementation should be used as a declarative structure to define the update graph for both input and output in the future


pkg/registry/graph.go
```diff
@@ -0,0 +1,34 @@
...
+	OperatorBundles []OperatorBundle
+	Head            BundleRef
+}
+
+type OperatorBundle struct {
+	BundlePath      string
+	Version         string // semver string
+	CsvName         string
+	ReplacesBundles []OperatorBundle
+	Replaces        []BundleRef
```
ecordell:
```suggestion
	Replaces []*OperatorBundle
```
Why not this?

pkg/registry/graph.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+type BundleRef struct {
+	BundlePath string
+	Version    string //semver string
```
ecordell: Why not type this as a semver field?

pkg/registry/graph.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Channel struct {
+	Name            string
+	OperatorBundles []OperatorBundle
+	Head            BundleRef
+}
+
+type OperatorBundle struct {
+	BundlePath      string
+	Version         string // semver string
```
ecordell: Why not typed?

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,173 @@
...
+			Head:            *head,
+		}
+
+		channels = append(channels, channel)
+	}
+
+	return channels, nil
+}
+
+func getBundle(bundles []registry.OperatorBundle, name string) *registry.OperatorBundle {
```
ecordell: it really seems like this could be a map?

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, b := range bundles {
+		if b.CsvName == name {
+			return &b
+		}
+	}
+	return nil
+}
+
+func getHeadBundleRefForChannel(bundles []registry.OperatorBundle) *registry.BundleRef {
```
ecordell: This assumes an already valid graph, which might not be true of the places you're intending to call this

pkg/sqlite/query.go
```diff
@@ -113,6 +113,76 @@ func (s *SQLQuerier) GetPackage(ctx context.Context, name string) (*registry.Pac
 	return pkg, nil
 }
 
+func (s *SQLQuerier) GetDefaultPackage(ctx context.Context, name string) (string, error) {
```
ecordell: nit: should this be `GetDefaultChannel`?

pkg/registry/graph.go
ecordell: After reviewing the rest, I think this could also be: 
```suggestion
	Replaces map[string]*OperatorBundle
```

or:

```suggestion
	Replaces map[BundleRef]*OperatorBundle
```

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,173 @@
...
+type GraphLoader interface {
+	Generate() (*registry.Package, error)
+}
+
+type SQLGraphLoader struct {
+	Querier     *SQLQuerier
+	PackageName string
+}
+
+// TODO
```
exdx: nit: remove

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return &registry.Package{
+		Name:           g.PackageName,
+		DefaultChannel: defaultChannel,
+		Channels:       channels,
+	}, nil
+}
+
+// TODO get bundle path and version
```
exdx: nit: remove

pkg/sqlite/graphloader_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"math/rand"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+func createLoadedTestDb(t *testing.T) (*sql.DB, func()) {
```
exdx: woo nice test 🎉

pkg/registry/graph.go
```diff
@@ -0,0 +1,34 @@
...
+	CsvName         string
```
ecordell: In some newer parts of this codebase, we have moved away from `CsvName` to `BundleName` or `Identifier`

pkg/sqlite/query.go
ecordell: Also - is the query interface updated with these?

pkg/registry/graph.go
kevinrizza: I think the intention of this type is to explicitly be an aggregate identifier. At some point `CsvName` could be deprecated entirely. For now they are equivalent though, and I think it's more likely that this field would be removed than refer to a different concept. Wdyt?

pkg/registry/graph.go
kevinrizza: great idea

pkg/sqlite/graphloader.go
kevinrizza: This specific implementation is reading from the sql layer. The assumption here is that the graph is valid because we wouldn't persist it in the first place otherwise.

That being said, I can totally as we flesh this out attempting to do some generation validation. Maybe that's what we should tackle next, and when the insert is written we include that as part of this function?

pkg/registry/graph.go
Bowenislandsong: WDYT if we replace the first 3 fields with bundleRef? This way we define how we refer to a bundle in one place?

pkg/registry/graph.go
```diff
@@ -0,0 +1,22 @@
...
+	Replaces map[BundleKey]map[BundleKey]struct{}
+}
+
+type BundleKey struct {
+	BundlePath string
+	Version    string //semver string
+	CsvName    string
+}
+
+func (b *BundleKey) IsEmpty() bool {
```
exdx: Since we're using the `BundleKey` struct as a key in the map, should we add another method to check if two bundle keys are identical? From https://blog.golang.org/maps (`where KeyType may be any type that is comparable`) .

I guess we don't have to for the purposes of defining the map (Go does this for us) but it may be a helpful utility function to have alongside `b.IsEmpty()`. `b.IsEqual(a)`.

kevinrizza: I think that's a fair thing to suggest, but perhaps it should just be included when something that needs to do the comparison is written?

pkg/sqlite/query.go
kevinrizza: latest updated includes the interface update

pkg/registry/types.go
```diff
@@ -76,6 +76,18 @@ type ChannelEntry struct {
 	Replaces    string
 }
 
+// ChannelEntryNode is a denormalized node in a channel graph annotated with additional entry level info
+type ChannelEntryNode struct {
```
exdx: I think the naming between `ChannelEntry` and `ChannelEntryNode` is a little confusing. Based on the comments they are similar, one just had more details in it. Maybe we can call it `ChannelEntryAnnotated`?

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,134 @@
...
+	"database/sql"
+	"fmt"
+
+	"github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+// GraphLoader generates a graph
+// GraphLoader supports multiple different loading schemes
+// GraphLoader from SQL, GraphLoader from old format (filesystem), GraphLoader from SQL + input bundles
+type GraphLoader interface {
```
exdx: Should be move this interface out of this package as part of this PR?

pkg/registry/graph.go
```diff
@@ -0,0 +1,22 @@
...
+
+type Package struct {
+	Name           string
+	DefaultChannel string
+	Channels       map[string]Channel
+}
+
+type Channel struct {
+	Head     BundleKey
+	Replaces map[BundleKey]map[BundleKey]struct{}
```
Bowenislandsong: What happened here? Cna we define this? like `ReplaceMap`?

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return &registry.Package{
+		Name:           g.PackageName,
+		DefaultChannel: defaultChannel,
+		Channels:       channels,
+	}, nil
+}
+
+// GraphFromEntries builds the graph from a set of channel entries
+func (g *SQLGraphLoader) GraphFromEntries(channelEntries []registry.ChannelEntryNode) (map[string]registry.Channel, error) {
```
exdx: I'm wondering if we should have`GraphFromEntries` be on the `GraphLoader` interface as well - we want all the different loaders to be able to provide this functionality as well right

exdx: Or maybe we can have a separate `Grapher` interface 😄

kevinrizza: I actually think we explicitly don't want other graphloaders to have this method. This is really specific to the sqlite layer, other loaders wouldn't be aware of this aggregation at all.

pkg/registry/graph.go
kevinrizza: What about `nodes` ?

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+type SQLGraphLoader struct {
+	Querier     *SQLQuerier
```
benluddy: Can this accept a registry.Query instead of a *SQLQuerier and still do its job? What's tying it to the SQLite implementation? Either way, I'd like to see the NewX funcs accept the querier directly, instead of coupling the graph loader to the way queriers are created.

kevinrizza: The reason is that this entire loader implentation is "get the aggregation and abstraction that the sql layer provides and convert it into a graph". Literally all of the functions here are directly an abstraction around that. If we made it generic, then the input from other functions would have to be in the channelentry form, but that abstraction is exactly the one we are attempting to hide through this interface.

kevinrizza: Another example loader would be "okay now take some bundle images as input and load them into this struct directly" which just has a different set of concepts as input

pkg/sqlite/graphloader.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return &registry.Package{
+		Name:           g.PackageName,
+		DefaultChannel: defaultChannel,
+		Channels:       channels,
+	}, nil
+}
+
+// GraphFromEntries builds the graph from a set of channel entries
+func (g *SQLGraphLoader) GraphFromEntries(channelEntries []registry.ChannelEntryAnnotated) (map[string]registry.Channel, error) {
```
exdx: suggestion: can we make this a function instead of a method of the `SQLGraphLoader`? the function doesn't use any of the `SQLGraphLoader` data. This makes the func easier to use in other pkgs

kevinrizza: seems reasonable

pkg/sqlite/graphloader.go
kevinrizza: circling back around to this, because I misunderstood and talked to ben offline:

updated the struct to use the query interface for mocking purposes

benluddy: Looks good, thanks!

PR #227, opened 2020-03-25
fix(dns): fix slow dns resolution of localhost for grpc health probe

ecordell:
Adds an nsswitch.conf to prefer local files over dns for resolution. 

see: https://github.com/golang/go/issues/35305

**Reviewer Checklist**

### DIFF
--- a/cmd/appregistry-server/main.go
+++ b/cmd/appregistry-server/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/api"
 	health "github.com/operator-framework/operator-registry/pkg/api/grpc_health_v1"
 	"github.com/operator-framework/operator-registry/pkg/appregistry"
+	"github.com/operator-framework/operator-registry/pkg/lib/dns"
 	"github.com/operator-framework/operator-registry/pkg/lib/log"
 	"github.com/operator-framework/operator-registry/pkg/registry"
 	"github.com/operator-framework/operator-registry/pkg/server"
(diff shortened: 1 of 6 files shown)


njhale:
Cool! Here's what I found:

1. I think we should add the file for all [generated Dockerfiles](https://github.com/operator-framework/operator-registry/blob/d2e885544104bf40d9f6c4e9e462a8f6e43d4616/pkg/containertools/dockerfilegenerator.go#L35), since we don't know if the given base image contains it. 
2. Looks like image builds are failing on [chmod to `/dev`](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/operator-framework_operator-registry/227/pull-ci-operator-framework-operator-registry-master-images/605/build-log.txt). I don't think the failure is related to any changes here. Since the blame is months old, my primary suspect is that the base image was swapped out under us in CI again.

ecordell:
> I think we should add the file for all generated Dockerfiles, since we don't know if the given base image contains it.

This is tricky, since the source image can vary on user input and there's no guarantee we'll have `echo`

njhale:
@ecordell, then do you think opm should just attempt to generate the file at runtime?

ecordell:
@njhale Updated with your suggestion.

In the downstream image, nsswitch is already populated correctly. I tested with the upstream image as well:

```sh
/go # cat /etc/nsswitch.conf
cat: can't open '/etc/nsswitch.conf': No such file or directory
/go # /registry-server &
/go # cat /etc/nsswitch.conf
hosts: files dns/go
```

PR #228, opened 2020-03-25
[WIP] feat: multi image graph loader

exdx:
Continuation of #220 

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -39,6 +39,21 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		return err
 	}
 
+	// use multi-image loader for add request with multiple image bundles
+	if len(request.Bundles) > 1 {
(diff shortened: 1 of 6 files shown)


pkg/lib/registry/registry.go
```diff
@@ -39,6 +39,21 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		return err
 	}
 
+	// use multi-image loader for add request with multiple image bundles
+	if len(request.Bundles) > 1 {
```
kevinrizza: should we just always use the multi image loader? why do we need a separate loader?

exdx: I agree, this looks kind of funny. I wanted to preserve the old code path, just in case, for now. But conceptually yeah I guess we should handle all `index add` calls through the new codepath

exdx:
Closed in favor of #285.

PR #229, opened 2020-03-26
[feature] Understanding packages from directory

Bowenislandsong:
This commit adds a feature to interpret bundle name and channels based on a given bundle directory and update bundle's annotation YAML. This feature is triggered when `opm alpha generate/build` is not supplying `packageName`, `channels`, and `defaultChannels` information.
This Bundle directory needs to be inside a Package directory that contains all bundles.

### DIFF
--- a/cmd/opm/alpha/bundle/build.go
+++ b/cmd/opm/alpha/bundle/build.go
@@ -44,40 +44,44 @@ func newBundleBuildCmd() *cobra.Command {
 		RunE: buildFunc,
 	}
 
-	bundleBuildCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "", "The directory where bundle manifests and metadata for a specific version are located")
+	bundleBuildCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "",
+		"The directory where bundle manifests and metadata for a specific version are located")
 	if err := bundleBuildCmd.MarkFlagRequired("directory"); err != nil {
 		log.Fatalf("Failed to mark `directory` flag for `build` subcommand as required")
 	}
 
-	bundleBuildCmd.Flags().StringVarP(&tagBuildArgs, "tag", "t", "", "The image tag applied to the bundle image")
+	bundleBuildCmd.Flags().StringVarP(&tagBuildArgs, "tag", "t", "",
+		"The image tag applied to the bundle image")
 	if err := bundleBuildCmd.MarkFlagRequired("tag"); err != nil {
 		log.Fatalf("Failed to mark `tag` flag for `build` subcommand as required")
 	}
 
-	bundleBuildCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "", "The name of the package that bundle image belongs to")
-	if err := bundleBuildCmd.MarkFlagRequired("package"); err != nil {
-		log.Fatalf("Failed to mark `package` flag for `build` subcommand as required")
-	}
+	bundleBuildCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "",
+		"The name of the package that bundle image belongs to "+
+			"(Required if `directory` is not pointing to a bundle in the nested bundle format)")
 
-	bundleBuildCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "", "The list of channels that bundle image belongs to")
-	if err := bundleBuildCmd.MarkFlagRequired("channels"); err != nil {
-		log.Fatalf("Failed to mark `channels` flag for `build` subcommand as required")
-	}
+	bundleBuildCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "",
+		"The list of channels that bundle image belongs to"+
+			"(Required if `directory` is not pointing to a bundle in the nested bundle format)")
 
-	bundleBuildCmd.Flags().StringVarP(&imageBuilderArgs, "image-builder", "b", "docker", "Tool to build container images. One of: [docker, podman, buildah]")
+	bundleBuildCmd.Flags().StringVarP(&imageBuilderArgs, "image-builder", "b", "docker",
+		"Tool to build container images. One of: [docker, podman, buildah]")
 
-	bundleBuildCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "", "The default channel for the bundle image")
+	bundleBuildCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "",
+		"The default channel for the bundle image")
 
-	bundleBuildCmd.Flags().BoolVarP(&overwriteArgs, "overwrite", "o", false, "To overwrite annotations.yaml locally if existed. By default, overwrite is set to `false`.")
+	bundleBuildCmd.Flags().BoolVarP(&overwriteArgs, "overwrite", "o", false,
+		"To overwrite annotations.yaml locally if existed. By default, overwrite is set to `false`.")
 
-	bundleBuildCmd.Flags().StringVarP(&outputDirArgs, "output-dir", "u", "", "Optional output directory for operator manifests")
+	bundleBuildCmd.Flags().StringVarP(&outputDirArgs, "output-dir", "u", "",
+		"Optional output directory for operator manifests")
 
 	return bundleBuildCmd
 }
 
 func buildFunc(cmd *cobra.Command, args []string) error {
-	err := bundle.BuildFunc(dirBuildArgs, outputDirArgs, tagBuildArgs, imageBuilderArgs,
-		packageNameArgs, channelsArgs, channelDefaultArgs, overwriteArgs)
+	err := bundle.BuildFunc(dirBuildArgs, outputDirArgs, tagBuildArgs, imageBuilderArgs, packageNameArgs,
+		channelsArgs, channelDefaultArgs, overwriteArgs)
 	if err != nil {
 		return err
 	}
--- a/cmd/opm/alpha/bundle/generate.go
+++ b/cmd/opm/alpha/bundle/generate.go
@@ -24,24 +24,25 @@ func newBundleGenerateCmd() *cobra.Command {
 		RunE: generateFunc,
 	}
 
-	bundleGenerateCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "", "The directory where bundle manifests for a specific version are located.")
+	bundleGenerateCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "",
+		"The directory where bundle manifests for a specific version are located.")
 	if err := bundleGenerateCmd.MarkFlagRequired("directory"); err != nil {
 		log.Fatalf("Failed to mark `directory` flag for `generate` subcommand as required")
 	}
 
-	bundleGenerateCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "", "The name of the package that bundle image belongs to")
-	if err := bundleGenerateCmd.MarkFlagRequired("package"); err != nil {
-		log.Fatalf("Failed to mark `package` flag for `generate` subcommand as required")
-	}
+	bundleGenerateCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "",
+		"The name of the package that bundle image belongs to")
 
-	bundleGenerateCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "", "The list of channels that bundle image belongs to")
-	if err := bundleGenerateCmd.MarkFlagRequired("channels"); err != nil {
-		log.Fatalf("Failed to mark `channels` flag for `generate` subcommand as required")
-	}
+	bundleGenerateCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "",
+		"The name of the package that bundle image belongs to "+
+			"(Required if `directory` is not pointing to a bundle in the nested bundle format)")
 
-	bundleGenerateCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "", "The default channel for the bundle image")
+	bundleGenerateCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "",
+		"The list of channels that bundle image belongs to"+
+			"(Required if `directory` is not pointing to a bundle in the nested bundle format)")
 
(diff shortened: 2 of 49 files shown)


Bowenislandsong:
adding test!

cmd/opm/alpha/bundle/build.go
```diff
@@ -55,14 +55,8 @@ func newBundleBuildCmd() *cobra.Command {
 	}
 
 	bundleBuildCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "", "The name of the package that bundle image belongs to")
```
kevinrizza: IMO we need to update this help message to let users know that it's required if they're not pointing to a directory in the nested bundle format

pkg/lib/bundle/generate.go
```diff
@@ -76,6 +76,24 @@ func GenerateFunc(directory, outputDir, packageName, channels, channelDefault st
 		return err
 	}
 
+	if channels == "" || packageName == "" || channelDefault == "" {
+		I, err := NewBundleDirInterperter(filepath.Join(directory, ".."))
```
kevinrizza: nit: use lowercase `i, err :=`

pkg/sqlite/graph_loader.go
```diff
@@ -0,0 +1,159 @@
+package sqlite
```
Bowenislandsong: DO not review this file. This is @exdx and @gallettilance 's part

pkg/registry/csv.go
```diff
@@ -53,6 +60,49 @@ type ClusterServiceVersion struct {
 	Spec json.RawMessage `json:"spec"`
 }
 
+func NewFromBundleDirectory(bundleDir string) (*ClusterServiceVersion, error) {
```
kevinrizza: Can we change this function name to be more descriptive? NewFromBundleDirectory describes the mechanics of what the function is doing, not the purpose or the output. Maybe `ReadCSVFromBundleDirectory`?

pkg/registry/csv.go
```diff
@@ -53,6 +60,49 @@ type ClusterServiceVersion struct {
...
+	files, err := ioutil.ReadDir(bundleDir)
+	if err != nil {
+		return nil, fmt.Errorf("error reading bundle directory %s, %v", bundleDir, err)
+	}
+
+	csvFile, err := func() (string, error) {
+		for _, f := range files {
```
kevinrizza: I think there are more clever ways we could find this file. For example we can check all the files in the directory to see if they're yaml files, and then for each yaml file attempt to serialize it into a CSV.

Otherwise this will fail if they don't follow the naming convention described here

pkg/lib/bundle/generate.go
```diff
@@ -76,6 +76,24 @@ func GenerateFunc(directory, outputDir, packageName, channels, channelDefault st
...
+		if err != nil {
+			return fmt.Errorf("error interpretating bundle from directory %s, %v", directory, err)
```
kevinrizza: spelling nit: `interpreting`

Also, it feels like the error messaging here is going to be unclear to users. Can we be more descriptive? Right now it seems like we don't differentiate between "this bundle isn't in a directory with our special helpful case" and "you didn't give us enough information to handle this request". For example if they just didn't specify a list of channels in a bundle that's in its own directory, we shouldn't surface "unable to interpret bundle from directory". we should tell them "missing input parameters".

There's also the case where a user just doesn't give us a default channel for example that we were already handling and I think this removes that ability as well

pkg/registry/directoryGraphLoader.go
```diff
@@ -0,0 +1,223 @@
...
+	loader := DirGraphLoader{
+		PackageDir: packageDir,
+	}
+
+	err = loader.loadBundleCsvPathMap(packageDir)
+	if err != nil {
+		return nil, fmt.Errorf("error geting CSVs from all bundles in the package directory, %v", err)
+	}
+
+	loader.loadReplaces()
```
kevinrizza: Why isn't this function just part of the `loader.Generate()`? It seems like this will make it harder to mock if some of the implementation is inside the constructor.

pkg/registry/directoryGraphLoader.go
```diff
@@ -0,0 +1,223 @@
...
+// defaultchannel, and head of all Channels. It returns parsing error if any.
+func (g *DirGraphLoader) parsePackageYAMLFile() (*Package, error) {
+	files, err := ioutil.ReadDir(g.PackageDir)
+	if err != nil {
+		return nil, fmt.Errorf("error reading bundle parent directory, %v", err)
+	}
+
+	ymlFile, err := func() (string, error) {
+		for _, f := range files {
+			if matched, err := path.Match("*.package.yaml", f.Name()); err != nil {
```
kevinrizza: same comment as the parser for the csv, i think we can serialize this rather than rely on a naming convention

pkg/sqlite/graph_loader.go
kevinrizza: this is already merged into master with a new file name fyi: https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/graphloader.go#L11

pkg/registry/csv.go
gallettilance: there is no hard requirement on the name of the file...

pkg/sqlite/graph_loader.go
kevinrizza: It's because you merged their commits into your repo and are pulling them along, but i squashed their commits and then renamed this file to get rid of the underscore in the name. you need to remove their commits and rebase.

cmd/opm/alpha/bundle/build.go
```diff
@@ -44,33 +45,41 @@ func newBundleBuildCmd() *cobra.Command {
...
+
+	bundleBuildCmd.Flags().StringVarP(&imageBuilderArgs, "image-builder", "b", "docker",
+		"Tool to build container images. One of: [docker, podman, buildah]")
 
-	bundleBuildCmd.Flags().StringVarP(&imageBuilderArgs, "image-builder", "b", "docker", "Tool to build container images. One of: [docker, podman, buildah]")
+	bundleBuildCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "",
+		"The default channel for the bundle image")
 
-	bundleBuildCmd.Flags().StringVarP(&channelDefaultArgs, "default", "e", "", "The default channel for the bundle image")
+	bundleBuildCmd.Flags().BoolVarP(&noChannelDefaultArgs, "no-default", "e", false,
```
kevinrizza: This seems like a not great user experience. We shouldn't add complexity here if it's not necessary. IMO we should just attempt to set the default channel like we were doing before if we can't find a packages.yaml file instead of asking the user if they want an error or not.

cmd/opm/alpha/bundle/build.go
```diff
@@ -44,33 +45,41 @@ func newBundleBuildCmd() *cobra.Command {
...
+		"The name of the package that bundle image belongs to " +
+		"(Required if `directory` is not pointing to a bundel in the nested bundel format)")
 
-	bundleBuildCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "", "The list of channels that bundle image belongs to")
-	if err := bundleBuildCmd.MarkFlagRequired("channels"); err != nil {
-		log.Fatalf("Failed to mark `channels` flag for `build` subcommand as required")
-	}
+	bundleBuildCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "",
+		"The list of channels that bundle image belongs to" +
+		"(Required if `directory` is not pointing to a bundel in the nested bundel format)")
```
kevinrizza: spelling nit: `bundle`

cmd/opm/alpha/bundle/build.go
```diff
@@ -44,33 +45,41 @@ func newBundleBuildCmd() *cobra.Command {
...
 		log.Fatalf("Failed to mark `tag` flag for `build` subcommand as required")
 	}
 
-	bundleBuildCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "", "The name of the package that bundle image belongs to")
-	if err := bundleBuildCmd.MarkFlagRequired("package"); err != nil {
-		log.Fatalf("Failed to mark `package` flag for `build` subcommand as required")
-	}
+	bundleBuildCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "",
+		"The name of the package that bundle image belongs to " +
+		"(Required if `directory` is not pointing to a bundel in the nested bundel format)")
```
kevinrizza: spelling nit: `bundle`

cmd/opm/alpha/bundle/build.go
```diff
@@ -44,33 +45,41 @@ func newBundleBuildCmd() *cobra.Command {
 		RunE: buildFunc,
 	}
 
-	bundleBuildCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "", "The directory where bundle manifests and metadata for a specific version are located")
+	bundleBuildCmd.Flags().StringVarP(&dirBuildArgs, "directory", "d", "",
+		"The directory where bundle manifests and metadata for a specific version are located " +
+		"(Can be used for interpreting bundle information if in the nested bundle format.)")
```
kevinrizza: I think this last line is not necessary for this specific parameter.

cmd/opm/alpha/bundle/generate.go
```diff
@@ -24,30 +24,35 @@ func newBundleGenerateCmd() *cobra.Command {
...
+	bundleGenerateCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "",
+		"The name of the package that bundle image belongs to")
 
-	bundleGenerateCmd.Flags().StringVarP(&channelsArgs, "channels", "c", "", "The list of channels that bundle image belongs to")
-	if err := bundleGenerateCmd.MarkFlagRequired("channels"); err != nil {
-		log.Fatalf("Failed to mark `channels` flag for `generate` subcommand as required")
-	}
+	bundleGenerateCmd.Flags().StringVarP(&packageNameArgs, "package", "p", "",
+		"The name of the package that bundle image belongs to "+
+			"(Required if `directory` is not pointing to a bundel in the nested bundel format)")
```
kevinrizza: Same spelling nits as the build cmd

pkg/registry/directoryGraphLoader.go
```diff
@@ -0,0 +1,225 @@
...
+// defaultchannel, and head of all Channels. It returns parsing error if any.
+func (g *DirGraphLoader) parsePackageYAMLFile() (*Package, error) {
+	files, err := ioutil.ReadDir(g.PackageDir)
+	if err != nil {
+		return nil, fmt.Errorf("error reading bundle parent directory, %v", err)
+	}
+
+	ymlFiles := []string{}
+	for _, f := range files {
+		if matched, err := path.Match("*.yaml", f.Name()); err != nil {
```
kevinrizza: can we just check to see if it's yaml? we do something similar when we unpack bundles: https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/conversion.go#L67

Bowenislandsong:
> Also, can you include some unit tests?

yep working on that now... thats why it is still wip.

cmd/opm/alpha/bundle/build.go
Bowenislandsong: ? if we can not find the package.yaml then we have run the command. By then we would either successfully get the channel or that we didn't. If we got it we are writing it to the annotation file. If we didn't we would have already returned something. 

what do you mean by `then` we ask if users want error or not?  There isn't a second step command.

I can only think of returning a warning.

kevinrizza: Because before we didn't attempt to look for it and tried to guess it. Why not retain that implicitly without adding an extra parameter?

pkg/lib/bundle/generate.go
```diff
@@ -76,6 +76,28 @@ func GenerateFunc(directory, outputDir, packageName, channels, channelDefault st
 		return err
 	}
 
+	// Channels and packageName are required fields where as default channel is automatically filled if unspecified
+	// and that either of the required field is missing. We are interpreting the bundle information through
+	// bundle directory embedded in the package folder.
+	if channels == "" || packageName == "" {
+		i, err := NewBundleDirInterperter(filepath.Join(directory, ".."))
+		if err != nil {
+			return fmt.Errorf("please manually input channels and packageName, "+
```
kevinrizza: 👍

pkg/registry/csv.go
```diff
@@ -53,6 +62,50 @@ type ClusterServiceVersion struct {
...
+		if err != nil {
+			continue
+		}
+
+		unstructuredCSV := unstructured.Unstructured{}
+		csvStruct := ClusterServiceVersion{}
+
+		decoder := yaml.NewYAMLOrJSONDecoder(csvReader, 30)
+		if err = decoder.Decode(&unstructuredCSV); err != nil {
+			return nil, fmt.Errorf("error unmarshalling CSV, %v", err)
```
kevinrizza: but won't this return an error if the directory contains yaml files that aren't csvs?

pkg/registry/csv.go
```diff
@@ -53,6 +62,50 @@ type ClusterServiceVersion struct {
...
+func ReadCSVFromBundleDirectory(bundleDir string) (*ClusterServiceVersion, error) {
+	files, err := ioutil.ReadDir(bundleDir)
+	if err != nil {
+		return nil, fmt.Errorf("error reading bundle directory %s, %v", bundleDir, err)
+	}
+
+	ymlFiles := []string{}
+	for _, f := range files {
+		if !f.IsDir() {
+			ymlFiles = append(ymlFiles, f.Name())
```
kevinrizza: shouldn't we use the yaml decoder here to see if the file is actually a yaml file?

pkg/registry/csv.go
Bowenislandsong: I will name this better. This just unmarshals to unstructured. It then tries to determin if it is CSV next 
` 		if unstructuredCSV.GetKind() != operators.ClusterServiceVersionKind {`

pkg/registry/csv.go
Bowenislandsong: This just grabs all files. I will rename the variables. This gets all files, then it unmarshals to yaml. then it tries to see if it is csv.

PR #230, opened 2020-03-27
[travis] pinning minikube version

Bowenislandsong:
This commit sets the minikube version to v1.17.3 and matches go.mod.

**Reviewer Checklist**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ jobs:
       chmod +x minikube
       sudo mv minikube /bin/
       minikube config set vm-driver none
+      minikube config set kubernetes-version v1.17.3
   - os: osx
     install: |
       brew install make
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,6 +12,7 @@ jobs:
       chmod +x minikube
       sudo mv minikube /bin/
       minikube config set vm-driver none
+      minikube config set kubernetes-version v1.17.3
```
benluddy: It would be cool to set this based on go.mod, so we don't need to remember to change this in the future. Maybe using go list?
```
$ go list -f '{{.Version}}' -m k8s.io/api
v0.17.3
```

benluddy:
```suggestion
      minikube config set kubernetes-version $(go list -f '{{.Version}}' -m k8s.io/api | sed 's/^v0\./v1./')
```

Bowenislandsong: nice !!!

Bowenislandsong:
/close
superseded by https://github.com/operator-framework/operator-registry/pull/234 to bring 2 test fixes together so that the PR can pass both tests.

PR #234, opened 2020-03-27
Fixing ci and travis tests 

Bowenislandsong:
Kube 1.18 is released recently, which reveals that `operator-registry` should not relying on the latest kube-version available. To avoid adjusting the minikube version, we, therefore, are pining the version to go.mod k8s/api.

Accidentally, at the same time, the ci-operator changed their way of testing image build by using buildah instead of docker. This change loses the sudo privileges and therefore the permission change for dev is no longer possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ jobs:
       chmod +x minikube
       sudo mv minikube /bin/
       minikube config set vm-driver none
+      minikube config set kubernetes-version $(go list -f '{{.Version}}' -m k8s.io/api | sed 's/^v0\./v1./')
   - os: osx
     install: |
       brew install make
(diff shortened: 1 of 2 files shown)


Bowenislandsong:
/cc @ecordell @kevinrizza @dinhxuanvu
I don't really see why we need to change the permission for the dev/ directory, and neither do I see where it is used.
This is the commit im talking about https://github.com/operator-framework/operator-registry/pull/234/commits/bba9fc7ce6d529ba55c50cb1b85b7c843d442112

ecordell:
@Bowenislandsong travis is passing now that master is fixed - can you clean up the commits here so that we can merge?

Bowenislandsong:
> @Bowenislandsong travis is passing now that master is fixed - can you clean up the commits here so that we can merge?

yes, doing that right now.

PR #235, opened 2020-03-27
Bug 1809320: fix(registry): change db load to get providedAPIs via bundle content

dinhxuanvu:
Currently, when bundle is loaded into the database, the providedAPIs
are extracted from CSV instead of bundle content. As a result, if
CRDs are dropped in the bundle but not referenced in CSV, they won't
get added into the DB. This ProvidedAPIs info in DB then gets queried
via registry API and causes the dependency resolution discrepancy
between 4.2 and 4.3+.

**Reviewer Checklist**

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -159,8 +159,7 @@ func TestGetBundle(t *testing.T) {
 		Version:   "0.9.2",
 		SkipRange: "< 0.6.0",
 	}
-
-	require.Equal(t, expected, bundle)
+	compareBundle(t, expected, bundle)
 }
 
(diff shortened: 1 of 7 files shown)


ecordell:
Couple of comments - 

We need to rebase this since the image loader fixes merged - in particular, provided and required apis are now associated with the bundle, not the channel entry (that should make this PR simpler as well). Happy to help with that in the morning!

Also, this doesn’t seem to address the issue for existing bundles in the database (which would require a migration). Is that simply because we think the issue is so rare that it’s not worth a migration?

dinhxuanvu:
No longer needed as the changes were included in #236

Issue #205, opened 2020-03-03
opm: Adding bundle to registry with new channel fails and deletes previous bundles


cdjohnson:
Using opm 1.5.11

I'm trying to create a scenario where I add a second channel to a registry:

|  Version | Replaces  | Default Channel |  Channels | 
|---|---|---|---|
|  1.0.0 |  | BETA | BETA |
|  1.0.1 | 1.0.0 | STABLE | BETA,STABLE |

Commands in order:
1. `operator-sdk bundle create quay.io/cdjohnson/testoperator:v1.0.0 --directory ../deploy/olm-catalog/testoperator/1.0.0 --package test-operator --channels BETA --default-channel BETA`
2. `docker push quay.io/cdjohnson/testoperator:v1.0.0`
3. `opm registry add -b quay.io/cdjohnson/testoperator:v1.0.0 -c docker`

The `bundles.db` shows:
--channel-entry--
BETA|test-operator|testoperator.v1.0.0
1|BETA|test-operator|testoperator.v1.0.0||0
--operatorbundle--
testoperator.v1.0.0|quay.io/cdjohnson/testoperator:v1.0.0

4. `operator-sdk bundle create quay.io/cdjohnson/testoperator:v1.0.1 --directory ../deploy/olm-catalog/testoperator/1.0.1 --package test-operator --channels BETA,STABLE --default-channel BETA`
5. `docker push quay.io/cdjohnson/testoperator:v1.0.1`
6. `opm registry add -b quay.io/cdjohnson/testoperator:v1.0.1 -c docker`
```
INFO[0000] adding to the registry                        bundles="[quay.io/cdjohnson/testoperator:v1.0.1]"
DEBU[0000] couldn't rollback - this is expected if the transaction committed  error="sql: transaction has already been committed or rolled back"
INFO[0000] running docker pull                           img="quay.io/cdjohnson/testoperator:v1.0.1"
DEBU[0000] [docker pull quay.io/cdjohnson/testoperator:v1.0.1]  img="quay.io/cdjohnson/testoperator:v1.0.1"
INFO[0001] running docker save                           img="quay.io/cdjohnson/testoperator:v1.0.1"
DEBU[0001] [docker save quay.io/cdjohnson/testoperator:v1.0.1 -o bundle_staging_537743893/bundle.tar]  img="quay.io/cdjohnson/testoperator:v1.0.1"
[...]

PR #236, opened 2020-03-28
fix(loader): Image Loader doesn't create the same channel entries as the other loaders.

ecordell:
This began as an investigation into #205 - odd behavior when adding a new bundle via `index add`

The issue there was the image loader was not properly handling the case (as described in #205) where a new bundle being added both replaces something in an existing channel, and starts a new channel as well. 

I removed the assumption that a `replace`s exists in the current channel, which was enough to prevent the bug in #205. I wrote a test in `image_test.go` to verify the behavior with the new graph representation.

That test made me realize that the image loader was not building graphs the same way as the other loaders - because the image loader doesn't synthesize nodes outside the current channel, the graph comes out missing some nodes. In particular, we would fail to answer correctly "does a replacement for X exist in channel A" even if there is an operator in channel A with a replaces field that says X. 

The fact that the behaviors diverge at all is somewhat concerning, so I looked at what it would take to have the image loader and the other loaders share the same graph building implementation. There were two blockers:

1. When the "old" loaders build the graph, they expect to be able to query the CSV for its `replaces` field, which is not always present anymore when using bundle images.
2. Provided and Required APIs are associated with a channel entry, not a bundle (this was done originally to remove a join from the most common query path). The simplest thing to do to keep the loaders the same was to update the graph by deleting the existing graph for a package and recalculating it - but deleting the graph would lose the provided and required api data, because they were tied to the graph nodes and not the bundle.

I started by addressing those two things:

- The first commit adds a migration that associates provided and required apis with the bundle directly
- The second commit adds a migration that extracts replaces / skips from the bundle when adding it.
- The third commit adds a test to verify the correct graph output for the #205 case, and refactors the image loader to re-use the existing load paths.

Follow up work:

There should be some indication that the migration may be incomplete if bundlepaths are used. I think this is somewhat okay at the moment because there are not a lot of bundle images being used. But existing indexes would need to `rm` the package and re-add all of the bundle images so that the replaces/skips fields are unpacked correctly. We should have some standard way to handle this going forward and surface options to the user for whether they want bundle pulls to occur for migrations.

**Reviewer Checklist**


ecordell:
Tested with some real bundles:

```sh
$ opm index add -c docker -b quay.io/shawn_hurley/testoperator:v1.0.0 --tag quay.io/ecordell/test-index:0
INFO[0000] building the index                            bundles="[quay.io/shawn_hurley/testoperator:v1.0.0]"
INFO[0000] running docker pull                           img="quay.io/shawn_hurley/testoperator:v1.0.0"
INFO[0000] running docker save                           img="quay.io/shawn_hurley/testoperator:v1.0.0"
INFO[0000] loading Bundle quay.io/shawn_hurley/testoperator:v1.0.0  img="quay.io/shawn_hurley/testoperator:v1.0.0"
INFO[0000] found annotations file searching for csv      dir=bundle_tmp215707306 file=bundle_tmp215707306/metadata load=annotations
INFO[0000] found csv, loading bundle                     dir=bundle_tmp215707306 file=bundle_tmp215707306/manifests load=bundle
INFO[0000] loading bundle file                           dir=bundle_tmp215707306/manifests file=testoperator.v1.0.0.clusterserviceversion.yaml load=bundle name=testoperator.v1.0.0
INFO[0000] Generating dockerfile                         bundles="[quay.io/shawn_hurley/testoperator:v1.0.0]"
INFO[0000] writing dockerfile: index.Dockerfile857028972  bundles="[quay.io/shawn_hurley/testoperator:v1.0.0]"
INFO[0000] running docker build                          bundles="[quay.io/shawn_hurley/testoperator:v1.0.0]"
INFO[0000] [docker build -f index.Dockerfile857028972 -t quay.io/ecordell/test-index:0 .]  bundles="[quay.io/shawn_hurley/testoperator:v1.0.0]"

$ docker push quay.io/ecordell/test-index:0

$ opm index add -c docker -f quay.io/ecordell/test-index:0 -b quay.io/shawn_hurley/testoperator:v1.0.1 --tag quay.io/ecordell/test-index:1
INFO[0000] building the index                            bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0000] Pulling previous image quay.io/ecordell/test-index:0 to get metadata  bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0000] running docker pull                           bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0000] Getting label data from previous image        bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0000] running docker inspect                        bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0000] running docker pull                           bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0001] running docker save                           bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0043] running docker pull                           img="quay.io/shawn_hurley/testoperator:v1.0.1"
INFO[0044] running docker save                           img="quay.io/shawn_hurley/testoperator:v1.0.1"
INFO[0044] loading Bundle quay.io/shawn_hurley/testoperator:v1.0.1  img="quay.io/shawn_hurley/testoperator:v1.0.1"
INFO[0044] found annotations file searching for csv      dir=bundle_tmp494931583 file=bundle_tmp494931583/metadata load=annotations
INFO[0044] found csv, loading bundle                     dir=bundle_tmp494931583 file=bundle_tmp494931583/manifests load=bundle
INFO[0044] loading bundle file                           dir=bundle_tmp494931583/manifests file=testoperator.v1.0.1.clusterserviceversion.yaml load=bundle name=testoperator.v1.0.1
INFO[0044] Generating dockerfile                         bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0044] writing dockerfile: index.Dockerfile925925129  bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0044] running docker build                          bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
INFO[0044] [docker build -f index.Dockerfile925925129 -t quay.io/ecordell/test-index:1 .]  bundles="[quay.io/shawn_hurley/testoperator:v1.0.1]"
```

Channel Entry table:

entry_id | channel_name | package_name | operatorbundle_name | replaces | depth
---------|---------------|----------------|-----------------------|----------|-------
1|BETA|test-operator|testoperator.v1.0.1|2|0
2|BETA|test-operator|testoperator.v1.0.0||1
3|STABLE|test-operator|testoperator.v1.0.1|4|0
4|STABLE|test-operator|testoperator.v1.0.0||1

pkg/sqlite/directory_test.go
```diff
@@ -231,3 +232,10 @@ func TestQuerierForDirectory(t *testing.T) {
 	require.NoError(t, err)
 	require.ElementsMatch(t, expectedDatabaseImages, dbImages)
 }
+
+func EqualBundles(t *testing.T, expected, actual api.Bundle) {
+	require.ElementsMatch(t, expected.ProvidedApis, actual.ProvidedApis)
+	require.ElementsMatch(t, expected.RequiredApis, actual.RequiredApis)
+	expected.RequiredApis, expected.ProvidedApis, actual.RequiredApis, actual.ProvidedApis = nil, nil, nil, nil
+	require.EqualValues(t, expected, actual)
+}
```
njhale: nit: missing endline

pkg/sqlite/migrations/006_associate_apis_with_bundle.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for requiredRows.Next() {
+		var group sql.NullString
+		var apiVersion sql.NullString
+		var kind sql.NullString
+		var name sql.NullString
+		var bundleVersion sql.NullString
+		var path sql.NullString
```
njhale: nit: block declaration

pkg/registry/interface.go
```diff
@@ -8,9 +8,9 @@ import (
 
 type Load interface {
 	AddOperatorBundle(bundle *Bundle) error
-	AddBundlePackageChannels(manifest PackageManifest, bundle Bundle) error
 	AddPackageChannels(manifest PackageManifest) error
-	RmPackageName(packageName string) error
+	AddBundlePackageChannels(manifest PackageManifest, bundle *Bundle) error
+	RemovePackage(packageName string) error
```
njhale: 👍

pkg/sqlite/migrations/007_replaces_skips.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+		_, err = tx.ExecContext(ctx, foreignKeyOn)
+		return err
+	},
+}
+
+func extractReplaces(ctx context.Context, tx *sql.Tx, name string) error {
+	addSql := `insert into operatorbundle(replaces, skips) values(?,?)`
+
+	// TODO: what should we do when there are bundle images?
```
njhale: Can we infer replaces and skips from the channel_entries table?

pkg/sqlite/migrations/007_replaces_skips_test.go
```diff
@@ -0,0 +1,58 @@
...
+	require.NoError(t, err)
+
+	// run down migration
+	err = migrator.Down(context.TODO(), migrations.Only(migrations.ReplacesSkipsMigrationKey))
+	require.NoError(t, err)
+
+	// check that bundle data is fine
+	var name sql.NullString
+	var csv sql.NullString
+	var bundle sql.NullString
```
njhale: nit: block declaration?

pkg/sqlite/load.go
```diff
@@ -454,74 +449,59 @@ func (s *SQLLoader) getCSV(tx *sql.Tx, csvName string) (*registry.ClusterService
...
+		return fmt.Errorf("cannot add apis for bundle with no name: %#v", bundle)
+	}
 	addAPI, err := tx.Prepare("insert or replace into api(group_name, version, kind, plural) values(?, ?, ?, ?)")
 	if err != nil {
 		return err
 	}
 	defer addAPI.Close()
 
-	addAPIProvider, err := tx.Prepare("insert into api_provider(group_name, version, kind, channel_entry_id) values(?, ?, ?, ?)")
+	addApiProvider, err := tx.Prepare("insert into api_provider(group_name, version, kind, operatorbundle_name, operatorbundle_version, operatorbundle_path) values(?, ?, ?, ?, ?, ?)")
```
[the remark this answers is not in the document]
ecordell: Good question - the graph loader assumes the combination of these three fields is unique, so I thought it would be good to match those assumptions here. I think it will come down to what "operatorbundle_name" means when we don't have CSVs in the bundle.

pkg/sqlite/migrations/007_replaces_skips.go
ecordell: Discussed offline, but persisting here. I was thinking we couldn't distinguish between replaces and skips in the channel entry table, but after some discussion we should be able to distinguish between them based on depth.

pkg/sqlite/migrations/006_associate_apis_with_bundle.go
```diff
@@ -0,0 +1,236 @@
...
+		`
+		_, err = tx.ExecContext(ctx, renameNewAndDropOld)
+		if err != nil {
+			return err
+		}
+
+		return err
+	},
+	Down: func(ctx context.Context, tx *sql.Tx) error {
+		// TODO
```
exdx: what happens if the migration fails? should we do anything with this `TODO`?

exdx:
Spent a good deal of time going over the PR with @ecordell - thanks for all the insight. Pulling out the skips and replaces (via the new schema updates) and rebuilding the image loader graph for each new bundle will also help with the bundle image commutativity work that we've been looking at @gallettilance @kevinrizza. 

/lgtm

PR #239, opened 2020-03-29
Revert "[feature] Understanding packages from directory"

Bowenislandsong:
Reverts operator-framework/operator-registry#229


ecordell:
/lgtm

force-merging to fix #229, then let's get CI fixed before anything else merges

PR #241, opened 2020-03-29
[feature] Understanding packages from bundle directory

Bowenislandsong:
This commit adds a feature to interpret bundle name and channels
based on a given bundle directory and update bundle's annotation
YAML. This feature is triggered when opm alpha generate/build is
not supplying packageName, channels, and defaultChannels information.
This Bundle directory needs to be inside a Package directory
that contains all bundles.

The previous reverted commit 13373cb43e72cb3b025e6f693fffccf94c13d817
accidentally changed opm command.
This commit removes the duplicated package flag.
This commit corrects bundle dir from pakcgae directory.
This commit adds an error message for getting empty channels for a bundle.
This commit skipes error for trying to parse anyfile as Json or Yaml format.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/generate.go
+++ b/pkg/lib/bundle/generate.go
@@ -76,6 +76,45 @@ func GenerateFunc(directory, outputDir, packageName, channels, channelDefault st
 		return err
 	}
 
+	// Channels and packageName are required fields where as default channel is automatically filled if unspecified
+	// and that either of the required field is missing. We are interpreting the bundle information through
+	// bundle directory embedded in the package folder.
+	if channels == "" || packageName == "" {
+		var notProvided []string
+		if channels == "" {
+			notProvided = append(notProvided, "channels")
+		}
+		if packageName == "" {
+			notProvided = append(notProvided, "package name")
+		}
+		log.Infof("Bundle %s information not provided, inferring from parent package directory",
+			strings.Join(notProvided, " and "))
+
+		i, err := NewBundleDirInterperter(directory)
+		if err != nil {
+			return fmt.Errorf("please manually input channels and packageName, "+
+				"error interpreting bundle from directory %s, %v", directory, err)
+		}
+
+		if channels == "" {
+			channels = strings.Join(i.GetBundleChannels(), ",")
+			if channels == "" {
+				return fmt.Errorf("error interpreting channels, please manually input channels instead")
+			}
+			log.Infof("Inferred channels: %s", channels)
+		}
+
+		if packageName == "" {
+			packageName = i.GetPackageName()
+			log.Infof("Inferred package name: %s", packageName)
+		}
+
+		if channelDefault == "" {
+			channelDefault = i.GetDefaultChannel()
+			log.Infof("Inferred default channel: %s", channelDefault)
+		}
+	}
+
 	log.Info("Building annotations.yaml")
 
 	// Generate annotations.yaml
--- a/pkg/lib/bundle/interpreter.go
+++ b/pkg/lib/bundle/interpreter.go
@@ -1,0 +1,54 @@
+package bundle
+
+import (
+	"fmt"
+	"path"
+	"sort"
+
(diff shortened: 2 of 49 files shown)


pkg/lib/bundle/interpreter.go
```diff
@@ -0,0 +1,54 @@
...
+	pkg           *registry.Package
+}
+
+func NewBundleDirInterperter(bundleDir string) (*bundleDirInterpreter, error) {
+	csv, err := registry.ReadCSVFromBundleDirectory(bundleDir)
+	if err != nil {
+		return nil, fmt.Errorf("error loading CSV from bundle directory, %v", err)
+	}
+
+	pkgDir, err := registry.NewPackageGraphLoaderFromDir(path.Join(bundleDir, ".."))
```
ecordell: I'm not sure this is what we want to do here.

This reads in the directory of files and generates a graph, and tells you if there are problems.

But the generate command is used for a single bundle. That single bundle may be applied to any target index, and has nothing to do with the bundles that happen to be in in the filesystem next to it.

I think this would be fine without the graph loader in this PR at all. Thoughts?

kevinrizza: But I thought the whole purpose of this though is to automagically decide what channels a given bundle is. How can we do that without attempting to build the graph for a given package?

Bowenislandsong: Same ask!

ecordell: Thanks for the explanation - commented before ☕️ !

I think there's a slight impedance mismatch here - this option is now only really useful if you're attempting convert an entire directory of the old format, right?

Should this really be part of the normal command for generating/buliding single bundles when it's so tied to the set?

kevinrizza: I think the trigger for it should be "if you didn't pass in these parameters explicitly, let's go see if we can find them for you" rather than it being an explicit opt in/opt out feature. IMO that's a better experience

Bowenislandsong: I second that this should be an experience enhancement type of feature based on the fact that you have the full-blown format with bundles nested in a package along with other bundles. Without it, it does not have any information we would be looking for. The `default channel` and `channels` are only obtainable in the `*.package.yaml` file. To figure out which channels a bundle belongs to require relationships of this bundle with respect to every channel heads and see if it is directly/transitively replaced by those channel heads to be included as part of that channel. Without the package and rest of the bundles, all we can be certain from a single bundle is the `package name`.

After all, this feature is just auto-filling `package name`, `channels`, and `default channels` for a bundle.

pkg/lib/bundle/generate.go
```diff
@@ -76,6 +76,45 @@ func GenerateFunc(directory, outputDir, packageName, channels, channelDefault st
...
+	// bundle directory embedded in the package folder.
+	if channels == "" || packageName == "" {
+		var notProvided []string
+		if channels == "" {
+			notProvided = append(notProvided, "channels")
+		}
+		if packageName == "" {
+			notProvided = append(notProvided, "package name")
+		}
+		log.Infof("Bundle %s information not provided, inferring from parent package directory",
```
Bowenislandsong, line 90: log/notification added at the info level to suggestion inference took place.

pkg/lib/bundle/generate.go
```diff
@@ -76,6 +76,45 @@ func GenerateFunc(directory, outputDir, packageName, channels, channelDefault st
...
+			return fmt.Errorf("please manually input channels and packageName, "+
+				"error interpreting bundle from directory %s, %v", directory, err)
+		}
+
+		if channels == "" {
+			channels = strings.Join(i.GetBundleChannels(), ",")
+			if channels == "" {
+				return fmt.Errorf("error interpreting channels, please manually input channels instead")
+			}
+			log.Infof("Inferred channels: %s", channels)
```
Bowenislandsong, line 104: The inference results are also posted.

PR #247, opened 2020-04-02
Semver index insert

kevinrizza:
(feat) Semver insert modes

This adds a new flag to `opm index add` and `opm registry add` which
defines two new insert modes to add bundles to the existing update graph

Rather than rely on parsing the CSV for the replaces field to define the
channel update graph explicitly, in these new modes `opm` implicitly
infers the update graph based on the version attached to the bundle and
semantic versioning rules (https://semver.org/#summary)

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -25,6 +25,7 @@ type AddToRegistryRequest struct {
 	SkipTLS       bool
 	InputDatabase string
 	Bundles       []string
+	Mode          registry.Mode
 }
 
 func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
@@ -44,6 +45,11 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		return err
 	}
 
+	graphLoader, err := sqlite.NewSQLGraphLoaderFromDB(db)
+	if err != nil {
+		return err
+	}
+
 	// TODO: Dependency inject the registry if we want to swap it out.
 	reg, destroy, err := containerdregistry.NewRegistry(
 		containerdregistry.SkipTLS(request.SkipTLS),
@@ -59,7 +65,7 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 
 	// TODO(njhale): Parallelize this once bundle add is commutative
 	for _, ref := range request.Bundles {
-		if err := populate(context.TODO(), dbLoader, reg, image.SimpleReference(ref)); err != nil {
+		if err := populate(context.TODO(), dbLoader, graphLoader, reg, image.SimpleReference(ref), request.Mode); err != nil {
 			err = fmt.Errorf("error loading bundle from image: %s", err)
 			if !request.Permissive {
 				r.Logger.WithError(err).Error("permissive mode disabled")
@@ -73,7 +79,7 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 	return utilerrors.NewAggregate(errs) // nil if no errors
 }
 
-func populate(ctx context.Context, loader registry.Load, reg image.Registry, ref image.Reference) error {
+func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, reg image.Registry, ref image.Reference, mode registry.Mode) error {
 	workingDir, err := ioutil.TempDir("./", "bundle_tmp")
 	if err != nil {
 		return err
@@ -88,9 +94,9 @@ func populate(ctx context.Context, loader registry.Load, reg image.Registry, ref
 		return err
 	}
 
-	populator := registry.NewDirectoryPopulator(loader, ref, workingDir)
+	populator := registry.NewDirectoryPopulator(loader, graphLoader, ref, workingDir)
 
-	return populator.Populate()
+	return populator.Populate(mode)
 }
 
 type DeleteFromRegistryRequest struct {
--- a/pkg/registry/bundlegraphloader.go
+++ b/pkg/registry/bundlegraphloader.go
@@ -1,0 +1,147 @@
+package registry
+
+import (
+	"fmt"
+
+	"github.com/blang/semver"
+)
+
+// BundleGraphLoader generates updated graphs by adding bundles to them, updating
+// the graph implicitly via semantic version of each bundle
+type BundleGraphLoader struct {
+}
+
+// AddBundleToGraph takes a bundle and an existing graph and updates the graph to insert the new bundle
+// into each channel it is included in
+func (g *BundleGraphLoader) AddBundleToGraph(bundle *Bundle, graph *Package, newDefaultChannel string, skippatch bool) (*Package, error) {
+	bundleVersion, err := bundle.Version()
+	if err != nil {
+		return nil, fmt.Errorf("Unable to extract bundle version from bundle %s, can't insert in semver mode", bundle.BundleImage)
+	}
+
+	versionToAdd, err := semver.Make(bundleVersion)
+	if err != nil {
+		return nil, fmt.Errorf("Bundle version %s is not valid", bundleVersion)
+	}
+
+	newBundleKey := BundleKey{
+		CsvName:    bundle.Name,
+		Version:    versionToAdd.String(),
+		BundlePath: bundle.BundleImage,
+	}
+
+	// initialize the graph if it started empty
+	if graph.Name == "" {
+		graph.Name = bundle.Package
+	}
+	if newDefaultChannel != "" {
+		graph.DefaultChannel = newDefaultChannel
+	}
+
+	// generate the DAG for each channel the new bundle is being insert into
+	for _, channel := range bundle.Channels {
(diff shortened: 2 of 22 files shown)


pkg/lib/registry/registry.go
```diff
@@ -20,9 +21,78 @@ type AddToRegistryRequest struct {
...
 
 func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
+	// TODO: Update private functions to pass along this db connection after migrating
+	db, err := sql.Open("sqlite3", request.InputDatabase)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	// TODO: decouple loader from migrator to call migrate separately
```
ecordell, line 35: fyi, migrator is already decoupled and `loader.Migrate` is just a convenience

pkg/registry/bundlegraphloader.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	newBundleKey := BundleKey{
+		CsvName:    bundle.Name,
+		Version:    versionToAdd.String(),
+		BundlePath: bundle.BundleImage,
+	}
+
+	// for each channel the new bundle is being insert into
+	for _, channel := range bundle.Channels {
```
ecordell, line 34: nit: this is getting long with lots of nested conditionals / loops, could use a bit of cleanup

pkg/registry/bundlegraphloader.go
```diff
@@ -0,0 +1,122 @@
...
+				channelGraph.Nodes = make(map[BundleKey]map[BundleKey]struct{}, 1)
+			}
+
+			lowestAhead := BundleKey{}
+			greatestBehind := BundleKey{}
+
+			// Iterate over existing nodes and compare the new node's version to find the
+			// lowest version above it and highest version below it (to insert between these nodes)
+			for node := range channelGraph.Nodes {
+				fmt.Println(node)
```
ecordell, line 50: nit: prints

pkg/sqlite/graphloader.go
```diff
@@ -9,55 +9,81 @@ import (
...
+	return fmt.Sprintf("Package %s not found in database", e.pkg)
+}
+
+func NewPackageNotInDatabaseError(pkg string) error {
+	return &PackageNotInDatabaseError{
+		pkg: pkg,
+	}
+}
+
+func IsPackageNotInDatabaseError(err error) bool {
```
ecordell, line 48: since we're on 1.13 now: https://golang.org/pkg/errors/#Is

pkg/sqlite/graphloader.go
```diff
@@ -77,6 +103,12 @@ func graphFromEntries(channelEntries []registry.ChannelEntryAnnotated) (map[stri
 			Version:    entry.Version,
 			CsvName:    entry.BundleName,
 		}
+
+		// skip synthetic channelentries that aren't pointers to actual bundles
+		if _, ok := existingBundles[key]; !ok {
```
ecordell, line 108: why are these skipped?

pkg/sqlite/load.go
```diff
@@ -113,39 +113,140 @@ func (s *SQLLoader) AddPackageChannels(manifest registry.PackageManifest) error
...
+			// It should only contain one real replacement, so let's find it and
+			// follow the chain. For the rest, they are fake entries and should be
+			// generated as synthetic replacements
+			for replace := range channel.Nodes[currentNode] {
+				if _, ok := channel.Nodes[replace]; !ok {
+					syntheticReplaces = append(syntheticReplaces, replace)
+				} else {
+					nextNode = replace
+				}
+			}
```
ecordell, line 188: we should catch when the graph isn't what we expect here - where we don't find any "real" replace in the graph

pkg/sqlite/graphloader.go
kevinrizza: This is specifically for `skips`. If the bundle is referenced but does not actually exist in the database on its own we shouldn't have a graph entry for it, just a reference in the replaces map.

pkg/sqlite/load.go
kevinrizza: Is there a difference between "no real" replaces and "no real replaces but has some synthetic replaces"? What about the case where there is no replacement but the first node in the graph has skips defined? Are we saying that's not valid?

pkg/registry/empty.go
```diff
@@ -80,6 +80,10 @@ func (EmptyQuery) GetBundlePathsForPackage(ctx context.Context, pkgName string)
 	return nil, errors.New("empty querier: cannot get images")
 }
 
+func (EmptyQuery) GetBundlesForPackage(ctx context.Context, pkgName string) (map[BundleKey]struct{}, error) {
+	return nil, errors.New("empty querier: cannot get images")
```
ecordell, line 84:
```suggestion
	return nil, errors.New("empty querier: cannot get bundles")
```

pkg/sqlite/load.go
ecordell: It looks to me like there is a difference -

We expect `channel.Nodes[replaces]` to exist, but not `channel.Nodes[skips]`

This block will find the `replaces` node if it exists, and add all the `skip` nodes if they exist - which is good. But if the `replaces` node doesn't exist, this will set `replaces` to whatever happens to be the last `skips` node, which would be incorrect (especially if elsewhere we assume `replaces` exists in `channel.Nodes`)

kevinrizza: I don't think it will. If the replacement node isn't in the graph, it will just never set `nextNode` at all and assume the graph ended (since just before this I initialized `nextNode` as an empty node). In semver insert mode, this implies that either we really are at the last node OR we aren't at the last node and the real replacement wasn't defined properly in the graph. In that case, we can check that the actual replacement chain is as long as the set of real nodes defined in the graph. If they aren't, it means we got to the end early.

But, when I was thinking about this I realized that this code is non-deterministic for graphs built with `skippatch` mode -- in that mode there are multiple real candidates for `nextNode` and the graph isn't being inserted properly. Right now this is definitely broken. So I also need to update the `nextNode` to a list of nodes to iterate over and insert

pkg/registry/bundlegraphloader_test.go
```diff
@@ -0,0 +1,345 @@
...
+							}`),
+				},
+				Channels: []string{"alpha", "stable"},
+			},
+			skipPatch: true,
+			expectedGraph: &Package{
+				Name:           "etcd",
+				DefaultChannel: "alpha",
+				Channels: map[string]Channel{
+					"alpha": {Head: BundleKey{CsvName: "etcdoperator.v0.9.3", Version: "0.9.3"},
```
ecordell, line 298: what's the right graph for skippatch mode? there shouldn't be 0.9.0 or 0.9.2 entry, right? 

Should alpha just be:

0.9.3 -> (0.9.0, 0.9.1, 0.9.2, 0.6.1)

0.9.3 skips 0.9.2, so it should skip 0.9.1 (skipped by 0.9.2) transitively
0.9.3 skips 0.9.0, so it should skip 0.6.1 (replaced by 0.9.0) transitively (I think this is arguable)

either way I don't think there should be an entry for 0.9.2, since it's definitively skipped in skippatch mode.

Does this definition of skippatch mode make sense: "If I add 0.Y.Z in skip patch mode, anything that would have updated to any 0.Y.W should instead update directly to 0.Y.Z"?

pkg/registry/bundlegraphloader_test.go
```diff
@@ -0,0 +1,345 @@
...
+func TestBundleGraphLoader(t *testing.T) {
+	tests := []struct {
+		fail              bool
+		graph             Package
+		bundle            Bundle
+		newDefaultChannel string
+		expectedGraph     *Package
+		skipPatch         bool
+	}{
+		{ // Add bundle to head of channels when replaces is defined
```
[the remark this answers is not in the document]
kevinrizza: Yeah it's a fair point. My intention was to ensure (via a codified test) that the bundle graph loader wasn't using other input (like the bundle image path) to derive the replaces field defined for existing operators. But I'm fine with just deleting this test given that the input is effectively the same with that assumption.

pkg/registry/bundlegraphloader_test.go
kevinrizza: Spoke a little about this with Evan, my opinion is that non-latest z versions should not be considered "real" nodes in the graph in skip-patch mode.

Updated the pr to reflect that in the bundle updater.

pkg/registry/channelupdateoptions.go
```diff
@@ -0,0 +1,26 @@
...
+
+func GetModeFromString(mode string) Mode {
+	switch strings.ToLower(mode) {
+	case "replaces":
+		return ReplacesMode
+	case "semver":
+		return SemVerMode
+	case "semver-skippatch":
+		return SkipPatchMode
+	default:
```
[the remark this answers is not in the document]
kevinrizza: I sort of disagree, because just not setting it gives you default behavior you describe. If you type `opm index add --mode=seemver` and it just attempts to insert in replaces mode I would be very confused.

PR #248, opened 2020-04-02
Fix panic in registry add

njhale:

- Fix panic during `registry add`
- Set missing bundle name to prevent `registry add` from failing


Bowenislandsong:
/hold 
for @awgreene review

PR #254, opened 2020-04-06
feat(registry): Add operator version dependency metadata

dinhxuanvu:
The operator-version based dependency information can be added to
a bundle via dependencies.yaml in /metadata. The dependencies info
will be parsed and is added to the DB in a new table named
`dependencies`. Each dependency entry is linked to a specific operator
bundle that it belongs to.

A new registry API `ListBundles` is added to return all available
bundles in the database. This information is to serve as input to
the new resolver.

**Reviewer Checklist**


pkg/sqlite/load.go
```diff
@@ -61,7 +61,7 @@ func (s *SQLLoader) AddOperatorBundle(bundle *registry.Bundle) error {
 }
 
 func (s *SQLLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error {
-	addBundle, err := tx.Prepare("insert into operatorbundle(name, csv, bundle, bundlepath, version, skiprange, replaces, skips) values(?, ?, ?, ?, ?, ?, ?, ?)")
+	addBundle, err := tx.Prepare("insert into operatorbundle(name, csv, bundle, bundlepath, version, skiprange, replaces, skips, dependencies) values(?, ?, ?, ?, ?, ?, ?, ?, ?)")
```
kevinrizza, line 64: Is this the right place to put this? I would like to see how the querier interface works here as well before we add it as just a serialized blob in a field on the bundle. Doesn't it make more sense to denormalize this somewhat and create a separate table on the database so that you can write queries explicitly about a set or subset of dependency information?

pkg/sqlite/image.go
```diff
@@ -0,0 +1,259 @@
+package sqlite
```
ecordell, line 1: this file was deleted from master, it now lives in the `registry` package

pkg/sqlite/migrations/008_dependencies.go
```diff
@@ -0,0 +1,47 @@
...
+			version TEXT,
+			kind TEXT,
+			operatorbundle_name TEXT,
+			operatorbundle_version TEXT,
+			operatorbundle_path TEXT,
+			FOREIGN KEY(operatorbundle_name, operatorbundle_version, operatorbundle_path) REFERENCES operatorbundle(name, version, bundlepath) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED,
+			FOREIGN KEY(type, package_name, group_name, version, kind) REFERENCES dependencies(type, package_name, group_name, version, kind) ON DELETE CASCADE
+		);
+		`
+		_, err := tx.ExecContext(ctx, sql)
```
ecordell, line 32: this will need a data migration as well, to populate this table with data derived from the existing entries (i.e. api_requirer -> GVK dependency)

ecordell:
@dinhxuanvu I think we probably also want some equivalent to the "api_provider" table for dependencies as well. Is this something that you have planned already?

pkg/sqlite/migrations/008_dependencies.go
dinhxuanvu: Kevin and I talked about this yesterday and our initial thought was to do this later. However, given your input, it seems it is best just do it now. I'll address this.

dinhxuanvu:
> @dinhxuanvu I think we probably also want some equivalent to the "api_provider" table for dependencies as well. Is this something that you have planned already?

We did talk about `provided` section in `dependencies.yaml` during planning but we never worked out the details. In the enhancement, we did ack the thought about this feature but said we deferred this to later release.

kevinrizza:
> > @dinhxuanvu I think we probably also want some equivalent to the "api_provider" table for dependencies as well. Is this something that you have planned already?
> 
> We did talk about `provided` section in `dependencies.yaml` during planning but we never worked out the details. In the enhancement, we did ack the thought about this feature but said we deferred this to later release.

Right now all of our dependencies are implicit in the packaging. If we want to define a "dependency providers" table now, it would just be derived from other stuff we are already including in the database. We definitely don't need to update the dependency.yaml spec to force users to offer that data again when they already have

pkg/registry/types.go
```diff
@@ -114,6 +114,71 @@ type Annotations struct {
...
+type Dependency struct {
+	// The type of dependency. It can be `olm.package` for operator-version based
+	// dependency or `olm.gvk` for gvk based dependency. This field is required.
+	Type string `json:"type" yaml:"type"`
+
+	// The name of dependency such as 'etcd'. This field is required.
+	Name string `json:"name" yaml:"name"`
+
+	// The group of GVK based dependency.
+	Group string `json:"group,omitempty" yaml:"group"`
```
kevinrizza, line 133: Should we have explicit fields here for each dependency type? Can we just have a `dependencyIdentifier` as a raw json object or something similar so that we don't have to hardcode all of these and ignore certain fields depending on the type of dependency?

pkg/registry/types.go
```diff
@@ -114,6 +114,71 @@ type Annotations struct {
...
+	Dependencies []Dependency `json:"dependencies" yaml:"dependencies"`
+}
+
+// Dependencies is a list of dependencies for a given bundle
+type Dependency struct {
+	// The type of dependency. It can be `olm.package` for operator-version based
+	// dependency or `olm.gvk` for gvk based dependency. This field is required.
+	Type string `json:"type" yaml:"type"`
+
+	// The name of dependency such as 'etcd'. This field is required.
```
kevinrizza, line 129: nit: `packageName`?

pkg/sqlite/query.go
```diff
@@ -752,3 +780,103 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
...
+		out.SkipRange = skipRange.String
+
+		provided, required, err := s.GetApisForEntry(ctx, entryID.Int64)
+		if err != nil {
+			return nil, err
+		}
+		out.ProvidedApis = provided
+		out.RequiredApis = required
+
+		dependencies, err := s.GetDependenciesForBundle(ctx, bundleName.String, version.String, bundlePath.String)
```
kevinrizza, line 835: Do we need to run a separate query here for each row? Can't we just do a join up front to get this table data as well?

pkg/registry/types.go
dinhxuanvu: I thought I removed `omitempty` part already (will do). I want a list of concrete fields that registry will parse and `opm` can validate. When the dependencies information are returned back along with api.Bundle, OLM simply retrieves the fields via .GetSomething() func instead of having to parse the JSON object.

pkg/sqlite/query.go
dinhxuanvu: We have `one to many` problem here with operatorbundle to multiple dependencies similar to the API table. The trade off will be a complex nested loop to read multiple rows of the same operatorbundle but with each dependency.

pkg/sqlite/migrations/008_dependencies.go
ecordell: It's probably best to do it now so that we don't have to worry about a db in a "half migrated' state

pkg/registry/types.go
kevinrizza: I'm not arguing that the input serialization of this object can't look like this (because that's what unioning these fields looks like on the yaml object), but it seems like we could translate it in the database. If for every single new type we add to the dependencies we have to add a set of fields, that database table will get really unwieldy really fast. Can we make it slightly more generic on the other end is what I'm really asking

pkg/sqlite/query.go
kevinrizza: From a performance perspective the cost is definitely going to be higher doing that many queries. Can't we just have the aggregation be a row per dependency and iterate over them one at a time? We have to do that bit anyway

pkg/registry/types.go
dinhxuanvu: I think we either stick with explicit fields or a big JSON blob. Doing it half-half will only force us to do more DB migration like we do now to get more info inside CSV into separate fields. A big JSON blob will limit our ability to query dependencies using its fields. I think the existing fields should serve us well for foreseeable future. I can see one more field `provided` and that should be it. If we end up with adding more fields, we can always split the big tables into smaller tables that handle a specific dependency type.

docs/design/operator-bundle.md
```diff
@@ -64,6 +65,25 @@ annotations:
...
+
+The dependency list will contain a `type` field for each item to specify what kind of dependency this is. There are two supported `type` of operator dependencies. It can be a package type (`olm.package`) meaning this is a dependency for a specific operator version. For `olm.package` type, the dependency information should include the `package` name and the `version` of the package in semver format. We use `blang/semver` library for semver parsing (https://github.com/blang/semver). For example, you can specify an exact version such as `0.5.2` or a range of version such as `>0.5.1` (https://github.com/blang/semver#ranges). In addition, the author can specify dependency that is similiar to existing CRD/API-based using `olm.gvk` type and then specify GVK information as how it is done in CSV. This is a path to enable operator authors to consolidate all dependencies (API or explicit version) to be in the same place.
```
Bowenislandsong, line 72: are the olm.gvk intended for specifying required CRD/APIs? AKA, CRD/APIs not provided by the bundle? Or is this a generic type for both `owned` and `required`?

pkg/api/registry.proto
```diff
@@ -37,6 +38,14 @@ message GroupVersionKind{
 	string plural = 4;
 }
 
+message Dependency{
```
ecordell, line 41: Since we will have multiple sets of fields based on the type of the dependency, we should do one of the following:

1. Have one type field, and another "value" field, that we parse in a second step, i.e.:

```proto
message Dependency{
  string type = 1;
  string value = 2;
}
```
```go
if dep.type == "olm.gvk" {
   gvktype = parseGVKValue(dep.value)
}
```

2. Use a union type to support future field variants, i.e.:

```proto
message Dependency {
  string type = 1;
  oneof deptype {
     GVKDependency gvk = 2;
     PackageDependency pkg = 3;
  }
}

message GVKDependency {
  string group = 1;
  string version = 2;
  string kind = 3;
}

message PackageDependency {
  string name = 1;
  string version = 2;
}
```

pkg/api/registry.proto
```diff
@@ -48,6 +57,7 @@ message Bundle{
 	repeated GroupVersionKind requiredApis = 8;
 	string version = 9;
 	string skipRange = 10;
+  repeated Dependency dependencies = 11;
```
ecordell, line 60: nit: spacing

pkg/registry/populator.go
```diff
@@ -46,30 +50,41 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
...
 	annotationsFile := &AnnotationsFile{}
+	dependenciesFile := &DependenciesFile{}
 	for _, f := range files {
 		err = decodeFile(filepath.Join(metadata, f.Name()), annotationsFile)
 		if err != nil || *annotationsFile == (AnnotationsFile{}) {
+			log.Info("found annotations file searching for csv")
 			continue
 		}
-		log.Info("found annotations file searching for csv")
+		if f.Name() == DependenciesFileName {
```
ecordell, line 65: Didn't we discuss identifying the dependency file based on some identifier in the file?

pkg/registry/types.go
```diff
@@ -114,6 +114,71 @@ type Annotations struct {
...
+	Name string `json:"packageName" yaml:"packageName"`
+
+	// The group of GVK based dependency
+	Group string `json:"group" yaml:"group"`
+
+	// The kind of GVK based dependency
+	Kind string `json:"kind" yaml:"kind"`
+
+	// The version of dependency in semver format
+	Version string `json:"version" yaml:"version"`
```
ecordell, line 139: Similar comments to the what I mentioned for protobuf. We either want a `value` bag to put things in, or we want a set of different types.

To highlight the problem: Version here is `semver` if `type=package` but `string` (specifically `kube apiversion string`) if `type=gvk`

pkg/server/server.go
```diff
@@ -31,6 +31,20 @@ func (s *RegistryServer) ListPackages(req *api.ListPackageRequest, stream api.Re
 	return nil
 }
 
+func (s *RegistryServer) ListBundles(req *api.ListBundlesRequest, stream api.Registry_ListBundlesServer) error {
+	bundles, err := s.store.ListBundles(stream.Context())
+	if err != nil {
+		return err
+	}
+	for _, b := range bundles {
+		if err := stream.Send(b); err != nil {
```
ecordell, line 40: +1 streaming

pkg/sqlite/migrations/008_dependencies.go
```diff
@@ -0,0 +1,111 @@
...
+
+var dependenciesMigration = &Migration{
+	Id: DependenciesMigrationKey,
+	Up: func(ctx context.Context, tx *sql.Tx) error {
+		sql := `
+		CREATE TABLE IF NOT EXISTS dependencies (
+			type TEXT,
+			package_name TEXT,
+			group_name TEXT,
+			version TEXT,
```
ecordell, line 25: This schema has the same issue as the protobuf and the go types

pkg/sqlite/migrations/008_dependencies.go
```diff
@@ -0,0 +1,111 @@
...
+
+		insertRequired := `INSERT INTO dependencies(type, package_name, group_name, version, kind, operatorbundle_name, operatorbundle_version, operatorbundle_path) VALUES (?, ?, ?, ?, ?, ?, ?, ?)`
+
+		bundleApis, err := getRequiredAPIs(ctx, tx)
+		if err != nil {
+			return err
+		}
+		for bundle, apis := range bundleApis {
+			for required := range apis.required {
+				_, err := tx.ExecContext(ctx, insertRequired, "olm.gvk", nil, required.Group, required.Version, required.Kind, bundle.CsvName, bundle.Version, bundle.BundlePath)
```
ecordell, line 46: nice!

pkg/sqlite/query.go
```diff
@@ -752,3 +780,157 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
...
+			dep.Version = depVersion.String
+			dep.Kind = depKind.String
+			dependencies = append(dependencies, dep)
+			out.Dependencies = dependencies
+
+			bundlesMap[bundleKey] = out
+		}
+	}
+
+	for _, v := range bundlesMap {
```
ecordell, line 877: instead of building a list of deps -> converting to a map to dedupe -> converting back to a list, why not build up the set of dependencies as a map to start with, and then do 1 converstion to a list before assigning to out.Dependencies?

pkg/sqlite/query.go
```diff
@@ -752,3 +780,157 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
...
+	}
+
+	return
+}
+
+func unique(deps []*api.Dependency) []*api.Dependency {
+	keys := make(map[string]bool)
+	list := []*api.Dependency{}
+	for _, entry := range deps {
+		depKey := fmt.Sprintf("%s/%s/%s/%s/%s", entry.Type, entry.Name, entry.Group, entry.Version, entry.Kind)
```
ecordell, line 892: this shouldn't be needed, you can use entry as a key directly

dinhxuanvu: Using a pointer as a key is not helpful for key lookup in this case. Not sure if I missed something but entry is a pointer.

pkg/sqlite/query.go
dinhxuanvu: The map is to dedupe `api.Bundle` as we have a one (bundle) to many (dependency) relationship. Each row has one bundle to one of many dependency. As a result, there are multiple rows of same bundle (one for reach dependency). Multiple bundles can have the same dependency so a map of dependencies is not helpful unless you meant using value as a list of bundles. I'm not sure if I misunderstood something here.

pkg/registry/populator.go
```diff
@@ -16,6 +17,10 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/image"
 )
 
+type Dependencies struct {
+	DependenciesStr []map[string]string `json:"dependencies" yaml:"dependencies"`
```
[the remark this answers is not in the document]
dinhxuanvu: I fixed the naming. 
Given the dependency syntax is a list of fields (not nested):
```
dependencies:
  - type: olm.package
    packgeName: prometheus
    version: >0.27.0
  - type: olm.gvk
    group: etcd.database.coreos.com
    kind: EtcdCluster
    version: v1beta2
```
I use `Dependencies` struct to get the JSON string of each item (which will be injected into `Value` in `DependenciesFile`) while using `DependenciesFile` to get the `Type`. This is to avoid using nested structure in the syntax. We already agree on the `Dependency` type only contains `Type` and `Value`. When client/user gets the string from `Value`, they can unmarshal into `GVKDependency` or `PackageDependency` type that I specify in `types.go` in `registry` package.

pkg/registry/populator_test.go
```diff
@@ -333,6 +339,82 @@ func TestImageLoading(t *testing.T) {
...
+	db, cleanup := CreateTestDb(t)
+	defer cleanup()
+
+	store, err := createAndPopulateDB(db)
+	require.NoError(t, err)
+
+	expectedDependencies := []*api.Dependency{
+		{
+			Type:  "olm.package",
+			Value: `{"packageName":"testoperator","type":"olm.package","version":"\u003e 0.2.0"}`,
```
ecordell, line 353: This looks like an encoding bug - why is it a unicode escape sequence?

pkg/registry/populator_test.go
```diff
@@ -333,6 +339,82 @@ func TestImageLoading(t *testing.T) {
...
+		{
+			name:    "etcdoperator.v0.9.2",
+			version: "0.9.2",
+			path:    "quay.io/test/etcd.0.9.2",
+		},
+		{
+			name:    "prometheusoperator.0.22.2",
+			version: "0.22.2",
+			path:    "quay.io/test/prometheus.0.22.2",
+		},
```
ecordell, line 377: Can you add test cases for bundles that you didn't add a dependencies.yaml for? 

We should be able to pick up existing api dependencies, since they should be migrated by the migration.

pkg/sqlite/directory_test.go
```diff
@@ -134,7 +134,8 @@ func TestQuerierForDirectory(t *testing.T) {
...
 			"{\"apiVersion\":\"apiextensions.k8s.io/v1beta1\",\"kind\":\"CustomResourceDefinition\",\"metadata\":{\"name\":\"etcdrestores.etcd.database.coreos.com\"},\"spec\":{\"group\":\"etcd.database.coreos.com\",\"names\":{\"kind\":\"EtcdRestore\",\"listKind\":\"EtcdRestoreList\",\"plural\":\"etcdrestores\",\"singular\":\"etcdrestore\"},\"scope\":\"Namespaced\",\"version\":\"v1beta2\"}}",
 		},
-		BundlePath: "",
+		BundlePath:   "",
+		Dependencies: []*api.Dependency{},
```
ecordell, line 138: This should be non-empty if the migration is working, right?

This test case also has:

```go
	RequiredApis: []*api.GroupVersionKind{
			{Group: "etcd.database.coreos.com", Version: "v1beta2", Kind: "EtcdCluster", Plural: "etcdclusters"},
			{Group: "etcd.database.coreos.com", Version: "v1beta2", Kind: "EtcdCluster", Plural: "etcdclusters"},
		},
```

pkg/sqlite/load.go
```diff
@@ -747,3 +753,27 @@ func (s *SQLLoader) AddBundlePackageChannels(manifest registry.PackageManifest,
 
 	return tx.Commit()
 }
+
+func (s *SQLLoader) addDependencies(tx *sql.Tx, bundle *registry.Bundle) error {
```
ecordell, line 757: This should also parse the CSV (if present) and find required apis (if present) and add them as dependencies

pkg/sqlite/migrations/008_dependencies.go
```diff
@@ -0,0 +1,119 @@
...
+		);
+		`
+		_, err := tx.ExecContext(ctx, sql)
+		if err != nil {
+			return err
+		}
+
+		insertRequired := `INSERT INTO dependencies(type, value, operatorbundle_name, operatorbundle_version, operatorbundle_path) VALUES (?, ?, ?, ?, ?)`
+
+		bundleApis, err := getRequiredAPIs(ctx, tx)
```
ecordell, line 38: based on the test cases, I think there are bugs with this

pkg/sqlite/migrations/008_dependencies.go
```diff
@@ -0,0 +1,119 @@
...
+					"type":    "olm.gvk",
+					"group":   required.Group,
+					"version": required.Version,
+					"kind":    required.Kind,
+				}
+				value, err := json.Marshal(valueMap)
+				if err != nil {
+					return err
+				}
+				_, err = tx.ExecContext(ctx, insertRequired, "olm.gvk", value, bundle.CsvName, bundle.Version, bundle.BundlePath)
```
ecordell, line 54: the dep type identifiers should probably be constants (olm.gvk, etc)

PR #256, opened 2020-04-06
feat: generate a bundle with configmap and secret objects

exdx:
Adds support for users putting configmap and secret objects inside their bundle. Part of the work to increase the flexibility of OLM for users. OLM knows how to install secrets from an install plan currently, but support for configmaps will be added as well. 

Currently no validation for these new objects is done (outside of them being valid kube objects) but they are whitelisted and will be installed when OLM installs the bundle. 

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/supported_resources.go
+++ b/pkg/lib/bundle/supported_resources.go
@@ -6,6 +6,7 @@ const (
 	SecretKind             = "Secret"
 	ClusterRoleKind        = "ClusterRole"
 	ClusterRoleBindingKind = "ClusterRoleBinding"
+	ConfigMapKind          = "ConfigMap"
 	ServiceAccountKind     = "ServiceAccount"
 	ServiceKind            = "Service"
 	RoleKind               = "Role"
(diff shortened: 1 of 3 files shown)


pkg/lib/bundle/supported_resources.go
```diff
@@ -32,6 +33,8 @@ func init() {
 	supportedResources[RoleBindingKind] = true
 	supportedResources[PrometheusRuleKind] = true
 	supportedResources[ServiceMonitorKind] = true
+	supportedResources[SecretKind] = true
+	supportedResources[ConfigMapKind] = true
```
njhale, line 37: nit: can this be a map literal and not in an init?

njhale: nit: I would also create a named type or some constants for the values here; e.g. `Namespaced, NotNamespaced`.

pkg/lib/bundle/testdata/validate/valid_bundle_additional_objects/manifests/etcdrestores.etcd.database.coreos.com.crd.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
njhale, line 1: What do you think about adding these manifests to existing valid bundles instead of creating a new directory for them?

kevinrizza: +1

exdx: Makes sense. I was being cautious with the idea "not to modify fixtures that existing tests rely on" but it's probably not relevant in this simple case.

Issue #258, opened 2020-04-06
opm 1.7.0: All channels deleted except for latest with opm registry add


cdjohnson:
When attempting to add additional channels, any previous channels are deleted.  Only the latest channel is preserved.

I believe this is a regression introduced in PR #236 as a result of issue #205 

1. Add a bundle to to the BETA channel:
```
operator-sdk bundle create quay.io/cdjohnson/brokenoperator:v1.0.0 --directory 1.0.0 --package test-operator --channels BETA --default-channel BETA
docker push quay.io/cdjohnson/brokenoperator:v1.0.0
opm registry add -b quay.io/cdjohnson/brokenoperator:v1.0.0 -c docker

sqlite3 bundles.db "select * from channel;"
BETA|test-operator|testoperator.v1.0.0

sqlite3 bundles.db "select * from channel_entry;"
1|BETA|test-operator|testoperator.v1.0.0||0

sqlite3 bundles.db "select name from operatorbundle;"
testoperator.v1.0.0
```

2. Add a second bundle to the STABLE channel where `olm.skipRange: <1.0.1`
```
operator-sdk bundle create quay.io/cdjohnson/brokenoperator:v1.0.1 --directory 1.0.1 --package test-operator --channels STABLE --default-channel STABLE
docker push quay.io/cdjohnson/brokenoperator:v1.0.1
opm registry add -b quay.io/cdjohnson/brokenoperator:v1.0.1 -c docker


sqlite3 bundles.db "select * from channel;"
STABLE|test-operator|testoperator.v1.0.1

sqlite3 bundles.db "select * from channel_entry;"
1|STABLE|test-operator|testoperator.v1.0.1||0

sqlite3 bundles.db "select name from operatorbundle;"
testoperator.v1.0.0
testoperator.v1.0.1
```

**Result:**  The BETA channel has been removed from both the `channel` and `channel_entry` table.

This SHOULD result in two separate channels and channel_entries.

[bundles.zip](https://github.com/operator-framework/operator-registry/files/4441177/bundles.zip)

ecordell:
Thanks for reporting this regression @cdjohnson - just cut 1.8.0 which should fix this issue.

narasimha-ntnx:
Hi, I am still facing this issue. I am using opm 1.23.

PR #260, opened 2020-04-07
fix(imageloader): image loader should consider existing channels during add

ecordell:
This queries the database for existing channels before adding an image and replacing its graph.

Also,  #213 moved the image loader to `registry` and renamed it, but didn't clean up the old image loader. I added the fix to the new location and cleaned up the old location.

This fixes the regression reported in #258

In addition to the tests added in this PR, I also verified manually with the manifests in the report:

```sh
$ opm registry add -b quay.io/ecordell/brokenoperator:v1.0.0 -c docker
$ sqlite3 bundles.db "select * from channel;"
beta|test-operator|testoperator.v1.0.0

$ opm registry add -b quay.io/ecordell/brokenoperator:v1.0.0 -c docker
$ sqlite3 bundles.db "select * from channel;"
beta|test-operator|testoperator.v1.0.0
stable|test-operator|testoperator.v1.0.1
```

**Reviewer Checklist**


pkg/registry/populator_test.go
```diff
@@ -1,84 +1,70 @@
-package sqlite
+package registry_test
 
 import (
 	"context"
 	"database/sql"
+	"fmt"
+	"github.com/operator-framework/operator-registry/pkg/registry"
```
njhale, line 7: nit: goimports

PR #272, opened 2020-04-15
fix(registry) Update the addDependencies to add requiredAPIs

dinhxuanvu:
RequiredAPIs information in CSV will be extracted and added to
dependencies table as gvk type dependency when operator is loaded
into the registry DB.

Signed-off-by: Vu Dinh <vdinh@redhat.com>

**Description of the change:**

**Motivation for the change:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #274, opened 2020-04-15
fix(query): ListBundles returns all bundles even without dependencies

dinhxuanvu:
ListBundles query uses INNER JOIN which only intersection meaning it only
returns bundles with dependencies. Switching to LEFT OUTER JOIN to return
all bundles with or without dependencies.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator_test.go
+++ b/pkg/registry/populator_test.go
@@ -441,6 +441,11 @@ func TestListBundles(t *testing.T) {
 	store, err := createAndPopulateDB(db)
 	require.NoError(t, err)
 
+	var count int
+	row := db.QueryRow("SELECT COUNT(*) FROM operatorbundle")
+	err = row.Scan(&count)
+	require.NoError(t, err)
+
 	expectedDependencies := []*api.Dependency{
 		{
 			Type:  "olm.package",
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
Waiting for #272 to merge first.

pkg/registry/populator_test.go
```diff
@@ -455,10 +460,16 @@ func TestListBundles(t *testing.T) {
...
-		dep := b.Dependencies
-		dependencies = append(dependencies, dep...)
+		count++
+		for _, d := range b.Dependencies {
+			if d.GetType() != "" {
+				dependencies = append(dependencies, d)
+			}
+		}
 	}
+	require.Equal(t, bundleCount, count)
```
ecordell, line 472: `requilre.Equal(t, count, len(bundles)` would work too

dinhxuanvu:
Actually, this PR can be merged independently without #274 so go ahead and remove `/hold` when ready. Thanks @ecordell @benluddy

benluddy:
/lgtm

Thanks!

dinhxuanvu:
Replaced by #280

PR #277, opened 2020-04-16
Add dependencies validation to opm

dinhxuanvu:

Add dependencies.yaml validation to opm

Add dependencies validation checks to bundle library in registry

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/validate.go
+++ b/pkg/lib/bundle/validate.go
@@ -56,7 +56,7 @@ func (i imageValidator) PullBundleImage(imageTag, directory string) error {
 // Outputs:
 // error: ValidattionError which contains a list of errors
 func (i imageValidator) ValidateBundleFormat(directory string) error {
-	var manifestsFound, metadataFound bool
+	var manifestsFound, metadataFound, annotationsFound, dependenciesFound bool
 	var metadataDir, manifestsDir string
 	var validationErrors []error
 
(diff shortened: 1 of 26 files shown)


pkg/lib/bundle/validate.go
```diff
@@ -165,6 +193,61 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
 		validationErrors = append(validationErrors, err)
 	}
 
+	// Validate dependencies if exists
+	for _, d := range dependenciesFile.Dependencies {
+		switch d.GetType() {
+		case "olm.gvk":
```
ecordell, line 199: don't we have constants for these?

pkg/lib/bundle/validate.go
```diff
@@ -165,6 +193,61 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
+			gvkDep := &registry.GVKDependency{}
+			if d.GetValue() == "" {
```
ecordell, line 201: this can be a followup, but I think this would be more readable if it were refactored a bit:

```go
ValidatorFor(d.GetType().Validate()
```

dinhxuanvu:
@ecordell Ready for final review

pkg/lib/bundle/validate.go
```diff
@@ -99,15 +99,43 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
+		}
 	}
 
-	var fileAnnotations AnnotationMetadata
+	if len(dependenciesFile.Dependencies) < 1 {
+		i.logger.Info("Could not find dependencies file")
+	}
+
+	if fileAnnotations.Annotations == nil {
+		validationErrors = append(validationErrors, fmt.Errorf("Could not find annotations file"))
```
benluddy, line 130: Is there supposed to be a `return` here? The next line logs that the annotations file was found.

dinhxuanvu: It was supposed to be an `else` statement there. Fixed.

pkg/lib/bundle/validate.go
```diff
@@ -99,16 +98,57 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
 	if err != nil {
-		fmtErr := fmt.Errorf("Unable to read annotations.yaml file: %s", err.Error())
-		validationErrors = append(validationErrors, fmtErr)
+		validationErrors = append(validationErrors, err)
+	}
+
+	// Look for the metadata and manifests sub-directories to find the annotations file
+	fileAnnotations := &AnnotationMetadata{}
+	dependenciesFile := &registry.DependenciesFile{}
+	for _, f := range files {
```
benluddy, line 110: This loop seems to be dependent on the guarantee that `ioutil.ReadDir` sorts its return value by filename (so annotations.yaml always comes before dependencies.yaml), and that fileAnnotations won't be modified by DecodeFile when passed a dependencies file. Can it be refactored a bit to be clearer and to require fewer assumptions?

pkg/lib/bundle/validate.go
```diff
@@ -160,16 +191,35 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
+	for _, d := range dependenciesFile.Dependencies {
+		dep := d.GetTypeValue()
+		errs := []error{}
+		if dep != nil {
+			switch d := dep.(type) {
+			case registry.GVKDependency:
+				errs = d.Validate()
+			case registry.PackageDependency:
+				errs = d.Validate()
+			}
```
benluddy, line 215: Does it make sense to add a default switch case that returns an error? I imagine that this is coupled with GetTypeValue and it would probably be a bug if GetTypeValue were extended without adding a corresponding case here.

PR #273, opened 2020-04-15
Add ListBundles to the client interface.

benluddy:

PR #280, opened 2020-04-16
fix(registry): Fix dependency load/query and add ListBundles to registry client

dinhxuanvu:
Combination PR for #272 #273 and #274 

Fix `addDependencies` method to load requiredAPIs from bundle.
Fix ListBundles query to return all bundles with or without dependencies
Add ListBundles() to registry client.

**Reviewer Checklist**


ecordell:
passed upstream ci, e2e-aws is broken at the moment.

PR #281, opened 2020-04-17
fix(index-add): docker index add failing

ecordell:
Fixes an error in `index add` with docker.

This is a quick fix, there are several issues with the current way we do this.

**Reviewer Checklist**


ecordell:
e2e is broken, upstream passed. merging.

PR #283, opened 2020-04-17
Bug 1825925: support deamonless export

ecordell:
`opm index extract` can now be used without specifying a containertool. podman and docker can still be used as a fallback.

This also includes a fix that pulls and unpacks the previous index image into a separate directory from where the database is ultimately modified and added to the new images

opm index extract is useful to run in image builds, so having an unprivileged way to extract enables that.

Index adds were excruciatingly slow, sending 1GB+ images into the build context.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/index/export.go
+++ b/cmd/opm/index/export.go
@@ -1,7 +1,6 @@
 package index
 
 import (
-	"fmt"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"k8s.io/kubectl/pkg/util/templates"
@@ -46,7 +45,7 @@ func newIndexExportCmd() *cobra.Command {
 		logrus.Panic("Failed to set required `package` flag for `index export`")
 	}
 	indexCmd.Flags().StringP("download-folder", "f", "downloaded", "directory where downloaded operator bundle(s) will be stored")
-	indexCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [none, docker, podman]")
+	indexCmd.Flags().StringP("container-tool", "c", "none", "tool to interact with container images (save, build, etc.). One of: [none, docker, podman]")
 	if err := indexCmd.Flags().MarkHidden("debug"); err != nil {
 		logrus.Panic(err.Error())
 	}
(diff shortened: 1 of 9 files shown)


cmd/opm/index/export.go
```diff
@@ -46,7 +45,7 @@ func newIndexExportCmd() *cobra.Command {
 		logrus.Panic("Failed to set required `package` flag for `index export`")
 	}
 	indexCmd.Flags().StringP("download-folder", "f", "downloaded", "directory where downloaded operator bundle(s) will be stored")
-	indexCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [none, docker, podman]")
+	indexCmd.Flags().StringP("container-tool", "c", "none", "tool to interact with container images (save, build, etc.). One of: [none, docker, podman]")
```
exdx, line 48: nit: maybe we could go with `""` instead of `none` for this case?

pkg/lib/indexer/indexer.go
```diff
@@ -115,151 +141,201 @@ type DeleteFromIndexRequest struct {
...
 	}
-	defer os.RemoveAll(tmpDir)
 
-	databaseFile, err := i.getDatabaseFile(tmpDir, request.FromIndex)
-	if err != nil {
+	// set a temp directory for unpacking an image
+	// this is in its own function context so that the deferred cleanup runs before we do a docker build
+	// which prevents the full contents of the previous image from being in the build context
+	var databasePath string
+	if err := func () error {
```
njhale, line 154: nit: if this is identical to the previous closure, could we just make it a helper function?

pkg/containertools/dockerfilegenerator.go
```diff
@@ -48,7 +48,7 @@ func (g *IndexDockerfileGenerator) GenerateIndexDockerfile(binarySourceImage, da
 	dockerfile += fmt.Sprintf("LABEL %s=%s\n", DbLocationLabel, DefaultDbLocation)
 
 	// Content
-	dockerfile += fmt.Sprintf("ADD %s/index.db %s\n", databaseFolder, DefaultDbLocation)
```
njhale, line 51: This was still just adding a single file, right? I'm not sure how that could have been including the extra context.

ecordell: It was including it in the docker build context, not the final image.

pkg/containertools/dockerfilegenerator.go
```diff
@@ -48,7 +48,7 @@ func (g *IndexDockerfileGenerator) GenerateIndexDockerfile(binarySourceImage, da
...
+	dockerfile += fmt.Sprintf("ADD %s %s\n", databasePath, DefaultDbLocation)
```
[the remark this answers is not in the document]
ecordell: I just left it as it was. We should probably be using COPY but that should be a separate change, IMO.

pkg/containertools/dockerfilegenerator.go
njhale: I see now, thanks.

pkg/containertools/dockerfilegenerator.go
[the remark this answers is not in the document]
njhale: That works.

PR #285, opened 2020-04-17
(feat): Bundle commutativity

kevinrizza:
`opm registry add` and `opm index add`, when presented with multiple
bundles, required that those bundles be in version order when inserting
in replaces mode. This requirement existed because bundles are inserted
into the database in order, and if one of these bundles replaces another
the db schema would reject the insert.

This commit updates the replaces mode insert logic to check the existing
package graph and attempt to insert bundles in order.

As a followup to this commit, once we define a method of inserting in
replaces mode based on a graph input, we should be able to reduce this
to a single db transaction by defining the graph as a set of the
existing db + new input bundles. This would also simplify error handling,
which would no longer need any database transactions.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"github.com/operator-framework/operator-registry/pkg/image"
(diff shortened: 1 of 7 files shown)


pkg/registry/populator.go
```diff
@@ -26,124 +27,91 @@ type DirectoryPopulator struct {
...
-		return fmt.Errorf("error checking provided apis in bundle %s: %s", bundle.Name, err)
-	}
-
+func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode) error {
 	switch mode {
 	case ReplacesMode:
-		err = i.loadManifestsReplaces(bundle, annotationsFile)
-		if err != nil {
-			return err
+		// TODO: This is relatively inefficient. Ideally, we should be able to use a replaces
```
ecordell, line 70: This comment is also my main comment - we have a lot of the lower level stuff in place to do this really efficiently now. 

But I don't think that needs to block

kevinrizza: yeah, looking at it it was just going to take a lot more time and effort to reimplement some of the sql layer to do that, so I opted for this initial implementation to get it out there and kicked the can down the road somewhat.

as for a test, not explicitly called out, but the root db populate now generates the database with out of order input: https://github.com/operator-framework/operator-registry/pull/285/files#diff-bac4eb64785fb97243acdc492eedd162R70

PR #289, opened 2020-04-20
Bug 1826766: fix(types): Fix GetDependencies func that returns duplicate items

dinhxuanvu:
The iterator loop reuses item pointer which leads to duplicate of the
same dependencies while other dependencies are missing out.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/types.go
+++ b/pkg/registry/types.go
@@ -196,7 +196,8 @@ func (pd *PackageDependency) Validate() []error {
 func (d *DependenciesFile) GetDependencies() []*Dependency {
 	var dependencies []*Dependency
 	for _, item := range d.Dependencies {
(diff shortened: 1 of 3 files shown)


pkg/registry/types.go
```diff
@@ -196,7 +196,9 @@ func (pd *PackageDependency) Validate() []error {
 func (d *DependenciesFile) GetDependencies() []*Dependency {
 	var dependencies []*Dependency
 	for _, item := range d.Dependencies {
-		dependencies = append(dependencies, &item)
+		dep := Dependency{}
+		dep = item
```
ecordell, line 200:
```suggestion
		dep := item
```

dinhxuanvu: Fixed.

PR #290, opened 2020-04-21
update the help info of the export subcommand

jianzhangbjz:

Update the help info. Remove this sentence: `This command requires access to docker or podman to complete successfully.` And change the flag `-o` to `-p`. After this change, it looks like:
```console
mac:operator-registry jianzhang$ ./opm index export --help
Export an operator from an index image into the appregistry format.

 This command will take an index image (specified by the --index option), parse it for the given operator (set by the --package option) and export the operator metadata into an appregistry compliant format (a package.yaml file).

 Note: the appregistry format is being deprecated in favor of the new index image and image bundle format.

Usage:
  opm index export [flags]

Flags:
  -c, --container-tool string    tool to interact with container images (save, build, etc.). One of: [none, docker, podman] (default "none")
  -f, --download-folder string   directory where downloaded operator bundle(s) will be stored (default "downloaded")
  -h, --help                     help for export
  -i, --index string             index to get package from
  -p, --package string           the package to export
```

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/index/export.go
+++ b/cmd/opm/index/export.go
@@ -14,7 +14,6 @@ var exportLong = templates.LongDesc(`
 
 	This command will take an index image (specified by the --index option), parse it for the given operator(s) (set by 
 	the --package option) and export the operator metadata into an appregistry compliant format (a package.yaml file). 
-	This command requires access to docker or podman to complete successfully.
 
 	Note: the appregistry format is being deprecated in favor of the new index image and image bundle format. 
 	`)


jianzhangbjz:
Related bug: https://bugzilla.redhat.com/show_bug.cgi?id=1825925

cmd/opm/index/export.go
```diff
@@ -40,7 +39,7 @@ func newIndexExportCmd() *cobra.Command {
 	if err := indexCmd.MarkFlagRequired("index"); err != nil {
 		logrus.Panic("Failed to set required `index` flag for `index export`")
 	}
-	indexCmd.Flags().StringP("package", "o", "", "the package to export")
+	indexCmd.Flags().StringP("package", "p", "", "the package to export")
```
kevinrizza, line 42: Can we make this an alias? If users are already using `-o` this will break them.

kevinrizza (dismissed):
Just had one comment about the flag change. Also, this pr needs the bugzilla reference in the header. Otherwise, this looks good.

cmd/opm/index/export.go
jianzhangbjz: @kevinrizza Thanks for your review! I'm sorry, but I'm not familiar with the [cobra](https://github.com/spf13/cobra), do you have any suggestion to add the `alias` for a flag? Something like below:
```console
Flags:
  -p, -o --package string           the package to export
```

njhale: @jianzhangbjz If you're unable to find a way to alias this, I suggest removing this client facing API change.

njhale:
/approve
/hold 

^until the user facing API change is removed

cmd/opm/index/export.go
jianzhangbjz: @njhale Ok, good idea, thanks!

jianzhangbjz:
@njhale Remove it, thank you!

jianzhangbjz:
@njhale Could you help add the `/lgtm` label? Thanks!

PR #295, opened 2020-04-23
Bug 1829945: support v1 CRDs in bundles and indexes

exdx:
Support loading v1 CRDs from a bundle into the index. 

**Reviewer Checklist**


pkg/registry/bundle.go
```diff
@@ -117,12 +128,25 @@ func (b *Bundle) ProvidedAPIs() (map[APIKey]struct{}, error) {
...
 	}
-	for _, crd := range crds {
-		for _, v := range crd.Spec.Versions {
-			provided[APIKey{Group: crd.Spec.Group, Version: v.Name, Kind: crd.Spec.Names.Kind, Plural: crd.Spec.Names.Plural}] = struct{}{}
-		}
-		if crd.Spec.Version != "" {
-			provided[APIKey{Group: crd.Spec.Group, Version: crd.Spec.Version, Kind: crd.Spec.Names.Kind, Plural: crd.Spec.Names.Plural}] = struct{}{}
+
+	for _, c := range crds {
+		crd, ok := c.(*apiextensionsv1.CustomResourceDefinition)
```
dinhxuanvu, line 133: I wonder if we can use a switch statement here like this (https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/bundle/validate.go#L215) so that we can support newer version of CRD in the future.

pkg/registry/bundle.go
```diff
@@ -262,23 +286,42 @@ func (b *Bundle) cache() error {
...
-		b.crds = []*v1beta1.CustomResourceDefinition{}
+	if b.v1beta1crds == nil {
+		b.v1beta1crds = []*apiextensionsv1beta1.CustomResourceDefinition{}
 	}
+	if b.v1crds == nil {
+		b.v1crds = []*apiextensionsv1.CustomResourceDefinition{}
+	}
+
 	for _, o := range b.Objects {
 		if o.GetObjectKind().GroupVersionKind().Kind == "CustomResourceDefinition" {
```
dinhxuanvu, line 297: Maybe should put this CRD string to be a constant as well.

pkg/registry/bundle.go
```diff
@@ -262,23 +286,42 @@ func (b *Bundle) cache() error {
...
-			// Marshal Unstructured and Decode as CustomResourceDefinition. FromUnstructured has issues
-			// converting JSON numbers to float64 for CRD minimum/maximum validation.
-			cb, err := o.MarshalJSON()
-			if err != nil {
-				return err
-			}
-			dec := serializer.NewCodecFactory(Scheme).UniversalDeserializer()
-			if _, _, err = dec.Decode(cb, nil, crd); err != nil {
-				return fmt.Errorf("error decoding CRD: %v", err)
+			if o.GetObjectKind().GroupVersionKind().Version == v1beta1CRDVersion {
```
dinhxuanvu, line 298: My same comment here to use `switch` statement instead.

pkg/registry/bundle.go
exdx: v1 is considered stable, and probably won't change for a very long time, but I think the code does look cleaner this way so lets change it

pkg/registry/bundle.go
exdx: I wonder if we need this at all - the `Bundle` struct has the uninitialized version of the CRD array attached to it when it gets created.

pkg/registry/bundle.go
dinhxuanvu: I would consider the switch statement here is cleaner as you can get rid of the `ok`. The `ok` check in else can be translated into `default` statement. On the other note, you can get rid of `continue` line in #138 as it doesn't do anything.

pkg/registry/bundle.go
```diff
@@ -262,23 +290,36 @@ func (b *Bundle) cache() error {
...
+				dec := serializer.NewCodecFactory(Scheme).UniversalDeserializer()
+				if _, _, err = dec.Decode(cb, nil, crd); err != nil {
+					return fmt.Errorf("error decoding CRD: %v", err)
+				}
+				b.v1crds = append(b.v1crds, crd)
+			case v1beta1CRDVersion:
+				crd := &apiextensionsv1beta1.CustomResourceDefinition{}
+				// Marshal Unstructured and Decode as CustomResourceDefinition. FromUnstructured has issues
+				// converting JSON numbers to float64 for CRD minimum/maximum validation.
+				cb, err := o.MarshalJSON()
```
kevinrizza, line 313: nit: can we move the marshaling and universal deserialization above the case statement? seems like this code is identical

njhale:
/lgtm cancel

Looks like there are some dependency issues. From Travis:

```
vendor/github.com/operator-framework/api/pkg/validation/internal/bundle.go:91:15: crd.GetName undefined (type interface {} is interface with no methods)
```

pkg/registry/bundle.go
```diff
@@ -22,7 +29,10 @@ func DefaultYAMLDecoder() runtime.Decoder {
 }
 
 func init() {
-	if err := v1beta1.AddToScheme(Scheme); err != nil {
+	if err := apiextensionsv1beta1.AddToScheme(Scheme); err != nil {
+		panic(err)
+	}
+	if err := apiextensionsv1.AddToScheme(Scheme); err != nil {
```
njhale, line 35: `k8s.io/apimachinery/pkg/util/runtime` has a `Must` function that's canonically used for this:

```golang
utilruntime.Must(apiextensionsv1.AddToScheme(Scheme))
```

pkg/registry/bundle.go
```diff
@@ -104,11 +115,18 @@ func (b *Bundle) Skips() ([]string, error) {
 	return b.csv.GetSkips()
 }
 
-func (b *Bundle) CustomResourceDefinitions() ([]*v1beta1.CustomResourceDefinition, error) {
+func (b *Bundle) CustomResourceDefinitions() ([]interface{}, error) {
```
njhale, line 118: I think `runtime.Object` may be more useful than the empty interface.

pkg/registry/bundle_test.go
```diff
@@ -0,0 +1,108 @@
...
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+)
+
+const (
+	manifestDir = "./testdata/v1crd_bundle/manifests"
+)
+
+// TestV1CRDsInBundle tests that adding a v1 and v1beta1 CRD to a bundle is successful.
+// The provided APIs and CRD objects in the created bundle are compared to those in a test manifest directory.
+func TestV1CRDsInBundle(t *testing.T) {
```
njhale, line 23: I think it might be simpler to tweak an existing test -- say in `populator_test.go` -- to contain a v1 CRD.

exdx:
> /lgtm cancel
> 
> Looks like there are some dependency issues. From Travis:
> 
> ```
> vendor/github.com/operator-framework/api/pkg/validation/internal/bundle.go:91:15: crd.GetName undefined (type interface {} is interface with no methods)
> ```

Yes this is a known issue from changing the type that the api dependency then imports (we will get around this by moving the bundle validation back to the registry). Should not be an issue once we do that but at this point the build fails and tests can't run.

pkg/registry/bundle_test.go
exdx: It could but since I'm changing `bundle.go` I wanted to write a `bundle_test.go`. Frankly surprised we didn't have unit tests there already.

pkg/lib/bundle/validate.go
```diff
@@ -306,18 +312,37 @@ func (i imageValidator) ValidateBundleContent(manifestDir string) error {
...
 					validationErrors = append(validationErrors, err)
+					continue
 				}
+
+				results := crdValidator.Validate(crd)
+				if len(results) > 0 {
+					for _, err := range results[0].Errors {
+						validationErrors = append(validationErrors, err)
+					}
+				}
```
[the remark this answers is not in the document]
exdx: This code may change as a result of moving the api CRD validation back to the registry. Will leave as-is for now

dinhxuanvu:
The dependency issue (with api) has been resolved and unit test should pass successfully now.
/hold cancel

PR #302, opened 2020-04-27
Drive default index base image with CI

kevinrizza:
Now that the upstream opm builder is built and tagged on each commit,
introduce the short hash of the commit as a build parameter. This drives
the default base image that opm index commands use in order to tie
a built binary to a specific version of the image rathern than always
using the latest version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,16 @@
 GO := GOFLAGS="-mod=vendor" go
+PKG := github.com/operator-framework/operator-registry
 CMDS := $(addprefix bin/, $(shell ls ./cmd))
 SPECIFIC_UNIT_TEST := $(if $(TEST),-run $(TEST),)
 
 .PHONY: all
 all: clean test build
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,12 +1,16 @@
...
+PKG := github.com/operator-framework/operator-registry
 CMDS := $(addprefix bin/, $(shell ls ./cmd))
 SPECIFIC_UNIT_TEST := $(if $(TEST),-run $(TEST),)
 
 .PHONY: all
 all: clean test build
 
+GIT_COMMIT := $(shell git rev-parse --short HEAD)
+
+$(CMDS): version_flags=-ldflags "-X $(PKG)/pkg/version.GitCommit=$(GIT_COMMIT)"
```
njhale, line 11: if both `extra_flags` and `version_flags` define `-ldflags`, we'll end up with two `-ldflags` post variable expansion -- does this work as expected?

njhale:
closing due to staleness

PR #310, opened 2020-04-29
Bug 1834935: Use SkipTLS in indexer, default on e2e tests

kevinrizza:
The e2e tests, after moving to a local container registry, are failing when certs are not set up. In order to take advantage of an insecure registry, update the indexer's api functions to all include the SkipTLS
parameter, and pass it along as an option to the containerd resolver constructor.

Additionally, update all the indexer requests in the e2e opm test suite to make use of that parameter.


test/e2e/e2e_suite_test.go
```diff
@@ -21,12 +23,17 @@ func TestE2E(t *testing.T) {
 
 var _ = BeforeSuite(func() {
 	// FIXME: Since podman login doesn't work with daemonless image pulling, we need to login with docker first so podman tests don't fail.
```
dinhxuanvu, line 25: random: This is sad.

.travis.yml
```diff
@@ -31,11 +31,23 @@ addons:
     - conntrack
 script: |
   "${MAKE}" unit build
-  if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-    sudo minikube start # vm-driver=none requires root
-    sudo chown -R "$USER" "$HOME/.kube" "$HOME/.minikube"
-    KUBECONFIG="$HOME/.kube/config" make e2e
+  if [[ "$TRAVIS_OS_NAME" != "linux" ]]; then
+    echo "skipping e2e tests for ${TRAVIS_OS_NAME}"
```
exdx, line 35: so we're publishing binaries for all 3 platforms but only running e2e tests against linux? The other platforms are supported but not tested against? Should we make a note of that in the docs.

.travis.yml
```diff
@@ -31,11 +31,23 @@ addons:
...
+    exit 0
   fi
+
+  if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
+    # no Quay.io credentials available for PR tests
+    echo "starting local test docker registry"
+    sudo docker run -d -p 5000:5000 --restart=always --name registry registry:2
+    export DOCKER_REGISTRY_HOST="localhost:5000"
+    printf "[registries.insecure]\nregistries = ['localhost:5000']\n" > /etc/containers/registries.conf
```
exdx, line 44: I ran into this issue when testing image polling on OLM, but there the openshift CI included the openshift registry so I was able to use that. I'm wondering if generating some certs on a per-CI run basis and passing those to the registry would make more sense than running everything with `skipTLScheck` since in production we would expect users to be hosting their images on a registry that has certs and our tests end up closer to the real thing

pkg/containertools/imagereader.go
```diff
@@ -53,7 +53,7 @@ func (b ImageLayerReader) GetImageData(image, outputDir string, opts ...GetImage
 		os.Mkdir(outputDir, 0777)
 	}
 
-	err := b.Cmd.Pull(image)
+	err := b.Cmd.Pull(image, options.SkipTLS)
```
gallettilance, line 56: skipsTLS is an option for the bundle pulling but not for the index pulling? Do I understand that code right?

gallettilance: nvm this PR is just about exposing that option to the tests

test/e2e/opm_test.go
```diff
@@ -251,14 +271,15 @@ var _ = Describe("opm", func() {
...
-			err = exportWith(containertools.NoneTool.String())
+			err = exportWith(containertools.NoneTool.String(), indexImage3)
 			Expect(err).NotTo(HaveOccurred())
 
 			By("loading manifests from a containerd-extracted directory")
 			err = initialize()
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		/* TODO: Reenable this test once the commutativity bug is fixed.
```
benluddy, line 282: Could you change this to https://pkg.go.dev/github.com/onsi/ginkgo?tab=doc#Skip instead?

pkg/containertools/command.go
```diff
@@ -41,9 +41,13 @@ func (r *ContainerCommandRunner) GetToolName() string {
 
 // Pull takes a container image path hosted on a container registry and runs the
 // pull command to download it onto the local environment
-func (r *ContainerCommandRunner) Pull(image string) error {
+func (r *ContainerCommandRunner) Pull(image string, skipTLS bool) error {
 	args := []string{"pull", image}
 
+	if r.containerTool.String() == "podman" {
```
benluddy, line 47: Shouldn't this check that `skipTLS` is true before adding the arg? Also, are the tests not running against Docker?

kevinrizza: Well, that's a good catch.

kevinrizza: They do run against docker, that particular parameter only exists (and is only needed) by podman.

benluddy: Oh, ok. Didn't see the config you're generating in the test script.

PR #314, opened 2020-04-30
Bug 1829994: Index generate dameonless

kevinrizza:
Enable daemonless index add when `--generate` is specified.

This also fixes a bug where, on multiple runs with --generate specified, the database would fail to build.

Doing this with a partial implementation to unblock podman users to be able to generate the database without shelling out to podman and hitting this bug https://github.com/containers/libpod/issues/5234. In the future, this implementation should be fleshed out so that all of the registry and index commands make this distinction between pull and build runners.

### DIFF
--- a/cmd/opm/index/add.go
+++ b/cmd/opm/index/add.go
@@ -55,7 +55,9 @@ func addIndexAddCmd(parent *cobra.Command) {
 	}
 	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling bundles")
 	indexCmd.Flags().StringP("binary-image", "i", "", "container image for on-image `opm` command")
-	indexCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
+	indexCmd.Flags().StringP("container-tool", "c", "", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
+	indexCmd.Flags().StringP("build-tool", "u", "", "tool to build container images. One of: [docker, podman]. Defaults to podman. Overrides part of container-tool.")
+	indexCmd.Flags().StringP("pull-tool", "p", "", "tool to pull container images. One of: [none, docker, podman]. Defaults to none. Overrides part of container-tool.")
 	indexCmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
 	indexCmd.Flags().Bool("permissive", false, "allow registry load errors")
 	indexCmd.Flags().StringP("mode", "", "replaces", "graph update mode that defines how channel graphs are updated. One of: [replaces, semver, semver-skippatch]")
@@ -96,15 +98,6 @@ func runIndexAddCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	containerTool, err := cmd.Flags().GetString("container-tool")
-	if err != nil {
-		return err
-	}
-
-	if containerTool == "none" {
-		return fmt.Errorf("none is not a valid container-tool for index add")
-	}
-
 	tag, err := cmd.Flags().GetString("tag")
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


cmd/opm/index/add.go
```diff
@@ -55,10 +55,12 @@ func addIndexAddCmd(parent *cobra.Command) {
 	}
 	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling bundles")
 	indexCmd.Flags().StringP("binary-image", "i", "", "container image for on-image `opm` command")
-	indexCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
+	indexCmd.Flags().StringP("build-tool", "c", "podman", "tool to build container images (build). One of: [docker, podman]")
+	indexCmd.Flags().StringP("pull-tool", "p", "none", "tool to pull container images (save, pull, etc.). One of: [none, docker, podman]")
 	indexCmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
 	indexCmd.Flags().Bool("permissive", false, "allow registry load errors")
 	indexCmd.Flags().StringP("mode", "", "replaces", "graph update mode that defines how channel graphs are updated. One of: [replaces, semver, semver-skippatch]")
+	indexCmd.Flags().String("container-tool", "", "container-tool has been deprecated. Please select a build and pull tool separately.")
```
njhale, line 63: Note for later: cobra has a way to mark options as deprecated:
```go
indexCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
if err := rootCmd.Flags().MarkDeprecated("container-tool", "ignored in favor of separate build-tool and pull-tool options"); err != nil {
		logrus.Panic(err.Error())
}
```

This has some nice properties like preserving the help when referenced, but omitting it from the standard help output.

cmd/opm/index/add.go
```diff
@@ -96,15 +98,29 @@ func runIndexAddCmdFunc(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	containerTool, err := cmd.Flags().GetString("container-tool")
+	if err != nil {
+		return err
+	}
+
+	if containerTool != "" {
```
njhale, line 120: Any thoughts on how backwards compatible we want to be? If we explode here we end up breaking existing clients. Though we can do a major version bump after to be dogmatic about semver.

cmd/opm/index/add.go
kevinrizza: Ah, nice. I'll update this to include it.

cmd/opm/index/add.go
kevinrizza: I was sort of back and forth on this. I think that's fair, maybe we want to just treat this as an alias and use it to overwrite defaults if the two new flags haven't been specified?

njhale: SGTM, but is there a convention for telling the difference between unset and set-to-default?

Bowenislandsong:
/lgtm
/hold
holding for second review

kevinrizza:
@ecordell updated based on your thoughts. ptal.

pkg/lib/indexer/indexer.go
```diff
@@ -339,8 +340,6 @@ func copyDatabaseTo(databaseFile, targetDir string) (string, error) {
 		if err := os.MkdirAll(targetDir, 0777); err != nil {
 			return "", err
 		}
-	} else {
-		return "", err
 	}
```
ecordell, line 343: Why is this removed?

pkg/lib/indexer/indexer.go
```diff
@@ -353,7 +352,7 @@ func copyDatabaseTo(databaseFile, targetDir string) (string, error) {
 	dbFile := path.Join(targetDir, defaultDatabaseFile)
 
 	// define the path to copy to the database/index.db file
-	to, err := os.OpenFile(dbFile, os.O_RDWR|os.O_CREATE, 0666)
+	to, err := os.OpenFile(dbFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
```
ecordell, line 355: Doesn't this change mean that if `copy` fails, you'll be left with an empty file instead of your previous index?

PR #315, opened 2020-05-01
fix(containerd): use the defined request scopes 

ecordell:
use the defined request scopes if the server doesn't respond with a scope

this works around a bug when talking to quay through an nginx proxy

**Reviewer Checklist**


benluddy:
It would be nice to have an automated check for vendored files that have been modified in-place -- easy to miss in review, especially if combined with many legitimate changes to the vendor directory.

ecordell:
@benluddy that would be a great check to have - I threw this up just to share the change last week, I should've put a hold on it.

PR #318, opened 2020-05-04
Bug 1827723: Make default channel optional

Bowenislandsong:
This commit address the bug where a bundle was released before the current default channel was created, the inference fails.


Bowenislandsong:
the new annotation for prometheus 0.14.0 in the e2e test:
```
annotations:
  operators.operatorframework.io.bundle.channel.default.v1: ""
  operators.operatorframework.io.bundle.channels.v1: stable
  operators.operatorframework.io.bundle.manifests.v1: manifests/
  operators.operatorframework.io.bundle.mediatype.v1: registry+v1
  operators.operatorframework.io.bundle.metadata.v1: metadata/
  operators.operatorframework.io.bundle.package.v1: prometheus
```

Bowenislandsong:
switched channels and default channel situation for 0.15.0 case so that this bundle is added in the middle between 0.14.0 and 0.22.2

manifests/prometheus/0.22.2/prometheusoperator.0.22.2.clusterserviceversion.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   annotations:
     alm-examples: '[{"apiVersion":"monitoring.coreos.com/v1","kind":"Prometheus","metadata":{"name":"example","labels":{"prometheus":"k8s"}},"spec":{"replicas":2,"version":"v2.3.2","serviceAccountName":"prometheus-k8s","securityContext": {}, "serviceMonitorSelector":{"matchExpressions":[{"key":"k8s-app","operator":"Exists"}]},"ruleSelector":{"matchLabels":{"role":"prometheus-rulefiles","prometheus":"k8s"}},"alerting":{"alertmanagers":[{"namespace":"monitoring","name":"alertmanager-main","port":"web"}]}}},{"apiVersion":"monitoring.coreos.com/v1","kind":"ServiceMonitor","metadata":{"name":"example","labels":{"k8s-app":"prometheus"}},"spec":{"selector":{"matchLabels":{"k8s-app":"prometheus"}},"endpoints":[{"port":"web","interval":"30s"}]}},{"apiVersion":"monitoring.coreos.com/v1","kind":"Alertmanager","metadata":{"name":"alertmanager-main"},"spec":{"replicas":3, "securityContext": {}}}]'
 spec:
-  replaces: prometheusoperator.0.15.0
+  replaces: prometheusoperator.0.14.0
```
kevinrizza, line 10: Why are we reordering these? I would prefer that, if we need manifests for a specific test, that we make a copy of them and place them in a separate test folder rather than modify our default examples.

pkg/lib/bundle/generate.go
```diff
@@ -12,6 +12,7 @@ import (
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/kubectl/pkg/util/slice"
```
kevinrizza, line 15: Is this external dependency really necessary? In general, I think that utils like this which are not included in the standard library and are trivial to write should just be implemented directly.

pkg/lib/bundle/generate.go
```diff
@@ -298,6 +302,10 @@ func ValidateChannelDefault(channels, channelDefault string) (string, error) {
 	var chanErr error
 	channelList := strings.Split(channels, ",")
 
+	if len(channelList) == 0 || slice.ContainsString(channelList, "", nil) {
+		return chanDefault, fmt.Errorf("invalid channels is provied: %s", channels)
```
kevinrizza, line 306: This error message seems confusing. Could we be more descriptive of what is actually going wrong here? It seems like we are actually checking two things that should have separate error messages:

1. That no channels were specified
2. That the list of channels included an empty string in the list I think?

Also spelling/grammar nit: use `are` instead of `is` for plurals, and `provided` instead of `provied`

pkg/lib/bundle/generate_test.go
```diff
@@ -207,7 +207,7 @@ func TestGenerateDockerfileFunc(t *testing.T) {
 		"LABEL operators.operatorframework.io.bundle.metadata.v1=%s\n"+
 		"LABEL operators.operatorframework.io.bundle.package.v1=test4\n"+
 		"LABEL operators.operatorframework.io.bundle.channels.v1=test5\n"+
-		"LABEL operators.operatorframework.io.bundle.channel.default.v1=test5\n\n"+
+		"LABEL operators.operatorframework.io.bundle.channel.default.v1=\n\n"+
```
kevinrizza, line 210: Rather than coopting existing tests, can we write an explicit test specifically for this edge case?

test/e2e/opm_test.go
```diff
@@ -75,7 +73,7 @@ func buildBundlesWith(containerTool string) error {
 		bundleTag3: bundlePath3,
 	} {
 		if err := inTemporaryBuildContext(func() error {
-			return bundle.BuildFunc(path, "", bundleImage+":"+tag, containerTool, packageName, channels, defaultChannel, false)
+			return bundle.BuildFunc(path, "", bundleImage+":"+tag, containerTool, "", "", "", false)
```
kevinrizza, line 76: Same comment here. Can we have a separate test for this rather than modifying an existing one?

pkg/lib/bundle/generate_test.go
Bowenislandsong: right, I agree with the previous judgment about the e2e. but the change does alter the outcome for this specific unit test.

pkg/lib/bundle/generate.go
Bowenislandsong: yeah, i see your point, just loved that library. getting rid of it.

pkg/lib/bundle/generate_test.go
kevinrizza: Ah, this was actually invalid and we weren't catching it. Gotcha

pkg/lib/bundle/generate.go
Bowenislandsong: ha, I copied that misspell and didn't even see it https://github.com/operator-framework/operator-registry/blob/a25687c90bbacce3c200a232376fa4952d22d307/pkg/lib/bundle/generate.go#L319

Bowenislandsong: https://play.golang.org/p/uemuPhlTisX found out that this split will always output an array with at least 1 length. My bad.

dinhxuanvu:
Looking good.
/approve
Just a quick question. Why do we need the aqua bundle for this fix?

Bowenislandsong:
> Looking good.
> /approve
> Just a quick question. Why do we need the aqua bundle for this fix?

Thanks, Kevin suggested to add a separate thing for e2e test that I am intereted and not to change existing package for my test purpose.

manifests/aqua/0.0.1/aquagateways.operator.aquasec.com.crd.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
kevinrizza, line 1: Rather than add something to the root of the `manifests/` folder, can we just add these to a `test_data` folder for a separate test?

Bowenislandsong: awww, I changed all those unit tests for it. Shouda thought about it... making the change now

PR #252, opened 2020-04-03
[feature] CSVless bundle

Bowenislandsong:
This work starts the support for csvless bundle by synthesizing a CSV based on Kube manifests in the bundle manifests folder and non-inferable information provided in the metadata folder. Amongst non-inferable information, we require a YAML file to supply most of the non-inferable information, an annotations.yaml to point to the directory names of manifests and metadata directory, and a dependencies.yaml to point out required APIs including CRDs and APIServices. Each provided APIService expects a description that contains at least the kind information in the non-inferable CSV info YAML file. This feature supports APISevrices and CRDs of all currently existing versions. This feature is provided by `generate-csv` command and will validate the bundle format and content to ensure correctness.

PR #322, opened 2020-05-10
pkg/registry: return error if no CSV object exists in a Bundle

estroz:
* pkg/registry: return an `ErrNoCSVInBundle` on operations on Bundles that require a CSV object be present if not

**Motivation for the change:** a Bundle lacking a CSV should not be considered a valid Bundle under current bundle semantics. This CSV check considers the eventual CSV-less bundle (#252).

**Reviewer Checklist**

/cc @dinhxuanvu @kevinrizza

### DIFF
--- a/pkg/registry/bundle.go
+++ b/pkg/registry/bundle.go
@@ -16,14 +16,30 @@ import (
 
 const (
 	v1beta1CRDVersion = "v1beta1"
-	v1CRDVersion = "v1"
-	CRDKind = "CustomResourceDefinition"
+	v1CRDVersion      = "v1"
+	CRDKind           = "CustomResourceDefinition"
 )
+
 // Scheme is the default instance of runtime.Scheme to which types in the Kubernetes API are already registered.
 var Scheme = runtime.NewScheme()
 
 // Codecs provides access to encoding and decoding for the scheme
 var Codecs = serializer.NewCodecFactory(Scheme)
+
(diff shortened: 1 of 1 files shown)


pkg/registry/bundle.go
```diff
@@ -40,15 +56,25 @@ type Bundle struct {
...
 }
 
 func NewBundle(name, pkgName string, channels []string, objs ...*unstructured.Unstructured) *Bundle {
-	bundle := &Bundle{Name: name, Package: pkgName, Channels: channels, cacheStale: false}
+	bundle := &Bundle{
+		Name:        name,
+		Package:     pkgName,
+		Channels:    channels,
+		cacheStale:  false,
+		mustHaveCSV: true,
```
dinhxuanvu, line 76: I see your reasoning for adding this flag. However, there must be a way to set this flag somewhere so we have control over this flag. It may be true for now but may not be in the future.

estroz: I am thinking the `Bundle` constructor can set this option in the future via a constructor parameter or inference on `Bundle.Objects`. For now it will always be true. I'm fine with just checking `Bundle.csv == nil` if you aren't sure a boolean option will be used to determine bundle type going forward.

estroz:
Closing since this should be added to the new [api `Bundle` definition](https://github.com/operator-framework/api/pull/35) first.

PR #324, opened 2020-05-11
Bug 1827723: Adding error message for more than one CSV in bundle dir

Bowenislandsong:
We assume a strict one CSV per bundle structure. Since there is a chance a bundle directly supplied contains more than one CSV, we output error messages to catch non-conforming bundles.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/csv.go
+++ b/pkg/registry/csv.go
@@ -65,8 +65,8 @@ type ClusterServiceVersion struct {
 	Spec json.RawMessage `json:"spec"`
 }
 
-// ReadCSVFromBundleDirectory tries to parse every YAML file in the directory and see if they are CSV.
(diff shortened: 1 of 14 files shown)


pkg/registry/csv.go
```diff
@@ -80,14 +80,15 @@ func ReadCSVFromBundleDirectory(bundleDir string) (*ClusterServiceVersion, error
 		}
 	}
 
+	csv := ClusterServiceVersion{}
```
dinhxuanvu, line 83: In line 69 above, it says `we will return the first file that is considered a CSV type`. That's no longer the behavior of this func as you will return error if there are multiple CSVs. Can you update the func comment?

PR #329, opened 2020-05-15
Use secure registry e2e

kevinrizza:
Run a container registry locally to manage e2e tests pushing and pulling container images.


njhale:
Closing b/c it's stale. I think we still want to do this when we have the chance.

PR #332, opened 2020-05-20
Bug 1827612: fix(server): make a copy of the db before migrating it

ecordell:
`opm registry server` and `registry-server` now make a copy of the db before migrating it

Note that this is currently implemented twice (once for each binary) to ease backports. I expect we will deprecate registry-server in favor of `opm registry serve` at some point, and the duplication will be removed.

if the server is run when it has read-only access to a database, it will
fail to migrate. this commit will cause the server to always copy the
database on start, so that it is always rw and can be migrated.

**Testing Notes**
We currently lack e2e testing that runs against the built binaries. I tested this by building the `Dockerfile` with this change (which has a non-root user), using that as a base image for a registry image, and running the image. 

Running it doesn't have the "migration error log"
```bash
$ docker run a7b0b3543fa4
time="2020-05-20T17:38:38Z" level=warning msg="unable to set termination log path" error="open /dev/termination-log: permission denied"
time="2020-05-20T17:38:38Z" level=info msg="serving registry" database=/bundles.db port=50051
```

and to confirm, I checked that the tmpdb file was created with the correct permissions:

```bash
$ docker exec -ti 19db7d0d829f /bin/bash
bash-4.2$ ls -al /tmp/db682566081
-rw------- 1 1001 root 118784 May 20 17:38 db682566081
```
Note that the file is owner `rw` with 1001, the userid set by the dockerfile.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/tmp/copy.go
+++ b/pkg/lib/tmp/copy.go
@@ -1,0 +1,54 @@
+package tmp
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+)
+
+// CopyTmpDB reads the file at the given path and copies it to a tmp directory, returning the copied file path or an err
+func CopyTmpDB(original string) (path string, err error) {
+	dst, err := ioutil.TempFile("", "db-")
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if cerr := dst.Close(); cerr != nil && err == nil {
+			err = cerr
+		}
+	}()
+
+	src, err := OpenRegularFile(original)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if cerr := src.Close(); cerr != nil && err == nil {
+			err = cerr
+		}
+	}()
+
+	_, err = io.Copy(dst, src)
+	if err != nil {
+		return "", err
+	}
+
+	return dst.Name(), nil
+}
+
+// OpenRegularFile opens the file at path and returns an error if it is not regular, does not exist, or cannot be opened
+func OpenRegularFile(path string) (*os.File, error) {
+	fd, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	fi, err := fd.Stat()
+	if err != nil {
+		return nil, err
+	}
+	if !fi.Mode().IsRegular() {
+		return nil, fmt.Errorf("%s is not a regular file", path)
+	}
+	return fd, nil
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 
 WORKDIR /src
 
+COPY OPM_VERSION OPM_VERSION
 COPY vendor vendor
 COPY cmd cmd
 COPY pkg pkg
--- a/cmd/opm/registry/serve.go
+++ b/cmd/opm/registry/serve.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"fmt"
 	"net"
+	"os"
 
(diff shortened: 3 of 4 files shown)


pkg/lib/tmp/copy.go
```diff
@@ -0,0 +1,49 @@
+package tmp
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
```
benluddy, line 9: I don't think it's worth aggregating errors, as long as we return the first error that was encountered.

pkg/lib/tmp/copy.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	if !srcStat.Mode().IsRegular() {
+		return fmt.Errorf("%s is not a regular file", srcName)
+	}
+	src, err := os.Open(srcName)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = utilerrors.NewAggregate([]error{err, src.Close()})
```
benluddy, line 38: We'd probably prefer to ignore it if the only error is on src.Close, at least for this use case.

pkg/lib/tmp/copy.go
exdx: I agree, it looks like any error in this codepath is terminal for the registry

pkg/lib/tmp/copy.go
```diff
@@ -0,0 +1,49 @@
...
+	}()
+	dst, err := os.Create(dstName)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err = utilerrors.NewAggregate([]error{err, dst.Close()})
+	}()
+	_, err = io.Copy(dst, src)
```
exdx, line 47: The one thing I'm thinking about is how much did we increase the vulnerability surface of the registry by running it as a user who has the ability to copy/manipulate random files. It may have been like this previously, but before we were only modifying one file during migration (the db) versus creating and copying files around. Maybe it's worth setting the name of the tempDB in the Dockerfile as an env argument and locking down the user permissions to only be able to modify those two files? Just a thought

ecordell: > running it as a user who has the ability to copy/manipulate random files.

To clarify, this is what the user running the server must have:

- read access to the database file
- write access to a new /tmp file

and /tmp is typically 777 + sticky bit.

pkg/lib/tmp/copy.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return dst.Name(), nil
+}
+
+// OpenRegularFile opens the file at path and returns an error if it is not regular, does not exist, or cannot be opened
+func OpenRegularFile(path string) (*os.File, error) {
+	fd, err := os.Open(path)
```
benluddy, line 42: `fd` isn't closed in the error case. Probably OK for the initial use case, since it's fatal anyway, but if this is a library package then someone may eventually use it along a non-fatal path and leak FDs.

pkg/lib/tmp/copy.go
```diff
@@ -0,0 +1,54 @@
...
+			err = cerr
+		}
+	}()
+
+	src, err := OpenRegularFile(original)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if cerr := src.Close(); cerr != nil && err == nil {
```
benluddy, line 27: IMO it's better for the registry server to tolerate errors from closing `src`, but that would either limit the general usefulness of this package or add some way to make that determination by inspecting the error value. Hopefully it's rare enough to not matter.

PR #334, opened 2020-05-21
Bug 1838756: fix(validation): Fix CRD v1beta int-float conversion error

dinhxuanvu:
During bundle content validation, manifest files are converted to
`unstructured` type and then convert back to CRD type which causes
conversion error between int to float. Shouldn't use `unstructured`
type as an intermediate type.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/testdata/validate/valid_bundle/manifests/etcdclusters.etcd.database.coreos.com.crd.yaml
+++ b/pkg/lib/bundle/testdata/validate/valid_bundle/manifests/etcdclusters.etcd.database.coreos.com.crd.yaml
@@ -14,3 +14,13 @@ spec:
     singular: etcdcluster
   scope: Namespaced
   version: v1beta2
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          type: object
+          properties:
+            test:
+              type: integer
+              minimum: 1
+              maximum: 32767
--- a/pkg/lib/bundle/validate.go
+++ b/pkg/lib/bundle/validate.go
@@ -313,10 +313,11 @@ func (i imageValidator) ValidateBundleContent(manifestDir string) error {
 				}
 			}
 		} else if gvk.Kind == CRDKind {
+			dec := k8syaml.NewYAMLOrJSONDecoder(strings.NewReader(string(data)), 30)
 			switch gv := gvk.GroupVersion().String(); gv {
 			case v1CRDapiVersion:
 				crd := &apiextensionsv1.CustomResourceDefinition{}
-				err := runtime.DefaultUnstructuredConverter.FromUnstructured(k8sFile.Object, crd)
+				err := dec.Decode(crd)
 				if err != nil {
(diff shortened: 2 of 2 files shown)


estroz:
@dinhxuanvu have you confirmed that this fixes the issue? I did some digging on the conversion error that this PR intends to fix:
```
cannot convert int64 to float64
```
I found a similar issue (https://github.com/vmware-tanzu/octant/issues/143) that linked the fix needed in `sigs.k8s.io/yaml`, an upstream fork of `github.com/ghodss/yaml` and the library the default unstructured converter uses. That fix has been [merged](https://github.com/kubernetes-sigs/yaml/pull/14) and exists in `sigs.k8s.io/yaml` v1.2.0. In the SDK we've made an effort to remove the direct `github.com/ghodss/yaml` dependency because it isn't maintained unlike the upstream fork. So perhaps this issue is caused by [use](https://github.com/operator-framework/operator-registry/blob/master/go.mod#L12) of the unmaintained dep, and even if it isn't now it could arise again in the future. I suggest switching to the upstream dep and seeing if that fixes your issue before using a different decoder.

dinhxuanvu:
> @dinhxuanvu have you confirmed that this fixes the issue? I did some digging on the conversion error that this PR intends to fix:
> 
> ```
> cannot convert int64 to float64
> ```
> 
> I found a similar issue ([vmware-tanzu/octant#143](https://github.com/vmware-tanzu/octant/issues/143)) that linked the fix needed in `sigs.k8s.io/yaml`, an upstream fork of `github.com/ghodss/yaml` and the library the default unstructured converter uses. That fix has been [merged](https://github.com/kubernetes-sigs/yaml/pull/14) and exists in `sigs.k8s.io/yaml` v1.2.0. In the SDK we've made an effort to remove the direct `github.com/ghodss/yaml` dependency because it isn't maintained unlike the upstream fork. So perhaps this issue is caused by [use](https://github.com/operator-framework/operator-registry/blob/master/go.mod#L12) of the unmaintained dep, and even if it isn't now it could arise again in the future. I suggest switching to the upstream dep and seeing if that fixes your issue before using a different decoder.

I can confirm this will fix the issue. The logic between using `unstructured` is to determine the GVK information. After that, it is not needed essentially. We should simply decode directly from source data (string) to the intended object type. It is always a better solution to avoid the "middle man".

estroz:
@dinhxuanvu whatever works. Do you still think it is reasonable to update the `yaml` dep across this and the api repo?

dinhxuanvu:
> @dinhxuanvu whatever works. Do you still think it is reasonable to update the `yaml` dep across this and the api repo?

Sure. If `github.com/ghodss/yaml` is no longer maintained, we should switch to a fork that is maintained which is `sigs.k8s.io/yaml` in this case as you pointed out. Also, just want to point out that we only use `github.com/ghodss/yaml` to do YAMLtoJSON conversion, not YAML to `unstructured` or anything like that. I will make PRs to update the `yaml` dep after this.

PR #337, opened 2020-05-26
Bug 1840370: Configure "podman build" to produce images with Docker manifests.

benluddy:
Podman's default is to produce OCI manifests and metadata, while
Docker (unsurprisingly) produces the Docker filetypes. This can lead
to behavior differences down the line, especially once container
registries are added to the mix.

### DIFF
--- a/pkg/containertools/factory_podman.go
+++ b/pkg/containertools/factory_podman.go
@@ -1,0 +1,33 @@
+package containertools
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+type PodmanCommandFactory struct{}
+
+func (p *PodmanCommandFactory) BuildCommand(o BuildOptions) (*exec.Cmd, error) {
+	args := []string{"build"}
+
+	if o.format != "" {
+		args = append(args, "--format", o.format)
+	} else {
+		args = append(args, "--format", "docker")
+	}
+
+	if o.dockerfile != "" {
+		args = append(args, "-f", o.dockerfile)
+	}
+
+	for _, tag := range o.tags {
+		args = append(args, "-t", tag)
+	}
+
+	if o.context == "" {
+		return nil, fmt.Errorf("context not provided")
+	}
+	args = append(args, o.context)
+
+	return exec.Command("podman", args...), nil
+}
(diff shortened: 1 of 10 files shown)


pkg/containertools/factory_podman.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"os/exec"
+)
+
+type PodmanCommandFactory struct{}
+
+func (p *PodmanCommandFactory) BuildCommand(o BuildOptions) (*exec.Cmd, error) {
+	args := []string{"build"}
+
+	if o.format != "" {
```
kevinrizza, line 13: Dumb thought: do we ever want to consider just passing arbitrary command arguments here rather than being explicit with flags? It might make for a simpler cli experience. Especially if we make the default for build fully daemonless

The counter is that we're on the hook for making all of these arbitrary flags work? 🤷

pkg/containertools/option_build.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (o *BuildOptions) SetDockerfile(dockerfile string) {
+	o.dockerfile = dockerfile
+}
+
+func (o *BuildOptions) SetContext(context string) {
+	o.context = context
+}
+
+func DefaultBuildOptions() BuildOptions {
```
kevinrizza, line 30: 👍

pkg/containertools/factory_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Name:    "podman multiple tags",
+			Factory: &PodmanCommandFactory{},
+			Options: BuildOptions{
+				context: ".",
+				tags:    []string{"foo", "bar"},
+			},
+			Args: []string{
+				"podman", "build", "--format", "docker", "-t", "foo", "-t", "bar", ".",
+			},
+		},
```
kevinrizza, line 136: Should we also include a test for the stub?

benluddy: Good catch, I'll add one.

pkg/containertools/factory_podman.go
benluddy: I think you could make a good argument for that, but I lean towards letting users directly invoke podman/whatever if they are in a circumstance that calls for it.

PR #338, opened 2020-05-27
Bug 1835884: opm bundle extract shouldn't validate annotations

kevinrizza:
Currently, the configmap writer attempts to validate the annotations
associated with the bundle. In reality, most of these annotations are
currently unused by the time the bundle is being unpacked and applied
to the cluster. There was an issue where default channel was no longer
required, but it was still validated by extract because it has a
separate validation function.

These annotations should be validated when building and
adding to the registry (as it exists now). This pr removes
annotation validation from the configmap writer in opm alpha bundle
extract.

### DIFF
--- a/pkg/configmap/configmap_test.go
+++ b/pkg/configmap/configmap_test.go
@@ -77,6 +77,19 @@ func TestLoad(t *testing.T) {
 				assert.Equal(t, 3, len(objects))
 			},
 		},
+		{
+			name:   "BundleWithNoDefaultChannel",
+			source: "testdata/bundle-with-no-default-channel.yaml",
+			assertFunc: func(t *testing.T, bundleGot *api.Bundle) {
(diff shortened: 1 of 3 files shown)


pkg/configmap/configmap_test.go
```diff
@@ -77,6 +77,19 @@ func TestLoad(t *testing.T) {
 				assert.Equal(t, 3, len(objects))
 			},
 		},
+		{
+			name:   "BundleWithNoDefaultChannel",
+			source: "testdata/bundle-with-no-default-channel.yaml",
+			assertFunc: func(t *testing.T, bundleGot *api.Bundle) {
+				csvGot := bundleGot.GetCsvJson()
+				assert.NotNil(t, csvGot)
```
benluddy, line 85: Nit: Pointless assertion, `csvGot` is a string. Actually, two tests above this one asserts nothing but this.

PR #348, opened 2020-06-04
Bug 1841178: fix(unpack): unpack union fs correctly when using podman/docker

njhale:
Correctly unpack union fs to avoid dropping whiteout files in the unpack
directory when using podman/docker.

### DIFF
--- a/pkg/containertools/imagereader.go
+++ b/pkg/containertools/imagereader.go
@@ -4,6 +4,7 @@ package containertools
 import (
 	"archive/tar"
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -11,6 +12,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/containerd/containerd/archive"
+	"github.com/containerd/containerd/archive/compression"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/containertools/imagereader_test.go
```diff
@@ -37,60 +34,44 @@ func TestReadImageLayersDocker(t *testing.T) {
...
-		expectedFilePath := filepath.Join(expectedFilePath, file)
-		expectedFile, err := ioutil.ReadFile(expectedFilePath)
-		require.NoError(t, err)
-
-		actualFilePath := filepath.Join(testOutputDir, file)
-		actualFile, err := ioutil.ReadFile(actualFilePath)
-		require.NoError(t, err)
-
-		require.Equal(t, string(expectedFile), string(actualFile))
+	tests := []struct {
```
kevinrizza, line 37: 👍

test/e2e/opm_test.go
```diff
@@ -288,6 +288,20 @@ var _ = Describe("opm", func() {
 				Expect(err).NotTo(HaveOccurred())
 			}
 
+			By("validating bundles")
```
kevinrizza, line 291: Can we potentially start separating these different things into separate tests? This is a test specifically about building bundles with `opm alpha bundle build` from a nested package manifest directory using inference for the package level parameters on each bundle.

njhale: Sure. I can move this to its own test, but it does seem like we would want to run validate in both of the existing tests too.

kevinrizza: That's probably fair. I still think these tests need to be slightly more targeted in general than just "Do everything and then throw an error when something goes wrong", otherwise we will quickly get into a place where any change will require updating one of these two tests -- I think it makes sense for all of our tests to have a sizable surface area, but I'm worried about them having too many different meanings.

pkg/containertools/imagereader.go
```diff
@@ -150,52 +153,28 @@ func extractBundleManifests(layerTarball, outputDir string, tarReader *tar.Reade
...
-		switch header.Typeflag {
-		case tar.TypeDir:
-			// Create the directory if it doesn't exist
-			directoryToWrite := filepath.Join(outputDir, header.Name)
-			if _, err := os.Stat(directoryToWrite); os.IsNotExist(err) {
-				os.Mkdir(directoryToWrite, 0777)
-			}
-		case tar.TypeReg:
-			manifestToWrite := filepath.Join(outputDir, header.Name)
+func adjustPerms(h *tar.Header) (bool, error) {
```
ecordell, line 169: this happens in two places now, right? should this be shared so that we unpack things the same way every time?

pkg/containertools/imagereader.go
```diff
@@ -150,52 +153,28 @@ func extractBundleManifests(layerTarball, outputDir string, tarReader *tar.Reade
...
+		if header.Typeflag == tar.TypeReg && header.Name == layerTarball {
+			// Found the embedded tarball for the layer
+			decompressed, err := compression.DecompressStream(tarReader)
+			if err != nil {
+				return err
 			}
-		}
-	}
-}
+			_, err = archive.Apply(context.TODO(), outputDir, decompressed, archive.WithFilter(adjustPerms))
```
ecordell, line 162: This is really just using containerd to unpack all images now. This is probably fine, but it is different behavior then creating a container + exporting the fs with each separate tool. (for example, the actual layered fs can preserve the original permissions entirely, while we have to make everything owner-writable to unpack)

test/e2e/opm_test.go
njhale: I agree, though I didn't want to diverge too much from the precedent set here in a bug fix.

pkg/containertools/imagereader.go
njhale: I think so. I expect us to move this logic into the `image` package in the future. I didn't know if doing that in a bug-fix would be overkill for such a small amount of code.

njhale:
Superseded by #351

PR #351, opened 2020-06-08
fix(images): use docker/podman create and cp for exec unpacking

njhale:
Use docker/podman create and cp commands to unpack bundle and index
images instead of a custom unpacking scheme. This fixes an issue where
whiteout files present in an image cause `opm alpha bundle validate` to
fail.


cmd/opm/alpha/bundle/validate.go
```diff
@@ -36,7 +42,25 @@ func validateFunc(cmd *cobra.Command, args []string) error {
 	logger := log.WithFields(log.Fields{"container-tool": imageBuilderArgs})
 	log.SetLevel(log.DebugLevel)
 
-	imageValidator := bundle.NewImageValidator(imageBuilderArgs, logger)
+	var (
+		registry image.Registry
+		err      error
+	)
+
+	tool := containertools.NewContainerTool(imageBuilderArgs, containertools.NoneTool)
```
kevinrizza, line 50: Right now the help text for this flag lists docker and podman and is a "build" tool. Can we update the help text to at least mention the `none` option? I think if we did that we should also probably attempt to rename the parameter, since we aren't building anything in bundle validate.

pkg/containertools/runner.go
```diff
@@ -101,6 +102,49 @@ func (r *ContainerCommandRunner) Save(image, tarFile string) error {
 	return nil
 }
 
+// Unpack copies a directory from a local container image to a directory in the local filesystem.
+func (r *ContainerCommandRunner) Unpack(image, src, dst string) error {
+	args := []string{"create", image, ""}
```
kevinrizza, line 107: 👍

pkg/containertools/runner.go
```diff
@@ -101,6 +102,49 @@ func (r *ContainerCommandRunner) Save(image, tarFile string) error {
 	return nil
 }
 
+// Unpack copies a directory from a local container image to a directory in the local filesystem.
+func (r *ContainerCommandRunner) Unpack(image, src, dst string) error {
```
ecordell, line 106: Shouldn't this replace `Save` entirely? I don't think we want to use save at all, unless I am missing a case where we want it.

cmd/opm/alpha/bundle/validate.go
njhale: 👍

pkg/containertools/runner.go
njhale: Good point; it should. I'll drop `Save` in favor of `Unpack`.

PR #356, opened 2020-06-10
Bug 1846100: fix(server): make a copy of the db before migrating it

ecordell:
Manual cherry-pick of #332


sferich888:
@ecordell are we planning on also manually creating a backport for 4.3? or will we try and cherry pick this back to 4.3?

PR #358, opened 2020-06-11
pkg/lib/bundle/generate.go: do not compare number of annotations

estroz:
* pkg/lib/bundle/generate.go: `ValidateAnnotations` does not compare number of annotations in existing vs expected

**Motivation for the change:** `ValidateAnnotations` should not error when arbitrary annotations exist in `annotations.yaml`.

**Reviewer Checklist**

/cc @ecordell @kevinrizza @dinhxuanvu


dinhxuanvu:
I forgot that this change may need some corrections in unit test. @estroz

estroz:
@dinhxuanvu are the tests you're referring to different from those in `make test-unit`?

PR #359, opened 2020-06-12
In permissive mode, allow pull to fail and to fallback on a local image

### LINKED ISSUE #226: opm: `index add` requires image be present remotely
Building an index image with a local-only image results in an authorization error, resulting from an image not being present remotely for pulling. The container tool used does not matter.

To reproduce:
```sh
$ operator-sdk new memcached-operator
$ operator-sdk add api --api-version example.com/v1 --kind Memcached
$ operator-sdk generate csv --csv-version 0.0.1
$ opm alpha bundle build \
	--tag quay.io/example/memcached-operator:v0.1.0 \
    --directory ./deploy/olm-catalog/memcached-operator \
    --package memcached-operator \
    --channels stable \
    --default stable
$ opm index add \
	--bundles quay.io/example/memcached-operator:v0.1.0 \
	--tag quay.io/example/memcached-operator-index:v1.0.0 \
	--container-tool docker
INFO[0000] building the index                            bundles="[quay.io/example/memcached-operator:v0.1.0]"
INFO[0000] running docker pull                           img="quay.io/example/memcached-operator:v0.1.0"
ERRO[0001] Error response from daemon: unauthorized: access to the requested resource is not authorized  img="quay.io/example/memcached-operator:v0.1.0"
FATA[0001] permissive mode disabled                      bundles="[quay.io/example/memcached-operator:v0.1.0]" error="error loading bundle from image: error pulling image: Error response from daemon: unauthorized: access to the requested resource is not authorized\n. exit status 1"
```

/kind bug

Jamstah:
Allow for pull to fail in permissive without skipping that bundle, if the bundle image has been pulled already.

Closes #226 

Ability to add packages without needing a remote registry.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -5,10 +5,11 @@ import (
 	"crypto/x509"
 	"database/sql"
 	"fmt"
-	"github.com/operator-framework/operator-registry/pkg/containertools"
-	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
 	"io/ioutil"
 	"os"
+
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
 
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -95,7 +96,7 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		simpleRefs = append(simpleRefs, image.SimpleReference(ref))
 	}
 
-	if err := populate(context.TODO(), dbLoader, graphLoader, dbQuerier, reg, simpleRefs, request.Mode); err != nil {
+	if err := populate(context.TODO(), dbLoader, graphLoader, dbQuerier, reg, simpleRefs, request.Mode, request.Permissive); err != nil {
 		err = fmt.Errorf("error loading bundle from image: %s", err)
 		if !request.Permissive {
 			r.Logger.WithError(err).Error("permissive mode disabled")
@@ -108,7 +109,7 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 	return utilerrors.NewAggregate(errs) // nil if no errors
 }
 
-func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode) error {
+func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode, permissive bool) error {
 	var errs []error
 
 	unpackedImageMap := make(map[image.Reference]string, 0)
@@ -122,7 +123,9 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 
 		if err = reg.Pull(ctx, ref); err != nil {
 			errs = append(errs, err)
-			continue
+			if !permissive {
+				continue
+			}
 		}
 
 		if err = reg.Unpack(ctx, ref, workingDir); err != nil {


pkg/lib/registry/registry.go
```diff
@@ -108,7 +109,7 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 	return utilerrors.NewAggregate(errs) // nil if no errors
 }
 
-func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode) error {
+func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode, permissive bool, logger *logrus.Entry) error {
```
kevinrizza, line 112: nit: Seems like the logger being passed in is not used

pkg/lib/registry/registry.go
```diff
@@ -122,7 +123,9 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 
 		if err = reg.Pull(ctx, ref); err != nil {
 			errs = append(errs, err)
-			continue
+			if !permissive {
+				continue
```
kevinrizza, line 127: Can we log a warning here?

pkg/lib/registry/registry.go
Jamstah: Sorry, yes, I originally logged a warning too.

pkg/lib/registry/registry.go
Jamstah: I did start like this, but then found that the pull error logs its own error, so thought it wasn't needed.

Jamstah:
> Right now, Permissive is used to describe whether or not the database should be allowed to use data that does not strictly follow certain schema conventions -- this is mostly a backwards compatibility feature so that certain users can continue using older manifests that don't conform to certain standards we added to the project later. Can we instead define an explicit flag for use-local-images or something similar? That way a user would be explicitly opting into the "don't do the pull" mechanism if that's the only thing they actually want (rather than to have potentially invalid manifests).

We could do this, can you point to documentation of the permissive field? I couldn't find it.

You say it's just for schema, but its already used to ignore pull problems in the case that not all bundles can be pulled.

kevinrizza:
> You say it's just for schema, but its already used to ignore pull problems in the case that not all bundles can be pulled.

There is unfortunately very little documentation for this particular feature, specifically because it's something that's used for specific user's backwards compatibility concerns. I would argue that ignoring pull failures should actually be separated from that, since it feels like somewhat of a regression.

At any rate though, continuing to overload even more meaning into that flag will just cause even more trouble if/when becoming more strict about the meaning. That's why I think making this an explicit option and separating that single concern would be better. Is there a reason why you feel it's not?

kevinrizza:
Also please remove the binary file from the commit.

Jamstah:
> Also please remove the binary file from the commit.

omg how embarrassing

Jamstah:
> > You say it's just for schema, but its already used to ignore pull problems in the case that not all bundles can be pulled.
> 
> There is unfortunately very little documentation for this particular feature, specifically because it's something that's used for specific user's backwards compatibility concerns. I would argue that ignoring pull failures should actually be separated from that, since it feels like somewhat of a regression.
> 
> At any rate though, continuing to overload even more meaning into that flag will just cause even more trouble if/when becoming more strict about the meaning. That's why I think making this an explicit option and separating that single concern would be better. Is there a reason why you feel it's not?

I wasn't asking for doc to be contentious, I really do think it would be better to separate concerns, and was just interested in what permissive mode was supposed to be, because I couldn't find it.

I have no objections to moving this to a new option, or even to split the other pull failure ignore out into the same/different new option too to clean up the meaning of permissive.

I did however work around this issue differently in my build env and am no longer invested in this PR, and doubt I'd get time to implement/test this as a new option in the near future.

Jamstah:
As this implementation approach is not ideal, I'm going to close this PR.

PR #363, opened 2020-06-19
feat(store): store provided apis as generic "properties"

ecordell:
Adds a properties table, and migrates provided apis to the properties table. Properties and dependencies share the same `type`s (currently `olm.gvk` and `olm.package`

Provided APIs and package/version info are now surfaced in the properties field of the bundle api.

this is to begin to support treating properties / dependencies as
generic attributes, instead of tying them to very specific hard-coded
notions like version and gvk.

**Reviewer Checklist**


pkg/sqlite/migrations/006_associate_apis_with_bundle_test.go
```diff
@@ -157,3 +154,41 @@ func oldGetChannelEntriesThatProvide(db *sql.DB, group, version, kind string) (e
 	}
 	return
 }
+
+func newGetChannelEntriesThatProvide(db *sql.DB, group, version, kind string) (entries []*registry.ChannelEntry, err error) {
```
ecordell, line 158: this replicates the previous query for the db tests, since the current one uses the properties field (that doesn't exist at this stage of migration)

pkg/sqlite/query.go
```diff
@@ -571,75 +613,153 @@ func (s *SQLQuerier) GetImagesForBundle(ctx context.Context, csvName string) ([]
...
+		})
+		groups[dep.Group] = struct{}{}
+		versions[dep.Version] = struct{}{}
+		kinds[dep.Kind] = struct{}{}
+	}
+	if err := requiredRows.Close(); err != nil {
+		return nil, nil, err
+	}
+
+	argsFor := func(s map[string]struct{}) string {
```
ecordell, line 694: this is a lot of extra work to grab the plurals from the api table - would like to argue that we don't need to bother with it, but for now I just made it completely backwards-compatible just in case.

pkg/api/registry.proto
```diff
@@ -54,7 +59,8 @@ message Bundle{
 	repeated GroupVersionKind requiredApis = 8;
 	string version = 9;
 	string skipRange = 10;
-      repeated Dependency dependencies = 11;
+    repeated Dependency dependencies = 11;
```
dinhxuanvu, line 62: nit: indentation

pkg/sqlite/load.go
```diff
@@ -725,3 +739,74 @@ func (s *sqlLoader) addDependencies(tx *sql.Tx, bundle *registry.Bundle) error {
 
 	return nil
 }
+
+func (s *sqlLoader) addPackageProperty(tx *sql.Tx, bundleName, pkg, version string) error {
```
dinhxuanvu, line 743: I can see `addPackageProperty` and `addGVKProperties` are for the 2 specific property types. In order to make this generic, should we add a generic func `addProperty` which will take a `type` and a `value`? Then, simply add those values to the `properties` table? Or this part will be addressed in subsequent PR?

pkg/api/registry.proto
dinhxuanvu, line 62: nit: Indentation

pkg/sqlite/query.go
```diff
@@ -571,75 +613,153 @@ func (s *SQLQuerier) GetImagesForBundle(ctx context.Context, csvName string) ([]
...
-	providedQuery := `SELECT DISTINCT api.group_name, api.version, api.kind, api.plural FROM api
-		 	  		  INNER JOIN channel_entry ON channel_entry.operatorbundle_name = api_provider.operatorbundle_name
-					  INNER JOIN operatorbundle ON operatorbundle.name=channel_entry.operatorbundle_name
-		 	  		  INNER JOIN api_provider ON (api.group_name=api_provider.group_name AND api.version=api_provider.version AND api.kind=api_provider.kind AND operatorbundle.name=api_provider.operatorbundle_name)
-			  		  WHERE channel_entry.entry_id=?`
+	groups := map[string]struct{}{}
+	kinds := map[string]struct{}{}
+	versions := map[string]struct{}{}
+
+	providedQuery := `SELECT DISTINCT properties.value FROM properties
```
benluddy, line 620: Would it be simpler not to collapse duplicate properties at this layer (or maybe at all)? My preference would be to keep properties as opaque as possible at the storage layer especially, and there's something unsatisfying about getting a different number of properties out on the read end than there were on the write end.

Besides, for a property type with JSON-encoded values, two strings could represent different but equivalent encodings of a single object.

pkg/sqlite/query.go
```diff
@@ -922,3 +1084,36 @@ func (s *SQLQuerier) GetDependenciesForBundle(ctx context.Context, name, version
...
+	if err != nil {
+		return nil, err
+	}
+	properties = []*api.Property{}
+	for rows.Next() {
+		var typeName sql.NullString
+		var value sql.NullString
+
+		if err := rows.Scan(&typeName, &value); err != nil {
+			return nil, err
```
benluddy, line 1104: Could return without closing `rows` here.

pkg/sqlite/query.go
dinhxuanvu: I'm not concerned about your second point because `json.Marshal` will sort the map keys. We can guarantee the JSON will be the same if the map has the same items.

pkg/server/server_test.go
```diff
@@ -526,6 +616,54 @@ func TestListBundles(t *testing.T) {
...
...,\"verbs\":[\"*\"]},{\"apiGroups\":[\"\"],\"resources\":[\"secrets\"],\"verbs\":[\"get\"]}],\"serviceAccountName\":\"etcd-operator\"}]},\"strategy\":\"deployment\"},\"keywords\":[\"etcd\",\"key value\",\"database\",\"coreos\",\"open source\"],\"labels\":{\"alm-owner-etcd\":\"etcdoperator\",\"operated-by\":\"etcdoperator\"},\"links\":[{\"name\":\"Blog\",\"url\":\"https://coreos.com/etcd\"},{\"name\":\"Documentation\",\"url\":\"https://coreos.com/operators/etcd/docs/latest/\"},{\"name\":\"etcd Operator Source Code\",\"url\":\"https://github.com/coreos/etcd-operator\"}],\"maintainers\":[{\"email\":\"support@coreos.com\",\"name\":\"CoreOS, Inc\"}],\"maturity\":\"alpha\",\"provider\":{\"name\":\"CoreOS, Inc\"},\"relatedImages\":[{\"image\":\"quay.io/coreos/etcd@sha256:3816b6daf9b66d6ced6f0f966314e2d4f894982c6b1493061502f8c2bf86ac84\",\"name\":\"etcd-v3.4.0\"},{\"image\":\"quay.io/coreos/etcd@sha256:49d3d4a81e0d030d3f689e7167f23e120abf955f7d08dbedf3ea246485acee9f\",\"name\":\"etcd-3.4.1\"}],\"replaces\":\"etcdoperator.v0.9.0\",\"selector\":{\"matchLabels\":{\"alm-owner-etcd\":\"etcdoperator\",\"operated-by\":\"etcdoperator\"}},\"skips\":[\"etcdoperator.v0.9.1\"],\"version\":\"0.9.2\"}}",
```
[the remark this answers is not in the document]
ecordell: Which part?

PR #365, opened 2020-06-22
Don't add empty related image references to index.

benluddy:
There have been a several instances of CSVs accidentally specifying the empty string in a `relatedImages` entry, which can be confusing to index readers down the line. This change omits the empty image references at index-add time.

Bundle validation to help authors detect this condition is also valuable, but not part of this change.

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	_ "github.com/mattn/go-sqlite3"
+	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/operator-framework/operator-registry/pkg/registry"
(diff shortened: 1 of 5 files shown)


pkg/sqlite/load.go
```diff
@@ -114,6 +116,14 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 		return err
 	}
 	for img := range imgs {
+		if img == "" {
+			// This condition is something that exists in
+			// the wild and must be resolved by operator
+			// authors, so it's not currently considered
+			// fatal at catalog build time.
+			s.logger.Warnf("warning: operator %q specifies an empty image reference")
```
kevinrizza, line 124: Should we be more descriptive here and potentially point the user to the actual bundle path that specifies the empty image reference if this is being loaded from a bundle image?

ecordell: Isn't this also not providing a value for the `%q` field?

Issue #367, opened 2020-06-22
Problem adding bundles after operator is removed when missing replaces


mvalahtv:
Creating index image from scratch
```
$ /tmp/operator-test/bin/opm index add -u docker --bundles quay.io/operator_testing/aqua-operator:v1.0.1,quay.io/operator_testing/aqua-operator:v1.0.2 --tag quay.io/operator_testing/upstream-community-operators-index:latest --skip-tls
INFO[0000] building the index                            bundles="[quay.io/operator_testing/aqua-operator:v1.0.1 quay.io/operator_testing/aqua-operator:v1.0.2]"
INFO[0000] resolved name: quay.io/operator_testing/aqua-operator:v1.0.1 
INFO[0000] fetched                                       digest="sha256:10f7ada4950f9f74ec90ff957caa35812262c9e975aacc9dbca7e628539fd790"
INFO[0001] fetched                                       digest="sha256:da517556988c3a2fa29320c67d1bf39dce9f5d3b852f0f54f15e8ff212acddde"
INFO[0001] fetched                                       digest="sha256:13a1ead6c4fcdb26c3b17c0690f6c9a6562f0838d6c79cfd6ba172ef4526f362"
INFO[0001] fetched                                       digest="sha256:c1ff09bbf7538894c683457efab44e93e8e7b63751b4d5614695462df570f4b8"
INFO[0002] unpacking layer: {application/vnd.docker.image.rootfs.diff.tar.gzip sha256:13a1ead6c4fcdb26c3b17c0690f6c9a6562f0838d6c79cfd6ba172ef4526f362 4445 [] map[] <nil>} 
INFO[0002] unpacking layer: {application/vnd.docker.image.rootfs.diff.tar.gzip sha256:da517556988c3a2fa29320c67d1bf39dce9f5d3b852f0f54f15e8ff212acddde 286 [] map[] <nil>} 
INFO[0002] resolved name: quay.io/operator_testing/aqua-operator:v1.0.2 
INFO[0002] fetched                                       digest="sha256:4d55cffc64b67a1559c13d7fe91a7a19f1916d3bf7adbecb7be9e643e9bb992b"
INFO[0003] fetched                                       digest="sha256:3ae2a56f3a776da7bdc7024cb261eac22051bf64cb6d3ae64c891df64149904f"
INFO[0003] fetched                                       digest="sha256:463d9e36ed63bb06c9dc12e99ac8fd5cdfea7141ca25330b9529c4777707a4e8"
[...]

PR #369, opened 2020-06-26
fix: registry rm orphaned bundles

kevinrizza:
In some cases, bundles can be orphaned from channel graphs (for example,
a new bundle could be added as the head of the channel that does not
specify a replacement). When that happens, registry/index rm commands in
opm cannot remove these bundles because the bundles themselves do not
have an association in the database with the package directly.

This commit modifies the operatorbundle table to include the package
name of the bundle so that bundles can be directly referenced when
calling registry rm, rather than being joined on the channel_entry table
(and thus beholden to the graph itself).

This commit includes a migration to update existing bundles to associate
the package name with bundles, however in the case that the bundle has
been orphaned we have no guarantee that a bundle should be associated
with a specific package name. As such, there could be existing indexes
that still include these orphaned bundles and cannot be reassociated
automatically.

Related to https://github.com/operator-framework/operator-registry/issues/367

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -67,7 +67,7 @@ func (s *sqlLoader) AddOperatorBundle(bundle *registry.Bundle) error {
 }
 
 func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error {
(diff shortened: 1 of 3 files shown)


njhale:
Closing due to staleness.

PR #370, opened 2020-06-29
Add timeout to opm registry serve

Jobava:
This relates to our project to serve operator metadata via pyxis:
CLOUDWF-1739

Add new flag `--timeout-seconds/-s` to `opm registry serve`
which allows specifying that the server should be shut down
after a number of seconds.

The server is gracefully shut down.

We are extracting information from the iib container and normally need this container up for a very short time
(1 min or less). Having a daemon complicates our setup (i.e. we'd need to handle server start/stop manually).

**Reviewer Checklist**

Some manual testing done:

```bash
[opm (master *%=)]$ ./bin/opm registry serve --database scratchpad/database/index.db -s 1
WARN[0000] unable to set termination log path            error="open /dev/termination-log: permission denied"
INFO[0000] Keeping server open for 1 seconds             database=scratchpad/database/index.db port=50051
INFO[0000] serving registry                              database=scratchpad/database/index.db port=50051
INFO[0001] Timeout expired. Gracefully stopping.         database=scratchpad/database/index.db port=50051

[opm (master *%=)]$ ./bin/opm registry serve --database scratchpad/database/index.db -s 2
WARN[0000] unable to set termination log path            error="open /dev/termination-log: permission denied"
INFO[0000] Keeping server open for 2 seconds             database=scratchpad/database/index.db port=50051
INFO[0000] serving registry                              database=scratchpad/database/index.db port=50051
INFO[0002] Timeout expired. Gracefully stopping.         database=scratchpad/database/index.db port=50051

[opm (master *%=)]$ ./bin/opm registry serve --database scratchpad/database/index.db
WARN[0000] unable to set termination log path            error="open /dev/termination-log: permission denied"
INFO[0000] Keeping server open for infinite seconds      database=scratchpad/database/index.db port=50051
INFO[0000] serving registry                              database=scratchpad/database/index.db port=50051
^C

[opm (master *%=)]$ ./bin/opm registry serve --database scratchpad/database/index.db --timeout-seconds infinite
WARN[0000] unable to set termination log path            error="open /dev/termination-log: permission denied"
INFO[0000] Keeping server open for infinite seconds      database=scratchpad/database/index.db port=50051
INFO[0000] serving registry                              database=scratchpad/database/index.db port=50051
^C

[opm (master *%=)]$ ./bin/opm registry serve --database scratchpad/database/index.db --timeout-seconds 1
WARN[0000] unable to set termination log path            error="open /dev/termination-log: permission denied"
INFO[0000] Keeping server open for 1 seconds             database=scratchpad/database/index.db port=50051
INFO[0000] serving registry                              database=scratchpad/database/index.db port=50051
INFO[0001] Timeout expired. Gracefully stopping.         database=scratchpad/database/index.db port=50051

[opm (master *%=)]$ ./bin/opm registry serve --database scratchpad/database/index.db
WARN[0000] unable to set termination log path            error="open /dev/termination-log: permission denied"
INFO[0000] Keeping server open for infinite seconds      database=scratchpad/database/index.db port=50051
INFO[0000] serving registry                              database=scratchpad/database/index.db port=50051
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ BUILD_DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 
 
 .PHONY: all
-all: clean test build sanity-check
+all: clean test build
 
 $(CMDS):
 	$(GO) build $(extra_flags) -o $@ ./cmd/$(notdir $@)
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ BUILD_DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 
 
 .PHONY: all
-all: clean test build
+all: clean test build sanity-check
 
 $(CMDS):
 	$(GO) build $(extra_flags) -o $@ ./cmd/$(notdir $@)
@@ -27,6 +27,18 @@ static: build
 .PHONY: unit
 unit:
 	$(GO) test $(SPECIFIC_UNIT_TEST) -count=1 -v -race ./pkg/...
+
+.PHONY: sanity-check
+sanity-check:
+	# Build a container with the most recent binaries for this project.
+	# Does not include the database, which needs to be added separately.
+	docker build -f upstream-builder.Dockerfile -t sanity-container .
+
+	# TODO: add more invocations of the opm binary here
+
+	# serve the container for a second, using the bundles.db in testdata
+	docker run --rm -it -v "$(shell pwd)"/pkg/lib/indexer/testdata/:/database sanity-container \
+		./bin/opm registry serve --database /database/bundles.db --timeout-seconds 1
 
 .PHONY: image
 image:
--- a/cmd/opm/registry/serve.go
+++ b/cmd/opm/registry/serve.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"strconv"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -42,8 +44,9 @@ func newRegistryServeCmd() *cobra.Command {
 	rootCmd.Flags().StringP("port", "p", "50051", "port number to serve on")
 	rootCmd.Flags().StringP("termination-log", "t", "/dev/termination-log", "path to a container termination log file")
 	rootCmd.Flags().Bool("skip-migrate", false, "do  not attempt to migrate to the latest db revision when starting")
-	return rootCmd
+	rootCmd.Flags().String("timeout-seconds", "infinite", "Timeout in seconds. This flag will be removed later.")
 
+	return rootCmd
 }
 
 func serveFunc(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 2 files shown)


cmd/opm/registry/serve.go
```diff
@@ -106,7 +109,27 @@ func serveFunc(cmd *cobra.Command, args []string) error {
...
+
+	timeout, err := cmd.Flags().GetString("timeout-seconds")
+	if err != nil {
+		return err
+	}
+
 	s := grpc.NewServer()
+	logger.Printf("Keeping server open for %s seconds", timeout)
+	if timeout != "infinite" {
+		timeoutSeconds, err := strconv.ParseInt(timeout, 10, 16)
```
Bowenislandsong, line 121: anything stopping people from inserting negative number?

dinhxuanvu: We should use `ParseUint` here to avoid negative numbers.

dinhxuanvu:
@kevinrizza @ecordell PTAL.

ecordell:
Is the [timeout](https://man7.org/linux/man-pages/man1/timeout.1.html) command not available to you on the target system? It's part of coreutils.

cmd/opm/registry/serve.go
Jobava: Fixed.

Jobava:
> Is the [timeout](https://man7.org/linux/man-pages/man1/timeout.1.html) command not available to you on the target system? It's part of coreutils.

Not from the context of an index container, which has just scratch, the go binaries and a sqlite db. Our use case requires us pulling the index images and we have no fancy shell features.

Jobava:
Per @gallettilance's request, I also added a test for CLI parameters. 

First I considered doing a unit test for the cobra invocations, but that got a bit complicated.

Instead I'm just doing an invocation of the opm binary with the parameters from the context of a container. I found a `bundles.db` file in the source and using that as a basis for `opm registry serve`.

kevinrizza:
@Jobava 

I'm having a hard time understanding the use case for this switch. 
1. Downstream the index containers are built on UBI and would have access to coreutils IIUC. 
2. That is what I imagine IIB should be using for the binary images, therefore the images that IIB is creating. 
3. What data is being pulled out of the image? I wonder if there is a way to expose that on the image instead of needing to run it for 2s?

kevinrizza:
That being said, there's nothing wrong with this code, I just want to make sure we're solving the problem the right way.

Jobava:
> @Jobava
> 
> I'm having a hard time understanding the use case for this switch.
> 
>     1. Downstream the index containers are built on UBI and would have access to coreutils IIUC.
> 
>     2. That is what I imagine IIB should be using for the binary images, therefore the images that IIB is creating.
> 
>     3. What data is being pulled out of the image? I wonder if there is a way to expose that on the image instead of needing to run it for 2s?

Our project requires us to collect this data and serve it from Pyxis. The data will be ultimately used by Catalog or others for display purposes.

As to why we want to serve it from pyxis and not just let Catalog collect it... that's because index data uses container registries for storage and pulling a whole index image is pretty expensive [1]. So we have to load every single index container, extract the operator metadata and store it. Having a timeout here just makes our job a bit easier (i.e. we can load data from kubernetes Jobs).

[1] - ~ 500 MiB of data (the docker image size) for 50KiB of actual index metadata (in uncompressed yaml form)

kevinrizza:
If `opm` had a command that could extract the contents of the index, wouldn't running the image at all be duplicate effort? `opm` already has `opm index extract` which fetches the manifests from all the bundles and puts them onto local disk. What if we just extended that to provide a serialized json/yaml blob that could be read by your tool directly, without the need for a kube cluster at all?

Jobava:
After talking to @kevinrizza , I removed the short version of the flag. We agreed that eventually opm will have a function to export data as json and after that our tooling will switch to using that. After such time we can also remove the --timeout-seconds flag.

/retest

Jobava:
Is this good for merging?

Makefile
```diff
@@ -9,7 +9,7 @@ BUILD_DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 
 
 .PHONY: all
-all: clean test build
+all: clean test build sanity-check
```
kevinrizza, line 12: Do we need this in `make all`? I'm okay with adding this addition make target, but I question the need to run it by default.

kevinrizza:
Aside from that one comment, I think this is fine to merge.

Makefile
Jobava: I will remove it from `make all`.

Jobava:
Hi again. Can we get an idea of when this can be merged?

Jobava:
@kevinrizza I see the test has failed, but can't tell why it is failing

PR #374, opened 2020-07-01
Separate build/pull tool for index rm

kevinrizza:
Like the index add command, index rm can separate the tool/lib that
pulls and unpacks index images from the tool that builds them. This
commit introduces that separation to index rm, defaulting pull tool to
none(using internal containerd libs for pulling) from the build tool
that defaults to podman.

### DIFF
--- a/cmd/opm/index/delete.go
+++ b/cmd/opm/index/delete.go
@@ -1,7 +1,6 @@
 package index
 
 import (
-	"fmt"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


cmd/opm/index/delete.go
```diff
@@ -75,15 +76,11 @@ func runIndexDeleteCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	containerTool, err := cmd.Flags().GetString("container-tool")
+	pullTool, buildTool, err := getContainerTools(cmd)
 	if err != nil {
 		return err
 	}
 
-	if containerTool == "none" {
```
Bowenislandsong, line 83: I know we have a lot of conversations around non-tool. I couldn't keep track of what's going on there, Is there any reason we are removing the check here? Thanks.

otherwise lgtm

kevinrizza: The concept of that check was just moved into the existing private method `getContainerTools()`, which I'm reusing the logic for:

https://github.com/operator-framework/operator-registry/blob/master/cmd/opm/index/add.go#L162

Bowenislandsong: ahh, nice. Thanks

/lgtm
/hold 
waiting for a second review

shawn-hurley:
/override ci/prow/e2e-aws 

Overriding because the test is flaking and this is critical fix for production. The Travis test passed is in verifying this change

PR #384, opened 2020-07-13
Check before insert

shawn-hurley:

* adding sanity check before adding a bundle to check it the package/csv/version is already provided or if the exact bundle is already added.
* exit code's now can be used to retrieve these specific error scenarios. 
* This is targeted at index add, but anything that uses the directory loader will get this change.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/main.go
+++ b/cmd/opm/main.go
@@ -5,13 +5,13 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/operator-framework/operator-registry/cmd/opm/alpha"
 	"github.com/operator-framework/operator-registry/cmd/opm/index"
 	"github.com/operator-framework/operator-registry/cmd/opm/registry"
 	"github.com/operator-framework/operator-registry/cmd/opm/version"
 	registrylib "github.com/operator-framework/operator-registry/pkg/registry"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 func main() {
@@ -37,14 +37,16 @@ func main() {
 	}
 
 	if err := rootCmd.Execute(); err != nil {
-		if agg, ok := err.(utilerrors.Aggregate); ok {
-			for _, e := range agg.Errors() {
-				if _, ok := e.(registrylib.BundleAlreadyAddedError); ok {
-					os.Exit(2)
-				}
-				if _, ok := e.(registrylib.BundleAlreadyInDatabaseError); ok {
-					os.Exit(3)
-				}
+		agg, ok := err.(utilerrors.Aggregate)
+		if !ok {
+			os.Exit(1)
+		}
+		for _, e := range agg.Errors() {
+			if _, ok := e.(registrylib.BundleImageAlreadyAddedErr); ok {
+				os.Exit(2)
+			}
+			if _, ok := e.(registrylib.PackageVersionAlreadyAddedErr); ok {
+				os.Exit(3)
 			}
 		}
 		os.Exit(1)
--- a/cmd/opm/main.go
+++ b/cmd/opm/main.go
@@ -10,6 +10,8 @@ import (
 	"github.com/operator-framework/operator-registry/cmd/opm/index"
 	"github.com/operator-framework/operator-registry/cmd/opm/registry"
 	"github.com/operator-framework/operator-registry/cmd/opm/version"
+	registrylib "github.com/operator-framework/operator-registry/pkg/registry"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 func main() {
@@ -35,6 +37,16 @@ func main() {
 	}
 
 	if err := rootCmd.Execute(); err != nil {
+		if agg, ok := err.(utilerrors.Aggregate); ok {
+			for _, e := range agg.Errors() {
+				if _, ok := e.(registrylib.BundleAlreadyAddedError); ok {
+					os.Exit(2)
+				}
+				if _, ok := e.(registrylib.BundleAlreadyInDatabaseError); ok {
+					os.Exit(3)
+				}
+			}
+		}
 		os.Exit(1)
 	}
 }
--- a/pkg/lib/indexer/indexer.go
+++ b/pkg/lib/indexer/indexer.go
@@ -85,6 +85,7 @@ func (i ImageIndexer) AddToIndex(request AddToIndexRequest) error {
 	// Add the bundles to the registry
 	err = i.RegistryAdder.AddToRegistry(addToRegistryReq)
 	if err != nil {
+		i.Logger.WithError(err).Debugf("unable to add bundle to registry")
 		return err
 	}
 
@@ -221,7 +222,6 @@ func (i ImageIndexer) PruneFromIndex(request PruneFromIndexRequest) error {
 
 	return nil
 }
-
 
 // extractDatabase sets a temp directory for unpacking an image
 func (i ImageIndexer) extractDatabase(buildDir, fromIndex string) (string, error) {
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -8,14 +8,13 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/operator-framework/operator-registry/pkg/containertools"
-	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
-
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
+	"github.com/operator-framework/operator-registry/pkg/containertools"
 	"github.com/operator-framework/operator-registry/pkg/image"
 	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
+	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
 	"github.com/operator-framework/operator-registry/pkg/registry"
 	"github.com/operator-framework/operator-registry/pkg/sqlite"
 )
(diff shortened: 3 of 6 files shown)


pkg/lib/indexer/indexer.go
```diff
@@ -85,6 +85,7 @@ func (i ImageIndexer) AddToIndex(request AddToIndexRequest) error {
 	// Add the bundles to the registry
 	err = i.RegistryAdder.AddToRegistry(addToRegistryReq)
 	if err != nil {
+		fmt.Printf("add to registry error: %#v", err)
```
gallettilance, line 88: How about:
```suggestion
		return fmt.Errorf("add to registry error: %#v", err)
```

pkg/lib/registry/registry.go
```diff
@@ -96,16 +95,17 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 	}
 
 	if err := populate(context.TODO(), dbLoader, graphLoader, dbQuerier, reg, simpleRefs, request.Mode); err != nil {
-		err = fmt.Errorf("error loading bundle from image: %s", err)
+		r.Logger.Debugf("unable to populate database: %s", err)
```
gallettilance, line 98: Not sure I understand why this change is needed

pkg/registry/populator.go
```diff
@@ -63,7 +63,54 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
...
+				errs = append(errs, err)
+				return utilerrors.NewAggregate(errs)
+			}
+			if bundle != nil && !isExactImage {
+				// raise error that this package + channel + csv combo is already in the db
+				errs = append(errs, BundleAlreadyInDatabaseError{ErrorString: "Bundle already added that provides package and csv"})
+				continue
+			}
+			if bundle != nil && isExactImage {
+				// raise error that this is already in the db
```
gallettilance, line 97:
```suggestion
				// raise error that this bundle and channel is already in the db
```

pkg/registry/populator.go
```diff
@@ -63,7 +63,54 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
...
+				errs = append(errs, BundleAlreadyAddedError{ErrorString: "Bundle already added"})
```
gallettilance, line 98:
```suggestion
				errs = append(errs, BundleAlreadyAddedError{ErrorString: fmt.Sprintf("Bundle %s already exists for channel %s", image.bundle.csv.GetName(), channel)})
```

cmd/opm/main.go
```diff
@@ -35,6 +37,16 @@ func main() {
 	}
 
 	if err := rootCmd.Execute(); err != nil {
+		if agg, ok := err.(utilerrors.Aggregate); ok {
+			for _, e := range agg.Errors() {
+				if _, ok := e.(registrylib.BundleAlreadyAddedError); ok {
+					os.Exit(2)
```
gallettilance, line 43: neat

pkg/registry/types.go
```diff
@@ -14,6 +14,24 @@ var (
 	ErrPackageNotInDatabase = errors.New("Package not in database")
 )
 
+// BundleAlreadyAddedError is an error that describes a bundle is already added
```
kevinrizza, line 17: What is the semantic difference between these two error types? Is there value in differentiating between that exact bundle was already added vs that specific csv was already added? Or is this doing something more subtle than what I'm reading?

Either way, the comments/naming here is confusing.

pkg/registry/populator.go
```diff
@@ -63,7 +63,55 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
...
+				isExactImage = true
+			}
+		}
+		for _, channel := range image.bundle.Channels {
+			bundle, err := i.querier.GetBundle(context.TODO(), image.bundle.Package, channel, image.bundle.csv.GetName())
+			if err != nil {
+				// Assume that if we can not find a bundle for the package, channel and or CSV Name that this is safe to add
+				continue
+			}
+			if bundle != nil && !isExactImage {
```
kevinrizza, line 92: This wouldn't really need to be per image though, right? Even in disparate channels, you can't add the same bundle to the registry twice.

pkg/registry/populator.go
```diff
@@ -63,7 +63,55 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
 	return nil
 }
 
+func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error {
```
ecordell, line 66: Just to clarify - these errors should already be returned by the normal loader, and this PR is specifically to get different return codes for some cases?

This seems fine for now, but IMO we should look at doing something similar to what we plan to do for bundle validation, and output a set of error codes (i.e. in json).

pkg/registry/populator.go
```diff
@@ -63,7 +63,55 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
...
+	var errs []error
+	images := make(map[string]bool)
+	for _, image := range imagesToAdd {
+		images[image.bundle.BundleImage] = true
+	}
+
+	for _, image := range imagesToAdd {
```
ecordell, line 73: this is making a lot of queries that could be wrapped up into larger sql queries, but it's not worth doing that now, IMO

pkg/registry/populator.go
shawn-hurley: +1 on adding that in the future.

pkg/registry/populator_test.go
```diff
@@ -426,6 +505,18 @@ func TestImageLoading(t *testing.T) {
 	}
 }
 
+func checkAggErr(aggErr, wantErr error) bool {
+	if a, ok := aggErr.(utilerrors.Aggregate); ok {
+		for _, e := range a.Errors() {
```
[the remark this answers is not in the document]
shawn-hurley: I don't think that is true, I am looking through each error to see if one of them is the error that I want, not that all the errors are the one that I want.

pkg/registry/types.go
shawn-hurley: the difference is

A bundle with a different pull spec providing the same package and csv in the same channels.

vs

The exact same bundle pull spec providing the same things.

WDYT of `PackageVersionAlreadyAdded` for the first case and `BundleAlreadyAdded` for the second?

pkg/registry/populator.go
kevinrizza: Agreed, we should probably just load the entire package in once, but it doesn't need to happen now.

pkg/registry/populator.go
shawn-hurley: meaning we should exit early once we find the same bundle path on line 83?

pkg/registry/populator.go
shawn-hurley: +1

pkg/registry/types.go
kevinrizza: I like `PackageVersionAlreadyAdded`, but for the second what about `BundleImageAlreadyAdded`?

pkg/registry/populator.go
kevinrizza: Yeah, I think if `isExactImage` is set to true you should just return the error for that entire bundle and move to the next one, no need to iterate over the channels at all.

pkg/registry/types.go
shawn-hurley: +1 on BundleImageAlreadyAdded changes coming

shawn-hurley: Adding `Err` on them as well

pkg/registry/populator.go
shawn-hurley: Do you think that we could also break out of the channel check loop once we know that one is failing?

pkg/registry/populator_test.go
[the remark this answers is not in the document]
ecordell: ah, I misread. `require.Contains` would work, but it's not a big deal.

pkg/registry/populator.go
```diff
@@ -63,7 +63,50 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
 	return nil
 }
 
+func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error {
+	var errs []error
+	images := make(map[string]bool)
+	for _, image := range imagesToAdd {
+		images[image.bundle.BundleImage] = true
+	}
```
estroz, line 71: Nit:
```suggestion
	images := make(map[string]struct{})
	for _, image := range imagesToAdd {
		images[image.bundle.BundleImage] = struct{}{}
	}
```

pkg/registry/types.go
```diff
@@ -14,6 +14,24 @@ var (
 	ErrPackageNotInDatabase = errors.New("Package not in database")
 )
 
+// BundleImageAlreadyAddedErr is an error that describes a bundle is already added
```
estroz, line 17: How about adding an `ExitCoder` interface to simplify/allow extension of `opm` error code usage: 
```go
type ExitCoder interface {
	ExitCode() int
}

type BundleImageAlreadyAddedErr struct {
	ErrorString string
}

...

func (e BundleImageAlreadyAddedErr) ExitCode() int {
	return 2
}
```

pkg/registry/populator_test.go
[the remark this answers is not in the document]
shawn-hurley: hum looks that this actually compares error strings and not error types:

```
    TestImageLoading/AddExactBundleAlreadyExists: populator_test.go:509:
                Error Trace:    populator_test.go:509
                                                        populator_test.go:489
                Error:          "[Bundle quay.io/prometheus/operator:0.14.0 already exists Bundle already added that provides package and csv]" does not contain ""
                Test:           TestImageLoading/AddExactBundleAlreadyExists
```

Just an FYI

pkg/registry/populator.go
kevinrizza: +1

pkg/registry/types.go
estroz: Disregard this comment, the exit code is bringing CLI problems into a non-CLI-specific library.

PR #388, opened 2020-07-15
Use go env to get architecture for building

Jamstah:
GOARCH is not an env var, so needs pulling from `go env` instead

Fix build for s390x

**Reviewer Checklist**


ecordell:
/ok-to-test
/lgtm
/approve

Thanks for noticing and fixing this!

PR #389, opened 2020-07-15
chore(deps): bump to kube 1.19

awgreene:


Bowenislandsong:
Does this need to import from the new API with kube1.19?

Bowenislandsong:
/hold 
Does this need to import from the new API with kube1.19?

kevinrizza:
@Bowenislandsong 

> Does this needs to import from the new API with kube1.19?

Go dependencies are filepath based, so bumping deps does not require updating imports (unless things moved, in which case there would be a compile error)

/approve

Bowenislandsong:
> @Bowenislandsong
> 
> > Does this needs to import from the new API with kube1.19?
> 
> Go dependencies are filepath based, so bumping deps does not require updating imports (unless things moved, in which case there would be a compile error)
> 
> /approve

right, same confusion. I actually meant operator-framework/api. I reckon we merge operator-fw/api first and add update that API here to test compatibility. Should not be a probelm, just want a cross integration test.

ecordell:
This will be available for reference if we need it, but I think we'll be skipping to 1.20 in the coming weeks - closing for now.

PR #390, opened 2020-07-15
Bug 1857502: fix(validate): pull bundle images before unpacking

njhale:
Pull bundle images before attempting to unpack in the bundle validator. Prevents validation from erroneously failing with "image not found".

### DIFF
--- a/pkg/image/containerdregistry/options.go
+++ b/pkg/image/containerdregistry/options.go
@@ -10,6 +10,7 @@ import (
 	"github.com/containerd/containerd/metadata"
 	"github.com/containerd/containerd/platforms"
 	"github.com/containerd/containerd/remotes"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/image/containerdregistry/options.go
```diff
@@ -99,7 +100,10 @@ func NewRegistry(options ...RegistryOption) (registry *Registry, err error) {
 		destroy:  destroy,
 		log:      config.Log,
 		resolver: resolver,
-		platform: platforms.Only(platforms.DefaultSpec()),
+		platform: platforms.Ordered(platforms.DefaultSpec(), specs.Platform{
```
njhale, line 103: This is a quick fix for getting this up and running on mac. I think we'll need to eventually match all architectures and unpack to different directories, though I don't know how well this will work for platforms w/ conflicting filesystems (e.g. different endianness)

exdx:
Thanks for the fix. I noticed that `validate` was broken recently so getting a fix out would be good.
/lgtm

PR #397, opened 2020-07-22
Adding ability to deprecate bundles

gallettilance:
Adding support marking versions as deprecated and truncating the update
graph of a given package. Deprecated versions will not be installable.
If a deprecated version is installed prior to upgrade appropriate alerts
will be fired on cluster.

Described in [this enhancement](https://github.com/operator-framework/enhancements/blob/master/enhancements/deprecated-bundles.md)

**Reviewer Checklist**

### DIFF
--- a/bundles/etcd.0.9.0/metadata/annotations.yaml
+++ b/bundles/etcd.0.9.0/metadata/annotations.yaml
@@ -1,4 +1,5 @@
 annotations:
   operators.operatorframework.io.bundle.package.v1: "etcd"
-  operators.operatorframework.io.bundle.channels.v1: "alpha,stable"
+  operators.operatorframework.io.bundle.channels.v1: "alpha,stable,beta"
   operators.operatorframework.io.bundle.channel.default.v1: "stable"
+  
--- a/cmd/opm/index/deprecate.go
+++ b/cmd/opm/index/deprecate.go
@@ -1,0 +1,133 @@
+package index
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"k8s.io/kubectl/pkg/util/templates"
+
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/lib/indexer"
+)
+
+var deprecateLong = templates.LongDesc(`
+	Deprecate and truncate operator bundles from an index.
+	
+	Deprecated bundles will no longer be installable. Bundles that are replaced by deprecated bundles will be removed enirely from the index.
+	
+	For example:
+
+		Given the update graph in quay.io/my/index:v1
+		1.4.0 -- replaces -> 1.3.0 -- replaces -> 1.2.0 -- replaces -> 1.1.0
+
+		Applying the command:
+		opm index deprecate --bundles "quay.io/my/bundle:1.3.0" --from-index "quay.io/my/index:v1" --tag "quay.io/my/index:v2"
+
+		Produces the following update graph in quay.io/my/index:v2
+		1.4.0 -- replaces -> 1.3.0 [deprecated]
+		
+	Deprecating a bundle that removes the default channel is not allowed. Changing the default channel prior to deprecation is possible by publishing a new bundle to the index.
+	`)
+
+func newIndexDeprecateTruncateCmd() *cobra.Command {
+	indexCmd := &cobra.Command{
+		Hidden: true,
+		Use:    "deprecatetruncate",
+		Short:  "Deprecate and truncate operator bundles from an index.",
+		Long:   deprecateLong,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if debug, _ := cmd.Flags().GetBool("debug"); debug {
+				logrus.SetLevel(logrus.DebugLevel)
+			}
+			return nil
+		},
+		RunE: runIndexDeprecateTruncateCmdFunc,
+	}
+
+	indexCmd.Flags().Bool("debug", false, "enable debug logging")
+	indexCmd.Flags().Bool("generate", false, "if enabled, just creates the dockerfile and saves it to local disk")
+	indexCmd.Flags().StringP("out-dockerfile", "d", "", "if generating the dockerfile, this flag is used to (optionally) specify a dockerfile name")
+	indexCmd.Flags().StringP("from-index", "f", "", "previous index to add to")
+	indexCmd.Flags().StringSliceP("bundles", "b", nil, "comma separated list of bundles to add")
+	if err := indexCmd.MarkFlagRequired("bundles"); err != nil {
+		logrus.Panic("Failed to set required `bundles` flag for `index add`")
+	}
+	indexCmd.Flags().StringP("binary-image", "i", "", "container image for on-image `opm` command")
+	indexCmd.Flags().StringP("container-tool", "c", "", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
+	indexCmd.Flags().StringP("build-tool", "u", "", "tool to build container images. One of: [docker, podman]. Defaults to podman. Overrides part of container-tool.")
+	indexCmd.Flags().StringP("pull-tool", "p", "", "tool to pull container images. One of: [none, docker, podman]. Defaults to none. Overrides part of container-tool.")
+	indexCmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
+	indexCmd.Flags().Bool("permissive", false, "allow registry load errors")
+	if err := indexCmd.Flags().MarkHidden("debug"); err != nil {
+		logrus.Panic(err.Error())
+	}
+
+	return indexCmd
+}
+
+func runIndexDeprecateTruncateCmdFunc(cmd *cobra.Command, args []string) error {
+	generate, err := cmd.Flags().GetBool("generate")
+	if err != nil {
+		return err
+	}
+
+	outDockerfile, err := cmd.Flags().GetString("out-dockerfile")
+	if err != nil {
+		return err
+	}
+
+	fromIndex, err := cmd.Flags().GetString("from-index")
+	if err != nil {
+		return err
+	}
+
+	bundles, err := cmd.Flags().GetStringSlice("bundles")
+	if err != nil {
+		return err
+	}
+
+	binaryImage, err := cmd.Flags().GetString("binary-image")
+	if err != nil {
+		return err
+	}
+
+	tag, err := cmd.Flags().GetString("tag")
+	if err != nil {
+		return err
+	}
+
+	permissive, err := cmd.Flags().GetBool("permissive")
+	if err != nil {
+		return err
+	}
+
+	pullTool, buildTool, err := getContainerTools(cmd)
+	if err != nil {
+		return err
+	}
+
+	logger := logrus.WithFields(logrus.Fields{"bundles": bundles})
+
+	logger.Info("deprecating bundles from the index")
+
+	indexDeprecator := indexer.NewIndexDeprecator(
+		containertools.NewContainerTool(buildTool, containertools.PodmanTool),
+		containertools.NewContainerTool(pullTool, containertools.NoneTool),
+		logger)
+
+	request := indexer.DeprecateFromIndexRequest{
+		Generate:          generate,
+		FromIndex:         fromIndex,
+		BinarySourceImage: binaryImage,
+		OutDockerfile:     outDockerfile,
+		Tag:               tag,
+		Bundles:           bundles,
+		Permissive:        permissive,
+	}
+
+	err = indexDeprecator.DeprecateFromIndex(request)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/lib/indexer/indexer.go
+++ b/pkg/lib/indexer/indexer.go
@@ -41,6 +41,7 @@ type ImageIndexer struct {
 	RegistryAdder       registry.RegistryAdder
 	RegistryDeleter     registry.RegistryDeleter
 	RegistryPruner      registry.RegistryPruner
+	RegistryDeprecator  registry.RegistryDeprecator
 	BuildTool           containertools.ContainerTool
 	PullTool            containertools.ContainerTool
 	Logger              *logrus.Entry
@@ -526,4 +527,62 @@ func generatePackageYaml(dbQuerier pregistry.Query, packageName, downloadPath st
 	}
 
 	return utilerrors.NewAggregate(errs)
+}
+
+// DeprecateFromIndexRequest defines the parameters to send to the PruneFromIndex API
+type DeprecateFromIndexRequest struct {
+	Generate          bool
+	Permissive        bool
+	BinarySourceImage string
+	FromIndex         string
+	OutDockerfile     string
+	Bundles           []string
+	Tag               string
+}
+
+// DeprecateFromIndex takes a DeprecateFromIndexRequest and deprecates the requested
+// bundles.
+func (i ImageIndexer) DeprecateFromIndex(request DeprecateFromIndexRequest) error {
+	buildDir, outDockerfile, cleanup, err := buildContext(request.Generate, request.OutDockerfile)
+	defer cleanup()
+	if err != nil {
+		return err
+	}
+
+	databasePath, err := i.extractDatabase(buildDir, request.FromIndex)
+	if err != nil {
+		return err
+	}
+
+	// Run opm registry prune on the database
+	deprecateFromRegistryReq := registry.DeprecateFromRegistryRequest{
+		Bundles:       request.Bundles,
+		InputDatabase: databasePath,
+		Permissive:    request.Permissive,
+	}
+
+	// Prune the bundles from the registry
+	err = i.RegistryDeprecator.DeprecateFromRegistry(deprecateFromRegistryReq)
+	if err != nil {
+		return err
+	}
+
+	// generate the dockerfile
+	dockerfile := i.DockerfileGenerator.GenerateIndexDockerfile(request.BinarySourceImage, databasePath)
+	err = write(dockerfile, outDockerfile, i.Logger)
+	if err != nil {
+		return err
+	}
+
+	if request.Generate {
+		return nil
+	}
+
+	// build the dockerfile with requested tooling
+	err = build(outDockerfile, request.Tag, i.CommandRunner, i.Logger)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -228,3 +228,37 @@ func (r RegistryUpdater) PruneFromRegistry(request PruneFromRegistryRequest) err
 
 	return nil
 }
+
+type DeprecateFromRegistryRequest struct {
+	Permissive    bool
+	InputDatabase string
+	Bundles       []string
+}
+
+func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequest) error {
+	db, err := sql.Open("sqlite3", request.InputDatabase)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	dbLoader, err := sqlite.NewSQLLiteLoader(db)
+	if err != nil {
+		return err
+	}
+	if err := dbLoader.Migrate(context.TODO()); err != nil {
+		return fmt.Errorf("unable to migrate database: %s", err)
+	}
+
+	deprecator := sqlite.NewSQLDeprecatorForBundles(dbLoader, request.Bundles)
+	if err := deprecator.Deprecate(); err != nil {
+		r.Logger.Debugf("unable to deprecate bundles from database: %s", err)
+		if !request.Permissive {
+			r.Logger.WithError(err).Error("permissive mode disabled")
+			return err
+		}
+		r.Logger.WithError(err).Warn("permissive mode enabled")
+	}
+
+	return nil
+}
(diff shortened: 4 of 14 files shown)


kevinrizza:
@shawn-hurley 

> We need to enhance add to undo the deprecation (add back deprecated bundle mark it's replaced bundle as deprecated)

Is that part necessary? I keep hearing about these requests from the pipelines that they need to have "undo" commands. But can't they just... point to a version of the index that didn't remove it yet?

gallettilance:
In the enhancement we punted on that command https://github.com/operator-framework/enhancements/blob/master/enhancements/deprecated-bundles.md#story-2 in favor of using images as @kevinrizza suggested

pkg/sqlite/query.go
```diff
@@ -1117,3 +1117,71 @@ func (s *SQLQuerier) GetPropertiesForBundle(ctx context.Context, name, version,
...
 	return
 }
+
+func (s *SQLQuerier) GetTailForBundleInChannel(ctx context.Context, csvName string) (bundleNames []string, err error) {
+	getDepthQuery := `SELECT DISTINCT channel_name, package_name, depth FROM channel_entry WHERE operatorbundle_name=?`
+
+	getTailQuery := `SELECT DISTINCT operatorbundle_name FROM channel_entry 
+				WHERE channel_name=?
+				AND package_name=?
+				AND depth > ?`
```
ecordell, line 1127: Without looking it up, I don't know if this is the right direction for `depth`. A unit test to prove it's correct would help 😄

pkg/lib/registry/registry.go
```diff
@@ -228,3 +228,89 @@ func (r RegistryUpdater) PruneFromRegistry(request PruneFromRegistryRequest) err
...
+		if err != nil || csvName == "" {
+			r.Logger.WithError(err).Warn("error getting csv for bundle ", bundlePath)
+			continue
+		}
+		version, err := dbQuerier.GetBundleVersion(context.TODO(), bundlePath)
+		if err != nil {
+			r.Logger.WithError(err).Warn("error getting version for bundle ", bundlePath)
+			continue
+		}
+		tailBundles, err := dbQuerier.GetTailForBundleInChannel(context.TODO(), csvName)
```
ecordell, line 302: I think we want this query to be based on the `(bundle, version, bundlepath)` to help future proof for a time when csvName is not unique

pkg/sqlite/load.go
```diff
@@ -805,3 +805,49 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
 
 	return nil
 }
+
+func (s *sqlLoader) rmChannelEntry(tx *sql.Tx, csvName string) error {
+	stmt, err := tx.Prepare("DELETE FROM channel_entry WHERE operatorbundle_name=?")
```
ecordell, line 810: can we use version and bundlepath here too, just to be safe?

pkg/sqlite/load.go
```diff
@@ -805,3 +805,49 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+	tx, err := s.db.Begin()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		tx.Rollback()
+	}()
+
+	for _, bundle := range tailBundles {
+		err := s.rmBundle(tx, bundle)
```
ecordell, line 833: same here - would be better to be more specific if we can

pkg/sqlite/query.go
```diff
@@ -1117,3 +1117,71 @@ func (s *SQLQuerier) GetPropertiesForBundle(ctx context.Context, name, version,
...
+		var depth sql.NullInt64
+		if err := rows.Scan(&channel, &pkg, &depth); err != nil {
+			return nil, err
+		}
+
+		tailRows, err := s.db.QueryContext(ctx, getTailQuery, channel, pkg, depth)
+		if err != nil {
+			return nil, err
+		}
+		defer tailRows.Close()
```
ecordell, line 1149: IIRC, it's racy to call `Close()` on a row object while another one (in this case, `rows`) is still not `Close`d.

pkg/lib/registry/registry.go
```diff
@@ -228,3 +228,93 @@ func (r RegistryUpdater) PruneFromRegistry(request PruneFromRegistryRequest) err
...
+	}
+
+	var reg image.Registry
+	var rerr error
+	switch request.ContainerTool {
+	case containertools.NoneTool:
+		reg, rerr = containerdregistry.NewRegistry(containerdregistry.SkipTLS(request.SkipTLS), containerdregistry.WithRootCAs(rootCAs))
+	case containertools.PodmanTool:
+		fallthrough
+	case containertools.DockerTool:
```
dinhxuanvu, line 279: Do we need a default case here to just return error in case they input something not usable?

pkg/sqlite/query.go
```diff
@@ -1117,3 +1117,71 @@ func (s *SQLQuerier) GetPropertiesForBundle(ctx context.Context, name, version,
 
 	return
 }
+
+func (s *SQLQuerier) GetTailForBundleInChannel(ctx context.Context, csvName string) (bundleNames []string, err error) {
```
dinhxuanvu, line 1121: Can you define what the `tail` means here? The tail means the oldest bundle?

pkg/sqlite/query.go
```diff
@@ -1117,3 +1117,71 @@ func (s *SQLQuerier) GetPropertiesForBundle(ctx context.Context, name, version,
...
+	getDepthQuery := `SELECT DISTINCT channel_name, package_name, depth FROM channel_entry WHERE operatorbundle_name=?`
```
dinhxuanvu, line 1122: Should you just use `ORDER BY` or something like that to just sort this by `depth`?

pkg/registry/interface.go
```diff
@@ -12,6 +12,7 @@ type Load interface {
 	AddPackageChannels(manifest PackageManifest) error
 	AddBundlePackageChannels(manifest PackageManifest, bundle *Bundle) error
 	RemovePackage(packageName string) error
+	DeprecateBundle(csvName, version, path string, tailBundles []string) error
```
dinhxuanvu, line 15: I can see the use case of wanting to query the list deprecated bundles for development reason or pipeline work.
For dependency resolution, should we modify the `ListBundles` query to exclude the deprecated bundles given those shouldn't considered for dependency resolution right? @ecordell @benluddy

gallettilance: I opened https://github.com/operator-framework/operator-lifecycle-manager/pull/1673 to filter deprecated bundles in the resolver

dinhxuanvu: I saw that PR but I wonder if we need to do that on resolver side when we can do it on the `ListBundles` query itself.

gallettilance: @ecordell made a point that it's easier to undo if it's on the resolver side

pkg/sqlite/query.go
```diff
@@ -1117,3 +1117,90 @@ func (s *SQLQuerier) GetPropertiesForBundle(ctx context.Context, name, version,
 
 	return
 }
+
+func (s *SQLQuerier) GetTailFromBundle(ctx context.Context, name string) (bundles []string, err error) {
```
kevinrizza, line 1121: The querier interface is generally to define the interface of the grpc API into requests against the database. If we aren't exposing this externally, shouldn't we put this in `loadprocs.go` or where ever it is actually used? It feels like we are abusing this interface this way.

pkg/sqlite/deprecate.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type SQLDeprecator interface {
+	Deprecate() error
+}
+
+// BundleDeprecator removes bundles from the database
+type BundleDeprecator struct {
+	store   registry.Load
+	querier registry.Query
```
kevinrizza, line 21: Why use `querier` and `store` here separately? That seems potentially dangerous over time.

gallettilance: can you elaborate on the danger?

pkg/registry/types.go
```diff
@@ -204,6 +212,11 @@ type PackageProperty struct {
 	Version string `json:"version" yaml:"version"`
 }
 
+type DeprecatedProperty struct {
+	// Whether the bundle is deprecated
+	Deprecated bool `json:"deprecated" yaml:"deprecated"`
```
benluddy, line 217: How would the ambiguity be resolved when a bundle has two properties: one with deprecated=true and one with deprecated=false? I would suggest that the presence of a "deprecated" property in the property set indicates deprecation, and remove the bool value altogether.

njhale:
```golang
    type DeprecatedProperty string
    
    func (DeprecatedProperty) String() string {
        return "deprecated"
    }
````

@benluddy something like this? or maybe just a string constant?

Comment (gallettilance) [reply pkg/registry/types.go:217]: That's a great point - thanks for catching that

Comment (gallettilance) [reply pkg/registry/types.go]: @njhale I missed your comment and implemented the property as an empty struct - let me know your thoughts

File: pkg/sqlite/load.go [hunk 13]
```diff
@@ -805,3 +805,176 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+	return "", "", registry.ErrBundleImageNotInDatabase
+}
+
+func (s *sqlLoader) DeprecateBundle(path string) error {
+	tx, err := s.db.Begin()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		tx.Rollback()
```
awgreene, line 947: For my betterment, does tx.Commit prevent this call from rollingback the transactions?

pkg/sqlite/load_test.go
```diff
@@ -232,3 +245,162 @@ func newBundle(t *testing.T, name, pkgName string, channels []string, objs ...*u
...
+							{
+								Name:           "alpha",
+								CurrentCSVName: "csv-a",
+							},
+							{
+								Name:           "stable",
+								CurrentCSVName: "csv-b",
+							},
+						},
+						DefaultChannelName: "alpha",
```
awgreene, line 321: Isn't alpha the the default channel? Why is csv-a which is the head of the alpha channel able to be removed?

pkg/sqlite/load_test.go
```diff
@@ -232,3 +245,162 @@ func newBundle(t *testing.T, name, pkgName string, channels []string, objs ...*u
...
+					"csv-b",
+					"csv-c",
+				},
+			},
+		},
+		{
+			description: "GetTailFromBundle/HandlesSkips",
+			fields: fields{
+				bundles: []*registry.Bundle{
+					newBundle(t, "csv-a", "pkg-0", []string{"alpha"}, newUnstructuredCSVwithSkips(t, "csv-a", "csv-b", "\"csv-d\",\"csv-e\",\"csv-f\"")),
```
awgreene, line 340: What does `"\"csv-d\",` mean? Does this represent skips?

gallettilance: "\"csv-d\",\"csv-e\",\"csv-f\"" is a list of skipped csvs

bundles/etcd.0.9.0/metadata/annotations.yaml
```diff
@@ -1,4 +1,4 @@
 annotations:
   operators.operatorframework.io.bundle.package.v1: "etcd"
-  operators.operatorframework.io.bundle.channels.v1: "alpha,stable"
+  operators.operatorframework.io.bundle.channels.v1: "alpha,stable,beta"
   operators.operatorframework.io.bundle.channel.default.v1: "stable"
```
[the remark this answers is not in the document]
gallettilance: I did not change this but I can add a new line :)

cmd/opm/index/deprecate.go
```diff
@@ -0,0 +1,132 @@
...
+		1.4.0 -- replaces -> 1.3.0 [deprecated]
+		
+	Deprecating a bundle that removes the default channel is not allowed.
+	`)
+
+func newIndexDeprecateCmd() *cobra.Command {
+	indexCmd := &cobra.Command{
+		Use:   "deprecate",
+		Short: "Deprecate operator bundles from an index.",
+		Long:  "",
```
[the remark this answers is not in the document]
gallettilance: oh good catch I missed that

cmd/opm/index/deprecate.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return err
+	}
+
+	logger := logrus.WithFields(logrus.Fields{"bundles": bundles})
+
+	logger.Info("deprecating bundles from the index")
+
+	indexDeprecator := indexer.NewIndexDeprecator(
+		containertools.NewContainerTool(buildTool, containertools.PodmanTool),
```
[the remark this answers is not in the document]
gallettilance: I'm confused - why?

pkg/lib/indexer/indexer.go
```diff
@@ -527,3 +528,61 @@ func generatePackageYaml(dbQuerier pregistry.Query, packageName, downloadPath st
...
+	}
+
+	// generate the dockerfile
+	dockerfile := i.DockerfileGenerator.GenerateIndexDockerfile(request.BinarySourceImage, databasePath)
+	err = write(dockerfile, outDockerfile, i.Logger)
+	if err != nil {
+		return err
+	}
+
+	if request.Generate {
```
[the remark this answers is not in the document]
gallettilance: we do above - this is to exit early (see index add)

pkg/lib/registry/registry.go
```diff
@@ -228,3 +228,39 @@ func (r RegistryUpdater) PruneFromRegistry(request PruneFromRegistryRequest) err
...
+		return err
+	}
+	defer db.Close()
+
+	dbLoader, err := sqlite.NewSQLLiteLoader(db)
+	if err != nil {
+		return err
+	}
+	if err := dbLoader.Migrate(context.TODO()); err != nil {
+		return fmt.Errorf("hello 3 %s", err)
```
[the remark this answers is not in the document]
gallettilance: oops

pkg/lib/registry/registry.go
```diff
@@ -228,3 +228,39 @@ func (r RegistryUpdater) PruneFromRegistry(request PruneFromRegistryRequest) err
...
+	deprecator := sqlite.NewSQLDeprecatorForBundles(dbLoader, request.Bundles)
+	if err := deprecator.Deprecate(); err != nil {
+		r.Logger.Debugf("unable to deprecate bundles from database: %s", err)
+		if !request.Permissive {
+			r.Logger.WithError(err).Error("permissive mode disabled")
+			return err
+		} else {
+			r.Logger.WithError(err).Warn("permissive mode enabled")
+		}
+		logrus.WithError(err).Warn("permissive mode enabled")
```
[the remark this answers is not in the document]
gallettilance: nice catch

pkg/sqlite/load.go
gallettilance: talked offline with @ecordell about this. If / when the csv name stops being unique we will make the change

pkg/sqlite/load.go
```diff
@@ -805,3 +805,176 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+	return nil
+}
+
+func getTailFromBundle(tx *sql.Tx, name string) (bundles []string, err error) {
+	getReplacesSkips := `SELECT replaces, skips FROM operatorbundle WHERE name=?`
+	isDefaultChannelHead := `SELECT head_operatorbundle_name FROM channel 
+							INNER JOIN package ON channel.name = package.default_channel 
+							WHERE channel.head_operatorbundle_name = ?`
+
+	nameQueue := make(chan string, 100)
```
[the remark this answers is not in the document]
gallettilance: +1

pkg/sqlite/load_test.go
gallettilance: Since csv-a replaces csv-b (head of stable channel), csv-a is marked as deprecated and csv-b + csv-c are removed (so the stable channel is removed).

pkg/sqlite/load.go
gallettilance: If tx.Commit is successful it would right? But isn't that what we want?

cmd/opm/index/deprecate.go
```diff
@@ -0,0 +1,132 @@
...
+		Given the update graph in quay.io/my/index:v1
+		1.4.0 -- replaces -> 1.3.0 -- replaces -> 1.2.0 -- replaces -> 1.1.0
+
+		Applying the command:
+		opm index deprecate --bundles "quay.io/my/bundle:1.3.0" --from-index "quay.io/my/index:v1" --tag "quay.io/my/index:v2"
+
+		Produces the following update graph in quay.io/my/index:v2
+		1.4.0 -- replaces -> 1.3.0 [deprecated]
+		
+	Deprecating a bundle that removes the default channel is not allowed.
```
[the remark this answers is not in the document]
gallettilance: if you want to change the default channel you should add a new version that specifies that change. It shouldn't be done during deprecation. We could have a separate command to change channel definitions around but that's not in scope

pkg/sqlite/load.go
awgreene: It is what we want - I just wanted to confirm this behavior.

cmd/opm/index/deprecate.go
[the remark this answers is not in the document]
awgreene: Sorry - I meant to ask why is the default container tool `PodmanTool`? I saw in other situations that the container tool was called with NoneTool (Example: `containertools.NewContainerTool(containerTool, containertools.NoneTool)`) and wanted to understand why.

cmd/opm/index/deprecate.go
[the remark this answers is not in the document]
awgreene: SGTM. Can we update the description to describe that to people?

pkg/lib/indexer/indexer.go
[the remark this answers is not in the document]
awgreene: Ahh - can we change [// build the dockerfile](https://github.com/operator-framework/operator-registry/pull/397/files#diff-66855dfefc79f095baf1c817a88f5fc9R581) to build the image?

pkg/sqlite/load_test.go
awgreene: Oh - I should have looked at the `newUnstructredCSV` method. I had assumed that c replaced b replaced a. Thank you.

cmd/opm/index/deprecate.go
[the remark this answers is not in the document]
awgreene: Apparently `containertools.NoneTool` specifies containrd

pkg/sqlite/load.go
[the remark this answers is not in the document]
gallettilance: +1 this shouldn't be 100. I set it back to 1 because this queue shouldn't contain more than one item at a time

cmd/opm/index/deprecate.go
```diff
@@ -0,0 +1,132 @@
...
+		Given the update graph in quay.io/my/index:v1
+		1.4.0 -- replaces -> 1.3.0 -- replaces -> 1.2.0 -- replaces -> 1.1.0
+
+		Applying the command:
+		opm index deprecate --bundles "quay.io/my/bundle:1.3.0" --from-index "quay.io/my/index:v1" --tag "quay.io/my/index:v2"
+
+		Produces the following update graph in quay.io/my/index:v2
+		1.4.0 -- replaces -> 1.3.0 [deprecated]
+		
+	Deprecating a bundle that removes the default channel is not allowed. Changing the default channel prior to deprecation is possible by publishing a new bundle to the index.
```
ecordell, line 28: > Changing the default channel prior to deprecation is possible by publishing a new bundle to the index.
it shouldn't block this PR, but it's been brought up at the past 2 working groups that we need a way to change this without adding a bundle.

pkg/registry/populator_test.go
```diff
@@ -670,3 +678,173 @@ func CheckBundlesHaveContentsIfNoPath(t *testing.T, db *sql.DB) {
...
+						"alpha",
+						"stable",
+					},
+					"prometheus": []string{
+						"preview",
+						"stable",
+					},
+				},
+			},
+		},
```
ecordell, line 788: this is looking pretty good, but doesn't quite convince me that there are no bugs - at the very least I think this should verify the full end state of the graph, not just the bundles/packages/channels.

I also want to make sure that this covers:

- bundle is in two channels
- bundle is in two packages
- bundle is skipped by another bundle
- bundle has children that are skipped that get removed

(it might already cover these implicitly, just need to check)

pkg/sqlite/load.go
```diff
@@ -805,3 +805,176 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+							WHERE channel.head_operatorbundle_name = ?`
+
+	nameQueue := make(chan string, 1)
+	tail := make(map[string]struct{})
+
+	nameQueue <- name
+
+	for {
+		select {
+		case next := <-nameQueue:
```
ecordell, line 866: this is not a big deal, but since we're not processing the queue concurrently this is setting up an unnecessary stack / associated locks when we could just use a slice.

pkg/sqlite/load_test.go
```diff
@@ -223,6 +223,19 @@ func newUnstructuredCSV(t *testing.T, name, replaces string) *unstructured.Unstr
 	return &unstructured.Unstructured{Object: out}
 }
 
+func newUnstructuredCSVwithSkips(t *testing.T, name, replaces, skips string) *unstructured.Unstructured {
```
ecordell, line 226: nonblocking/nit: might be useful to have `skips ...string`?

dinhxuanvu:
lgtm so far. There are still a new comments from Evan need to be address and then this PR should be good to go.

pkg/sqlite/load.go
gallettilance: actually since this is a channel with a buffer size of 1 I can just use a single variable 🤦

cdjohnson:
@gallettilance I realize this was merged quite some time ago, but why wasn't the `deprecatetruncate` command added to the `opm registry` commands?

PR #400, opened 2020-07-23
[release-4.5] Bug 1845588: fix(images): use docker/podman create and cp for exec unpacking

njhale:
Manual cherry-pick of interdependent fixes #351 and #390


shawn-hurley:
/hold

I think we need to determine if this is the fix we want to make

njhale:
> /hold
> 
> I think we need to determine if this is the fix we want to make

Could you elaborate a little on why we wouldn't want to backport this fix? alternatives?

(I figure there might be some discussion I missed or maybe some connection to our version skew discussions that I'm not making)

njhale:
/hold cancel

@shawn-hurley, following up on our Slack discussion, I've removed the unintentional inclusion of the `none` container tool option from this patch. Nice catch!

PTAL

eparis:
not sure what happened to tide over teh weekend, but this didn't merge, so pulling the label

PR #402, opened 2020-07-24
Bug 1866437: skip TLS option for pulling indexes

ankitathomas:
Extending skipTLS option to allow opm to fetch index images from insecure registries with containerd

When using an insecure registry in a dev environment, opm passes the skipTLS option only for pulling of bundles from the registry. Any processing of pre-existing index images requires using docker as a pull tool, since it allows pulling from insecure registries as a default. This PR passes the skipTLS option to pulling of the index images as well, removing the need for having docker as a requirement for working with insecure registries.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/index/add.go
+++ b/cmd/opm/index/add.go
@@ -53,7 +53,7 @@ func addIndexAddCmd(parent *cobra.Command) {
 	if err := indexCmd.MarkFlagRequired("bundles"); err != nil {
 		logrus.Panic("Failed to set required `bundles` flag for `index add`")
 	}
-	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling bundles")
+	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling bundles or index")
 	indexCmd.Flags().StringP("binary-image", "i", "", "container image for on-image `opm` command")
 	indexCmd.Flags().StringP("container-tool", "c", "", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
 	indexCmd.Flags().StringP("build-tool", "u", "", "tool to build container images. One of: [docker, podman]. Defaults to podman. Overrides part of container-tool.")
--- a/cmd/opm/index/add.go
+++ b/cmd/opm/index/add.go
@@ -108,9 +108,13 @@ func runIndexAddCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	skipTLS, err := cmd.Flags().GetBool("skip-tls")
-	if err != nil {
-		return err
+	var skipTLS *bool
+	if cmd.Flags().Changed("skip-tls") {
+		skipTLSVal, err := cmd.Flags().GetBool("skip-tls")
+		if err != nil {
+			return err
+		}
+		skipTLS = &skipTLSVal
 	}
 
 	mode, err := cmd.Flags().GetString("mode")
--- a/cmd/opm/index/add.go
+++ b/cmd/opm/index/add.go
@@ -53,7 +53,6 @@ func addIndexAddCmd(parent *cobra.Command) {
 	if err := indexCmd.MarkFlagRequired("bundles"); err != nil {
 		logrus.Panic("Failed to set required `bundles` flag for `index add`")
 	}
-	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling bundles or index")
 	indexCmd.Flags().StringP("binary-image", "i", "", "container image for on-image `opm` command")
 	indexCmd.Flags().StringP("container-tool", "c", "", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
 	indexCmd.Flags().StringP("build-tool", "u", "", "tool to build container images. One of: [docker, podman]. Defaults to podman. Overrides part of container-tool.")
@@ -108,13 +107,9 @@ func runIndexAddCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	var skipTLS *bool
-	if cmd.Flags().Changed("skip-tls") {
-		skipTLSVal, err := cmd.Flags().GetBool("skip-tls")
-		if err != nil {
-			return err
-		}
-		skipTLS = &skipTLSVal
+	skipTLS, err := cmd.Flags().GetBool("skip-tls")
+	if err != nil {
+		return err
 	}
 
 	mode, err := cmd.Flags().GetString("mode")
--- a/cmd/opm/index/delete.go
+++ b/cmd/opm/index/delete.go
@@ -41,6 +41,7 @@ func newIndexDeleteCmd() *cobra.Command {
 	indexCmd.Flags().StringP("pull-tool", "p", "", "tool to pull container images. One of: [none, docker, podman]. Defaults to none. Overrides part of container-tool.")
 	indexCmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
 	indexCmd.Flags().Bool("permissive", false, "allow registry load errors")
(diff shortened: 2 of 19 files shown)


njhale:
/hold

It looks like this fails to compile.

cmd/opm/index/add.go
```diff
@@ -108,9 +108,13 @@ func runIndexAddCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	skipTLS, err := cmd.Flags().GetBool("skip-tls")
-	if err != nil {
-		return err
+	var skipTLS *bool
+	if cmd.Flags().Changed("skip-tls") {
```
ecordell, line 112: Why is this necessary for this flag? Is this used to check when a child command has overridden the flag from a parent command?

ankitathomas: We have different default behavior with docker and podman for the tls verify. I'm not sure if we have any usage where we omit the flag because we expect that default behavior. skipping TLS by default isn't something we should encourage, this is mainly for development purposes. But enabling TLS by default means that any scripts that rely on docker enabling insecure pulls by default will fail because they lack the option.

So this was just a way to add a third "stick to the tool's default behavior" option

njhale: IMO, we should make `SkipTLSVerify=false` the default across the board. I'm more concerned about this being understandable than breaking existing UX.

exdx:
Is it worth adding a one line log output when running  opm commands with the `tls-skip` flag telling the user this is insecure? 

I suppose their intent is already known, based on them including the flag in their command, but maybe its worth reinforcing that this is insecure and for development purposes only?

pkg/containertools/runner.go
```diff
@@ -22,14 +22,75 @@ type CommandRunner interface {
...
+	default:
+	}
+	cmdArgs = append(cmdArgs, args...)
+	return cmdArgs
+}
+
+func defaultConfig(toolName string) *RunnerConfig {
+	switch toolName {
+	case "docker":
+		// docker disables tls verify by default, mimic that behavior
```
exdx, line 72: Maybe I'm misunderstanding, but docker disables tls verify when the client is communicating with the daemon, ie when communicating over http to the daemon. It enables tls verify when pushing/pulling images, which I believe is the context here. Pushing to an insecure registry will by default cause docker to throw an error, without the flag set

pkg/containertools/runner.go
```diff
@@ -22,14 +22,75 @@ type CommandRunner interface {
...
+		return &RunnerConfig{
+			SkipTLS: true,
+		}
+	case "podman":
```
exdx, line 76: As @njhale mentioned I'm not sure we want to have different default tls-skip behavior based on the container building tool the user chooses.

pkg/containertools/runner.go
njhale: So, does setting `SkipTLSVerify=false` at the top-level in `opm` prevent the docker variant from communicating with the docker daemon?

ankitathomas: I was mistaken about this part, docker isn't blocked by the presence/absence of this option, it needs the --tls flag for client side auth, which we don't need to do.

ankitathomas: > Pushing to an insecure registry will by default cause docker to throw an error, without the flag set

Docker doesn't have any issues with an insecure local registry. Any remote registries should have tls on them, so it's a non-issue

pkg/containertools/runner.go
ankitathomas: Modified this to a global default with tls-skip being false.

cmd/opm/index/add.go
ankitathomas: This was a misunderstanding on my part about docker's flags, it checks tls by default for any remote repository. The pointer flag isn't necessary.

cmd/opm/index/prunestranded.go
```diff
@@ -36,6 +36,7 @@ func newIndexPruneStrandedCmd() *cobra.Command {
 	indexCmd.Flags().StringP("binary-image", "i", "", "container image for on-image `opm` command")
 	indexCmd.Flags().StringP("container-tool", "c", "podman", "tool to interact with container images (save, build, etc.). One of: [docker, podman]")
 	indexCmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
+	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling index")
```
njhale, line 39: For future reference, if this option is shared by all subcommands, you can place it on the root command (see [persistent flags](https://github.com/spf13/cobra/blob/master/README.md#persistent-flags)).

pkg/lib/indexer/indexer.go
```diff
@@ -176,6 +179,8 @@ type PruneStrandedFromIndexRequest struct {
 	FromIndex         string
 	OutDockerfile     string
 	Tag               string
+	CaFile            string
+	SkipTLS           bool
```
njhale, line 183: This has got me thinking: do we want to set a custom CA and TLS settings per-request, or for all requests an indexer makes? I don't think we need to change this here, but worth remember for future refactoring.

pkg/image/containerdregistry/certs.go
```diff
@@ -0,0 +1,24 @@
+package containerdregistry
+
+import (
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+)
+
+func RootCAs(CaFile string) (*x509.CertPool, error) {
```
njhale, line 9: Nice! but does this need to be tightly coupled to `containerd`?

pkg/containertools/runner.go
```diff
@@ -22,14 +22,53 @@ type CommandRunner interface {
...
+	}
+}
+
+func (r *RunnerConfig) apply(options []RunnerOption) {
+	for _, option := range options {
+		option(r)
+	}
+}
+
+func (r *ContainerCommandRunner) argsForCmd(cmd string, args ...string) []string {
```
njhale, line 46: 👍

pkg/containertools/runner.go
```diff
@@ -22,14 +22,53 @@ type CommandRunner interface {
...
+	cmdArgs := []string{cmd}
+	switch r.containerTool {
+	case PodmanTool:
```
njhale, line 49: Don't we also need to do this for docker?

pkg/containertools/runner.go
njhale: an alternative approach would be to have a `CommandBuilder` abstraction to incrementally build out a command and execute it.

pkg/image/containerdregistry/certs.go
ankitathomas: Not necessarily, it's something we can move to `pkg/lib/`

pkg/containertools/runner.go
ankitathomas: IThe docker daemon itself maintains a list of insecure remote registries it can talk to, so we'd have to re-run the daemon with a new list of registries if we wanted to support those. Insecure registries on localhost are fine though.

cmd/opm/index/delete.go
```diff
@@ -41,6 +41,7 @@ func newIndexDeleteCmd() *cobra.Command {
 	indexCmd.Flags().StringP("pull-tool", "p", "", "tool to pull container images. One of: [none, docker, podman]. Defaults to none. Overrides part of container-tool.")
 	indexCmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
 	indexCmd.Flags().Bool("permissive", false, "allow registry load errors")
+	indexCmd.Flags().Bool("skip-tls", false, "skip TLS certificate verification for container image registries while pulling index")
```
kevinrizza, line 44: Is this necessary when the parent command passes the flag along?

ankitathomas: I've removed the extra flag

PR #404, opened 2020-07-25
feat(registry): add graceful shutdown to grpc server

ecordell:
When the server receives `SIGTERM`, this calls `GracefulStop` on the grpc server.

This is better for clients, and may improve e2e test time for OLM.

**Reviewer Checklist**


exdx:
Should we maybe listen for `SIGINT` as well? If I recall correctly kubelet sends SIGINT first and then waits 30 seconds before calling `SIGTERM`

exdx:
Nope, nevermind -- mixed up `SIGTERM` with `SIGKILL` 
/lgtm

PR #396, opened 2020-07-22
feat(bundle): Add dependency constraints support

dinhxuanvu:
Add dependency contraints that support discrete GVK and package
dependency constraints and the new compound dependency or set of
constraints (gvk or package).

Signed-off-by: Vu Dinh <vdinh@redhat.com>

**Description of the change:**

**Motivation for the change:**

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #405, opened 2020-07-26
Fix listbundles property bug

ecordell:
This fixes a bug where only one property would be returned for a bundle.

**Reviewer Checklist**


kevinrizza:
+1 on this, I actually fixed this same issue in my commit here https://github.com/operator-framework/operator-registry/pull/396/files#diff-d08cc6a068ece62fe78d1e82a80d3c52R1018

/lgtm

PR #406, opened 2020-07-28
feat(registry): Add label dependency constraint and property

dinhxuanvu:
1. Introduce a new type of dependency constraint named `olm.label` which
represents the constraints that are based on CSV labels.
2. The new label property is parsed from CSV labels.

**Reviewer Checklist**


manifests/etcd/0.9.2/etcdoperator.v0.9.2.clusterserviceversion.yaml
```diff
@@ -4,6 +4,9 @@ kind: ClusterServiceVersion
 metadata:
   name: etcdoperator.v0.9.2
   namespace: placeholder
+  labels:
+    olm.label: testlabel
+    olm.label.1: testlabel1
```
kevinrizza, line 9: Should we be putting invalid values into the canonical examples in our `/manifests` folder?

pkg/lib/bundle/validate_test.go
```diff
@@ -61,6 +61,14 @@ func TestValidateBundleDependencies(t *testing.T) {
...
 			},
 		},
+		{
+			description: "registryv1 bundle/invalid label type",
+			mediaType:   RegistryV1Type,
+			directory:   "./testdata/validate/invalid_dependencies_bundle/invalid_label_dependency/",
+			errs: []error{
+				fmt.Errorf("couldn't parse dependency of type olm.label"),
+			},
+		},
```
kevinrizza, line 71: Should we also include a test for a valid and invalid label property?

dinhxuanvu: This validation is for dependency given dependency information is provided via a separate file. All of the properties are extracted from CSV hence I added the labels into the example CSV in `/manifests`.

manifests/etcd/0.9.2/etcdoperator.v0.9.2.clusterserviceversion.yaml
dinhxuanvu: All of the test cases that are for property extraction using manifests from this folder so that's why I added it here. This is just another folder for test manifests. I don't see an issue with changing. Plus, additional properties to this operator is harmless given it only matters if someone requires this property.

kevinrizza: It feels to me like the `/manifests` folder should be a good example of what manifests should be, and if we want something with invalid data, it should be saved somewhere else.

docs/design/operator-bundle.md
```diff
@@ -69,21 +69,24 @@ annotations:
...
       packageName: prometheus
       version: >0.27.0
   - type: olm.gvk
     value:
       group: etcd.database.coreos.com
       kind: EtcdCluster
       version: v1beta2
+  - type: olm.gvk
+    value:
+      label: lts
```
benluddy, line 89: Whoops?

pkg/lib/bundle/testdata/validate/invalid_dependencies_bundle/invalid_label_dependency/manifests/etcdbackups.etcd.database.coreos.com.crd.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
benluddy, line 1: I'm not singling out this PR in particular, but I am concerned about the overhead of managing all of the test fixtures and golden manifests that operator-registry has been accumulating. On master, there are well over _three hundred_ manifests like this.

I'd like to have an async discussion about this, separate from this PR.

pkg/registry/types.go
```diff
@@ -219,6 +230,15 @@ func (gd *GVKDependency) Validate() []error {
 	return errs
 }
 
+// Validate will validate GVK dependency type and return error(s)
+func (ld *LabelDependency) Validate() []error {
+	errs := []error{}
+	if ld.Label == "" {
```
benluddy, line 236: Why is this an error? IMO it could be invalid to omit the label field entirely from the dependency specification, but it should be OK to set it to any string, even the empty string.

pkg/sqlite/load.go
```diff
@@ -803,5 +803,23 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
 		}
 	}
 
+	// Add label properties
+	if csv, err := bundle.ClusterServiceVersion(); err == nil {
+		for k, v := range csv.GetLabels() {
```
benluddy, line 808: I don't think Kube labels are the place to represent operator labels. That makes operator labels subject to the same name restrictions as Kube labels. We also don't expect operator labels to necessarily be tied to CSV selection/aggregation on-cluster.

docs/design/operator-bundle.md
dinhxuanvu: Whoops indeed ;)

pkg/sqlite/load.go
dinhxuanvu: If it is not in the labels then where do you think it should be? Annotations? I use Labels because the name of the type is label. That only makes sense to have it in labels if we can it `label` type. Plus, any CSV selection/aggregation will need to select labels based on keys. As long as we preserve the `olm.labels` key to use for this purpose then we should be fine.

pkg/registry/types.go
dinhxuanvu: I'm fine with the suggestion. Will address this.

manifests/etcd/0.9.2/etcdoperator.v0.9.2.clusterserviceversion.yaml
dinhxuanvu: The labels above are totally valid.

pkg/lib/bundle/testdata/validate/invalid_dependencies_bundle/invalid_label_dependency/manifests/etcdbackups.etcd.database.coreos.com.crd.yaml
dinhxuanvu: The test manifests have multiple CRDs (usually 3) which are not needed. So we can build a fake operator with one CRD and use it for the test fixtures to reduce the number of files.

manifests/etcd/0.9.2/etcdoperator.v0.9.2.clusterserviceversion.yaml
dinhxuanvu: Also, just a note. The manifests inside `/manifests` are hardly good examples given one of them has multiple CSV under same directory and one operator uses flatten format that we don't use anymore. The manifests in `/bundles` are actually better as examples.

pkg/registry/types.go
dinhxuanvu: Well, given this `label` field is a string. A empty string == there is no `label` field at all.

pkg/sqlite/load.go
dinhxuanvu: @ecordell FYI

pkg/registry/types.go
benluddy: If the struct field is the empty string, then either the label field was omitted or the label field was included and explicitly set to the empty string. It's impossible to distinguish between both cases here.

dinhxuanvu: That's what I meant. So either way, omitting the field or set the field to empty string is invalid from the validation standpoint. Plus, why should we allow an empty string as a label?

pkg/lib/bundle/testdata/validate/invalid_dependencies_bundle/invalid_label_dependency/manifests/etcdbackups.etcd.database.coreos.com.crd.yaml
benluddy: I would like to see fewer tests that require full test fixtures rather than requiring fewer test fixtures per test. Let's discuss elsewhere so that this one topic doesn't distract from the focus of this change.

dinhxuanvu: Agree.

pkg/registry/types.go
benluddy: > why should we allow an empty string as a label
My point is that disallowing the empty string is an arbitrary rule. Why is the empty string different from any other string? I always prefer to have fewer special cases as a rule.

I won't die on this hill, and I don't expect it to ruin anyone's day if they can't use the empty string, but this is an undocumented behavior without a test case.

pkg/lib/bundle/testdata/validate/valid_bundle/metadata/dependencies.yaml
```diff
@@ -7,3 +7,5 @@ dependencies:
       group: test.coreos.com
       kind: testapi
       version: v1
+  - type: olm.label
```
benluddy, line 10: Is the olm.package dependency in this fixture wrong? That's nice ammunition for my complaining about unmaintainable test fixtures.

pkg/registry/types.go
dinhxuanvu: The zero value of a string is an empty string. That is a difference between empty string vs other strings in my perspective. You assign a string an empty string is equivalent to not assigning anything to that string at all. As you said before, there is impossible to distinguish between both cases. I think it is better to error out when they forget to assign a value to string or the value that is empty because why would anyone wants to do so in reality. We can revisit this issue if there is a demand to have an empty string as a value for a label. Plus, we will find out if someone asks for it when bundle validation will surface this up in the pipeline anyway.

pkg/lib/bundle/testdata/validate/valid_bundle/metadata/dependencies.yaml
dinhxuanvu: The ones that are wrong here are the gvk and label ones. They need to under `value:` like package. This is my fault. Also this bundle is used for manifests validation AKA CRD and CSV, not metadata so I remove this file altogether to avoid confusion.

dinhxuanvu:
I wonder if this e2e-aws failure is valid or not. The error message doesn't seem to be OLM-related.

pkg/registry/types.go
benluddy: I don't think I've explained my problem with this clearly enough, but it's not very significant, so I'll resolve this thread.

pkg/sqlite/load.go
```diff
@@ -803,6 +804,34 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
 		}
 	}
 
+	// Add label properties
+	if csv, err := bundle.ClusterServiceVersion(); err == nil {
+		annotations := csv.ObjectMeta.GetAnnotations()
+		if v, ok := annotations[registry.PropertyKey]; ok {
+			var props []registry.Property
```
anik120, line 811:
```suggestion
			var properties []registry.Property
```

pkg/sqlite/load.go
```diff
@@ -803,6 +804,34 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+			if err := json.Unmarshal([]byte(v), &props); err == nil {
+				for _, prop := range props {
```
anik120, line 813:
```suggestion
				for _, property := range properties {
```

docs/design/operator-bundle.md
```diff
@@ -69,21 +69,24 @@ annotations:
...
 
 An example of a `dependencies.yaml` that specifies Prometheus operator and etcd CRD dependencies:
 
 ```
 dependencies:
   - type: olm.package
-    value:
```
anik120, line 79: `nit`: Noise.

anik120:
Looks like it's been lgtm-ed, if approved please resolve my comments they're very minor.

PR #408, opened 2020-07-29
fix(validation): Remove default channel validation from bundle lib

dinhxuanvu:
It is not possible to determine if the default channel value is valid
or not given we can "infer" default channel by using an emptry string.
Also, author can add a default channel that is not listed in the channel
list in some cases. As a result, the default channel is subjective
and there is no methodology to validate it. Therefore, removing this
validation will avoid any false positve.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/generate.go
+++ b/pkg/lib/bundle/generate.go
@@ -293,32 +293,6 @@ func ValidateAnnotations(existing, expected []byte) error {
 	return utilerrors.NewAggregate(errs)
 }
 
-// ValidateChannelDefault validates provided default channel to ensure it exists in
-// provided channel list.
-func ValidateChannelDefault(channels, channelDefault string) (string, error) {
-	var chanDefault string
-	var chanErr error
-	channelList := strings.Split(channels, ",")
-
-	if containsString(channelList, "") {
-		return chanDefault, fmt.Errorf("invalid channels are provided: %s", channels)
-	}
-
-	if channelDefault != "" {
-		for _, channel := range channelList {
-			if channel == channelDefault {
-				chanDefault = channelDefault
-				break
-			}
-		}
-		if chanDefault == "" {
-			chanDefault = channelList[0]
-			chanErr = fmt.Errorf(`The channel list "%s" doesn't contain channelDefault "%s"`, channels, channelDefault)
-		}
-	}
-	return chanDefault, chanErr
-}
-
 // GenerateAnnotations builds annotations.yaml with mediatype, manifests &
 // metadata directories in bundle image, package name, channels and default
 // channels information.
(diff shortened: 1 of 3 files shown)


pkg/lib/bundle/generate.go
```diff
@@ -379,7 +343,7 @@ func GenerateDockerfile(mediaType, manifests, metadata, copyManifestDir, copyMet
 	fileContent += fmt.Sprintf("LABEL %s=%s\n", MetadataLabel, metadata)
 	fileContent += fmt.Sprintf("LABEL %s=%s\n", PackageLabel, packageName)
 	fileContent += fmt.Sprintf("LABEL %s=%s\n", ChannelsLabel, channels)
-	fileContent += fmt.Sprintf("LABEL %s=%s\n\n", ChannelDefaultLabel, chanDefault)
```
estroz, line 382: If `channelDefault` is an empty string, this label will be empty. Is that ok/desirable?

pkg/lib/bundle/generate.go
```diff
@@ -355,11 +324,6 @@ func GenerateAnnotations(mediaType, manifests, metadata, packageName, channels,
 func GenerateDockerfile(mediaType, manifests, metadata, copyManifestDir, copyMetadataDir, workingDir, packageName, channels, channelDefault string) ([]byte, error) {
 	var fileContent string
 
-	chanDefault, err := ValidateChannelDefault(channels, channelDefault)
```
estroz, line 358: If `channelDefault` is not an empty string, we still want to verify that it exists in `channels`.

pkg/lib/bundle/generate.go
dinhxuanvu: I think it is OK to not have a defaultChannel value AKA empty string because in annotation generation function, there is a case where inferred defaultChannel is set to empty string. @Bowenislandsong Can you double-confirm this?

kevinrizza: Wouldn't it be better to just not set this label at all?

Bowenislandsong: yes. Later when added to Index, the default behavior will kick in (right now the default behavior is to add the first channel as default).

pkg/lib/bundle/generate.go
dinhxuanvu: There are cases when that is totally fine to have default channel not on the list. @shawn-hurley Can you explain this further?

pkg/lib/bundle/generate.go
estroz: In that case, wouldn't it be better/more obvious to the user to write a label containing the first channel as default if a default isn't explicitly set?

pkg/lib/bundle/generate.go
kevinrizza: Default channel mentioned in the bundle != Default channel on the package

PR #413, opened 2020-08-06
Bug 1866930: Clean up stranded bundles

kevinrizza:
Currently, it is possible in opm to strand one or more bundles if a new
bundle is added to a channel without replaces set. When this happens,
`registry/index rm` can have unexpected behavior, because of the loss of
the channel entry table rows linking the bundle to a package.

To resolve this problem for, this commit introduces a new command
`prune-stranded` which explicity removes all stranded bundles from
the database. In addition, it updates `rm` to remove all stranded
bundles whenever a package is being removed.


cmd/opm/index/prunestranded.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/lib/indexer"
+)
+
+func newIndexPruneStrandedCmd() *cobra.Command {
+	indexCmd := &cobra.Command{
+		Use:   "prune-stranded",
```
dinhxuanvu, line 15: Would `prune` be sufficient? Or we consider having different types of pruning?

pkg/sqlite/stranded_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"testing"
+
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/registry"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStrandedBundleRemover(t *testing.T) {
```
dinhxuanvu, line 15: I assume because of FOREIGN KEY with ON DELETE CASCADE, all dependencies + properties associated with the stranded operators should be removed. However, it would be nice if this test case can verify that as well.

cmd/opm/index/prunestranded.go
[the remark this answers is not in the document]
kevinrizza: Unfortunately there is already a `prune` command that does something else

cmd/opm/registry/prunestranded.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/operator-framework/operator-registry/pkg/lib/registry"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newRegistryPruneStrandedCmd() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "prune-stranded",
+		Short: "prune an operator registry DB of stranded bundles",
```
ecordell, line 13: nit: I believe docker uses the word `dangling` for this, which might help users understand the issue. But doesn't need to change.

pkg/sqlite/load.go
```diff
@@ -1001,3 +1001,57 @@ func (s *sqlLoader) DeprecateBundle(path string) error {
...
+		}
+		if name.Valid {
+			strandedBundles = append(strandedBundles, fmt.Sprintf(`"%s"`, name.String))
+		}
+	}
+
+	if len(strandedBundles) == 0 {
+		return nil, nil
+	}
+
```
ecordell, line 1045: doesn't `rows` need to be closed before this transaction (right now it's deferred to the end)?

pkg/sqlite/stranded_test.go
```diff
@@ -0,0 +1,76 @@
...
+	require.NoError(t, err)
+	require.Equal(t, 1, len(packageBundles))
+
+	rows, err := db.QueryContext(context.TODO(), "select * from operatorbundle")
+	require.NoError(t, err)
+	require.Equal(t, 2, rowCount(rows))
+	require.NoError(t, rows.Close())
+
+	// prune the orphaned bundle
+	removedBundles, err := store.RemoveStrandedBundles()
```
ecordell, line 53: I think this should test the case where multiple bundles are stranded since that's a big part of the pruning

pkg/sqlite/load.go
kevinrizza: ah, yep, making a quick change

kevinrizza:
@dinhxuanvu @ecordell ptal

PR #416, opened 2020-08-11
Add test to verify empty bundles in addition does not break

shawn-hurley:
Test to verify the behavior of adding an empty list of bundles

Prevent backwards incompatible changes.

**Reviewer Checklist**


njhale:
/lgtm

tide should just be waiting for a merge window now

shawn-hurley:
@ecordell can you override with valid bug? I believe the change is just a test change and is needed to make sure that we don't break pipeline functionality

ecordell:
this a safe, test-only change, marking with `valid-bug`

PR #417, opened 2020-08-11
Bug 1867848: Return empty properties and dependencies in ListBundles responses.

benluddy:
Bundles that don't have any properties/dependencies should return nil
or [] for those fields, but due to an error in handling NULL column
values, a single zero-valued entry was being returned.


pkg/sqlite/query.go
```diff
@@ -1028,29 +1047,29 @@ func (s *SQLQuerier) ListBundles(ctx context.Context) (bundles []*api.Bundle, er
...
 			list = append(list, entry)
 		}
 	}
 	return list
 }
 
 func uniqueProps(props []*api.Property) []*api.Property {
-	keys := make(map[string]bool)
-	list := []*api.Property{}
+	keys := make(map[string]struct{})
```
kevinrizza, line 1067: 👍

pkg/sqlite/query_test.go
```diff
@@ -0,0 +1,289 @@
...
+	"testing"
+
+	"github.com/operator-framework/operator-registry/pkg/api"
+	"github.com/operator-framework/operator-registry/pkg/sqlite"
+	"github.com/operator-framework/operator-registry/pkg/sqlite/sqlitefakes"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestListBundles(t *testing.T) {
+	const query = `SELECT DISTINCT channel_entry.entry_id, operatorbundle.bundle, operatorbundle.bundlepath,
```
kevinrizza, line 17: Am I misreading something or is this const just not used? Is it injected into something somehow?

benluddy: Nope, good catch. The ListBundles impl calls out to other methods on SQLQuerier, which in turn make other database queries, so I was originally going to set the stub behavior based on the query arg. I ended up stubbing based on invocation count instead and didn't use it.

pkg/sqlite/query.go
```diff
@@ -1,3 +1,5 @@
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -o sqlitefakes/fake_rowscanner.go . RowScanner
```
njhale, line 1: 👍

pkg/sqlite/query.go
```diff
@@ -918,23 +942,25 @@ func (s *SQLQuerier) ListBundles(ctx context.Context) (bundles []*api.Bundle, er
 	}
 	defer rows.Close()
 
-	bundles = []*api.Bundle{}
+	var bundles []*api.Bundle
 	bundlesMap := map[string]*api.Bundle{}
 	for rows.Next() {
-		var entryID sql.NullInt64
```
njhale, line 924: 👍

pkg/sqlite/query_test.go
```diff
@@ -0,0 +1,278 @@
...
+	"context"
+	"database/sql"
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/operator-framework/operator-registry/pkg/api"
+	"github.com/operator-framework/operator-registry/pkg/sqlite"
+	"github.com/operator-framework/operator-registry/pkg/sqlite/sqlitefakes"
+	"github.com/stretchr/testify/assert"
```
njhale, line 13: nit: import order

pkg/sqlite/query.go
```diff
@@ -1028,29 +1047,29 @@ func (s *SQLQuerier) ListBundles(ctx context.Context) (bundles []*api.Bundle, er
...
 	}
 
-	return
+	return bundles, nil
 }
 
 func unique(deps []*api.Dependency) []*api.Dependency {
-	keys := make(map[string]bool)
-	list := []*api.Dependency{}
+	keys := make(map[string]struct{})
```
njhale, line 1054: 👍

PR #429, opened 2020-08-31
Bug 1880127: Parallelize opm index export

harishsurf:
This PR modifies `opm index export` to ensure exporting is done in parallel with an intent to cut down the overall time taken. It is achieved by the following two changes:
1. Make `--package` flag of the `opm index export` optional and allow comma-separated values
2. Parallelize the `ExportFromIndex` API

With this change, the export is possible by just specifying the index-image

Unpacking and exporting of community operators takes >2hrs as it is done synchronously

Signed-off-by: Harish <hgovinda@redhat.com>

### DIFF
--- a/pkg/lib/indexer/indexer.go
+++ b/pkg/lib/indexer/indexer.go
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
@@ -26,12 +29,13 @@ import (
 )
 
 const (
-	defaultDockerfileName = "index.Dockerfile"
-	defaultImageTag       = "operator-registry-index:latest"
-	defaultDatabaseFolder = "database"
-	defaultDatabaseFile   = "index.db"
-	tmpDirPrefix          = "index_tmp_"
-	tmpBuildDirPrefix     = "index_build_tmp"
+	defaultDockerfileName     = "index.Dockerfile"
+	defaultImageTag           = "operator-registry-index:latest"
+	defaultDatabaseFolder     = "database"
+	defaultDatabaseFile       = "index.db"
+	tmpDirPrefix              = "index_tmp_"
+	tmpBuildDirPrefix         = "index_build_tmp"
+	concurrencyLimitForExport = 10
 )
 
 // ImageIndexer is a struct implementation of the Indexer interface
@@ -467,7 +471,7 @@ func write(dockerfileText, outDockerfile string, logger *logrus.Entry) error {
 // ExportFromIndexRequest defines the parameters to send to the ExportFromIndex API
 type ExportFromIndexRequest struct {
 	Index         string
-	Package       string
+	Packages      []string
 	DownloadPath  string
 	ContainerTool containertools.ContainerTool
 	CaFile        string
@@ -497,14 +501,20 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
 	defer db.Close()
 
 	dbQuerier := sqlite.NewSQLLiteQuerierFromDb(db)
+
+	// fetch all packages from the index image if packages is empty
+	if len(request.Packages) == 0 {
+		request.Packages, err = dbQuerier.ListPackages(context.TODO())
+		if err != nil {
+			return err
+		}
+	}
+
+	bundles, err := getBundlesToExport(dbQuerier, request.Packages)
 	if err != nil {
 		return err
 	}
 
-	bundles, err := getBundlesToExport(dbQuerier, request.Package)
-	if err != nil {
-		return err
-	}
 	i.Logger.Infof("Preparing to pull bundles %+q", bundles)
 
 	// Creating downloadPath dir
@@ -513,40 +523,68 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
 	}
 
 	var errs []error
-	for _, bundleImage := range bundles {
-		// try to name the folder
-		folderName, err := dbQuerier.GetBundleVersion(context.TODO(), bundleImage)
-		if err != nil {
-			return err
-		}
-		if folderName == "" {
-			// operator-registry does not care about the folder name
-			folderName = bundleImage
-		}
-		exporter := bundle.NewExporterForBundle(bundleImage, filepath.Join(request.DownloadPath, folderName), request.ContainerTool)
-		if err := exporter.Export(); err != nil {
-			err = fmt.Errorf("error exporting bundle from image: %s", err)
-			errs = append(errs, err)
-		}
+	var wg sync.WaitGroup
+	wg.Add(len(bundles))
+	var mu = &sync.Mutex{}
+
+	sem := make(chan struct{}, concurrencyLimitForExport)
+
+	for bundleImage, bundleDir := range bundles {
+		go func(bundleImage string, bundleDir bundleDirPrefix) {
+			defer wg.Done()
+
+			sem <- struct{}{}
+			defer func() {
+				<-sem
+			}()
+
+			// generate a random folder name if bundle version is empty
+			if bundleDir.bundleVersion == "" {
+				bundleDir.bundleVersion = strconv.Itoa(rand.Intn(10000))
+			}
+			exporter := bundle.NewExporterForBundle(bundleImage, filepath.Join(request.DownloadPath, bundleDir.pkgName, bundleDir.bundleVersion), request.ContainerTool)
+			if err := exporter.Export(); err != nil {
+				err = fmt.Errorf("exporting bundle image:%s failed with %s", bundleImage, err)
+				mu.Lock()
+				errs = append(errs, err)
+				mu.Unlock()
+			}
+		}(bundleImage, bundleDir)
 	}
-	if err != nil {
-		errs = append(errs, err)
+	// Wait for all the go routines to finish export
+	wg.Wait()
+
+	if errs != nil {
 		return utilerrors.NewAggregate(errs)
 	}
 
-	err = generatePackageYaml(dbQuerier, request.Package, request.DownloadPath)
-	if err != nil {
-		errs = append(errs, err)
+	for _, packageName := range request.Packages {
+		err := generatePackageYaml(dbQuerier, packageName, filepath.Join(request.DownloadPath, packageName))
+		if err != nil {
+			errs = append(errs, err)
+		}
 	}
 	return utilerrors.NewAggregate(errs)
 }
 
-func getBundlesToExport(dbQuerier pregistry.Query, packageName string) ([]string, error) {
-	bundles, err := dbQuerier.GetBundlePathsForPackage(context.TODO(), packageName)
-	if err != nil {
-		return nil, err
+type bundleDirPrefix struct {
+	pkgName, bundleVersion string
+}
+
+func getBundlesToExport(dbQuerier pregistry.Query, packages []string) (map[string]bundleDirPrefix, error) {
+	bundleMap := make(map[string]bundleDirPrefix)
+
+	for _, packageName := range packages {
+		bundlesForPackage, err := dbQuerier.GetBundlesForPackage(context.TODO(), packageName)
+		if err != nil {
+			return nil, err
+		}
+		for k, _ := range bundlesForPackage {
+			bundleMap[k.BundlePath] = bundleDirPrefix{pkgName: packageName, bundleVersion: k.Version}
+		}
 	}
-	return bundles, nil
+
+	return bundleMap, nil
 }
 
 func generatePackageYaml(dbQuerier pregistry.Query, packageName, downloadPath string) error {
--- a/pkg/lib/indexer/indexer_test.go
+++ b/pkg/lib/indexer/indexer_test.go
@@ -30,10 +30,16 @@ func TestGetBundlesToExport(t *testing.T) {
 		t.Fatalf("creating querier: %s", err)
 	}
 
-	bundleImages, err := getBundlesToExport(dbQuerier, "etcd")
+	bundleMap, err := getBundlesToExport(dbQuerier, []string {"etcd"})
 	if err != nil {
 		t.Fatalf("exporting bundles from db: %s", err)
 	}
+
+	var bundleImages []string
+	for bundlePath, _ := range bundleMap {
+		bundleImages = append(bundleImages, bundlePath)
+	}
+
 	sort.Strings(bundleImages)
 
 	if !reflect.DeepEqual(expected, bundleImages) {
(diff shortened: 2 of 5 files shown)


pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +511,68 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
-		}
-		exporter := bundle.NewExporterForBundle(bundleImage, filepath.Join(request.DownloadPath, folderName), request.ContainerTool)
-		if err := exporter.Export(); err != nil {
-			err = fmt.Errorf("error exporting bundle from image: %s", err)
-			errs = append(errs, err)
-		}
-	}
-	if err != nil {
-		errs = append(errs, err)
+		go func(bundleImage string, errs []error) {
```
benluddy, line 522: Should there be a concurrency limit?

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +511,68 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+	}
+
+	// Wait for all the go routines to finish export
+	wg.Wait()
+
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	if errs != nil {
```
ecordell, line 555: no need for a nil check, `NewAggregate` will return nil if passed nil

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +511,68 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+				return
+			}
+			if folderName == "" {
+				// operator-registry does not care about the folder name
+				folderName = bundleImage
+			}
+			exporter := bundle.NewExporterForBundle(bundleImage, filepath.Join(request.DownloadPath, folderName), request.ContainerTool)
+			if err := exporter.Export(); err != nil {
+				err = fmt.Errorf("error exporting bundle from image: %s", err)
+				errs = append(errs, err)
```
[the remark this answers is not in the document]
harishsurf: Removed `errs` being passed as an argument inside of go routine

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +511,68 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+			defer wg.Done()
+			// try to name the folder
+			folderName, err := dbQuerier.GetBundleVersion(context.TODO(), bundleImage)
+			select {
+			case <-ctx.Done():
+				return // Error in one of the go routine, terminate
+			default: // Default is must to avoid blocking
+			}
+			if err != nil {
+				fmt.Errorf("error getting bundle version from image: %s", err)
```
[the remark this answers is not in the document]
harishsurf: No. I changed it to log using `logrus`

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
 	}
 
 	var errs []error
+	var g errgroup.Group
```
benluddy, line 526: Can `errs` and `mu` be dropped altogether? The original error handling behavior is suspect (especially considering original L532, which I think is actually looking at the `err` that was assigned on original L504?), and I would be easier to understand if this method were to only return the first error it encounters (which is consistent with errgroup's behavior).

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+	var mu = &sync.Mutex{}
+
+	concurrentGoroutines := make(chan struct{}, concurrencyLimitForExport)
```
benluddy, line 529: Nit: `sem`? I think you can expect future readers to be familiar with semaphores, and it's a reasonably common Go idiom to use buffered channels this way.

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+				return err
+			}
+			if bundleVersion == "" {
+				// operator-registry does not care about the folder name
+				bundleVersion = bundleImage
+			}
+			// fetch bundle name from bundle image
+			pkgName := strings.FieldsFunc(bundleImage, splitBundlePath)[2]
+			exporter := bundle.NewExporterForBundle(bundleImage, filepath.Join(request.DownloadPath, pkgName, bundleVersion), request.ContainerTool)
+			if err := exporter.Export(); err != nil {
```
benluddy, line 551: I noticed that you removed the cancellation behavior from an earlier revision. You might consider using https://pkg.go.dev/golang.org/x/sync/errgroup?tab=doc#WithContext, and propagating it to `Export()` (and further to https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/bundle/exporter.go#L62 and friends) so that things like image pulls can be interrupted.

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+				err = fmt.Errorf("error exporting bundle from image: %s", err)
```
benluddy, line 552: It may be helpful to include `bundleImage` in this error.

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
-			errs = append(errs, err)
-		}
+		})
 	}
-	if err != nil {
-		errs = append(errs, err)
+
+	// Wait for all go routines to finish export
+	if err := g.Wait(); err != nil {
+		i.Logger.Errorf("Error in one of the go routines while unpacking the bundle: %s", err)
```
benluddy, line 563: Nit: I'm not convinced that logging here is helpful for users.

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+			if err != nil {
+				i.Logger.Errorf("error getting bundle version from image: %s", err)
+				return err
+			}
+			if bundleVersion == "" {
+				// operator-registry does not care about the folder name
+				bundleVersion = bundleImage
+			}
+			// fetch bundle name from bundle image
+			pkgName := strings.FieldsFunc(bundleImage, splitBundlePath)[2]
```
benluddy, line 549: String manipulation like this has a tendency to be brittle. Could you use something from https://godoc.org/github.com/docker/distribution/reference?

pkg/lib/indexer/indexer_test.go
```diff
@@ -30,7 +30,8 @@ func TestGetBundlesToExport(t *testing.T) {
 		t.Fatalf("creating querier: %s", err)
 	}
 
-	bundleImages, err := getBundlesToExport(dbQuerier, "etcd")
+	packages := []string {"etcd"}
+	bundleImages, err := getBundlesToExport(dbQuerier, packages)
```
benluddy, line 34:
```suggestion
	bundleImages, err := getBundlesToExport(dbQuerier, []string {"etcd"})
```

pkg/lib/indexer/indexer.go
kevinrizza: +1, it should probably not refer to the implementation

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+	var g errgroup.Group
+	var mu = &sync.Mutex{}
+
+	concurrentGoroutines := make(chan struct{}, concurrencyLimitForExport)
+
 	for _, bundleImage := range bundles {
-		// try to name the folder
-		folderName, err := dbQuerier.GetBundleVersion(context.TODO(), bundleImage)
-		if err != nil {
+		bundleImage := bundleImage // to ensure every go routine gets a it's own copy of closure variable
```
kevinrizza, line 532: comment nit: `to ensure every go routine gets its own copy of closure variable`

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+
+		g.Go(func() error {
+			concurrentGoroutines <- struct{}{}
+			defer func() { <-concurrentGoroutines }()
```
kevinrizza, line 536: nit: can you explode this out?

```go
defer func() {
    <-concurrrentGoroutines
}()
```

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+
+			// get the bundle version to be used in naming the folder
+			bundleVersion, err := dbQuerier.GetBundleVersion(context.TODO(), bundleImage)
```
kevinrizza, line 539: Do we need to make a separate transaction to the database for each package? Can't we just get this all at once when we do the initial package query above?

pkg/lib/indexer/indexer.go
```diff
@@ -513,38 +523,72 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+			if err != nil {
+				i.Logger.Errorf("error getting bundle version from image: %s", err)
+				return err
+			}
+			if bundleVersion == "" {
+				// operator-registry does not care about the folder name
```
kevinrizza, line 545: This comment feels odd, why not just describe what we're doing rather than why it is happening? It feels like a comment that would quickly go out of date if another module changed behavior. It also probably means that `bundleVersion` is the wrong variable to set here (either `bundleVersion` should actually be `dirPrefix` or something like that, or we should set the value of the variable further down)

pkg/lib/indexer/indexer.go
harishsurf: modified code doesn't have this

pkg/lib/indexer/indexer.go
harishsurf: modified the code

pkg/lib/indexer/indexer.go
harishsurf: modified code doesn't use this anymore

pkg/lib/indexer/indexer.go
harishsurf: modified code doesn't use this anymore

pkg/lib/indexer/indexer.go
harishsurf: Not using `errgroup` since the team's input was to collect errors instead of failing it on the first `err`. Switched back to `sync.WaitGroup`

pkg/lib/indexer/indexer.go
harishsurf: same comment as [here](https://github.com/operator-framework/operator-registry/pull/429#discussion_r485815618)

pkg/lib/indexer/indexer.go
harishsurf: I don't think this check is needed in the newly modified code (maybe I am wrong?)

pkg/lib/indexer/indexer.go
benluddy: Ok. That makes sense if the idea is to allow everything to complete despite any errors.

pkg/lib/indexer/indexer.go
harishsurf: Set the concurrency limit to 10

pkg/lib/indexer/indexer.go
kevinrizza: ah, you are mostly right. the problem is the case where bundleVersion is not set -- technically version is an optional field on the CSV. in that case, you still need to generate something for the value of the folder.

mvalahtv:
@harishsurf is it possible to test it?

harishsurf:
@mvalahtv, I have added an e2e test that uses the community operator's catsrc image. If you had to test it locally, once you pull down this code, you could run `opm index export --index quay.io/operatorhubio/catalog:latest` to see it in action

test/e2e/opm_test.go
```diff
@@ -44,6 +44,7 @@ var (
 	indexImage1 = "quay.io/olmtest/e2e-index:" + indexTag1
 	indexImage2 = "quay.io/olmtest/e2e-index:" + indexTag2
 	indexImage3 = "quay.io/olmtest/e2e-index:" + indexTag3
+	indexImage4 = "quay.io/operatorhubio/catalog:latest"
```
kevinrizza, line 47: Can we use one of the test indexes that we already build rather than rely on an external dependency that we don't control?

harishsurf: changed it to use `indexImage2`

pkg/lib/indexer/indexer.go
harishsurf: Added the check to generate a random name when empty

mvalahtv:
@harishsurf it is working fine for me on local machine. I can export catalog in 2 mins. Great job !!!

pkg/lib/indexer/indexer.go
```diff
@@ -513,40 +523,68 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+
+	sem := make(chan struct{}, concurrencyLimitForExport)
+
+	for bundleImage, bundleDir := range bundles {
+		go func(bundleImage string, bundleDir bundleDirPrefix) {
+			defer wg.Done()
+
+			sem <- struct{}{}
+			defer func() {
+				<-sem
```
njhale, line 538: to make this airtight, I think we want a select that checks a context for cancellation, so don't wait forever for slow exports:

```go
select {
case <- ctx.Done():
case <- sem:
}
```

or we could just use [x/sync/semaphore](https://godoc.org/golang.org/x/sync/semaphore#Weighted.Acquire):

```go
import "golang.org/x/sync/semaphore"
// ...
sem := semaphore.NewWeighted(concurrencyLimitForExport)
//...
if err := sem.Acquire(ctx, 1); err != nil {
    return err
}
//...
```

njhale (commented):
Looking good. I think we want to attach a BZ so we can get this merged. It improves UX considerably from what we have today.

pkg/lib/indexer/indexer.go
```diff
@@ -26,12 +29,13 @@ import (
...
-	defaultDatabaseFile   = "index.db"
-	tmpDirPrefix          = "index_tmp_"
-	tmpBuildDirPrefix     = "index_build_tmp"
+	defaultDockerfileName     = "index.Dockerfile"
+	defaultImageTag           = "operator-registry-index:latest"
+	defaultDatabaseFolder     = "database"
+	defaultDatabaseFile       = "index.db"
+	tmpDirPrefix              = "index_tmp_"
+	tmpBuildDirPrefix         = "index_build_tmp"
+	concurrencyLimitForExport = 10
```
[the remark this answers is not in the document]
benluddy: I don't think so, it's mostly I/O bound.

dinhxuanvu (approved):
/lgtm

PR #434, opened 2020-09-04
Bug 1866861: 4.5 index images should work with 4.6 catalog operator.

benluddy:
The ListBundles endpoint was introduced to the registry API in 4.5, but was not consumed by any clients. Several necessary changes to ListBundles were identified as part of changes to dependency resolution in 4.6. Since 4.5.z indexes must work on 4.6 clusters, ListBundles must be minimally functional in 4.5.


ecordell (approved):
/lgtm
/approve

/hold 
for second lgtm

PR #439, opened 2020-09-10
Use multi-stage build for registry image.

benluddy:
The upstream-registry-builder image is bigger than it needs to be, and
about half of its size seems to be coming from unused build
artifacts. Putting all of that in a separate builder stage cuts the
size of the final image by about half.

### DIFF
--- a/upstream-builder.Dockerfile
+++ b/upstream-builder.Dockerfile
@@ -13,7 +13,7 @@ RUN GRPC_HEALTH_PROBE_VERSION=v0.3.2 && \
     wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-$(go env GOARCH) && \
     chmod +x /bin/grpc_health_probe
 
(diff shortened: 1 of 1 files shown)


upstream-builder.Dockerfile
```diff
@@ -12,8 +12,15 @@ RUN make static
 RUN GRPC_HEALTH_PROBE_VERSION=v0.3.2 && \
     wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-$(go env GOARCH) && \
     chmod +x /bin/grpc_health_probe
-RUN cp /build/bin/opm /bin/opm && \
-    cp /build/bin/initializer /bin/initializer && \
-    cp /build/bin/appregistry-server /bin/appregistry-server && \
-    cp /build/bin/configmap-server /bin/configmap-server && \
-    cp /build/bin/registry-server /bin/registry-server
+
+FROM golang:1.13-alpine
```
joelanford, line 16: Is it necessary to include the `go` tooling in the image?

```suggestion
FROM alpine:3
```

PR #442, opened 2020-09-14
disable cross compilation while build issues are resolved

ecordell:

**Reviewer Checklist**


ecordell:
/hold 

this should only be merged if it's not possible to ship on golang 1.14 downstream

shawn-hurley:
Overriding valid bug because this change is to fix the builds that are currently failing.

PR #447, opened 2020-09-17
(fix) adding sanity check to catch empty default channel

gallettilance:

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -83,6 +83,7 @@ func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error 
 				continue
 			}
 		}
+		containsDefault := false
 		for _, channel := range image.bundle.Channels {
 			bundle, err := i.querier.GetBundle(context.TODO(), image.bundle.Package, channel, image.bundle.csv.GetName())
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/registry/populator.go
```diff
@@ -94,6 +95,15 @@ func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error
...
 				break
 			}
+			if channel == image.annotationsFile.GetDefaultChannelName() {
+				containsDefault = true
+			}
+		}
+		if !containsDefault {
+			_, err := i.querier.GetBundleForChannel(context.TODO(), image.bundle.Package, image.annotationsFile.GetDefaultChannelName())
+			if err != nil {
+				errs = append(errs, EmptyDefaultChannel{ErrorString: "Cannot set an empty default channel"})
```
njhale, line 105: If this error string isn't changing, shouldn't it be hard-coded in the `Error() string` method?

pkg/registry/populator.go
```diff
@@ -94,6 +95,15 @@ func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error
 				errs = append(errs, PackageVersionAlreadyAddedErr{ErrorString: "Bundle already added that provides package and csv"})
 				break
 			}
+			if channel == image.annotationsFile.GetDefaultChannelName() {
+				containsDefault = true
+			}
+		}
+		if !containsDefault {
```
njhale, line 102: historically, we've inferred the default channel when |channels|=1 -- is this happening at some other level (e.g. when getting the annotations file)? did we drop this behavior at some point?

gallettilance: yep that still happens today and is what is returned by the `GetDefaultChannel` function

pkg/registry/populator.go
gallettilance: I forgot to add `err` to the string - my bad

hasbro17:
@gallettilance Just checking in, since this PR has been sitting for a while now. Do you still have time to rebase and address the review? If not, I can take over.

gallettilance:
@hasbro17 Please take over if this is still needed

ankitathomas:
https://github.com/operator-framework/operator-registry/commit/809651742864f61db6099163eb1b2539d6d92cd3#diff-5dd7ba96a38b29d88c95d131248cc0dc6bf9b3b62adbe50612975076d1d3b96d

With the semver based channel head/default channel selection, we can close this out

PR #448, opened 2020-09-18
Bug 1880501: Adding overwrite-latest flag to index add

gallettilance:

Allow index maintainers to overwrite the latest bundle in a package on `opm index add` by adding an `--overwrite-latest` option.

**Caveats**

- valid only when using `replaces` mode
- each overwrite pulls all images for the package in order to rebuild the graph

**Benefits**

- image pulls are parallelized
- supports overwrites of the update graph

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -31,10 +31,11 @@ type AddToRegistryRequest struct {
 	Bundles       []string
 	Mode          registry.Mode
 	ContainerTool containertools.ContainerTool
+	Overwrite     bool
 }
 
 func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
-	db, err := sql.Open("sqlite3", request.InputDatabase)
+	db, err := sql.Open("sqlite3", "file:"+request.InputDatabase+"?_foreign_keys=on")
 	if err != nil {
 		return err
 	}
@@ -84,51 +85,115 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		simpleRefs = append(simpleRefs, image.SimpleReference(ref))
 	}
 
-	if err := populate(context.TODO(), dbLoader, graphLoader, dbQuerier, reg, simpleRefs, request.Mode); err != nil {
+	if err := populate(context.TODO(), dbLoader, graphLoader, dbQuerier, reg, simpleRefs, request.Mode, request.Overwrite); err != nil {
 		r.Logger.Debugf("unable to populate database: %s", err)
 
 		if !request.Permissive {
 			r.Logger.WithError(err).Error("permissive mode disabled")
 			return err
-		} else {
-			r.Logger.WithError(err).Warn("permissive mode enabled")
 		}
+		r.Logger.WithError(err).Warn("permissive mode enabled")
 	}
 
 	return nil
 }
 
-func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode) error {
+func unpackImage(ctx context.Context, reg image.Registry, ref image.Reference) (image.Reference, string, func(), error) {
 	var errs []error
+	workingDir, err := ioutil.TempDir("./", "bundle_tmp")
+	if err != nil {
+		errs = append(errs, err)
+	}
 
-	unpackedImageMap := make(map[image.Reference]string, 0)
-	for _, ref := range refs {
-		workingDir, err := ioutil.TempDir("./", "bundle_tmp")
-		if err != nil {
-			errs = append(errs, err)
-			continue
+	if err = reg.Pull(ctx, ref); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err = reg.Unpack(ctx, ref, workingDir); err != nil {
+		errs = append(errs, err)
+	}
+
+	cleanup := func() {
+		if err := os.RemoveAll(workingDir); err != nil {
+			logrus.Error(err)
 		}
-		defer os.RemoveAll(workingDir)
-
-		if err = reg.Pull(ctx, ref); err != nil {
-			errs = append(errs, err)
-			continue
-		}
-
-		if err = reg.Unpack(ctx, ref, workingDir); err != nil {
-			errs = append(errs, err)
-			continue
-		}
-
-		unpackedImageMap[ref] = workingDir
 	}
 
 	if len(errs) > 0 {
-		return utilerrors.NewAggregate(errs)
+		return nil, "", cleanup, utilerrors.NewAggregate(errs)
+	}
+	return ref, workingDir, cleanup, nil
+}
+
+func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode, overwrite bool) error {
+	unpackedImageMap := make(map[image.Reference]string, 0)
+	for _, ref := range refs {
+		to, from, cleanup, err := unpackImage(ctx, reg, ref)
+		if err != nil {
+			return err
+		}
+		unpackedImageMap[to] = from
+		defer cleanup()
 	}
 
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap)
+	overwriteImageMap := make(map[string]map[image.Reference]string, 0)
+	if overwrite {
+		// find all bundles that are attempting to overwrite
+		for to, from := range unpackedImageMap {
+			img, err := registry.NewImageInput(to, from)
+			if err != nil {
+				return err
+			}
+			overwritten, err := querier.GetBundlePathIfExists(ctx, img.Bundle.Name)
+			if err != nil {
+				if err == registry.ErrBundleImageNotInDatabase {
+					continue
+				}
+				return err
+			}
+			if overwritten != "" {
+				// get all bundle paths for that package - we will re-add these to regenerate the graph
+				bundles, err := querier.GetBundlesForPackage(ctx, img.Bundle.Package)
+				if err != nil {
+					return err
+				}
+				cleanups := make(chan func(), 1)
+				errs := make(chan error, 1)
+				for bundle := range bundles {
+					if _, ok := overwriteImageMap[img.Bundle.Package]; !ok {
+						overwriteImageMap[img.Bundle.Package] = make(map[image.Reference]string, 0)
+					}
+					// parallelize image pulls
+					go func(bundle registry.BundleKey, img *registry.ImageInput) {
+						if bundle.CsvName != img.Bundle.Name {
+							to, from, cleanup, err := unpackImage(ctx, reg, image.SimpleReference(bundle.BundlePath))
+							if err != nil {
+								errs <- err
+							}
+							cleanups <- cleanup
+							overwriteImageMap[img.Bundle.Package][to] = from
+						} else {
+							overwriteImageMap[img.Bundle.Package][to] = from
+							delete(unpackedImageMap, to)
+						}
+					}(bundle, img)
+				}
+				for i := 0; i < len(bundles)-1; i++ {
+					select {
+					case err := <-errs:
+						return err
+					default:
+						cleanup := <-cleanups
+						defer cleanup()
+					}
+				}
+			} else {
+				return fmt.Errorf("index add --overwrite-latest is only supported when using bundle images")
+			}
+		}
+	}
 
+	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap, overwriteImageMap, overwrite)
 	return populator.Populate(mode)
 }
 
--- a/pkg/registry/bundle.go
+++ b/pkg/registry/bundle.go
@@ -73,7 +73,6 @@ func NewBundleFromStrings(name, pkgName string, channels []string, objs []string
 func (b *Bundle) Size() int {
 	return len(b.Objects)
 }
-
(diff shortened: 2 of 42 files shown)


pkg/registry/populator.go
```diff
@@ -63,14 +65,21 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
 	return nil
 }
 
-func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error {
+func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput, mode Mode) error {
+
+	if i.overwrite && mode != ReplacesMode {
+		return OverwriteErr{ErrorString: "overwrite-latest flag is only supported in Replaces mode"}
+	}
```
njhale, line 72: should we also return an error when `len(imagesToAdd) > 1 && i.overwrite`?

dinhxuanvu:
Can we have a description/summary on what we are trying to accomplish with this? Perhaps mentioning the use case/issue here would be helpful.

pkg/registry/populator.go
```diff
@@ -119,6 +156,44 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode)
 		// same package linearly.
 		var err error
 		var validImagesToAdd []*ImageInput
+		overwriteBundle := ImageInput{}
+
+		// find the image to overwrite if it exists
+		if i.overwrite {
+			for j, image := range imagesToAdd {
```
njhale, line 163: are we no longer constraining our input to a single bundle when `--overwrite-latest` has been specified?

pkg/registry/populator.go
```diff
@@ -119,6 +156,44 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode)
...
+						break
+					}
+				}
+			}
+		}
+
+		// Add the overwriting bundle first
+		if overwriteBundle != (ImageInput{}) {
+			// Remove existing csv
+			err = i.loader.RmCsv(overwriteBundle.bundle.csv.GetName())
```
njhale, line 185: would it be easier to make `loadManifestsReplaces` perform an "upsert", since we've already checked for the conflict cases in the sanity check?

pkg/registry/populator.go
```diff
@@ -119,6 +156,42 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode)
...
+					imagesToAdd = append(imagesToAdd[:j], imagesToAdd[j+1:]...)
+					break
+				}
+			}
+		}
+
+		// Add the overwriting bundle first
+		if overwriteBundle != (ImageInput{}) {
+			// Remove existing csv
+			err = i.loader.RmCsv(overwriteBundle.bundle.csv.GetName())
```
njhale, line 183: This should only happen when `i.overwrite == true`. there are tests that depend on the standard failure behavior which fail now because this always overwrites.

pkg/registry/testdata/overwrite/prometheus.0.14.0/manifests/servicemonitor.crd.yaml
```diff
@@ -0,0 +1,224 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
njhale, line 2: we don't need CRDs and dependencies to ensure that overwriting works, right?

pkg/registry/populator_test.go
```diff
@@ -868,3 +871,250 @@ func TestDeprecateBundle(t *testing.T) {
...
+				bundlePaths = append(bundlePaths, strings.Join([]string{bundle.BundlePath, bundle.ChannelName}, "/"))
+			}
+			fmt.Println("remaining", bundlePaths)
+			require.ElementsMatch(t, tt.expected.remainingBundles, bundlePaths)
+
+			// Ensure remaining channels and default channel match
+			packages, err := query.ListPackages(context.Background())
+			require.NoError(t, err)
+
+			for _, pkg := range packages {
```
njhale, line 1101: this doesn't catch when a package in `remainingPkgChannels` doesn't exist in the database

pkg/registry/populator.go
```diff
@@ -119,6 +160,47 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode)
...
+					// get head of the channel
+					bundle, err := i.querier.GetBundleForChannel(context.TODO(), image.bundle.Package, channel)
+					if err != nil {
+						return err
+					}
+					if bundle != nil && bundle.GetCsvName() == image.bundle.csv.GetName() {
+						// found the overwriting bundle
+						overwriteBundle = *image
+						// remove the overwriting bundle from the images to add since we will add that bundle first
+						imagesToAdd = append(imagesToAdd[:j], imagesToAdd[j+1:]...)
```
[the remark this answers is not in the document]
gallettilance: you're right - at the time I didn't know the adding would be identical so I separated it.

pkg/registry/testdata/overwrite/prometheus.0.14.0/manifests/servicemonitor.crd.yaml
gallettilance: We may want to test those cases too

pkg/lib/registry/registry.go
```diff
@@ -84,51 +86,89 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
...
+					if _, ok := overwriteImageMap[img.Bundle.Package]; !ok {
+						overwriteImageMap[img.Bundle.Package] = make(map[image.Reference]string, 0)
+					}
+					if bundle.CsvName != img.Bundle.Csv.GetName() {
+						to, from, cleanup, err := unpackImage(ctx, reg, image.SimpleReference(bundle.BundlePath))
+						if err != nil {
+							return err
+						}
+						defer cleanup()
+						overwriteImageMap[img.Bundle.Package][to] = from
```
njhale, line 164: nit: log the errors that come out of `cleanup`

pkg/sqlite/remove_test.go
```diff
@@ -30,7 +31,8 @@ func TestRemover(t *testing.T) {
 			query,
 			map[image.Reference]string{
 				image.SimpleReference("quay.io/test/" + name): "../../bundles/" + name,
-			}).Populate(registry.ReplacesMode)
+			},
+			make(map[string]map[image.Reference]string, 0), false).Populate(registry.ReplacesMode)
```
njhale, line 35: send nil instead?

pkg/registry/populator_test.go
```diff
@@ -868,3 +871,258 @@ func TestDeprecateBundle(t *testing.T) {
...
+
+func TestOverwrite(t *testing.T) {
+	type args struct {
+		firstAdd   map[image.Reference]string
+		secondAdd  map[image.Reference]string
+		overwrites map[string]map[image.Reference]string
+	}
+	type pkgChannel map[string][]string
+	type expected struct {
+		err                      error
```
njhale, line 883: unused. It seems we're missing negative cases; e.g. "I should get an error when overwriting a non-latest bundle"

pkg/registry/populator.go
```diff
@@ -119,28 +167,38 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode)
 		// same package linearly.
 		var err error
 		var validImagesToAdd []*ImageInput
+
+		for pkg := range i.overwriteDirMap {
+			err := i.loader.RemovePackage(pkg)
```
njhale, line 172: we lose all "inline" bundles not provided by bundle images

pkg/lib/registry/registry.go
```diff
@@ -84,51 +86,89 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
...
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap)
+	overwriteImageMap := make(map[string]map[image.Reference]string, 0)
+	if overwrite {
+		// find all bundles that are attempting to overwrite
+		for to, from := range unpackedImageMap {
+			img, err := registry.NewImageInput(to, from)
+			if err != nil {
+				return err
+			}
+			overwritten, err := querier.GetBundlePathIfExists(ctx, img.Bundle.Csv.GetName())
```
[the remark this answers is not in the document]
gallettilance: Great idea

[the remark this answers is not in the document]
njhale: after our discussion -- offline -- it sounded like we should refuse to update index images with inline content when `--overwrite-latest` is given.

pkg/registry/bundle.go
```diff
@@ -41,7 +41,7 @@ type Bundle struct {
 	Package      string
 	Channels     []string
 	BundleImage  string
-	csv          *ClusterServiceVersion
+	Csv          *ClusterServiceVersion
```
ecordell, line 44: why did this change to public?

pkg/registry/imageinput.go
```diff
@@ -16,9 +16,9 @@ type ImageInput struct {
 	metadataDir      string
 	to               image.Reference
 	from             string
-	annotationsFile  *AnnotationsFile
+	AnnotationsFile  *AnnotationsFile
```
ecordell, line 19: same: why did these become public?

pkg/registry/populator.go
ecordell, line 172: I think we want the remove / add to happen in a single transaction. 

It would also be nice to not need to thread through the "overwriteImages" separately everywhere. is there a way we can we push the `overwrite` option down to `loadManifestReplaces` (or call `loadManifestsReplaceOverwrite` if overwrite is set, etc), so that we only need to worry about overwriting when we try to insert and it fails because it already exists?

pkg/registry/populator.go
```diff
@@ -23,18 +23,22 @@ type Dependencies struct {
...
 	return &DirectoryPopulator{
-		loader:      loader,
-		graphLoader: graphLoader,
-		querier:     querier,
-		imageDirMap: imageDirMap,
+		loader:          loader,
+		graphLoader:     graphLoader,
+		querier:         querier,
+		imageDirMap:     imageDirMap,
+		overwriteDirMap: overwriteDirMap,
```
ecordell, line 40: do we need to distinguish which ones are "overwrites" up front?

pkg/lib/registry/registry.go
```diff
@@ -84,51 +85,98 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
...
+		return nil, "", cleanup, utilerrors.NewAggregate(errs)
 	}
+	return ref, workingDir, cleanup, nil
+}
 
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap)
+func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode, overwrite bool) error {
+	unpackedImageMap := make(map[image.Reference]string, 0)
+	for _, ref := range refs {
+		to, from, cleanup, err := unpackImage(ctx, reg, ref)
```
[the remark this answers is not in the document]
gallettilance: good point

pkg/registry/bundle.go
gallettilance: https://github.com/operator-framework/operator-registry/pull/448/files#diff-08f655f8aa9db090b9d078c7c11450a3R141 we need to unpack the images here and in order to know which images to unpack we need to access these fields outside the registry package

pkg/registry/imageinput.go
gallettilance: same as https://github.com/operator-framework/operator-registry/pull/448#discussion_r493115614

pkg/registry/populator.go
gallettilance: yeah I think it makes sense to do it upfront so we only pull the images when we need to. I think more of the registry package needs to be made public in order to make that possible but it shouldn't be too bad

pkg/lib/registry/registry.go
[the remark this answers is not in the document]
gallettilance: any objections to parallelizing it?

pkg/registry/populator.go
gallettilance: we can but that requires exposing the image registry package to the registry package which today can't easily be done because of cyclic imports...

pkg/registry/bundle.go
ecordell: You shouldn't need `Csv.GetName` - that's what `Name` is, which is already public

maneeshmehra:
@gallettilance had already marked the bugzilla item linked to this PR as critical. Asking all reviewers to review this with highest priority as it blocks our operator verification pipelines. TIA.

pkg/lib/registry/registry.go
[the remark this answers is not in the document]
njhale: I'm fine with this as is for now.

pkg/registry/populator.go
njhale: I think you could avoid importing the image/registry package with a `ReplacePackage(pkg string, bundle ...Bundle)` method on the loader -- still getting a single transaction at the DB level. 

However, I do think that at this point, we should prioritize getting something working and merged over long-term support; partly b/c the declarative index config will supersede this work in the near future.

pkg/registry/bundle.go
gallettilance: gotcha! will revert that change then

njhale:
All pending comments addressed offline.

/lgtm

PR #451, opened 2020-09-23
Bug 1878213: Update index add func to handle optional default channel

dinhxuanvu:
The default channel annotation is no longer required as default
channel information is now optional. Users may provide default
channel if they intend to update default channel information in
the index/DB.

The default channel may still be infered if package.yaml is present.
If the default channel cannot be infered, then the default channel
will be omitted.

Bundle validation will no longer error out if default channel
annotation is missing.

If default channel is not provided in annotations, the existing
default channel will remain the same.

For the case of the first bundle being added and the default channel
isn't provided, it will be inferred from provided channel list if
there is only one channel. Otherwise, the index add will fail until
the default channel is provided.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -161,22 +161,12 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, mode Mode)
 }
 
 func (i *DirectoryPopulator) loadManifestsReplaces(bundle *Bundle, annotationsFile *AnnotationsFile) error {
-	channels, err := i.querier.ListChannels(context.TODO(), annotationsFile.GetName())
-	existingPackageChannels := map[string]string{}
-	for _, c := range channels {
-		current, err := i.querier.GetCurrentCSVNameForChannel(context.TODO(), annotationsFile.GetName(), c)
-		if err != nil {
-			return err
-		}
-		existingPackageChannels[c] = current
-	}
-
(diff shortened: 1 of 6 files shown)


dinhxuanvu:
@joelanford Slight change of plan. I planned to make 2 PRs for validation and for index add change but the PR is not that complex. So I keep everything in one. Plus, we need a BZ for every PR due to master is close which is a bit of unnecessary hassle.

pkg/registry/populator.go
```diff
@@ -389,9 +387,29 @@ func translateAnnotationsIntoPackage(annotations *AnnotationsFile, csv *ClusterS
...
+		PackageName: annotations.GetName(),
+		Channels:    channels,
+	}
+
+	defaultChan := annotations.GetDefaultChannelName()
+	if defaultChan != "" {
+		if _, found := existingChannels[defaultChan]; found {
+			manifest.DefaultChannelName = annotations.GetDefaultChannelName()
+		} else {
+			return manifest, fmt.Errorf("Default channel doesn't exist: %s", defaultChan)
```
exdx, line 399: nit: since this error is user facing I think it could be more descriptive. Users can run into this issue and its not very intuitive why. 

`fmt.Errorf("channel %s, labelled as default in annotations.yaml, not found in existing package channels", defaultChan)`

exdx (commented):
/lgtm

PR #455, opened 2020-09-24
Bug 1882479: erroring when replacement not found

gallettilance:

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -302,15 +302,11 @@ func (i *DirectoryPopulator) getNextReplacesImagesToAdd(imagesToAdd []*ImageInpu
 		// no new images can be added, the current iteration aggregates all the
 		// errors that describe invalid bundles
 		if pkgFoundImages == 0 && pkgRemainingImages > 0 {
-			errs = append(errs, utilerrors.NewAggregate(pkgErrs))
+			errs = append(errs, pkgErrs...)
 		}
 	}
 
-	if len(errs) > 0 {
-		return nil, nil, utilerrors.NewAggregate(errs)
-	}
-
-	return foundImages, remainingImages, nil
+	return foundImages, remainingImages, utilerrors.NewAggregate(errs)
 }
 
 func (i *DirectoryPopulator) loadManifestsSemver(bundle *Bundle, annotations *AnnotationsFile, skippatch bool) error {
--- a/pkg/registry/populator_test.go
+++ b/pkg/registry/populator_test.go
@@ -686,6 +686,39 @@ func CheckBundlesHaveContentsIfNoPath(t *testing.T, db *sql.DB) {
 	}
 }
 
+func TestDirectoryPopulator(t *testing.T) {
+	db, cleanup := CreateTestDb(t)
+	defer cleanup()
+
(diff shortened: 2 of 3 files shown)


pkg/registry/populator_test.go
```diff
@@ -631,6 +631,37 @@ func TestListBundles(t *testing.T) {
...
+	}
+	add := map[image.Reference]string{
+		image.SimpleReference("quay.io/test/etcd.0.9.2"):        "../../bundles/etcd.0.9.2",
+		image.SimpleReference("quay.io/test/prometheus.0.22.2"): "../../bundles/prometheus.0.22.2",
+	}
+	expectedErr := errors.NewAggregate([]error{
+		fmt.Errorf("Invalid bundle %s, bundle specifies a non-existent replacement %s", "etcdoperator.v0.9.2", "etcdoperator.v0.9.0"),
+		fmt.Errorf("Invalid bundle %s, bundle specifies a non-existent replacement %s", "prometheusoperator.0.22.2", "prometheusoperator.0.15.0"),
+	})
+	require.ElementsMatch(t, expectedErr, populate(add))
```
njhale, line 664: I don't think `TestListBundles` is the appropriate spec for this. It probably needs its own.

pkg/registry/replacesgraphloader.go
```diff
@@ -15,11 +15,21 @@ func (r *ReplacesGraphLoader) CanAdd(bundle *Bundle, graph *Package) (bool, erro
 		return false, fmt.Errorf("Invalid content, unable to parse bundle")
 	}
 
+	// adding the first bundle in the graph
+	if replaces == "" {
+		return true, nil
+	}
+	csvName := bundle.Name
```
njhale, line 22: nit: `csvName` is only used once, so we should probably omit the local variable in favor of `bundle.Name`.

pkg/registry/replacesgraphloader.go
```diff
@@ -15,11 +15,21 @@ func (r *ReplacesGraphLoader) CanAdd(bundle *Bundle, graph *Package) (bool, erro
 		return false, fmt.Errorf("Invalid content, unable to parse bundle")
 	}
 
+	// adding the first bundle in the graph
+	if replaces == "" {
```
njhale, line 19: this is duplicated

gallettilance: nice catch - must have happened when I reverted this file...

pkg/registry/populator.go
```diff
@@ -302,15 +302,11 @@ func (i *DirectoryPopulator) getNextReplacesImagesToAdd(imagesToAdd []*ImageInpu
 		// no new images can be added, the current iteration aggregates all the
 		// errors that describe invalid bundles
 		if pkgFoundImages == 0 && pkgRemainingImages > 0 {
```
njhale, line 304: nit: you could always use the lengths of `remainingImages` and `foundImages` instead of these counters

gallettilance: +1 just didn't want to change too much about the add when implementing the overwrite fix

ecordell (approved):
/lgtm

PR #459, opened 2020-09-28
Bug 1883329: Bundle validation shouldn't fail if default channel is not provided

dinhxuanvu:
Default channel annotation is now optional. It is only needed if
there is an intention to update default channel in the index.
As a result, bundle validation will allow default channel annotation
to be missing.

**Reviewer Checklist**


joelanford:
LGTM

(not sure if I have \<slash>lgtm access, but even if I did, I'd rather leave it for another OLM'er to review and merge :slightly_smiling_face: )

PR #460, opened 2020-09-29
Bug 1883560: Clean up /tmp/directory in the operator-registry image

akram:
Clean up /tmp/directory in the operator-registry image

The image is 320MB with 265MB just in /tmp

PR #461, opened 2020-09-29
Bug 1883560: fix(dockerfile): clean out tmp directory after binaries are placed

ecordell:
Remove bins from /tmp after they've been moved into the correct location

The image was much larger than it needed to be.

**Reviewer Checklist**


ecordell:
closing in favor of #460

hasbro17 (approved):
/lgtm

PR #463, opened 2020-09-29
feat: index image opm compatibility e2e test

exdx:
This is to help ensure that no changes are introduced into operator-registry that cause a problem for indexes.
The test measures startup success (meaning that the latest production index data can be properly handled by `opm registry serve`)

Test requirements:
* REDHAT_REGISTRY_USERNAME and REDHAT_REGISTRY_PASSWORD environmental variables be added to registry CI. I have tokens associated with my personal redhat account that we could use, or make new ones. 
* Due to https://github.com/moby/moby/issues/35987 we must be root user when copying some read-only files out of the 4.6 index image (which is based off the UBI image). This means having to use `sudo -E <env variables>` when running this test locally, which isn't great. Hopefully in CI we already run the tests as root user. 

**Reviewer Checklist**

### DIFF
--- a/test/e2e/opm_test.go
+++ b/test/e2e/opm_test.go
@@ -8,13 +8,15 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/otiai10/copy"
-	"github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/util/rand"
-
+	command "github.com/operator-framework/operator-registry/cmd/opm/registry"
+	"github.com/operator-framework/operator-registry/pkg/client"
 	"github.com/operator-framework/operator-registry/pkg/containertools"
 	"github.com/operator-framework/operator-registry/pkg/image"
 	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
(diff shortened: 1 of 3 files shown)


test/e2e/opm_test.go
```diff
@@ -435,3 +439,92 @@ var _ = Describe("opm", func() {
...
+			// dial server via gRPC client to ensure successful startup
+			// by default server listens on port 50051 for requests
+			var client *rpc.Client
+			Eventually(func() error {
+				const addr = "127.0.0.1:50051"
+				client, err = rpc.Dial("tcp", addr)
+				GinkgoT().Logf("rpc dial error: %s", err)
+				return err
+			}).Should(Succeed())
+
```
ecordell, line 516: Can we add some basic sanity checks here?

Maybe:
 - the list bundles api returns something? 
 - the listimages api returns a list of images? (we use it for disconnected mirroring)
 - is there a package we're pretty sure will always exist that we could use? (amq?)

ecordell: I realize that this part of the test will be slightly more brittle, but hopefull if it fails it will always be clear what the issue is, and if the test needs to be updated

exdx: I think that it's a good idea. I propose using the healthcheck and list bundles endpoints since those together demonstrate the index is running and is the most future-proof (as we move to the declarative-based index format). 

The registry interface exposes these methods so its easy to implement in the test as well.

ecordell:
/lgtm
/hold

holding until upstream tests pass

exdx:
Looking at the upstream test results looks like the e2e failure was unrelated to this PR https://travis-ci.com/github/operator-framework/operator-registry/jobs/399778402. 

Looks like we don't run the opm e2e test in Travis because the credentials are not provided (the tests are skipped and therefore pass). So I'm currently working on making sure the new test runs downstream

PR #464, opened 2020-09-29
Bug 1883905: (fix) concurrent writes during image pulls

gallettilance:

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -157,35 +157,37 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 				if err != nil {
 					return err
 				}
-				cleanups := make(chan func(), 1)
-				errs := make(chan error, 1)
+				type unpackedImage struct {
+					to      image.Reference
+					from    string
+					cleanup func()
+					err     error
+				}
+				unpacked := make(chan unpackedImage)
 				for bundle := range bundles {
-					if _, ok := overwriteImageMap[img.Bundle.Package]; !ok {
-						overwriteImageMap[img.Bundle.Package] = make(map[image.Reference]string, 0)
-					}
 					// parallelize image pulls
 					go func(bundle registry.BundleKey, img *registry.ImageInput) {
 						if bundle.CsvName != img.Bundle.Name {
 							to, from, cleanup, err := unpackImage(ctx, reg, image.SimpleReference(bundle.BundlePath))
-							if err != nil {
-								errs <- err
-							}
-							cleanups <- cleanup
-							overwriteImageMap[img.Bundle.Package][to] = from
+							unpacked <- unpackedImage{to: to, from: from, cleanup: cleanup, err: err}
 						} else {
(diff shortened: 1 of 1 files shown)


pkg/lib/registry/registry.go
```diff
@@ -157,35 +157,36 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
...
-							if err != nil {
-								errs <- err
-							}
-							cleanups <- cleanup
-							overwriteImageMap[img.Bundle.Package][to] = from
+							unpacked <- unpackedImage{to: to, from: from, cleanup: cleanup, err: err}
 						} else {
-							overwriteImageMap[img.Bundle.Package][to] = from
+							unpacked <- unpackedImage{to: to, from: from, cleanup: func() { return }, err: nil}
+							// safe because this case is only hit once
```
ecordell, line 175: it's only hit once per goroutine, but it's still concurrent access to a map

gallettilance: what I meant is there is only one go routine that will ever hit this case

pkg/lib/registry/registry.go
```diff
@@ -157,35 +157,37 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
...
 					// parallelize image pulls
 					go func(bundle registry.BundleKey, img *registry.ImageInput) {
 						if bundle.CsvName != img.Bundle.Name {
 							to, from, cleanup, err := unpackImage(ctx, reg, image.SimpleReference(bundle.BundlePath))
-							if err != nil {
-								errs <- err
-							}
-							cleanups <- cleanup
-							overwriteImageMap[img.Bundle.Package][to] = from
+							unpacked <- unpackedImage{to: to, from: from, cleanup: cleanup, err: err}
```
ecordell, line 172: if `unpackImage` were updated to return an `unpackedImage`, this whole goroutine could just be:

```go
go func(bundle registry.BundleKey, img *registry.ImageInput) {
  unpacked <- unpackImage(ctx, reg, image.SimpleReference(bundle.BundlePath))
}
```

gallettilance: I think we want to skip the one we already unpacked

pkg/lib/registry/registry.go
```diff
@@ -157,35 +157,37 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
...
+				for i := 0; i < len(bundles); i++ {
+					unpack := <-unpacked
+					if unpack.err != nil {
+						return unpack.err
+					}
+					overwriteImageMap[img.Bundle.Package][unpack.to] = unpack.from
+					if _, ok := unpackedImageMap[unpack.to]; ok {
+						delete(unpackedImageMap, unpack.to)
 					}
+					defer unpack.cleanup()
```
ankitathomas, line 190: Shouldn't we cleanup bundles if they have errors?

gallettilance: ideally yes but it depends on the error if there is anything to clean up

PR #465, opened 2020-09-30
Bug 1883377: Enable default channel inference if only one channel is provided

dinhxuanvu:
Improve the error handle for missing default channel.

Infer default channel for the first bundle of the package if not
provided via annotation. This only works if only one channel is
provided in channel list.

**Reviewer Checklist**


dinhxuanvu:
Note: The default channel mechanism is the same as it is in `replaces` mode.

jmrodri:
If this keeps the foreign key violation from happening, I am all for it.

pkg/registry/bundlegraphloader_test.go
```diff
@@ -9,14 +9,20 @@ import (
 )
 
 func TestBundleGraphLoader(t *testing.T) {
+	empty := &AnnotationsFile{}
+	alpha := &AnnotationsFile{}
+	beta := &AnnotationsFile{}
```
ecordell, line 14: is beta used anywhere?

dinhxuanvu: It isn't. I'll remove it.

dinhxuanvu:
@ecordell Updated. PTAL

ecordell (approved):
/lgtm

PR #468, opened 2020-10-01
Bug 1881527: Add e2e tests for export directory structure

ankitathomas:
The multi-package export in https://github.com/operator-framework/operator-registry/pull/429 changed the bundle export path, breaking pipelines depending on the old path format. Switching to the old behavior for single package exports.

### DIFF
--- a/cmd/opm/index/export.go
+++ b/cmd/opm/index/export.go
@@ -39,6 +39,7 @@ func newIndexExportCmd() *cobra.Command {
 	if err := indexCmd.MarkFlagRequired("index"); err != nil {
 		logrus.Panic("Failed to set required `index` flag for `index export`")
 	}
+	indexCmd.Flags().StringP("", "o", "", "the package to export. Deprecated, please use --package flag")
 	indexCmd.Flags().StringSliceP("package", "p", nil, "comma separated list of packages to export")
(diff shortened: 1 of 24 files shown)


kevinrizza (commented):
Didn't we also want to revert the change to the parameter?

ankitathomas:
> Didn't we also want to revert the change to the parameter?

There are other opm index commands that use -p for the package flag, I thought we should have uniformity across our commands for flags

kevinrizza:
Yeah, you're right. Let's try to unify everything later, and get this fix in for now.

/approve

shawn-hurley:
/hold

When discussed with the pipeline, they asked us to keep the new behavior. 

Is anyone else needing this to stay the same? If we need to keep it the same, we have to coordinate the release of this update with the changes in IIB to revert back.

ankitathomas:
> Didn't we also want to revert the change to the parameter?

I've added back `-o` as a deprecated flag

gallettilance (approved):
/lgtm

njhale:
since we need to keep the new behavior, I'm going to close this out in favor of a flag alias PR: https://github.com/operator-framework/operator-registry/pull/485

PR #474, opened 2020-10-12
Make the inability to create /etc/nsswitch.conf non-fatal.

benluddy:
Images run as non-root were not allowed to write to
/etc/nsswitch.conf, which had been considered required. Instead, the
scratch/non-glibc images published by the Operator Framework project
now include a default nsswitch config file, and runtime creation of a
default nsswitch config file is now best-effort.


ecordell (approved):
/lgtm

dinhxuanvu (approved):
/lgtm

PR #479, opened 2020-10-14
Bug 1885425: fix(indexing): include skips in bulk add order

njhale:

When adding more than one bundle to an index in replaces mode, topologically sort the input using the skips field (in the absence of replaces). This ensures a deterministic update graph for authors who rely solely on skips to generate edges.

Some package maintainers rely on the `skip` field to generate graph edges and omit `replaces` entirely. Because of this, no ordering exists between input bundles on bulk index add operations This can lead to non-deterministic update graphs when the list of input bundles isn't given in the desired order for `opm index add`; e.g. OCP 4.6 index refresh jobs which bulk add an opaque set of bundles.


pkg/registry/populator_test.go
```diff
@@ -1272,19 +1271,14 @@ func TestOverwrite(t *testing.T) {
...
-				// comprise an aggregate error isn't
-				// important to the tested behaviors,
-				// so sort them:
-				errs := agg.Errors()
-				sort.Slice(errs, func(i, j int) bool {
-					return errs[i].Error() < errs[j].Error()
-				})
-				popErr = utilerrors.NewAggregate(errs)
+
+			err = populate(tt.args.secondAdd, tt.args.overwrites)
```
njhale, line 1275: I ran into a similar issue, but needed to make it even less dependent on a specific error type, since I ended up flattening the array that was being compared under the hood.

njhale:
CI should be back

/retest

pkg/registry/input_stream.go
```diff
@@ -0,0 +1,209 @@
...
+		stream.packages[pkg][image.Bundle.Name] = image
+	}
+
+	// Validate each package, dropping any invalid packages so that the stream can still be used with remaining packages.
+	var errs []error
+	for pkg, images := range stream.packages {
+		// Multiple input images require some ordering between them -- using skips and replaces -- to ensure an index add operation is deterministic
+		connected, err := graphConnected(images)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("Error determining connectedness of package graph %s: %s", pkg, err))
```
kevinrizza, line 35: Isn't there slightly more to this? Don't we need to divide them up into a graph per channel? You could, presumably, add two bundles that aren't connected at all. In that case, you could add them in any order and it shouldn't return an error.

kevinrizza (changes requested):

pkg/registry/input_stream_test.go
```diff
@@ -0,0 +1,305 @@
...
+			return false
+		}
+		if next != nil {
+			t.Errorf("expected stream to end after the sequence was exhausted")
+			return false
+		}
+
+		return true
+	}
+	if err := quick.Check(f, nil); err != nil {
```
ecordell, line 302: 👍

pkg/registry/input_stream.go
njhale: > Don't we need to divide them up into a graph per channel?

I don't think we _need_ to. I imposed strict rules to simplify our logic and the edge cases we need to deal with. How often will we be adding bundles to a package with diverging branches?

kevinrizza: I think it's actually a fairly common case at the moment: you build release channels based on minor or major versions, then use a skiprange in the higher channel to determine what set of things can be upgraded from one to the next.

njhale: you're right, I'll need to rethink this.

njhale: Ah, I think the key here is that if either bundle defines a different default channel, then the order does matter.

njhale:
Superseded by #503

PR #483, opened 2020-10-20
Bug 1889388: Set replaces in ListBundles query result using channel entries.

benluddy:
The "replaces" column in ListBundles query result rows was being populated with the "replaces" from the respective bundle. For registries built using semver mode, the source of truth for "replaces" is the "replaces" column on each channel entry row.

### DIFF
--- a/pkg/sqlite/query.go
+++ b/pkg/sqlite/query.go
@@ -924,19 +924,20 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
 	return "", nil
 }
 
-func (s *SQLQuerier) ListBundles(ctx context.Context) ([]*api.Bundle, error) {
-	query := `SELECT DISTINCT channel_entry.entry_id, operatorbundle.bundle, operatorbundle.bundlepath,
-	channel_entry.operatorbundle_name, channel_entry.package_name, channel_entry.channel_name, operatorbundle.replaces, operatorbundle.skips,
-	operatorbundle.version, operatorbundle.skiprange,
-	dependencies.type, dependencies.value,
-	properties.type, properties.value
-	FROM channel_entry
-	INNER JOIN operatorbundle ON operatorbundle.name = channel_entry.operatorbundle_name
-	LEFT OUTER JOIN dependencies ON dependencies.operatorbundle_name = channel_entry.operatorbundle_name
-	LEFT OUTER JOIN properties ON properties.operatorbundle_name = channel_entry.operatorbundle_name
-	INNER JOIN package ON package.name = channel_entry.package_name`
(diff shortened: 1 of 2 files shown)


pkg/sqlite/query.go
```diff
@@ -924,19 +924,20 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
...
-	channel_entry.operatorbundle_name, channel_entry.package_name, channel_entry.channel_name, operatorbundle.replaces, operatorbundle.skips,
-	operatorbundle.version, operatorbundle.skiprange,
-	dependencies.type, dependencies.value,
-	properties.type, properties.value
-	FROM channel_entry
-	INNER JOIN operatorbundle ON operatorbundle.name = channel_entry.operatorbundle_name
-	LEFT OUTER JOIN dependencies ON dependencies.operatorbundle_name = channel_entry.operatorbundle_name
-	LEFT OUTER JOIN properties ON properties.operatorbundle_name = channel_entry.operatorbundle_name
-	INNER JOIN package ON package.name = channel_entry.package_name`
+const listBundlesQuery = `SELECT DISTINCT channel_entry.entry_id, operatorbundle.bundle, operatorbundle.bundlepath,
```
exdx, line 927: My only question is does this change to the query impact non-semver bundles in any way? Is that query behavior the same as before. Or are we saying in the latest versions of opm we expect the bundles to be built in semver mode?

benluddy: For replaces bundles, the existing query should work the same either way, since the replaces column in channel_entry should be synthesized from the bundle (https://github.com/operator-framework/operator-registry/blob/e144a5769a022a36f03130f8d2c1a7196a56a794/pkg/sqlite/load.go#L273). The existing query is only problematic for semver indices, since in that case the bundle and channel_entry "replaces" column aren't necessarily the same.

@kevinrizza Please check my understanding. I put a hold on this to wait for confirmation.

kevinrizza: You're right @benluddy, in both cases the channel_entry table is correct -- the source of truth should be the edges listed in the channel_entry table. The only diff is that in semver mode the replaces field is ignored when deciding what to add. In replaces mode that table gets rebuilt from scratch every time, so it should always be correct.

pkg/sqlite/query.go
```diff
@@ -924,19 +924,20 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
...
+channel_entry.operatorbundle_name, channel_entry.package_name, channel_entry.channel_name, replaced_entry.operatorbundle_name, operatorbundle.skips,
```
kevinrizza, line 928: This is definitely better, but in theory we might want to synthesize *all* of the channel edges directly from just the channel_entry table given that these input fields are disparate and may or may not be relevant when deciding what edges are actually real. That's the intention of the channel_entry table itself.

That being said, I think that's something we can worry about later and for now this is better.

Issue #482, opened 2020-10-19
TestListPackages is flaky


benluddy:
Observed in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/operator-framework_operator-registry/481/pull-ci-operator-framework-operator-registry-master-unit/1318308372407128064#1:build-log.txt%3A1492:

``` === RUN   TestListPackages
time="2020-10-19T21:55:10Z" level=info msg="loading Bundles" dir=../../manifests
time="2020-10-19T21:55:10Z" level=info msg=directory dir=../../manifests file=manifests load=bundles
time="2020-10-19T21:55:10Z" level=info msg=directory dir=../../manifests file=etcd load=bundles
time="2020-10-19T21:55:10Z" level=info msg=directory dir=../../manifests file=0.6.1 load=bundles
time="2020-10-19T21:55:10Z" level=info msg="found csv, loading bundle" dir=../../manifests file=etcdoperator.clusterserviceversion.yaml load=bundles
time="2020-10-19T21:55:10Z" level=info msg="loading bundle file" dir=../../manifests/etcd/0.6.1 file=etcdcluster.crd.yaml load=bundle name=etcdoperator.v0.6.1
time="2020-10-19T21:55:10Z" level=info msg="loading bundle file" dir=../../manifests/etcd/0.6.1 file=etcdoperator.clusterserviceversion.yaml load=bundle name=etcdoperator.v0.6.1
time="2020-10-19T21:55:12Z" level=info msg=directory dir=../../manifests file=0.9.2 load=bundles
time="2020-10-19T21:55:12Z" level=info msg="found csv, loading bundle" dir=../../manifests file=etcdoperator.v0.9.0.yaml load=bundles
time="2020-10-19T21:55:12Z" level=info msg="loading bundle file" dir=../../manifests/etcd/0.9.2 file=etcdbackup.crd.yaml load=bundle name=etcdoperator.v0.9.0
time="2020-10-19T21:55:12Z" level=info msg="loading bundle file" dir=../../manifests/etcd/0.9.2 file=etcdcluster.crd.yaml load=bundle name=etcdoperator.v0.9.0
time="2020-10-19T21:55:12Z" level=info msg="loading bundle file" dir=../../manifests/etcd/0.9.2 file=etcdoperator.v0.9.0.yaml load=bundle name=etcdoperator.v0.9.0
[...]

PR #485, opened 2020-10-20
Bug 1881527: fix(opm): properly deprecate export flag '-o'

njhale:
Add a hidden alias for the `-p` flag of opm index export. This properly
deprecates the `-o` shorthand, making the cli backwards compatible for
clients that haven't switched over to `-p`.

Note: there's not really a good native option for deprecating flag shorthands without defining a new flag, thus the given implementation.

### DIFF
--- a/cmd/opm/index/export.go
+++ b/cmd/opm/index/export.go
@@ -1,6 +1,8 @@
 package index
 
 import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


benluddy:
/test unit

https://github.com/operator-framework/operator-registry/issues/482

PR #486, opened 2020-10-20
Bug 1889721: Check prerelease version for semver-skippatch mode

ankitathomas:
Switch to using semver's compare to decide semver-skippatch candidates

### DIFF
--- a/pkg/registry/bundlegraphloader.go
+++ b/pkg/registry/bundlegraphloader.go
@@ -153,6 +153,9 @@ func (g *BundleGraphLoader) AddBundleToGraph(bundle *Bundle, graph *Package, ann
 	return graph, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/registry/bundlegraphloader.go
```diff
@@ -154,5 +154,5 @@ func (g *BundleGraphLoader) AddBundleToGraph(bundle *Bundle, graph *Package, ann
 }
 
 func isSkipPatchCandidate(version, toCompare semver.Version) bool {
```
estroz, line 156:
```suggestion
// isSkipPatchCandidate returns true if version is equal to toCompare
// in major and minor positions and strictly greater in all others,
// indicating that toCompare can be skipped over to version.
func isSkipPatchCandidate(version, toCompare semver.Version) bool {
```

PR #504, opened 2020-11-03
Bug 1889714: ensure replaces and skips field on exported bundles

ankitathomas:
Semver mode ignores replaces and skips field on csv if set, so package upgrade path is defined in the index.
Ensure these fields are exported based on index value to allow appregistry to correctly read the upgrade graph
from the exported nested bundle format

### DIFF
--- a/pkg/lib/indexer/indexer.go
+++ b/pkg/lib/indexer/indexer.go
@@ -5,23 +5,24 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"github.com/blang/semver"
-	yaml3 "github.com/ghodss/yaml"
-	"github.com/operator-framework/api/pkg/operators"
 	"io"
 	"io/ioutil"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"math/rand"
 	"os"
 	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 
+	"github.com/blang/semver"
+	yaml3 "github.com/ghodss/yaml"
+	"github.com/operator-framework/api/pkg/operators"
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 
@@ -539,8 +540,8 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
 
 	sem := make(chan struct{}, concurrencyLimitForExport)
 
-	for bundleImage, bundleDir := range bundles {
-		go func(bundleImage string, bundleDir bundleDirPrefix) {
+	for bundleDir, bundleInfo := range bundles {
+		go func(bundleDir bundleDirPrefix, bundleInfo bundleExportInfo) {
 			defer wg.Done()
 
 			sem <- struct{}{}
@@ -553,21 +554,21 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
 				bundleDir.bundleVersion = strconv.Itoa(rand.Intn(10000))
 			}
 			downloadPath := filepath.Join(request.DownloadPath, bundleDir.pkgName, bundleDir.bundleVersion)
-			exporter := bundle.NewExporterForBundle(bundleImage, downloadPath, request.ContainerTool)
+			exporter := bundle.NewExporterForBundle(bundleInfo.bundlePath, downloadPath, request.ContainerTool)
 			if err := exporter.Export(); err != nil {
-				err = fmt.Errorf("exporting bundle image:%s failed with %s", bundleImage, err)
+				err = fmt.Errorf("exporting bundle image:%s failed with %s", bundleInfo.bundlePath, err)
 				mu.Lock()
 				errs = append(errs, err)
 				mu.Unlock()
 				return
 			}
-			if err := ensureCSVFields(downloadPath, bundleDir); err != nil {
-				err = fmt.Errorf("exporting bundle image:%s : cannot ensure required fields on csv: %v", bundleImage, err)
+			if err := ensureCSVFields(downloadPath, bundleInfo); err != nil {
+				err = fmt.Errorf("exporting bundle image:%s : cannot ensure required fields on csv: %v", bundleInfo.bundlePath, err)
 				mu.Lock()
 				errs = append(errs, err)
 				mu.Unlock()
 			}
-		}(bundleImage, bundleDir)
+		}(bundleDir, bundleInfo)
 	}
 	// Wait for all the go routines to finish export
 	wg.Wait()
@@ -586,12 +587,18 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
 }
 
 type bundleDirPrefix struct {
-	pkgName, name, bundleVersion, replaces string
-	skips                                  map[string]struct{}
+	pkgName, bundleVersion string
 }
 
-func getBundlesToExport(dbQuerier pregistry.Query, packages []string) (map[string]bundleDirPrefix, error) {
-	bundleMap := make(map[string]bundleDirPrefix)
+type bundleExportInfo struct {
+	name       string
+	bundlePath string
+	replaces   string
+	skips      map[string]struct{}
+}
+
+func getBundlesToExport(dbQuerier pregistry.Query, packages []string) (map[bundleDirPrefix]bundleExportInfo, error) {
+	bundleMap := make(map[bundleDirPrefix]bundleExportInfo)
 	graphLoader := &sqlite.SQLGraphLoader{
 		Querier: dbQuerier,
 	}
@@ -604,23 +611,34 @@ func getBundlesToExport(dbQuerier pregistry.Query, packages []string) (map[strin
 		for _, channel := range bundleGraph.Channels {
 			var replaceDepth map[string]int64
 			for node, replaceNodes := range channel.Nodes {
+				bundleKey := bundleDirPrefix{
+					pkgName:       packageName,
+					bundleVersion: node.Version,
+				}
 				var replaces string
 				var skips map[string]struct{}
-				if bundleInfo, ok := bundleMap[node.BundlePath]; ok {
+				if bundleInfo, ok := bundleMap[bundleKey]; ok {
+					if bundleInfo.name != node.CsvName || bundleInfo.bundlePath != node.BundlePath {
+						return nil, fmt.Errorf("cannot export bundle version %s for package %s: multiple CSVs found", node.Version, packageName)
+					}
 					replaces = bundleInfo.replaces
 					skips = bundleInfo.skips
 				}
 				if skips == nil {
 					skips = map[string]struct{}{}
 				}
-				skips[replaces] = struct{}{}
+				if len(replaces) > 0 {
+					skips[replaces] = struct{}{}
+				}
 				for replace := range replaceNodes {
 					if replace.CsvName == replaces {
 						continue
 					}
+
+					skips[replace.CsvName] = struct{}{}
+
 					if _, ok := channel.Nodes[replace]; !ok {
 						//synthetic entry
-						skips[replace.CsvName] = struct{}{}
 						continue
 					}
 					if len(replaces) == 0 {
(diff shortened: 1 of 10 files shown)


pkg/lib/indexer/indexer.go
```diff
@@ -696,3 +776,124 @@ func (i ImageIndexer) DeprecateFromIndex(request DeprecateFromIndexRequest) erro
...
 	return nil
 }
+
+func ensureCSVFields(bundleDir string, bundleInfo bundleDirPrefix) error {
+	dirContent, err := ioutil.ReadDir(bundleDir)
+	if err != nil {
+		return fmt.Errorf("error reading bundle directory %s, %v", bundleDir, err)
+	}
+
+	var foundCSV bool
```
exdx, line 786: suggestion: instead of using this bool we could just break when we find a CSV? Then the rest of the code can be outside the for loop and potentially in its own function.

There shouldn't be a case where there is more than one CSV in a bundle that is to be exported from an index. It's possible if the user uses entirely their own tooling, but I'm not sure we need to worry about it here.

pkg/lib/indexer/indexer.go
```diff
@@ -696,3 +776,124 @@ func (i ImageIndexer) DeprecateFromIndex(request DeprecateFromIndexRequest) erro
...
+			modified = true
+		}
+
+		if modified {
+			unstructuredCSV, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&csv)
+			if err != nil {
+				return fmt.Errorf("unable to convert modified csv: %v", err)
+			}
+
+			// remove creationTimestamp: null
```
exdx, line 869: aren't these just plain yaml manifests in a bundle? why would there be a creationTimestamp in the metadata of the CSV if that field is set by the api-server

pkg/registry/csv.go
```diff
@@ -340,3 +340,50 @@ func (csv *ClusterServiceVersion) GetOperatorImages() (map[string]struct{}, erro
 
 	return images, nil
 }
+
+// setCSVField sets a top level field on the CSV spec
+func (csv *ClusterServiceVersion) setCSVSpecField(key string, newValue interface{}) (err error) {
+	var objmap map[string]*json.RawMessage
+	if err := json.Unmarshal(csv.Spec, &objmap); err != nil {
+		return err
```
exdx, line 348: nit: shadowing the return parameter with a new declaration inside the function

seems like not having a named error return type makes this function more clear, if not using a bare return

exdx (commented):
Great job! Just left some general feedback.

Definitely some heavy lifting for the sake of backwards compatibility

pkg/lib/indexer/indexer.go
```diff
@@ -570,19 +586,83 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
 		}
-		for k, _ := range bundlesForPackage {
-			bundleMap[k.BundlePath] = bundleDirPrefix{pkgName: packageName, bundleVersion: k.Version}
+		versionMap := make(map[string]string)
+		for _, channel := range bundleGraph.Channels {
+			var replaceDepth map[string]int64
+			for node, replaceNodes := range channel.Nodes {
+				var replaces string
+				var skips map[string]struct{}
+				if bundleInfo, ok := bundleMap[node.BundlePath]; ok {
```
benluddy, line 609: What if one bundle appears in two channels with different skips/replaces? I don't think bundlepath is a precise-enough identifier.

pkg/lib/indexer/indexer.go
```diff
@@ -570,19 +586,83 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+					if replacesErr == nil && newErr == nil && !replacesSemver.EQ(newSemver) {
+						if replacesSemver.LT(newSemver) {
+							replaces = replace.CsvName
+							versionMap[replaces] = replace.Version
+						} else {
+							skips[replace.CsvName] = struct{}{}
+						}
+						continue
+					}
+					return nil, fmt.Errorf("multiple replaces on package %s, bundle version %s: %s and %s", packageName, node.CsvName, replaces, replace.CsvName)
```
benluddy, line 656: It looks like no tests are covering this section (https://codecov.io/gh/operator-framework/operator-registry/pull/504/diff#D1-632).

benluddy, line 656: Is this a misleading error in the case when the error is the result of failing to parse a bundle version as a semver?

pkg/lib/indexer/indexer.go
```diff
@@ -570,19 +586,83 @@ func (i ImageIndexer) ExportFromIndex(request ExportFromIndexRequest) error {
...
+						} else {
+							skips[replace.CsvName] = struct{}{}
+						}
+						continue
+					}
+
+					replacesSemver, replacesErr := semver.Make(versionMap[replaces])
+					newSemver, newErr := semver.Make(replace.Version)
+					if replacesErr == nil && newErr == nil && !replacesSemver.EQ(newSemver) {
+						if replacesSemver.LT(newSemver) {
```
benluddy, line 648: Are these version-based assumptions valid for an index that was built in replaces mode?

pkg/lib/indexer/indexer_test.go
```diff
@@ -84,3 +85,71 @@ func TestGeneratePackageYaml(t *testing.T) {
...
+				},
+			},
+			csv:  `{"kind":"ClusterServiceVersion","metadata":{"name":"testoperator.v0.0.1"},"spec":{"installModes":[{"supported":true,"type":"OwnNamespace"}],"replaces":"v0.0.3","skips":["v0.0.0","v0.0.2"],"version":"1.0.0"}}`,
+			want: `{"kind":"ClusterServiceVersion","metadata":{"name":"testoperator.v0.0.1"},"spec":{"installModes":[{"supported":true,"type":"OwnNamespace"}],"replaces":"v0.0.3","skips":["v0.0.0","v0.0.1","v0.0.2"],"version":"1.0.0"}}`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f, err := os.OpenFile("testdata/testoperator/testoperator.csv.yaml", os.O_WRONLY|os.O_TRUNC, 0)
```
benluddy, line 136: If the function under test writes to files in the target directory, the test should set up a temporary file/directory to work under. As a reader, I'm going to expect anything in testdata to be a static test fixture. Alternatively, you could refactor the function under test so that it doesn't directly deal with files.

pkg/lib/indexer/indexer_test.go
```diff
@@ -84,3 +85,71 @@ func TestGeneratePackageYaml(t *testing.T) {
...
+			require.NoError(t, err)
+			_, err = f.Write([]byte(tt.csv))
+			require.NoError(t, err)
+			f.Close()
```
benluddy, line 140: Nit: This file descriptor will leak if any of the above assertions fail.

pkg/registry/csv.go
```diff
@@ -340,3 +340,50 @@ func (csv *ClusterServiceVersion) GetOperatorImages() (map[string]struct{}, erro
...
+	csv.Spec = updatedSpec
+
+	return nil
+}
+
+// GetSkips returns the name of the older ClusterServiceVersion objects that
+// are skipped by this ClusterServiceVersion object.
+//
+// If not defined, the function returns an empty string.
+func (csv *ClusterServiceVersion) SetSkips(newSkips []string) (err error) {
```
benluddy, line 375: Nit:

```suggestion
func (csv *ClusterServiceVersion) SetSkips(newSkips []string) error {
```

pkg/sqlite/query.go
```diff
@@ -1168,3 +1168,25 @@ func (s *SQLQuerier) GetBundlePathIfExists(ctx context.Context, bundleName strin
...
 	return
 }
+
+func (s *SQLQuerier) GetBundleReplacesDepth(ctx context.Context, packageName, bundleName string) (replaceChain map[string]int64, err error) {
+	query := `SELECT replaces.operatorbundle_name, min(channel_entry.depth)
+			FROM channel_entry
+			inner JOIN channel_entry replaces ON channel_entry.replaces = replaces.entry_id
+			inner JOIN operatorbundle replacesbundle ON replaces.operatorbundle_name = replacesbundle.name
+			WHERE channel_entry.package_name = ? AND channel_entry.operatorbundle_name = ?
+			GROUP BY replaces.operatorbundle_name;`
```
benluddy, line 1178: This query will be confused by bundles that appear in multiple channels with different skips/replaces.

pkg/lib/indexer/indexer.go
ankitathomas: We'd only get to this error if we have two real replace bundles with different csv names, at the same depth on the replaces chain (through multiple channels) which can't be ranked according to semver. opm should prevent this from happening, since it has a unique package+csv constraint.

One way of exporting a bundle with this ambiguity in the replaces would be to add the replaces/skips chain for the bundle's replacements to its own skips list, but that doesn't seem like expected behavior

pkg/lib/indexer/indexer.go
ankitathomas: [CreationTimestamp](https://github.com/operator-framework/operator-registry/blob/36dc231df42d611494a69fa5b4d1868ce06f2100/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L196) isn't a pointer field on the CSV struct, it gets set to a zero timestamp. This would be exported unless the field was removed.

We can also switch to https://github.com/clarketm/json for the export, that would omit all zero-valued fields.

pkg/registry/csv.go
njhale: in this case, the shadowed variable is scoped only to the if statement, but I agree with the point on clarity

pkg/lib/indexer/indexer.go
ankitathomas: For an index built in replaces mode, the depth for both replaces on the chain would be different.

ankitathomas: For an index built in replaces mode, the depth for both replaces on the chain would be different.

pkg/lib/indexer/indexer.go
```diff
@@ -696,3 +776,124 @@ func (i ImageIndexer) DeprecateFromIndex(request DeprecateFromIndexRequest) erro
...
+		defer csvFile.Close()
+		unstructuredCSV := unstructured.Unstructured{}
+
+		decoder := k8syaml.NewYAMLOrJSONDecoder(csvFile, 30)
+		if err = decoder.Decode(&unstructuredCSV); err != nil {
+			continue
+		}
+
+		if unstructuredCSV.GetKind() != operators.ClusterServiceVersionKind {
+			continue
```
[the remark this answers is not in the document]
ankitathomas: This does hold them open till the function exits. However, it should only be 2 files at most

pkg/sqlite/query.go
ankitathomas: This is used to aggregate all the valid replaces entries by the least depth in all channels. From what I can see, it looks like this depth is unique for bundles added in replaces mode. 

semver mode requires ordering by versions instead, if there is ambiguity with this ordering, the exporter uses semver for choosing a replaces bundle

pkg/lib/indexer/indexer.go
ankitathomas: We aggregate the skips/replaces across channels, since we can't have the same version twice in the nested bundle format.

pkg/lib/indexer/indexer.go
```diff
@@ -573,16 +590,96 @@ type bundleDirPrefix struct {
...
+							versionMap[replaces] = replace.Version
+						} else {
+							skips[replace.CsvName] = struct{}{}
+						}
+						continue
+					}
+
+					replacesSemver, replacesErr := semver.Make(versionMap[replaces])
+					newSemver, newErr := semver.Make(replace.Version)
+					if replacesErr == nil && newErr == nil && !replacesSemver.EQ(newSemver) {
```
exdx, line 664: this makes sense but is a bit confusing -- it combines the case where there is an issue with making the semver representation of the version and if the old/new semver representations do not match. it's also a bit repetitive, since if there is an error making a semver type we can almost be sure the types won't match. WDYT about handling the error case explicitly?

exdx (commented):
nice job with the tests!
/lgtm

pkg/lib/indexer/indexer.go
ankitathomas: That makes sense, it would make the error a lot clearer

dinhxuanvu (approved):
/approve

pkg/lib/indexer/indexer.go
```diff
@@ -696,3 +800,152 @@ func (i ImageIndexer) DeprecateFromIndex(request DeprecateFromIndexRequest) erro
...
+		}
+		csvReplaces, err := csv.GetReplaces()
+		if err != nil {
+			return err
+		}
+		csvSkips, err := csv.GetSkips()
+		if err != nil {
+			return err
+		}
+		sort.Strings(csvSkips)
```
kevinrizza, line 843: Why is this sort necessary? Shouldn't we be able to do this in any arbitrary order?

pkg/sqlite/query.go
kevinrizza: I don't believe this stands to reason. For bundles with synthetic channel entries (those that have skips set at multiple levels or those inserted with --skip-patch mode), there will be duplicates of depth. As far as I know, depth isn't used to calculate the importance or truth of any of the replacements at all.

kevinrizza (commented):
A few small comments. This looks really close!

pkg/lib/indexer/indexer.go
```diff
@@ -573,16 +590,103 @@ type bundleDirPrefix struct {
...
+						if replaceDepth[replaces] > replaceDepth[replace.CsvName] {
+							replaces = replace.CsvName
+							versionMap[replaces] = replace.Version
+						}
+						continue
+					}
+
+					// 2 replaces versions at the same depth on replaces chains of different channels
+					// Choose the newer semver as the replaces
+					replacesSemver, err := semver.Make(versionMap[replaces])
```
benluddy, line 664: As far as I know, there shouldn't be any ambiguity in the edges stored in the index that needs to be resolved by something like version comparison.

pkg/lib/indexer/indexer.go
```diff
@@ -696,3 +800,152 @@ func (i ImageIndexer) DeprecateFromIndex(request DeprecateFromIndexRequest) erro
...
+		}
+
+		return nil
+	}
+	return fmt.Errorf("no ClusterServiceVersion object found in %s", bundleDir)
+}
+
+// getSkipsReplaces produces a merged skips/replaces for a bundle, using contents of its csv
+// and the upgrade graph in the index.
+func getSkipsReplaces(csvInfo, indexInfo bundleExportInfo) ([]string, string) {
```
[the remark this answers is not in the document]
kevinrizza: The only time you would want to consider the old csv's values is if they have something set but they should not be. It seems like it's worth parsing them all just to confirm that's the case, but we should always just assume that the index is the source of truth and completely overwrite the existing csvs.

ankitathomas:
This isn't an issue with the release of 4.8. Closing this out

PR #505, opened 2020-11-03
Bug 1889721: add skippatch unit test

ankitathomas:
Adding unit tests for isSkipPatchCandidate

### DIFF
--- a/pkg/registry/bundlegraphloader_test.go
+++ b/pkg/registry/bundlegraphloader_test.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"encoding/json"
+	"github.com/blang/semver"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


pkg/registry/bundlegraphloader_test.go
```diff
@@ -312,3 +313,71 @@ func TestBundleGraphLoader(t *testing.T) {
...
+			expected: false,
+		},
+		{
+			name: "accept larger patch version",
+			added: "0.0.1",
+			compare: "0.0.0",
+			expected: true,
+		},
+		{
+			name: "accept patch version without pre-release",
```
kuiwang02, line 349: @ankitathomas if it is the following data,
```console
{
			name: "do not accept patch version with pre-release",
			added: "0.0.0-1",
			compare: "0.0.0",
			expected: false,
		},
```
is it correct? I think so. thanks

pkg/registry/bundlegraphloader_test.go
```diff
@@ -312,3 +313,71 @@ func TestBundleGraphLoader(t *testing.T) {
...
+			added: "0.0.0",
+			compare: "0.0.0",
+			expected: false,
+		},
+		{
+			name: "do not accept different major/minor version",
+			added: "0.0.0",
+			compare: "0.0.0",
+			expected: false,
+		},
```
benluddy, line 335: It looks like you meant to change the inputs for this case?

pkg/registry/bundlegraphloader_test.go
```diff
@@ -312,3 +313,71 @@ func TestBundleGraphLoader(t *testing.T) {
...
+			expected: true,
+		},
+		{
+			name: "accept lexicographically larger pre-release",
+			added: "0.0.1-11",
+			compare: "0.0.1-2",
+			expected: true,
+		},
+		{
+			name: "accept lexicographically larger pre-release",
```
benluddy, line 367: NIt: It might be nice to give this subtest a distinct name to play nicely with `-run` and have clearer test run outputs.

pkg/registry/bundlegraphloader_test.go
benluddy: @ankitathomas To add on to this, if `f(a, c) -> !f(c, a)` and `!f(a, c) -> f(c, a)`, then this test could automatically verify both for each case.

ankitathomas: I've added checking for the absence of a commutative edge

pkg/registry/bundlegraphloader_test.go
```diff
@@ -312,3 +313,73 @@ func TestBundleGraphLoader(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			added, err := semver.Make(tt.added)
+			require.NoError(t, err)
+			compare, err := semver.Make(tt.compare)
+			require.NoError(t, err)
+			actual := isSkipPatchCandidate(added, compare)
+			assert.Equal(t, tt.expected, actual)
+
+			if !tt.commutative {
```
benluddy, line 379: Now that I see it (and recognize that there are cases that don't have this property), I'm not sure that adding a second assertion to the test is better than explicitly writing these cases. I won't ask you to change it again if you disagree, though.

benluddy (commented):
/approve

pkg/registry/bundlegraphloader_test.go
ankitathomas: We can make this assertion whenever we have a valid skippatch pair. `f(a, c) -> !f(c, a)` holds true, so this should cover all the cases we need

PR #509, opened 2020-11-04
copy .git directory for build tag info

### LINKED ISSUE #507: upstream-opm-builder image does not populate the version in opm.
When a release is cut for operator-registry, a version of `opm` is added to that release, and the the builder image is updated in quay:  https://quay.io/repository/operator-framework/upstream-opm-builder with a matching tag for the release.

Unfortunately, the version of `opm` that's embedded in this image is not exactly the same image that's in the release and running `opm version` on the version within the upstream-opm-builder image doesn't show ANY version (it's an empty string).

It's therefore difficult to verify that the opm version is the correct version other than trusting the tag.

Ideally, the version of opm would be IDENTICAL to the linux-amd64 version in the release tgz.  
Minimally, the version and build should MATCH the version in the release tag.

ankitathomas:
Copy .git directory to intermediary build container to provide make with the tag and sha for an opm build.

Closes #507

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,10 +5,7 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 
 WORKDIR /src
 
-COPY vendor vendor
-COPY cmd cmd
-COPY pkg pkg
-COPY Makefile go.mod go.sum ./
+COPY . .
 RUN make build cross
 
 # copy and build vendored grpc_health_probe
--- a/appr-registry.Dockerfile
+++ b/appr-registry.Dockerfile
@@ -3,11 +3,7 @@ FROM golang:1.13-alpine as builder
 RUN apk update && apk add sqlite build-base git mercurial bash
 WORKDIR /go/src/github.com/operator-framework/operator-registry
 
-COPY vendor vendor
-COPY cmd cmd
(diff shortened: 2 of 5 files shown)


benluddy:
Is there any reason not to copy the entire build context into the builder stage rather than enumerating files/directories like this?

ankitathomas:
> Is there any reason not to copy the entire build context into the builder stage rather than enumerating files/directories like this?

If the build context is clean, then we could just copy the entire build context.

njhale (approved):
/approve

Dockerfile
```diff
@@ -5,10 +5,7 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 
 WORKDIR /src
 
-COPY vendor vendor
-COPY cmd cmd
-COPY pkg pkg
-COPY Makefile go.mod go.sum ./
+COPY . .
```
dinhxuanvu, line 8: Is this COPY all a bit excessive?

dinhxuanvu (approved):
/lgtm

Dockerfile
ankitathomas: It's only to the intermediary container, our final image size won't change.

PR #512, opened 2020-11-05
Bug 1889388: Reconstruct replaces and skips for ListBundles from channel_entry.

benluddy:
The channel_entry table is authoritative on incoming upgrade edges to
a bundle, so the "skips" and "replaces" fields in bundles returned by
the ListBundles endpoint should be derived from channel_entry rather
than from operatorbundle.

### DIFF
--- a/pkg/sqlite/query.go
+++ b/pkg/sqlite/query.go
@@ -924,17 +924,80 @@ func (s *SQLQuerier) GetCurrentCSVNameForChannel(ctx context.Context, pkgName, c
 	return "", nil
 }
 
-const listBundlesQuery = `SELECT DISTINCT channel_entry.entry_id, operatorbundle.bundle, operatorbundle.bundlepath,
-channel_entry.operatorbundle_name, channel_entry.package_name, channel_entry.channel_name, replaced_entry.operatorbundle_name, operatorbundle.skips,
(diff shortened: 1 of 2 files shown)


exdx:
Took a deep dive into this PR, and ran `opm registry serve` against a local copy of the 4.6 prod index DB with both the new query and the old query for amq-streams. I then compared the results via 
```
grpcurl -plaintext  localhost:50051 api.Registry/ListBundles | jq -c "{csvName,packageName,channelName,replaces,skips}" | grep amq-streams
```
https://gist.github.com/exdx/491587bb06864c9bec750597f0673f0e
The results look consistent. 

One thing we noticed is that the skips field is being returned as `null` with this patch versus `[""]` in master. This was fixed in 4.5 but apparently not in master -- this PR incorporates this fix additionally. This is due to the skips and replaces field in `operatorbundle` being set as empty text versus null. The resolver handles both cases the same but this is an additional bonus in this PR.

ankitathomas (commented):
/lgtm

dinhxuanvu (approved):
/lgtm

exdx:
myself and @benluddy were still looking into this. not sure if we want to merge as is. will hold for now, feel free to cancel

/hold

PR #503, opened 2020-11-03
Bug 1885425: fix(indexing): order bulk add by version field

njhale:
When adding more than one bundle to an index in replaces mode, sort the input by order of ascending version field (semver). This ensures a deterministic update graph in the absence of a replaces field.

The replaces field takes priority.

PR #517, opened 2020-11-13
Bug 1868125: [opm] Add valid bundles to index when `permissive` enabled

anik120:

When the `opm index add` was used with the `permissive` flag, the index
was being created, but the valid bundles from the list of bundles provided
with the command was not being added. Instead, an empty index was being
created. This PR fixes the issue by adding the valid bundles to the index,
while ignoring the invalid ones.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -29,9 +29,10 @@ type DirectoryPopulator struct {
 	imageDirMap     map[image.Reference]string
 	overwriteDirMap map[string]map[image.Reference]string
 	overwrite       bool
+	permissive      bool
 }
 
-func NewDirectoryPopulator(loader Load, graphLoader GraphLoader, querier Query, imageDirMap map[image.Reference]string, overwriteDirMap map[string]map[image.Reference]string, overwrite bool) *DirectoryPopulator {
+func NewDirectoryPopulator(loader Load, graphLoader GraphLoader, querier Query, imageDirMap map[image.Reference]string, overwriteDirMap map[string]map[image.Reference]string, overwrite, permissive bool) *DirectoryPopulator {
 	return &DirectoryPopulator{
 		loader:          loader,
 		graphLoader:     graphLoader,
@@ -39,6 +40,7 @@ func NewDirectoryPopulator(loader Load, graphLoader GraphLoader, querier Query, 
 		imageDirMap:     imageDirMap,
 		overwriteDirMap: overwriteDirMap,
 		overwrite:       overwrite,
+		permissive:      permissive,
 	}
 }
 
@@ -154,7 +156,7 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, imagesToRe
 	if err != nil {
 		return err
 	}
-
+	var errs []error
 	switch mode {
 	case ReplacesMode:
 		// TODO: This is relatively inefficient. Ideally, we should be able to use a replaces
@@ -165,7 +167,6 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, imagesToRe
 		// Additionally, it would be preferrable if there was a single database transaction
 		// that took the updated graph as a whole as input, rather than inserting bundles of the
 		// same package linearly.
-		var err error
 		var validImagesToAdd []*ImageInput
 
 		for pkg := range i.overwriteDirMap {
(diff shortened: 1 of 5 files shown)


pkg/registry/populator.go
```diff
@@ -29,16 +29,18 @@ type DirectoryPopulator struct {
 	imageDirMap     map[image.Reference]string
 	overwriteDirMap map[string]map[image.Reference]string
 	overwrite       bool
+	permissive      bool
 }
 
-func NewDirectoryPopulator(loader Load, graphLoader GraphLoader, querier Query, imageDirMap map[image.Reference]string, overwriteDirMap map[string]map[image.Reference]string, overwrite bool) *DirectoryPopulator {
+func NewDirectoryPopulator(loader Load, graphLoader GraphLoader, querier Query, imageDirMap map[image.Reference]string, overwriteDirMap map[string]map[image.Reference]string, overwrite, permissive bool) *DirectoryPopulator {
```
njhale, line 35: I don't think we need to propagate this down so long as we always aggregate errors and continue instead of exiting immediately.

njhale (changes requested):
One comment on the design.

Can we also add a test to make sure this works as expected?

pkg/registry/populator.go
anik120: @njhale I'm trying to think why we might prefer to aggregate errors instead of propagating it all the way down. 
The reason I originally decided to take it all the way down is coz

a) If we have introduced a concept of `permissive` for `opm` it seems natural to have the same concept for our `DirectoryPopulator` too. Basically if some other implementation instead of `opm` want to use `DirectoryPopulator` it should be aware/have access to `permissive` too.

b) If you scale this to n bundles where n is large, it seems more efficient to fail fast if `permissive` isn't enabled instead of churning all the way through.

pkg/registry/populator.go
```diff
@@ -284,14 +286,17 @@ func (i *DirectoryPopulator) getNextReplacesImagesToAdd(imagesToAdd []*ImageInpu
 				pkgFoundImages++
 				foundImages = append(foundImages, pkgImage)
 			} else {
-				pkgRemainingImages++
-				remainingImages = append(remainingImages, pkgImage)
+				if !i.permissive {
+					pkgRemainingImages++
+					remainingImages = append(remainingImages, pkgImage)
```
estroz, line 291: Are `remainingImages` not used by callers who want permissive population?

pkg/registry/populator.go
estroz: It seems like the error-permissive concept would benefit from a non-permissive error type, since there are error instances that shouldn't be permitted in any case.

```go
type ErrNonPermissive struct {
	err error
}

func foo(l string) error {
	if err = nonPermissiveErrorReturner(l); err != nil {
		return ErrNonPermissive{err}
	}
	if err = permissiveErrorReturner(l); err != nil {
		return err
	}
	return nil
}

func bar(list []string, permissive bool) error {
	var errs []error
	for _, l := range list {
		err := foo(l)
		if err != nil {
			if errors.As(err, &ErrNonPermissive{}) || !permissive {
				return err
			}
			errs = append(errs, err)
		}
	}
}
```

Then you wouldn't have to propagate `permissive` down as far, and you could have any procedure that returns an error "know" about permissiveness without implementing (too much) of the handling logic itself

njhale: > If you scale this to n bundles where n is large, it seems more efficient to fail fast if permissive isn't enabled instead of churning all the way through.

fair enough, but these function signatures are already unwieldy as they are today. For the most part, I think we could benefit from not copying all this data around by making `populate` a method of `DirectoryPopulator` (unless it's used by more than one type of populator).

I'm also of the opinion that a lot of this code will get tossed soon anyway, so it may not be super important to keep things pretty now, but we should definitely think about it for the next iteration of operator-registry.

pkg/registry/populator.go
anik120: @estroz `remainingImages` are the `pkgImage` s that are not "valid bundles" and could not be added to [`foundImages`](https://github.com/operator-framework/operator-registry/blob/0778a098b6827e60812eeecaf36d4eba25aa6587/pkg/registry/populator.go#L304). The reason we need to not add "invalid bundles" to `remainingImages` when permissive mode is enabled is to make sure that we don't [loop infinitely because `imagesToAdd`](https://github.com/operator-framework/operator-registry/blob/0778a098b6827e60812eeecaf36d4eba25aa6587/pkg/registry/populator.go#L185-L186) always has those "invalid images" in them

anik120:
#503 and #522 fixes the issue.

PR #522, opened 2020-11-18
fix(opm-index-add) Throw error when a bundle is invalid

anik120:

With the introduction of #503, when a bundle from the list of bundles
being added with the `opm index add` command was invalid, opm was not
throwing an error, and was building the index with the valid bundles
instead, even when the `permissive` mode was not enabled. This PR
fixes the issue.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/input_stream.go
+++ b/pkg/registry/input_stream.go
@@ -142,7 +142,7 @@ func (r *ReplacesInputStream) Next() (*ImageInput, error) {
 				delete(r.packages, pkg)
 			}
 
-			return image, nil
+			return image, utilerrors.NewAggregate(errs)
 		}
 
 		// No viable bundle found in the package, can't parse it any further


pkg/registry/input_stream.go
```diff
@@ -142,7 +142,7 @@ func (r *ReplacesInputStream) Next() (*ImageInput, error) {
 				delete(r.packages, pkg)
 			}
 
-			return image, nil
+			return image, utilerrors.NewAggregate(errs)
```
dinhxuanvu, line 145: Should we add `packageErrs` into the `errs` before return?

dinhxuanvu (commented):
Looking good but just a quick comment.

pkg/registry/input_stream.go
njhale, line 145: this fix looks incorrect
the way this should work is, as long as we find a bundle that works for at least one package, don't return an error 
eventually next will have no bundles, or bundles that cause errors on add

njhale (changes requested):

pkg/registry/input_stream.go
njhale: I've opened #523, which I believe fixes the regression. It also adds a test that fails before/succeeds after the change.

@anik120 PTAL, and let me know if you agree with the fix.

anik120:
Closing in favor of #523

PR #523, opened 2020-11-19
fix(indexing): respect strict mode

njhale:
Keep packages in the input stream regardless of errors attempting to add
bundles of that package to an index. This prevents invalid bundles from
being elided before all valid bundles have been exhausted, and fixes a
regression that effectively broke strict (non-permissive) mode.

Supersedes #522

### DIFF
--- a/pkg/registry/input_stream.go
+++ b/pkg/registry/input_stream.go
@@ -90,7 +90,7 @@ func (r *ReplacesInputStream) canAdd(bundle *Bundle, packageGraph *Package) erro
 
 	if replaces != "" && !packageGraph.HasCsv(replaces) {
 		// We can't add this until a replacement exists
-		return fmt.Errorf("Invalid bundle %s, bundle specifies a non-existent replacement %s", bundle.Name, replaces)
+		return fmt.Errorf("Invalid bundle %s, replaces nonexistent bundle %s", bundle.Name, replaces)
 	}
 
 	images, ok := r.packages[packageGraph.Name]
@@ -145,11 +145,8 @@ func (r *ReplacesInputStream) Next() (*ImageInput, error) {
 			return image, nil
 		}
 
-		// No viable bundle found in the package, can't parse it any further
-		if len(packageErrs) > 0 {
-			delete(r.packages, pkg)
-			errs = append(errs, packageErrs...)
-		}
+		// No viable bundle found in the package, can't parse it any further, so return any errors
+		errs = append(errs, packageErrs...)
 	}
 
 	// We've exhausted all valid input bundles, any errors here indicate invalid input of some kind
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -196,7 +196,6 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, imagesToRe
 				errs = append(errs, err)
 				break
 			}
-
 		}
 
 		if len(errs) > 0 {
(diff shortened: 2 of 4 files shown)


pkg/registry/input_stream.go
```diff
@@ -90,7 +90,7 @@ func (r *ReplacesInputStream) canAdd(bundle *Bundle, packageGraph *Package) erro
 
 	if replaces != "" && !packageGraph.HasCsv(replaces) {
 		// We can't add this until a replacement exists
-		return fmt.Errorf("Invalid bundle %s, bundle specifies a non-existent replacement %s", bundle.Name, replaces)
+		return fmt.Errorf("Invalid bundle %s, replaces nonexistent bundle %s", bundle.Name, replaces)
```
anik120, line 93: This feels more like a personal choice and unrelated to the fix the PR is trying to address.

pkg/registry/input_stream.go
```diff
@@ -147,7 +147,6 @@ func (r *ReplacesInputStream) Next() (*ImageInput, error) {
 
 		// No viable bundle found in the package, can't parse it any further
 		if len(packageErrs) > 0 {
```
anik120, line 149: Do we need this check anymore? Cant we just append the errors regardless?

pkg/registry/populator.go
```diff
@@ -196,7 +196,6 @@ func (i *DirectoryPopulator) loadManifests(imagesToAdd []*ImageInput, imagesToRe
 				errs = append(errs, err)
 				break
 			}
-
```
anik120, line 199: nit: Seems like some noise in a file not being touched by this PR.

anik120 (changes requested):
@njhale looking good, had a couple of comments and a nit

pkg/registry/input_stream.go
njhale: sure, but the existing wording doesn't make sense.

pkg/registry/input_stream.go
njhale: Yup, looks like that would work for the edge cases: https://play.golang.org/p/_SNa9QLWEhQ

pkg/registry/populator.go
njhale: I did this on purpose, didn't like the extra space; I'm pretty sure I introduced this in the previous patch too.

pkg/registry/input_stream.go
```diff
@@ -145,11 +145,8 @@ func (r *ReplacesInputStream) Next() (*ImageInput, error) {
 			return image, nil
 		}
 
-		// No viable bundle found in the package, can't parse it any further
-		if len(packageErrs) > 0 {
-			delete(r.packages, pkg)
-			errs = append(errs, packageErrs...)
-		}
+		// No viable bundle found in the package, can't parse it any further, so return any errors
```
anik120, line 148: nit: Is this comment still relevant?

pkg/registry/input_stream.go
anik120: If we feel that was we can probably open another housekeeping PR that's titled `fix log message`. I'm just not sure if changing log message goes with `respect strict mode`

pkg/registry/populator.go
anik120: I'd put forth the same argument as [this](https://github.com/operator-framework/operator-registry/pull/523#discussion_r527022248) comment, but if you feel strongly about it anyway it's a small PR so I guess it's not worth delaying this PR because of this.

pkg/registry/input_stream.go
njhale: I feel strongly that we shouldn't tax CI for trivial changes. I think the k8s docs also suggest that trival changes should go into larger PRs for that project.

pkg/registry/input_stream.go
njhale: yes

PR #528, opened 2020-12-02
Bug 1902824: fix(index): account for default channel in index add order

njhale:
Ensure the default channel referenced by a bundle exists, either by that
bundle or in the package, before returning it as the next element of an
input stream. Adding this check prevents smaller versions from breaking
index add when the `--overwrite-latest` flag is set; e.g. if a z-stream
bundle was originally added _after_ a larger y-stream and references a
default channel only defined in that y-stream.


gallettilance (approved):
/lgtm

PR #536, opened 2020-12-09
chore: move upstream tests from travisCI

exdx:
Move upstream tests to Github Actions and remove TravisCI configuration entirely. 

Upstream actions were added as well as an update to the release docs. The existing release CI was not changed but the `unit` job was consolidated as part of the upstream tests. The only functional change was `codecov` reported errors did not result in a CI failure. 

Prerequisite for getting the opm e2e tests to run as expected again 

Travis changed their policy around supporting open source repositories 

**Reviewer Checklist**


exdx:
The upstream e2e tests now match the e2e results locally - all but one test is skipped and the remaining test fails. This will be addressed in follow-up PRs but I believe this PR fulfills its original goal of moving upstream CI to Github Actions

.github/workflows/upstreams-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - run: make unit
+      - run: sed -i'' "s:^github.com/$GITHUB_REPOSITORY/::" coverage.out
+      - uses: codecov/codecov-action@v1
+        with:
+          file: coverage.out
+          fail_ci_if_error: false
+  run-e2e-minikube:
+    runs-on: ubuntu-latest
```
ankitathomas, line 27: we should add e2e tests for macos-latest and windows-latest too

ankitathomas (approved):
/lgtm

.github/workflows/upstreams-tests.yml
exdx: that's definitely a good idea especially for `opm`
but I think first we should work on getting the e2e tests to pass 😄 
then add additional os's as a followup

exdx:
This PR will be merged manually since there is no accompanying BZ and it only introduces net-new CI. The e2e CI has now been moved to GitHub Actions as intended -- the fact that kind CI is red is intentional and does not block any registry PRs (these Actions are not required for prow to merge). The failing CI should bring attention to the fact that the registry e2e tests are not working as intended, 10/11 tests are being skipped as they were on TravisCI. 

Subsequent work will be done to address the skipped e2e tests and the issue that causes the `Launch bundle` test to fail on kind -- since this test builds images on the host and then uses them within the kind cluster there is no way to share these images without an overhaul of the test.

PR #540, opened 2020-12-18
WIP (e2e) regression check with existing catalogs

gallettilance:

Integration tests with the Red Hat Operators. Ensure that no regression is present in OPM from the perspective of:

1. adding new bundles to indexes
2. overwriting the latest bundles
3. batch adding bundles to an index

All the above should produce graphs that are consistent with the graphs that have been published so far.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/opm_test.go
+++ b/test/e2e/opm_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"reflect"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -23,6 +24,8 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/lib/indexer"
 	"github.com/operator-framework/operator-registry/pkg/registry"
 	"github.com/operator-framework/operator-registry/pkg/sqlite"
+
+	lregistry "github.com/operator-framework/operator-registry/pkg/lib/registry"
 )
 
 var (
@@ -45,6 +48,9 @@ var (
 	indexImage1 = "quay.io/olmtest/e2e-index:" + indexTag1
 	indexImage2 = "quay.io/olmtest/e2e-index:" + indexTag2
 	indexImage3 = "quay.io/olmtest/e2e-index:" + indexTag3
+
+	ocpVersion      = "v4.6"
+	redhatOperators = "registry.redhat.io/redhat/redhat-operator-index:" + ocpVersion
 )
 
 type bundleLocation struct {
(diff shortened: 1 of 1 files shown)


test/e2e/opm_test.go
```diff
@@ -45,6 +48,9 @@ var (
 	indexImage1 = "quay.io/olmtest/e2e-index:" + indexTag1
 	indexImage2 = "quay.io/olmtest/e2e-index:" + indexTag2
 	indexImage3 = "quay.io/olmtest/e2e-index:" + indexTag3
+
+	ocpVersion      = "v4.6"
```
dinhxuanvu, line 52: Can you explain why we hard code in OCP version here? Are we gonna update this version for every release?
Given master is 4.7 so we use 4.6 here to essentially get 4.6 and check if it is compatible with 4.7?

dinhxuanvu (commented):
Just a comment.
Can you have a commit message to explain what you are trying to accomplish with this PR?
Thanks.

test/e2e/opm_test.go
```diff
@@ -425,6 +431,91 @@ var _ = Describe("opm", func() {
 			err = buildIndexWith(containerTool, indexImage, nextIndex, bundles[4:].images(), registry.ReplacesMode, true) // 1.0.1-overwrite
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		It("is consistent with existing update graphs", func() {
```
exdx, line 435: This looks good. I'm wondering if this can take the place of #463 a PR I had open for a while that attempts to do something similar but didn't really go anywhere

test/e2e/opm_test.go
```diff
@@ -45,6 +48,9 @@ var (
 	indexImage1 = "quay.io/olmtest/e2e-index:" + indexTag1
 	indexImage2 = "quay.io/olmtest/e2e-index:" + indexTag2
 	indexImage3 = "quay.io/olmtest/e2e-index:" + indexTag3
+
+	ocpVersion      = "v4.6"
+	redhatOperators = "registry.redhat.io/redhat/redhat-operator-index:" + ocpVersion
```
exdx, line 53: This is a private image -- I think we will run into the issue where we cannot pass in the registry credentials on forks of operator-registry -- meaning this test will only be able to run on merges to master

gallettilance: the image is public because it's the one used by all OCP clusters - right?

test/e2e/opm_test.go
gallettilance: oh cool I didn't realize you had started that! I'll take a look to see what I can steal from there :) thanks!

test/e2e/opm_test.go
gallettilance: oh that's because I was lazy. I need to figure out how we run it for all supported OCP versions. Not sure how not to hardcode that if you have any suggestions

test/e2e/opm_test.go
exdx: I think you still need auth to access `registry.redhat.io` images -- OCP clusters have these credentials baked in by default somewhere I suppose. 

For example, pulling the index image from a minikube node (with its own docker daemon) results in a 401 unauthorized:
```
docker@minikube:~$ docker pull registry.redhat.io/redhat/redhat-operator-index:v4.6
Error response from daemon: Get https://registry.redhat.io/v2/redhat/redhat-operator-index/manifests/v4.6: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication
```

test/e2e/opm_test.go
exdx: That's a good idea -- I think this PR goes one step further by actually running integration tests with the index, my original PR just checks to see that `opm registry serve <index>` works. I think closing my original PR when this one merges makes sense.

exdx:
Closes #463

test/e2e/opm_test.go
exdx: That being said, checking in registry credentials to github actions and only having this test run on commits to master is still valuable

dinhxuanvu:
Was included in #571

PR #544, opened 2021-01-05
Inherit replaced versions when skipping patches

jeloba:

Inherit the replaced versions from deleted versions when skipping patches with `semver-skippatch`. This ensures that it is possible to build a complete "replaces" path to previous versions, that are not skipped/removed from the channel.

In addition to this, skipped version from previous skips, but on other minor versions are trimmed if they where skipped. These are "synthectic" versions and not part of the "single upgrade path". They will be recreated later as synthetic versions in `sqlLoader.AddPackageChannelsFromGraph`.

When adding a series of bundles with versions `0.0.1`, `0.1.0`, `0.1.1` with

```shell
opm index add --mode semver-skippatch ...
```

The following error will be thrown:

> Invalid graph: some (non-bottom) nodes defined in the graph were not mentioned as replacements of any node

This fixes the error by enabling `sqlLoader.AddPackageChannelsFromGraph` to build a complete graph from the channel head to the oldest version in the channel.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/bundlegraphloader.go
+++ b/pkg/registry/bundlegraphloader.go
@@ -130,6 +130,12 @@ func (g *BundleGraphLoader) AddBundleToGraph(bundle *Bundle, graph *Package, ann
 		// the new node. If we didn't find a node semantically ahead, the new node is
 		// the new channel head
 		if !lowestAhead.IsEmpty() {
+
+			// Inherit replacements form the version replacing the new node.
+			for replaced, val := range channelGraph.Nodes[lowestAhead] {
+				replaces[replaced] = val
+			}
+
 			channelGraph.Nodes[lowestAhead] = map[BundleKey]struct{}{
 				newBundleKey: struct{}{},
 			}
--- a/pkg/registry/bundlegraphloader.go
+++ b/pkg/registry/bundlegraphloader.go
@@ -87,6 +87,19 @@ func (g *BundleGraphLoader) AddBundleToGraph(bundle *Bundle, graph *Package, ann
 					node.CsvName, node.Version, node.BundlePath)
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/registry/bundlegraphloader.go
```diff
@@ -150,6 +153,37 @@ func (g *BundleGraphLoader) AddBundleToGraph(bundle *Bundle, graph *Package, ann
...
+					}
+				}
+			}
+
+			for _, node := range allSkips {
+				delete(channelGraph.Nodes, node)
+			}
+		}
+	}
+
```
timflannagan, line 186: Super minor nit, but I think we can just reverse the conditional check and exit early, which will allow us to deflate the block iterating over the channels in the graph:

```go
if !skippath {
    return graph, nil
}

for channel := range graph.Channels {
    ...
}

return graph, nil
```

jeloba:
@timflannagan anything else I can do here to get this merged? I have no idea why the coverage check turned red after the logic adjustment... :confused:

jeloba:
@jmrodri @kevinrizza is there anything else I can do to get this reviewed?

jmrodri:
@jeloba I'll look at it tomorrow and get someone from olm to look as well. Thanks for the heads up sorry for the delay.

ankitathomas:
/hold

Inferring edges does sound useful, but this is a behavior change that can leave dropped/unexpected edges. 
`skipped version from previous skips, but on other minor versions are trimmed if they where skipped` seems like it would leave anyone on a previously skipped version stranded, even if the skipped version is explicitly given in the csv. 

v1.18.1+ also has support for file based catalogs with separate channel objects, which should give a lot more flexibility to define upgrade graphs better.

jeloba:
I'm closing this in favor of the new file based catalog definitions.

PR #545, opened 2021-01-07
Check if package exists before removing from registry

timflannagan:

Before, when attempting to delete a package from an existing registry,
no error was logged when that package did not exist.

Update pkg/sqlite/load.go and check if the package we're currently
processing exists in the registry before removing it.
```bash
make bin/opm
./bin/opm index add -c podman --bundles quay.io/operator-framework/operator-bundle-prometheus:0.14.0,quay.io/operator-framework/operator-bundle-prometheus:0.15.0,quay.io/operator-framework/operator-bundle-prometheus:0.22.2,quay.io/tflannag/bundle:v4.6,quay.io/tflannag/bundle:v4.7 --tag quay.io/tflannag/index:latest
...
$ podman create quay.io/tflannag/index:latest
5bbe07c6785da4885c59d739c8ab234994112df4dd46abd7e81fb3abb5828735
$ podman cp 5bbe07c6785da4885c59d739c8ab234994112df4dd46abd7e81fb3abb5828735:/database/index.db test.db
```

The test.db contains several channels for the `prometheus` and
`metering-ocp` packages:

```bash
sqlite> select * from channel;
name    package_name  head_operatorbundle_name
------  ------------  -------------------------
beta    prometheus    prometheusoperator.0.22.2
stable  prometheus    prometheusoperator.0.22.2
4.6     metering-ocp  metering-operator.v4.6.0
4.7     metering-ocp  metering-operator.v4.7.0
sqlite> select * from package;
name          default_channel
------------  ---------------
prometheus    stable
metering-ocp  4.7
sqlite>
```

This produces the following error when attempting to remove the
`prometheus2` package from the empty sqlite registry:

```bash
$ ./bin/opm registry rm -d test.db -o prometheus2
INFO[0000] removing from the registry                    packages="[prometheus2]"
INFO[0000] deleting packages                             pkg=prometheus2
INFO[0000] input has been sanitized                      pkg=prometheus2
INFO[0000] packages: [prometheus2]                       pkg=prometheus2
FATA[0000] permissive mode disabled                      error="error deleting packages from database: error removing operator package prometheus2: package does not exist"
```

Open Questions:
- Should this be a warning message instead?
- Does this need to be backported?

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	_ "github.com/mattn/go-sqlite3"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -612,12 +613,47 @@ func (s *sqlLoader) addAPIs(tx *sql.Tx, bundle *registry.Bundle) error {
 	return nil
 }
 
+// checkIfPackageExists is a method responsible for determing
+// whether the @packageName package exists in the package
+// table and returns an error if that package is non-existent.
+func (s *sqlLoader) checkIfPackageExists(tx *sql.Tx, packageName string) error {
+	id, err := tx.Prepare(`
+	SELECT DISTINCT name
+	FROM package
+	WHERE name=?
+	`)
+	if err != nil {
+		return err
+	}
+	defer id.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+
+	r, err := id.QueryContext(ctx, packageName)
+	if err != nil {
+		return err
+	}
+	if !r.Next() {
+		return fmt.Errorf("package does not exist")
+	}
+	err = r.Close()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// getCSVNames is a method responsible for querying the
+// channel_entry table and returning any distinct CSV
+// names that are associated with the @packageName package.
 func (s *sqlLoader) getCSVNames(tx *sql.Tx, packageName string) ([]string, error) {
 	getID, err := tx.Prepare(`
 	  SELECT DISTINCT channel_entry.operatorbundle_name
 	  FROM channel_entry
-	  WHERE channel_entry.package_name=?`)
-
+	  WHERE channel_entry.package_name=?
+	`)
 	if err != nil {
 		return nil, err
 	}
@@ -645,6 +681,9 @@ func (s *sqlLoader) getCSVNames(tx *sql.Tx, packageName string) ([]string, error
 	return csvNames, nil
 }
 
+// RemovePackage is a method responsible for removing a
+// single package and it's associated CSVs, channels, etc.
+// from the registry database.
 func (s *sqlLoader) RemovePackage(packageName string) error {
 	if err := func() error {
 		tx, err := s.db.Begin()
@@ -655,36 +694,47 @@ func (s *sqlLoader) RemovePackage(packageName string) error {
 			tx.Rollback()
 		}()
 
+		var errors []error
+		err = s.checkIfPackageExists(tx, packageName)
+		if err != nil {
+			errors = append(errors, err)
+		}
+
 		csvNames, err := s.getCSVNames(tx, packageName)
 		if err != nil {
-			return err
+			errors = append(errors, err)
 		}
 		for _, csvName := range csvNames {
 			if err := s.rmBundle(tx, csvName); err != nil {
-				return err
+				errors = append(errors, err)
 			}
 		}
 
 		deletePackage, err := tx.Prepare("DELETE FROM package WHERE package.name=?")
 		if err != nil {
-			return err
+			errors = append(errors, err)
 		}
 		defer deletePackage.Close()
 
 		if _, err := deletePackage.Exec(packageName); err != nil {
-			return err
+			errors = append(errors, err)
 		}
 
 		deleteChannel, err := tx.Prepare("DELETE FROM channel WHERE package_name = ?")
 		if err != nil {
-			return err
+			errors = append(errors, err)
 		}
 		defer deleteChannel.Close()
 
 		if _, err := deleteChannel.Exec(packageName); err != nil {
-			return err
+			errors = append(errors, err)
 		}
-		return tx.Commit()
+
+		if err = tx.Commit(); err != nil {
+			errors = append(errors, err)
+		}
+
+		return utilerrors.NewAggregate(errors)
 	}(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


timflannagan1:
I'll add some unit tests
/hold

pkg/sqlite/load.go
```diff
@@ -642,6 +642,13 @@ func (s *sqlLoader) getCSVNames(tx *sql.Tx, packageName string) ([]string, error
 		return nil, err
 	}
 
+	// Check if the query returned zero rows.
+	// In this case, we can make the assumption that
+	// the package we're querying for doesn't exist.
+	if len(csvNames) == 0 {
```
njhale, line 648: hmm, I think we still need to clean up the other tables that represent a package; e.g. `package`. It looks like the `RemovePackage` method below would return before cleaning up the other table(s).

timflannagan1: Converting this to a warning message may be the most straightforward solution to allow existing cleanup functions to run in `RemovePackage` and it still improves the UX. I'm still trying to wrap my head around what situations we would care about cleanup when attempting to removing a package that doesn't exist.

anik120: The errors for each package from the list of packages is aggregated in the calling site, so I'm not sure going with either an error or warn really changes things. Either should work I think. 
  
However, I don't think `getCSVNames` guarantees that if no CSV exists for a package, the package itself is non-existent,  which is what I think @njhale's point is too. 
Maybe using the result of the [Exec](https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/load.go#L673) statement to see if `RowsAffected` is 0 is a better time to throw the error/warning.

anik120 (commented):
Looking good @timflannagan1. Had a suggestion.

pkg/sqlite/load.go
```diff
@@ -612,12 +612,39 @@ func (s *sqlLoader) addAPIs(tx *sql.Tx, bundle *registry.Bundle) error {
...
 
+// checkIfPackageExists is a method responsible for determing
+// whether the @packageName package exists in the package
+// table and returns an error if that package is non-existent.
+func (s *sqlLoader) checkIfPackageExists(packageName string) error {
+	query := `
+	SELECT DISTINCT name
+	FROM package
+	WHERE name=$1
+	`
```
timflannagan1, line 623: It appears that are several tables that all track the various package names (e.g. `channel`, `channel_entry`, etc.) and I ended up going with the `package` table for this information. I'm not entirely sure what table should serve as the source of truth for packages, but this appears to be a sufficient check in most/all cases.

pkg/sqlite/load.go
timflannagan1: I ended up moving the logic of checking if a package exists into its own method, but I'm fine with deflating that logic back into `getCSVNames` if we don't think there's value going forward in maintaining that method. That method simply queries the `package` table for whether that `@packageName` parameter exists in the name column, and if it doesn't (i.e. no rows were returned by the `db.Query(...)` then return an error. I [made a quick review comment](https://github.com/operator-framework/operator-registry/pull/545#discussion_r554469440) on whether that table should serve as the source of truth for package existence.

Updated the `RemovePackage` to avoid exiting early if we encounter a package that doesn't exist in the registry database in order to ensure that the other tables that may be tracking bundles or channels associated with that package can have a chance to be cleaned up just in the case.

timflannagan1: I also wasn't aware of that `utilerrors.Aggregate` package until I dived into the registry code, but at a glance, that package is definitely a nice interface for handling errors and much cleaner than just appending errors and returned that joined error string using something `fmt.Errorf(...)`.

pkg/sqlite/load.go
```diff
@@ -654,37 +683,51 @@ func (s *sqlLoader) RemovePackage(packageName string) error {
 		tx.Rollback()
 	}()
 
+	var errors []error
+	// In the case that the @packageName package does not exist,
+	// log an error but don't exit early to ensure that the proper
+	// tables can still cleaned up properly.
+	err = s.checkIfPackageExists(packageName)
```
anik120, line 690: Since database operations are typically expensive, I wonder if it's worth connecting to the database just to log an error, especially when we can possibly do that by checking the results returned from the other exec statements (i.e the number returned as a result of the delete exec statement later that indicates the number of rows affected)

pkg/sqlite/load.go
```diff
@@ -645,6 +672,8 @@ func (s *sqlLoader) getCSVNames(tx *sql.Tx, packageName string) ([]string, error
 	return csvNames, nil
 }
 
+// RemovePackage is a method responsible for removing a
```
anik120, line 675: `nit`: Might be worth mentioning that the package is removed in it's entirety, i.e channels/CSVs etc removed from other tables too.

anik120 (changes requested):

pkg/sqlite/load.go
anik120: FYI my thoughts on `checkIfPackageExists` are in [this comment](https://github.com/operator-framework/operator-registry/pull/545#discussion_r555039538).

pkg/sqlite/load.go
timflannagan1: I think it's reasonable to point out the cost implications of performing an additional SQL statement during this method, and whether that logic can be deflated to check the rows affected from the other `delete from ...` statements. That said, if we want to keep this method, I would argue that we could simply cache this query as a prepared statement (or just avoid using the query method when interacting with the underlying database) and now we get relatively optimized performance.

timflannagan1: Hmm, it looks like `db.Query` _may_ create a new prepared statement, but it's difficult to say confidently that's what's happening here, and it's likely that behavior is driver-specific.

timflannagan1: Thinking more about this, we _could_ just check the rows affected for a delete call, but then we would likely need to do that for both of the delete calls being made, and in the case that method is refactored or extended in the future, we may need to continue to make that comparison, just in case the various tables are out of sync somehow (unlikely looking at the code, but still possible).

If we're fine with that approach, I can update the implementation to check the rows affected for both of the delete calls, and if the number of rows affected is non-zero, i.e. we can make a reasonable assumption that package name is present, then we can make a `packageExists` bool as true and append an error only when that boolean variable is false at the end of the control flow.

@njhale Any thoughts on how to best tackle this? I think there's value in creating a new method just for checking package existence, but it may be overkill as we have alternatives in place that mirror that same logic without too much additional code.

pkg/sqlite/load.go
timflannagan1: Yeah, makes sense - done in the last round of force pushes.

pkg/sqlite/load.go
```diff
@@ -655,36 +694,47 @@ func (s *sqlLoader) RemovePackage(packageName string) error {
...
 		}
 		defer deleteChannel.Close()
 
 		if _, err := deleteChannel.Exec(packageName); err != nil {
-			return err
+			errors = append(errors, err)
 		}
-		return tx.Commit()
+
+		if err = tx.Commit(); err != nil {
```
benluddy, line 733: Do we still want to commit the transaction if an error was returned from one of the statements? I am concerned that this will make it possible to perform a partial removal.

benluddy (commented):
My read of the bug report is:
1. The helptext for `opm index rm` doesn't clearly communicate that the values passed to `-o`/`--operators` should be package names, as opposed to bundle/CSV names.
1. If a user misinterprets `-o`, there isn't any feedback that would indicate the mistake.

Can this PR avoid changing control flow, clarify the helptext and add logging that indicates how many things were removed? It's technically a breaking change to make it an error to rm a nonexistent package -- someone may be relying on opm exiting with a zero status in a script somewhere.

PR #558, opened 2021-01-19
Fix upstream example dockerfile

### LINKED ISSUE #472: Error in building in Image with upstream-example.Dockerfile
On the master branch when you do `docker build -t test-registry-image -f upstream-example.Dockerfile .` you see:
```
Step 5/10 : COPY --from=builder /build/bundles.db /bundles.db
COPY failed: stat /var/lib/docker/overlay2/e7f6a0ede48483e52a6e0b05b18967b1596997852306cbc1efca29ed316a8ee5/merged/build/bundles.db: no such file or directory
```
I tried to debug this and found that the `bundles.db` file exist in `/` and not in `/build/` in `BUILDER` image, So by changing [line](https://github.com/operator-framework/operator-registry/blob/master/upstream-example.Dockerfile#L7) to `COPY --from=builder /bundles.db /bundles.db` solved the problem. Let me know if this is true :), I'll be more than happy to send a PR for this issue :)

ecordell:
Fixes #472

### DIFF
--- a/upstream-example.Dockerfile
+++ b/upstream-example.Dockerfile
@@ -5,7 +5,7 @@ RUN /bin/initializer -o ./bundles.db
 
 FROM scratch
 COPY ["nsswitch.conf", "/etc/nsswitch.conf"]
-COPY --from=builder /build/bundles.db /bundles.db
+COPY --from=builder /bundles.db /bundles.db
 COPY --from=builder /bin/registry-server /registry-server
 COPY --from=builder /bin/grpc_health_probe /bin/grpc_health_probe
 EXPOSE 50051


estroz (approved):
/lgtm

upstream-example.Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN /bin/initializer -o ./bundles.db
 
 FROM scratch
 COPY ["nsswitch.conf", "/etc/nsswitch.conf"]
-COPY --from=builder /build/bundles.db /bundles.db
```
PratikMahajan, line 8: Removing `/build` can break this for people using an older version of podman. 
A better fix will be adding `WORKDIR /build` on Line 2

PR #570, opened 2021-01-27
Bug 1921763: add missing closes

rphillips:
There are a few places where Close() is forgotten to be called. Over time these could cause memory issues.

4.7 has 4 opm processes running that I have seen grow to 500 MB over time.They are the top 4 memory consumers on the machine.

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -495,6 +495,7 @@ func (s *sqlLoader) getBundleSkipsReplacesVersion(tx *sql.Tx, bundleName string)
 		err = rerr
 		return
 	}
+	defer rows.Close()
 	if !rows.Next() {
 		err = fmt.Errorf("no bundle found for bundlename %s", bundleName)
 		return
@@ -535,6 +536,7 @@ func (s *sqlLoader) getBundlePathIfExists(tx *sql.Tx, bundleName string) (bundle
 		err = rerr
 		return
 	}
+	defer rows.Close()
 	if !rows.Next() {
 		// no bundlepath set
 		return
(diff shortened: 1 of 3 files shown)


pkg/sqlite/load.go
```diff
@@ -627,6 +629,7 @@ func (s *sqlLoader) getCSVNames(tx *sql.Tx, packageName string) ([]string, error
 	if err != nil {
 		return nil, err
 	}
+	defer rows.Close()
```
ecordell, line 632: `rows.Close()` is called a few lines below (not deferred to capture the error)

pkg/sqlite/load.go
```diff
@@ -1024,12 +1029,15 @@ func getTailFromBundle(tx *sql.Tx, name string) (bundles []string, err error) {
...
 				if err != nil {
+					rows.Close()
 					return nil, err
 				}
 				if defaultChannelHead.Valid {
+					rows.Close()
 					return nil, registry.ErrRemovingDefaultChannelDuringDeprecation
 				}
 			}
+			rows.Close()
```
ecordell, line 1040: rows.Close() can return an error

pkg/sqlite/load.go
rphillips: fixed this up

pkg/sqlite/load.go
rphillips: fixed this.

rphillips:
/retest

this should be ready

dinhxuanvu (approved):
/approve

PR #571, opened 2021-01-28
Bug 1921778: Add build metadata version precedence

njhale:

Add a semver version preference to [build IDs](https://semver.org/#spec-item-10) to accommodate existing published content making use of them to differentiate between successive releases of the same version.

Rule: when two versions are semver equivalent, compare their build IDs as if they were [prereleases](https://semver.org/#spec-item-9).

e.g. `v1.2.3+1.2.3` and `v.1.2.3+1.1.55` are evaluated to `v1.0.0-1.2.3` and `v1.0.0-1.1.55` respectively and compared. This yields  `v1.0.0-1.2.3 < v1.0.0-1.1.55` which is then mapped back to `v1.2.3+1.2.3 < v.1.2.3+1.1.55` 

When the `--overwrite-latest` option is used on index add, for each bundle given two things happen:

1. All bundles in the given package are removed from the index
2. All bundles removed - old latest bundle + new latest bundle are re-added for the package

Bundles are still added one by one -- to save on refactoring -- so when we arrive at the add for bundles with equal versions but separate build IDs, we choke finding the local maxima and throw an error. Providing ordering for build IDs, will allow us to select the _correct_ bundle in these scenarios.

See https://hackmd.io/73MxgpJgTgGXjAk-CmMOTw?edit for more details.


pkg/registry/helper_test.go
```diff
@@ -0,0 +1,325 @@
...
+	} {
+		t.Run(tt.description, func(t *testing.T) {
+			fieldVersion, err := semver.Parse(tt.fields.version)
+			require.NoErrorf(t, err, "bad testcase")
+
+			argVersion, err := semver.Parse(tt.args.version)
+			require.NoErrorf(t, err, "bad testcase")
+
+			c := bundleVersion{version: fieldVersion}.compare(bundleVersion{version: argVersion})
+			switch tt.expect.order {
```
ecordell, line 225: you could have the test class embed the comparator directly?

pkg/registry/populator.go
```diff
@@ -379,18 +380,43 @@ func (i *DirectoryPopulator) loadOperatorBundle(manifest PackageManifest, bundle
...
+
+// compare returns a value less than one if the receiver arg is less smaller the given version, greater than one if it is larger, and zero if they are equal.
+// This comparison follows typical semver precedence rules, with one addition: a simple ordering is defined for build metadata (+<build-metadata>) corresponding to whole numbers such that they are in ascending order; e.g. 1.0.0 < 1.0.0+1 < 1.0.0+2 ...
+func (b bundleVersion) compare(v bundleVersion) int {
+	if c := b.version.Compare(v.version); c != 0 {
+		return c
+	}
+
+	// Attempt to compare based on build metadata
+	return buildInt(b.version).Cmp(buildInt(v.version))
```
bparees, line 399: this comparison needs to be more sophisticated.  we should follow the logic semver uses for comparing prereleases, which as i understand it is to break the prerelease into dotted segments, and compare each segment until it finds a difference.

Perhaps you can take the buildID, and append it to a "v1.0.0" version as a prerelease, and do that comparison using semver libs, to determine which buildID is greater.

e.g.:  v1.2.3+1.2.3 vs v.1.2.3+1.1.55, you extract the buildid as 1.2.3 and 1.1.55, then do a semver comparison of "v1.0.0-1.2.3" vs "v1.0.0-1.1.55", and whichever one of those is greater, means that buildID is greater.

pkg/registry/populator.go
```diff
@@ -379,18 +380,43 @@ func (i *DirectoryPopulator) loadOperatorBundle(manifest PackageManifest, bundle
...
+}
+
+func buildInt(v semver.Version) *big.Int {
+	build := big.NewInt(0)
+	if len(v.Build) != 1 {
+		// Can only be an integer build-meta with a single element, otherwise treat as zero
+		return build
+	}
+	if b, _ := build.SetString(v.Build[0], 10); b != nil {
```
bparees, line 408: buildIDs don't have to be integers so if i'm understanding this correctly, this seems likely to fail.

njhale: yup -- the intent is to treat everything outside the domain (positive integers) as zero

pkg/registry/populator.go
njhale: sgtm -- I'll prototype it out.

ecordell: I think we can make this work for most cases by:

- extract buildId
- make a new semver string (or build it directly), but set major/minor/patch to `0` and set `prerelease` to the buildid
- compare both of those constructed versions

That is basically saying "apply the prerelease comparison rules to buildid" which I think covers a lot of cases (when buildid is not valid semver, etc). 

did a quick poc in the playground: https://play.golang.org/p/jMDNbBlP_Ib

test/e2e/opm_test.go
```diff
@@ -405,6 +421,62 @@ var _ = Describe("opm", func() {
...
+			for _, pkg := range packages {
+				existing, err := graphLoader.Generate(pkg)
+				Expect(err).NotTo(HaveOccurred(), "Error generating graph for package %s", pkg, existing)
+
+				for _, ch := range existing.Channels {
+					request := lregistry.AddToRegistryRequest{
+						Permissive:    false,
+						SkipTLS:       false,
+						InputDatabase: dbFile,
+						Bundles:       []string{ch.Head.BundlePath},
```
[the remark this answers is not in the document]
bparees: @kevinrizza @gallettilance i forget, was that already a known constraint over the `overwrite-latest` flag?  That it only works for the latest version, not just "any head of a channel"?

it seems likely to be a problematic limitation for teams that maintain z-stream branches of their operators, if i'm understanding the limitation correctly.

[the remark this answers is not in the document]
bparees: (but if it's not a new limitation being introduced by this PR, we can hash that out elsewhere)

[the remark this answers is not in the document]
kevinrizza: It is not a new limitation, it was a major limitation at the time that we came up with `overwrite-latest` because we did not have a way of inserting a bundle into the middle of a channel. It only solves a very very specific use case for retesting and republishing the same version over and over in stage before you release your operator and was never intended as a way to arbitrarily insert over a previous version.

That's also why that command is hidden in opm. It's (imo) a half baked feature that we threw in because "the pipeline just had to have it" at the very very very last minute of 4.6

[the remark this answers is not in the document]
bparees: ok thanks, will take the rest of the discussion offline.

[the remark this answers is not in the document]
gallettilance: @njhale we need to modify this test case to only test overwrite-latest for channel heads that are latest. That code is part of the overwrite-latest code path here: https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/populator.go#L123

test/e2e/opm_test.go
```diff
@@ -405,6 +421,62 @@ var _ = Describe("opm", func() {
...
+						Mode:          registry.ReplacesMode,
+						ContainerTool: tool,
+						Overwrite:     true,
+					}
+
+					err = adder.AddToRegistry(request)
+					if err != nil {
+						errs = append(errs, fmt.Errorf("Error overwriting bundles for package %s: %s", pkg, err))
```
gallettilance, line 466: @njhale looks like even if overwrite-latest fails we are just accumulating the errs here without raising them or doing anything with them.

test/e2e/opm_test.go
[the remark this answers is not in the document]
njhale: @gallettilance done.

kevinrizza:
Generally I think this is in a good state. One thing that needs to happen as a result of this is that we need to write a doc that explicitly describes our versioning preferences (given that we are no longer following pure semantic versioning). That being said, there's not a good place to add existing rules and I don't want to block this pr for a request to write a doc like that.

The other thing I was thinking about here is that while we are adding this build version comparison to the insert preference, we are explicitly *not* adding build metadata to all comparison, specifically with respect to semver mode as well as skiprange resolution on cluster. I think we need a more complete story there and we need to revisit this shortly.

/lgtm

PR #582, opened 2021-02-17
feat(declarative-config): opm add creates declarative configs

anik120:

This PR implements a new opm add command that accepts bundle 
images as inputs, and creates declarative configs for those bundles 
that are added to a collection of existing configs (an empty collection
is also considered an existing collection of configs)

https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md

**Reviewer Checklist**


pkg/containertools/dockerfilegenerator.go
```diff
@@ -8,14 +8,15 @@ import (
 )
 
 const (
-	defaultBinarySourceImage = "quay.io/operator-framework/upstream-opm-builder"
-	DefaultDbLocation        = "/database/index.db"
-	DbLocationLabel          = "operators.operatorframework.io.index.database.v1"
+	defaultBinarySourceImage    = "quay.io/operator-framework/upstream-opm-builder"
```
timflannagan, line 11: Is there any reason we can't export this default const variable like we did with the others?

pkg/lib/indexconfig/indexconfig.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		pkgJSON, err := json.MarshalIndent(configPackage, " ", "   ")
+		if err != nil {
+			return err
+		}
+		f.Write(pkgJSON)
+		f.Write([]byte("\n"))
+	} else {
+		return err
+	}
```
timflannagan, line 120: I'd recommend moving this to a helper function or moving around some of this logic to deflate any if/else if/else clause type patterns. It looks like you can just check if the err is nil, then open that file and write some bytes to it and return early, else check if the error is due to the file not existing and handle it from there.

pkg/containertools/dockerfilegenerator.go
anik120: @timflannagan this is a generated file 
```
go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . DockerfileGenerator
```

pkg/lib/indexconfig/indexconfig_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestPrintConfigCreatesNewConfig(t *testing.T) {
+
+	// create bundle from manifests
+	bundle, err := testBundle()
+	require.NoError(t, err)
+	// create temporary dir for test
+	tmpDir, err := ioutil.TempDir("./", tmpTestDirPrefix)
+	defer os.RemoveAll(tmpDir)
+	require.NoError(t, err)
```
timflannagan, line 33: This seems like something that abstracted away a bit as a test setup helper function, where you could pass a testing.T parameter and uses that new `t.Cleanup` functionality that was added in 1.14.

pkg/lib/indexconfig/indexconfig.go
timflannagan: Looking at some of the other parts of the registry codebase, it seems like we have to perform this kind of check in a couple of different places. Probably not something we can generalize unless it's just checking for file existence and creating that file as needed, or handling errors as needed.

pkg/lib/indexconfig/indexconfig.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"os"
+
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
+	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
+	"github.com/operator-framework/operator-registry/pkg/lib/certs"
+	registry "github.com/operator-framework/operator-registry/pkg/lib/registry"
+	regi "github.com/operator-framework/operator-registry/pkg/registry"
```
anik120, line 15: Trying to think of a better name for this alias

pkg/lib/indexconfig/indexconfig_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	manifestDir      = "./testdata/bundle/manifests"
+	tmpTestDirPrefix = "tmp_test_"
+	expectedConfig   = "{\n    \"schema\": \"olm.package\",\n    \"name\": \"lib-bucket-provisioner\",\n    \"defaultChannel\": \"alpha\",\n    \"icon\": {\n       \"base64data\": \"\",\n       \"mediatype\": \"\"\n    },\n    \"channels\": [\n       \"alpha\",\n       \"beta\"\n    ],\n    \"description\": \"\"\n }\n{\n    \"schema\": \"olm.bundle\",\n    \"name\": \"lib-bucket-provisioner.v1.0.0\",\n    \"package\": \"lib-bucket-provisioner\",\n    \"image\": \"\",\n    \"version\": \"1.0.0\",\n    \"properties\": [\n       {\n          \"type\": \"olm.channel\",\n          \"value\": {\n             \"name\": \"alpha\",\n             \"replaces\": \"\"\n          }\n       },\n       {\n          \"type\": \"olm.channel\",\n          \"value\": {\n             \"name\": \"beta\",\n             \"replaces\": \"\"\n          }\n       },\n       {\n          \"type\": \"skips\",\n          \"value\": \"\"\n       },\n       {\n          \"type\": \"skipsRange\",\n          \"value\": \"\"\n       }\n    ],\n    \"relatedImages\": []\n }\n"
```
exdx, line 22: Can we put the expected config fixture (and I'm sure there will be many) in a separate `testdata/bundle/golden` directory as a json-file instead? The plus is since they are human-readable it's easier to understand the test as well.

pkg/lib/indexconfig/configdefs.go
```diff
@@ -0,0 +1,111 @@
+package indexconfig
```
exdx, line 1: One other observation I have is regarding the word "config" everywhere -- I know we call it the declarative index config, but I'm wondering what value the word config provides. Go packages are all namespaced scoped. I'm wondering if we had something like `pkg/lib/declarative/...` instead and then remove the word config from the names altogether.

pkg/lib/indexconfig/indexconfig.go
exdx: maybe call the `pkg/lib/registry` package `libregistry` and then call `regi` `registry` instead?

pkg/lib/indexconfig/indexconfig_test.go
anik120: So I'm on 1.13, and I'm wondering if just `t.Cleanup` is enough justification to tell a system on 1.13 you can't run the tests unless you upgrade to 1.14.

pkg/lib/indexconfig/configdefs.go
anik120: I'm not sure `declarative` by itself conveys anything honestly. I've tried to convey the message in such a way that someone who's read the enhancement once can read the code easily. The word declarative isn't used more than 2-3 times in the enhancement itself too btw.

pkg/lib/indexer/indexer.go
```diff
@@ -291,25 +306,27 @@ func (i ImageIndexer) PruneFromIndex(request PruneFromIndexRequest) error {
 	return nil
 }
 
-// ExtractDatabase sets a temp directory for unpacking an image
-func (i ImageIndexer) ExtractDatabase(buildDir, fromIndex, caFile string, skipTLS bool) (string, error) {
+// ExtractDatabases sets a temp directory for unpacking an image's sqllite database and package configs folder
+func (i ImageIndexer) ExtractDatabases(buildDir, fromIndex, caFile string, skipTLS bool) (string, string, error) {
```
njhale, line 310: I don't think we want to couple database extraction with config extraction. I feel it would be cleaner for each to have a dedicated method for extraction (or for there to be an interface for deserialization that we can switch between)

At a higher level, we should probably be treating these formats as equivalents (i.e. `sqlite=config`); defining a common intermediate representation for interacting with data, manipulate it, then convert to the target from our IR directly.

njhale (commented):
@anik120 thank you for working on this!

I left a good deal of notes/questions/comments here for when folks get the chance.

I've also been thinking about whether we first want the schema for the declarative config written down somewhere (e.g. jsonschema) for clients to reference

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
+	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
+	"github.com/operator-framework/operator-registry/pkg/lib/certs"
+	libregistry "github.com/operator-framework/operator-registry/pkg/lib/registry"
+	registry "github.com/operator-framework/operator-registry/pkg/registry"
+
+	"github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
anik120: I could swap them, but does it really matter which group's listed first as long as they are separate?

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
...
+		rootCAs, err := certs.RootCAs(request.CaFile)
+		if err != nil {
+			return fmt.Errorf("failed to get RootCAs: %v", err)
+		}
+		reg, rerr = containerdregistry.NewRegistry(containerdregistry.SkipTLS(request.SkipTLS), containerdregistry.WithRootCAs(rootCAs))
+	case containertools.PodmanTool:
+		fallthrough
+	case containertools.DockerTool:
+		reg, rerr = execregistry.NewRegistry(request.ContainerTool, c.Logger, containertools.SkipTLS(request.SkipTLS))
+	}
```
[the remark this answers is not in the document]
anik120: I can see why we might want to think about that. However, other interfaces used here also accept the parameters needed to construct the registry and construct it within themselves. And the constructions vary for example based on whether it's using the `buildTool` or the `pullTool` etc. I think it's totally reasonable to investigate the possibility of constructing the appropriate registries for the appropriate interface and inject them, but it sounds like the subject of a separate PR.

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
+package declarative
```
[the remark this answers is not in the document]
anik120: Thanks for clarifying this looking at the code base this distinction isn't apparent probably coz it already breaks the rule like you said. I've moved the package out.

pkg/containertools/dockerfilegenerator.go
```diff
@@ -8,14 +8,16 @@ import (
...
+	DefaultDbLocation           = "/database/index.db"
+	DefaultConfigFolderLocation = "/configs/"
+	DbLocationLabel             = "operators.operatorframework.io.index.database.v1"
+	ConfigsLocationLabel        = "operators.operatorframework.io.configs.v1"
 )
 
 // DockerfileGenerator defines functions to generate index dockerfiles
 type DockerfileGenerator interface {
-	GenerateIndexDockerfile(string, string) string
+	GenerateIndexDockerfile(string, string, string) string
```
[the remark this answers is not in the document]
anik120: Ref: https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md#migration-plan

When the declarative index feature is delivered we'll just be generating an index with only a config path. This change is only a stop gap measure so that master is not broken while new features are being developed. It should be changed back to accept only one before the release of the new binary.

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
...
+		if err := printConfig(img.Bundle, request.ConfigFolder); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func printConfig(bundle *registry.Bundle, configFolder string) error {
+
```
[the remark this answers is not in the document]
anik120: I am not aware of any style guideline for this enforced by `gofmt/goimport`, although I have removed the newline here and in other places I've introduced.

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, ref := range simpleRefs {
+		to, from, cleanup, err := libregistry.UnpackImage(context.TODO(), reg, ref)
+		if err != nil {
+			return err
+		}
+		defer cleanup()
+
+		img, err := registry.NewImageInput(to, from)
+		if err := printConfig(img.Bundle, request.ConfigFolder); err != nil {
```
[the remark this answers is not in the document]
anik120: Ref: https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md#migration-plan

Ultimately we will be sticking with only `AddToConfig` so we won't need an interface that can swap out destination formats.

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// ConfigEditor is an implementation of IndexConfig
+type ConfigEditor struct {
+	Logger *logrus.Entry
+}
+
+// AddToConfig persists the bundle in it's package config
+func (c ConfigEditor) AddToConfig(request AddConfigRequest) error {
```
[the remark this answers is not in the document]
anik120: Ref: https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md#migration-plan

This will be the ONLY way to add bundle in the new binary release.

pkg/lib/declarative/indexconfig.go
```diff
@@ -0,0 +1,114 @@
...
+		img, err := registry.NewImageInput(to, from)
+		if err := printConfig(img.Bundle, request.ConfigFolder); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func printConfig(bundle *registry.Bundle, configFolder string) error {
```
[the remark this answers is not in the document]
anik120: Isn't `Indexer` that interface?

pkg/lib/declarative/interfaces.go
```diff
@@ -0,0 +1,30 @@
...
+	ConfigFolder  string
+	ContainerTool containertools.ContainerTool
+	CaFile        string
+	SkipTLS       bool
+}
+
+// IndexConfig represents the content packages of an index
+// in a declarative way using config files
+type IndexConfig interface {
+	AddToConfig(AddConfigRequest) error
```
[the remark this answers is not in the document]
anik120: Are you sure we won't write an alternate/mock implementations in the future?

pkg/lib/indexer/indexer.go
```diff
@@ -98,8 +102,19 @@ func (i ImageIndexer) AddToIndex(request AddToIndexRequest) error {
...
 	}
 
+	// Add the bundles declarative config
+	addToConfigReq := declarative.AddConfigRequest{
+		Bundles:       request.Bundles,
+		ConfigFolder:  configFolderPath,
+		SkipTLS:       request.SkipTLS,
+		ContainerTool: i.PullTool,
+	}
+	if err := i.IndexConfig.AddToConfig(addToConfigReq); err != nil {
```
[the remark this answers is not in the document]
anik120: As mentioned above we'll only be maintaining the configs folder. This is just a stop gap measure to unblock the other feature work in the pipeline, but we'll only be adding to the config by the time the new binary is released.

pkg/lib/indexer/indexer.go
anik120: Isn't `extract contents of the index` one task? I'd argue that the database and the config are one as `contents` of the index, not separate concerns. And that becomes apparent when I think about writing two routines for extracting them: the routines are almost identical, expect for what file path they're targeting inside the container.

anik120:
> I've also been thinking about whether we first want the schema for the declarative config written down somewhere (e.g. jsonschema) for clients to reference

@njhale thank you for the valuable feedbacks! There have been discussions about this, and the decision was to have the schema introduced in this PR for ease of review, but eventually move it out to the api repo that'll have all the functionality required for client to reference it. Fyi it's planned as part of the index validation with declarative config feature.

pkg/containertools/dockerfilegenerator.go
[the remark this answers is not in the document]
kevinrizza: Does it make more sense for us to create some new interfaces rather than modify existing ones? It would probably make it more obvious so that we know what needs to get cleaned up afterwards

pkg/containertools/dockerfilegenerator.go
```diff
@@ -46,9 +48,13 @@ func (g *IndexDockerfileGenerator) GenerateIndexDockerfile(binarySourceImage, da
 
 	// Labels
 	dockerfile += fmt.Sprintf("LABEL %s=%s\n", DbLocationLabel, DefaultDbLocation)
+	dockerfile += fmt.Sprintf("LABEL %s=%s\n", ConfigsLocationLabel, DefaultConfigFolderLocation)
 
 	// Content
 	dockerfile += fmt.Sprintf("ADD %s %s\n", databasePath, DefaultDbLocation)
+	if configFolderPath != "" {
```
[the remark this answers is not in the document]
kevinrizza: I think it could make sense in the future, but without the tooling being there I think it's outside the scope of the related enhancement

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
njhale: well it's the convention we've been following everywhere.

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
njhale: > However, other interfaces used here also accept the parameters needed to construct the registry and construct it within themselves

I know there's precedent for this, but IMO not dependency injecting these makes the command harder to test; this is net new and I would like to avoid incurring the same issues from the start.

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
njhale: thanks!

pkg/containertools/dockerfilegenerator.go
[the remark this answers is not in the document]
njhale: I second what Kevin is saying. I prefer we define the interfaces we want up front.

pkg/containertools/dockerfilegenerator.go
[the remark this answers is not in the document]
njhale: I'm suggesting that the MVP doesn't require any image building, which shouldn't require tooling.

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
njhale: I'm not sure I understand the reasoning here. 

In the interim, adding such an interface gives us:
- code deduplication
- better testing

In the long term:
- better testing
- room for new formats should they arise

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
njhale: As a side-effect of having a config file as input, I'd imagine it's possible to manually add a bundle to a config. I'm basically wondering if _reading_ a config and serving its content are a better place to start than adding to an existing config.

pkg/lib/declarative/interfaces.go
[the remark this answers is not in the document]
njhale: We could, but how useful would they be? to what clients?

I think this interface is abstracting the wrong thing for this problem; the "how" to add and not the "what" to add. IIRC, we haven't used the other interfaces of the other types of this pattern for mocks or alternate implementations as well.

pkg/containertools/dockerfilegenerator.go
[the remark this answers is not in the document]
anik120: The reason I'm having trouble understanding the request for a new interface is because ultimately, we do need the same `DockerfileGenerator` interface even when we will only include the configs and not the database. 

But, that will be possible only when the ability to serve the apis with the configs before available. Ref: https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md#usage-of-the-json-representation-of-packages-by-olm-components
In which case we generate the Dockerfile in a way that only adds the configs and serves those configs. 

As for how it stands now, if we have an implementation for `DockerfileGenerator` that adds the configs (but not the database), the container image will not have the ability to serve the apis. Therefore the database must be included (and served) in the container.

pkg/lib/indexer/indexer.go
njhale: >  Isn't extract contents of the index one task?

The tasks share the same goal, but may achieve them differently -- I think you're picking up on the need for an abstraction over the extraction of content from different types of catalogs to the filesystem.

Consider what happens to this method signature when we deprecate sqlite indexes; removal of a return value will require us to update the logic in all code that depends on it. Similarly, if we ever add a new type of index, we'd potentially add a new return value, also requiring an update. Abstracting over a shared method signature gives us the opportunity to reduce the number of places we need to make changes (think dependency injection) and increases the opportunities for code reuse.

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
anik120: @njhale is this what you had in mind https://github.com/operator-framework/operator-registry/pull/582/commits/5b18699efa188c9f408481aa5f56d0f0adfb2c4a
(I shall merge the commits if it is what we had in mind)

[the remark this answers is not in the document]
njhale: yup! that's what I was looking for.

pkg/lib/declarative/interfaces.go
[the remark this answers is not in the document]
anik120: @njhale if we have run into this question of "what" to add now, how can we be certain we will absolutely never run into the question of "how" to add(and therefore how are we certain "how" to add is the wrong question?)

pkg/declarative/indexconfig.go
```diff
@@ -0,0 +1,95 @@
...
+type ConfigEditor struct {
+	Logger      *logrus.Entry
+	ImgRegistry image.Registry
+}
+
+// AddToConfig persists the bundle in it's package config
+func (c ConfigEditor) AddToConfig(request AddConfigRequest) error {
+
+	defer func() {
+		if err := c.ImgRegistry.Destroy(); err != nil {
```
njhale, line 25: We should probably destroy the registry in the same scope that it was created in. This way, we remove at least one side effect of running `AddToConfig`.

pkg/lib/indexer/indexer.go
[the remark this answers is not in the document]
njhale: As-is, if a user were to add to an existing index it seems like we'll end up with a mismatch between the content of the sqlite database and the config. Is there a step I've overlooked that populates the config with the content in the existing sqlite database? Maybe there's some reason this doesn't matter that I'm missing?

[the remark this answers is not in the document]
anik120: Right, the ability to _read_ the config files is necessary to perform the same calculations we do now before adding to the database `channel_entry` table. This is currently planned with the work that'll serve the database (_read_ the files to serve content) and once that becomes available and we're able to perform the same checks/calculations we do now with the `querier`, the content in the db will be identical to the content in the configs. 

(Also why the db is not being removed with this PR)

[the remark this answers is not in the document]
anik120: (Which I suppose is a really good argument for the work to serve the content should go in first)

pkg/containertools/dockerfilegenerator.go
```diff
@@ -46,9 +49,17 @@ func (g *IndexDockerfileGenerator) GenerateIndexDockerfile(binarySourceImage, da
 
 	// Labels
 	dockerfile += fmt.Sprintf("LABEL %s=%s\n", DbLocationLabel, DefaultDbLocation)
+	dockerfile += fmt.Sprintf("LABEL %s=%s\n", ConfigsLocationLabel, DefaultConfigFolderLocation)
 
 	// Content
-	dockerfile += fmt.Sprintf("ADD %s %s\n", databasePath, DefaultDbLocation)
+	sortedsrc := make([]string, 0)
```
njhale, line 55:
```suggestion
	var sortedsrc []string
```

pkg/containertools/dockerfilegenerator.go
```diff
@@ -46,9 +49,17 @@ func (g *IndexDockerfileGenerator) GenerateIndexDockerfile(binarySourceImage, da
 
 	// Labels
 	dockerfile += fmt.Sprintf("LABEL %s=%s\n", DbLocationLabel, DefaultDbLocation)
+	dockerfile += fmt.Sprintf("LABEL %s=%s\n", ConfigsLocationLabel, DefaultConfigFolderLocation)
 
 	// Content
-	dockerfile += fmt.Sprintf("ADD %s %s\n", databasePath, DefaultDbLocation)
```
njhale, line 51: I like how flexible the new method is, but it also drops the guarantee that the `databasePath` is placed at `DefaultDbLocation`.

It is now less clear which of the inputs is used as the database, and the onus is on the method client to know to add a mapping to `DefaultDbLocation`.

anik120: @njhale not sure how we feel about documenting that in the method doc to make it more clear? It also does feel like it might be overkill when only one type of stored format (sqlilite OR configs OR anything else) is being used though. Do we feel like reverting to the previous design that extended the number of method arguments (which was slightly unclean but simple to read) is not favorable over ensuring future extensibility of this method?   
It should be noted again that extending the number of method argument will only be a stop gap measure, till the feature for migrating old indexes to new format is implement, in which case we'll be back to only one format. 

The other option is leave the method untouched, so this PR would create the configs, but not load them to the container image. And when the migration/serve is ready, a new PR would switch the data being loaded to configs instead of the database.

pkg/lib/indexer/indexer.go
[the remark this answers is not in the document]
joelanford: If it helps to have the read side in first, perhaps a PR that just lays down the `pkg/declarative` types and load functions would be good to get in? Then we could build out in parallel from there?

pkg/lib/indexer/indexer.go
```diff
@@ -140,15 +151,27 @@ func (i ImageIndexer) DeleteFromIndex(request DeleteFromIndexRequest) error {
 		return err
 	}
 
-	databasePath, err := i.ExtractDatabase(buildDir, request.FromIndex, request.CaFile, request.SkipTLS)
+	fromIndex := image.SimpleReference(request.FromIndex)
+
+	// Get the old index image's dbLocationLabel to find this path
+	labels, err := i.ImgRegistry.Labels(context.TODO(), fromIndex)
```
joelanford, line 157: Nit: is it possible to add a `context.Context` parameter to this function signature or do we have an API guarantee to maintain? It'd be nice to pipe a real context all the way in from cobra.

pkg/lib/indexer/indexer.go
```diff
@@ -253,15 +290,27 @@ func (i ImageIndexer) PruneFromIndex(request PruneFromIndexRequest) error {
...
+	if err != nil {
+		return err
+	}
+
+	dbLocation, ok := labels[containertools.DbLocationLabel]
+	if !ok {
+		return fmt.Errorf("index image %s missing label %s", request.FromIndex, containertools.DbLocationLabel)
+	}
+	extractedDbLocation, err := i.ExtractIndexContent(fromIndex, dbLocation, filepath.Join(buildDir, defaultDatabaseFolder))
```
joelanford, line 305: It looks like this section of code is repeated in several places. Can we put it in a separate function?

pkg/lib/indexconfig/indexconfig_test.go
joelanford: FWIW, Go 1.16 is out now, so Go 1.13 is pretty out of date. On many of the projects I've worked on we've bumped Go versions soon after each release and I haven't heard contributors complain. Go is _super_ stable and rarely causes problems. Also, I'm fairly sure Go 1.15 is being used to build OLM's packages for OpenShift builds, so we may as well use at least that.

pkg/lib/indexconfig/configdefs.go
joelanford: Before I saw this PR, I had named a package for a similar purpose `declcfg`. I'm good with `declarative` too but figured I'd throw out what my first instinct was.

pkg/lib/declarative/indexconfig.go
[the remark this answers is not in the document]
joelanford: Plug for https://github.com/AanZee/goimportssort. I have it configured (along with goimports and go fmt) in my IDE's on-save config and I never think about it anymore.

pkg/lib/declarative/configdefs.go
```diff
@@ -0,0 +1,111 @@
...
+	Mediatype  string `json:"mediatype"`
+}
+
+type configPackage struct {
+	Schema         string   `json:"schema"`
+	Name           string   `json:"name"`
+	DefaultChannel string   `json:"defaultChannel"`
+	Icon           icon     `json:"icon"`
+	Channels       []string `json:"channels"`
+	Description    string   `json:"description"`
```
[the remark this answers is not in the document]
joelanford: Another possible downside of a list of freeform properties is the ambiguity that arises when there may be an expectation for a certain count of a certain type of property.

Would it be possible to write a schema that we could use to validate against that says "allow exactly 0 or 1 item in the list of properties to have `"type": "skipRange"`"?

If we have a schema but can't encode constraints like that, then what? Do we have a schema and a separate doc about those extra constraints and maintain custom validators after we've parsed into Go structs?

I _definitely_ don't see the big picture or understand the long-term vision for OLM yet, but it feels like if we have enshrined some of these properties as "first class fields" in our APIs and docs, then maybe it's more user friendly to make them structured and easier for users to grok.

pkg/containertools/dockerfilegenerator.go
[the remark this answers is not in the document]
joelanford: This may have been discussed more in the last week, but it seems like we'll have the config server done and merged in the next week or two. I would agree with Kevin and Nick that it seems like it will be unnecesssary for us to build images that contain both databases and configs.

pkg/declcfg/addTocfg.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		m, err := registry.CreateModelFromBundle(img.Bundle)
+		if err != nil {
+			return fmt.Errorf("Error creating model from bundle:%s", err)
+		}
+		bundleConfig := ConvertFromModel(*m)
+		if !existingPackage(decCfg, img.Bundle.Package) {
+			b.Logger.Infof("adding new package metadata to configs")
+			decCfg.Packages = append(decCfg.Packages, bundleConfig.Packages...)
+		}
```
joelanford, line 64: I'm curious if you considered converting the `registry.Bundle` directly to a `declcfg.Bundle` and skipping the model altogether. What are the pros/cons of each approach?

pkg/declcfg/addTocfg.go
```diff
@@ -0,0 +1,104 @@
...
+		b.Logger.Infof("adding bundle metadata to configs")
+		decCfg.Bundles = append(decCfg.Bundles, bundleConfig.Bundles...)
+	}
+	os.RemoveAll(request.OutDir) // this is required because WriteDir only writes to an empty directory
+	os.Mkdir(request.OutDir, os.ModePerm)
+	b.Logger.Infof("Writing modified configs to %s", request.OutDir)
+	return WriteDir(*decCfg, request.OutDir)
```
joelanford, line 71: This will always rewrite config dirs into our "preferred" structure. There are probably some benefits to that, but I could also see users wanting us to not touch their existing structure.

One potential option would be to search the configs directory for the file containing the `olm.package` for this bundle, and write a new file with just this `olm.bundle` blob.

pkg/declcfg/addTocfg.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		bundleConfig := ConvertFromModel(*m)
+		if !existingPackage(decCfg, img.Bundle.Package) {
+			b.Logger.Infof("adding new package metadata to configs")
+			decCfg.Packages = append(decCfg.Packages, bundleConfig.Packages...)
+		}
+		b.Logger.Infof("adding bundle metadata to configs")
+		decCfg.Bundles = append(decCfg.Bundles, bundleConfig.Bundles...)
+	}
+	os.RemoveAll(request.OutDir) // this is required because WriteDir only writes to an empty directory
```
joelanford, line 68: We probably need to do the equivalent of this:
1. `mv configs configs.bak`
2. write updated configs
3. `if $(2); then rm -r configs.bak; else mv configs.bak configs; fi`

That way we preserve the original configs if there's an error writing the new configs.

pkg/registry/csv.go
```diff
@@ -341,3 +352,48 @@ func (csv *ClusterServiceVersion) GetOperatorImages() (map[string]struct{}, erro
 
 	return images, nil
 }
+
+type Icon struct {
+	MediaType  string `json:"mediatype"`
+	Base64data string `json:"base64data"`
```
joelanford, line 358: If you use `Base64data []byte`, the json unmarshaller will automatically decode the base64 string into the binary []byte data.

Since this is a new type and it is presumably only being used in the new `opm add` command, I think we could start enforcing that any newly added bundles must have valid icon data, and error out if they don't.

The only catch is that if index maintainers want to rebuild existing indexes in this format from existing bundles that have this issue, that won't work. In that case, perhaps we could log a warning and skip setting the icon?

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+	mIcon := &model.Icon{
+		// looks like the CSV icon is a slice of objects https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go#L283
+		// Any better idea here than just taking the first element of the slice and working with that? Since I'm assuming in most cases users only pass in one element (icon)
+		// in the CSV anyway
+		MediaType: i[0].MediaType,
+		Data:      []byte(i[0].Base64data),
+	}
+	pkg := &model.Package{
+		Name:        b.Annotations.PackageName,
+		Description: desc,
```
joelanford, line 30: Is the `declcfg.Package.Description` field meant to hold the huge long csv description? (I need to double check to see what I did when converting from the sqlite DB into the model).

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+		Icon:        mIcon,
+	}
+	mb, err := registryBundleToModelBundle(b)
+	mb.Package = pkg
+	if err != nil {
+		return nil, err
+	}
+	mbs := make(map[string]*model.Bundle)
+	mbs[b.csv.Name] = mb
+	annotationsChannels := strings.Split(b.Annotations.Channels, ",")
```
joelanford, line 40: We might want to trim any whitespace before or after each comma-separated channel, and probably ignore empty strings.

So basically, this: `"      ch1, ch2   , ,, ch3   "` should turn into `[]string{"ch1","ch2","ch3"}`

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	for _, p := range providedGVKs {
+		out = append(out, property.MustBuildGVKProvided(p.Group, p.Version, p.Kind, p.Plural))
+		out = append(out, property.MustBuildGVK(p.Group, p.Version, p.Kind, ""))
+	}
+
+	for _, p := range requiredGVKs {
+		out = append(out, property.MustBuildGVKRequired(p.Group, p.Version, p.Kind, p.Plural))
+	}
```
joelanford, line 212: A bunch of this looks duplicated (and some of this code is stale since the `plural` field was removed from the `serve` PR.

Can this be refactored? If necessary and it makes sense to do, we could consider putting any remaining duplicated code in a centralized place.

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+
+	return out, nil
+}
+
+func convertToModelRelatedImages(csv *ClusterServiceVersion) ([]model.RelatedImage, error) {
```
joelanford, line 217: I have a similar implementation here: https://github.com/joelanford/operator-registry/blob/f81871a10201ea7e096f75361701f12fc53220c1/pkg/api/api_to_model.go#L128-L149

If we can find a place to centralize, perhaps we should?

joelanford (commented):
Looks like a great start! Have some comments and suggestions.

cmd/opm/root/add.go
```diff
@@ -0,0 +1,108 @@
...
+				logrus.SetLevel(logrus.DebugLevel)
+			}
+			return nil
+		},
+
+		RunE: addFunc,
+	}
+
+	rootCmd.Flags().Bool("debug", false, "enable debug logging")
+	rootCmd.Flags().StringP("mode", "", "replaces", "graph update mode that defines how channel graphs are updated. One of: [replaces, semver, semver-skippatch]")
```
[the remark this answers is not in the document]
anik120: Thanks for noticing this, missed stripping this out after the conversation.

pkg/registry/registry_to_model.go
anik120: It is

pkg/declcfg/addTocfg.go
anik120: @joelanford this takes us back to our discussion about `WriteFile` vs `writeFile` in the `declCfg` package. I thought the decision was to only expose `WriteFile`, which would translate to: If you use any tool that uses the `declCfg` package, your directory will get rewritten. 

TBH I am in favor of controlling how the directory is laid out, to enforce some hygiene.

pkg/declcfg/declcfg.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type RelatedImage struct {
+	Name  string `json:"name"`
+	Image string `json:"image"`
+}
+
+type Meta struct {
+	Schema  string
+	Package string
```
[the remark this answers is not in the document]
joelanford: You'll probably want to put these comments in #601, since that's where the `pkg/declcfg` originates.

But to answer this question... `Package` is optional in the meta blob, and is helpful to align package-specific blobs with their `olm.package`. There's a blurb about this in the EP [here](https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md#filesystem-structure).

pkg/declcfg/declcfg.go
```diff
@@ -0,0 +1,83 @@
...
+
+	Blob json.RawMessage
+}
+
+func (m Meta) MarshalJSON() ([]byte, error) {
+	return m.Blob, nil
```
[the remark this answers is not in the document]
joelanford: `m.Blob` is just the raw json that we found when we unmarshalled it. So it will contain whatever first-class fields are present.

pkg/declcfg/declcfg.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (m *Meta) UnmarshalJSON(blob []byte) error {
+	type tmp struct {
+		Schema     string              `json:"schema"`
+		Package    string              `json:"package,omitempty"`
+		Properties []property.Property `json:"properties,omitempty"`
```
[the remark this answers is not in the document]
joelanford: `Properties` is optional. If a JSON blob contains a `properties` field, it must parse as a list of `property.Property`.

pkg/declcfg/declcfg_to_model.go
```diff
@@ -0,0 +1,108 @@
+package declcfg
+
+import (
+	"fmt"
+
+	"github.com/operator-framework/operator-registry/pkg/model"
+	"github.com/operator-framework/operator-registry/pkg/property"
+)
+
+func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
```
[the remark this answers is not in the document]
joelanford: I haven't thought _too_ much about it, but since we're dealing with concrete types on both sides of the conversion, nothing is coming to mind.

pkg/declcfg/declcfg_to_model.go
```diff
@@ -0,0 +1,108 @@
+package declcfg
```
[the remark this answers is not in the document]
joelanford: There's a separate `model_to_declcfg.go`. We could combine both of these files into a `model.go` file. But I generally like to keep files in a package scoped somewhat small to make navigation and discovery easier. But I don't have any really strong feelings on this.

cmd/opm/root/add.go
```diff
@@ -0,0 +1,108 @@
...
+	rootCmd.Flags().StringP("out", "o", "", "the output directory where the modifed configs are written. Defaults to the input configs directory.")
+	rootCmd.Flags().StringP("ca-file", "", "", "the root Certificates to use with this command")
+	rootCmd.Flags().StringP("pull-tool", "p", "", "tool to pull container images. One of: [none, docker, podman]. Defaults to none. Overrides part of container-tool.")
```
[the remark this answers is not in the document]
anik120: Updating with `none` tool as the only option.

cmd/opm/root/add.go
```diff
@@ -0,0 +1,108 @@
...
+		return errors.New("number of positional argument mismatch error: please see help for more instructions on correct usage of the command")
+	}
+
+	pullTool, err := cmd.Flags().GetString("pull-tool")
+	if err != nil {
+		return err
+	}
+	if pullTool == "" {
+		pullTool = "none"
+	}
```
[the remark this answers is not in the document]
anik120: Removed the flag.

pkg/declcfg/addTocfg.go
```diff
@@ -0,0 +1,104 @@
...
+	cleanup := func() {
+		if err := os.RemoveAll(workingDir); err != nil {
+			logrus.Error(err)
+		}
+	}
+	if len(errs) > 0 {
+		return nil, "", cleanup, utilerrors.NewAggregate(errs)
+	}
+	return ref, workingDir, cleanup, nil
+}
```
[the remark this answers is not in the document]
anik120: Looks like the image registry I'm creating above has an Unpack implementation so I don't need to include this here!

pkg/declcfg/addTocfg.go
```diff
@@ -0,0 +1,104 @@
...
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/registry"
```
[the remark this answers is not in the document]
anik120: >I see the functionality present in this file as more of a client of pkg/declcfg

That is a really good point 

>Here are some alternate suggestions:
./cmd/opm/add/request.go
./pkg/action/add_bundles_to_configs.go
With option 1, we're basically saying this functionality is an implementation detail of the CLI, so it belongs alongside the CLI code in the cmd/opm/add package.
With option 2, we're saying a library of "actions" that the CLI uses, but other clients could use those actions too. This is how the Helm CLI is setup, and is a pattern I followed in the kubectl operator code.

Both the options sounds really good, so it's hard to choose one over the other just based on the meaning of choosing each option that's laid out here. In lieu of that, going with option 2 in the interest of standardizing design across projects unless I hear any objections.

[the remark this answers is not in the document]
joelanford: Thinking more, if we want to align actions to CLI commands, perhaps just `./pkg/action/add.go`?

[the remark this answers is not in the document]
njhale: I've actually seen `pkg/cmd` used in a lot of kubectl plugins/utilities build with cobra. It seems relatively clean to me.

e.g. https://github.com/njhale/kubectl-stash/blob/master/pkg/cmd/get.go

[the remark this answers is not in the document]
joelanford: That works too! I don't have a strong opinion on what package it ends up in. For me the main points are:
1. This package is completely standalone from the CLI implementation (i.e. it shouldn't mention cobra or flags, IMO, that's what `cmd/opm/add.go` is for).
2. It's separate for the underlying building blocks it uses (e.g. `pkg/declcfg`, `pkg/registry`, etc.)
3. It's designed such that it can be used as a standalone library. Come at it from the mindset of someone who needs this functionality vendored into their other non-operator-registry tool.

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	i, err := b.csv.GetIcons()
+	if err != nil {
+		return nil, fmt.Errorf("Could not get icon from bundle CSV:%s", err)
+	}
+	mIcon := &model.Icon{
+		// looks like the CSV icon is a slice of objects https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go#L283
+		// Any better idea here than just taking the first element of the slice and working with that? Since I'm assuming in most cases users only pass in one element (icon)
+		// in the CSV anyway
+		MediaType: i[0].MediaType,
```
joelanford, line 25: Picking the first one seems fine.

Also, check `len(i) > 0`?

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/operator-framework/operator-registry/pkg/model"
+	"github.com/operator-framework/operator-registry/pkg/property"
+)
+
+func CreateModelFromBundle(b *Bundle) (*model.Model, error) {
```
joelanford, line 12: Move logic from here to `model.Model.AddBundles([]model.Bundle)`?

We need a slice of `model.Bundle` because a `model.Bundle` is channel-specific, and a registry.Bundle can be a member of multiple channels.

Maybe we want `convertRegistryBundleToModelBundles(b *Bundle) ([]model.Bundle, error)`

pkg/registry/registry_to_model.go
joelanford: Disregard. We talked about how this code is necessary for now because we want conversion functions from the various types to also go into the model.

Unfortunately, we can't re-use the api conversion method because it takes `api.Bundle`, not `registry.Bundle`.

In the future, perhaps we can look at converting a bundle image directly to `model.Bundle`

pkg/registry/registry_to_model.go
joelanford: Need to remove the `plural` stuff, but disregard the other refactoring suggestion. Not easy to centralize when the type we're converting from is different (i.e. registry.Bundle vs api.Bundle)

pkg/declcfg/declcfg_to_model.go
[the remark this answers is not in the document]
anik120: I really like the break down to `model_to_declcfg.go` and `declcfg_to_model.go` instead of munging everything into one file for the same reason Joe mentioned.

pkg/registry/registry_to_model.go
anik120: Going towards `convertRegistryBundleToModelBundles(b *Bundle) ([]model.Bundle, error)`

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,241 @@
...
+	relatedImages, err := convertToModelRelatedImages(csv)
+	if err != nil {
+		return nil, fmt.Errorf("Could not get Related images from bundle: %v", err)
+	}
+
+	return &model.Bundle{
+		Name:          csv.Name,
+		Image:         b.BundleImage,
+		Replaces:      replaces,
+		Skips:         skips,
```
joelanford, line 92: I kind of wonder if we should specifically ignore `replaces` (and maybe `skips`) values from the CSV when adding bundles.

If we _don't_ ignore them, we're effectively making `opm add` build upgrade graphs in `replaces` mode, aren't we?

If we _do_ ignore them, a follow-on tool _must_ be used to insert the added bundles into the upgrade graph.

anik120: I don't think adding `replaces` from the CSV here is essentially "building" the graph. It's just adding the metadata, i.e expressing the *intent* to build the graph. I don't think it's possible to say you must use another tool always to effectively make your configs make sense.

pkg/declcfg/addTocfg.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+func (b BundleAdder) AddToConfig(request AddConfigRequest) error {
+	b.Logger.Infof("loading configs from directory")
+	decCfg, err := LoadDir(request.ConfigsDir)
+	if err != nil {
+		return fmt.Errorf("Error loading directory:%s", request.ConfigsDir)
+	}
+	for _, bundle := range request.Bundles {
```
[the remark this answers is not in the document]
anik120: Added `model.AddBundle(model.Bundle)`

anik120:
waiting on #617 to remove WIP tag

cmd/opm/root/add.go
```diff
@@ -0,0 +1,113 @@
...
+		}()
+		err = wait.PollImmediate(retryInterval, timeout, func() (done bool, e error) {
+			a.logger.Infof("Pulling bundle %q", simpleRef.String())
+			e = reg.Pull(cmd.Context(), simpleRef)
+			if err != nil {
+				a.logger.Info("Error pulling bundle %q. Trying to pull again.", simpleRef.String())
+				return false, err
+			}
+			return true, nil
+		})
```
joelanford, line 96: This will run until the timeout if the error is not retriable (e.g. the image reference is not found in the registry). I'd suggest something like this: https://pkg.go.dev/k8s.io/client-go/util/retry#OnError

cmd/opm/root/add.go
```diff
@@ -0,0 +1,113 @@
...
+		tmpDir, err := ioutil.TempDir("./", "bundle_tmp")
+		if err != nil {
+			return fmt.Errorf("error creating temp directory to unpack bundle image %q in:%v", simpleRef.String(), err)
+		}
+		defer func() {
+			a.logger.Infof("Removing temp directory %q bundle was unpacked in", tmpDir)
+			if err := os.RemoveAll(tmpDir); err != nil {
+				a.logger.Errorf("error removing temp directory %q bundle was unpacked in: %v", tmpDir, err)
+			}
+		}()
```
joelanford, line 87: Using defer in a for loop means that a bunch of deferred functions will be added to the defer stack and they only run when this entire function exits.

It might be better to encapsulate the bundle extract logic into its own function so that we clean up after ourselves after each bundle instead of after all bundles.

cmd/opm/root/add.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		a.logger.Infof("Unpacking bundle %q into %q", simpleRef.String(), tmpDir)
+		err = reg.Unpack(cmd.Context(), simpleRef, tmpDir)
+		if err != nil {
+			return fmt.Errorf("error unpacking image %q: %v", simpleRef.String(), err)
+		}
+		bundles = append(bundles, action.InputBundle{Dir: tmpDir, ImgRef: simpleRef})
+	}
+	request := action.AddConfigRequest{
+		Bundles:    bundles,
```
joelanford, line 108: IMO, all the logic to extract the bundles should be encapsulated in the `AddConfigRequest.AddToConfig()` function somewhere. Otherwise other theoretical clients of this package would have to do all this work themselves.

So I think we should just pass `args[1:]` into the AddConfigRequest as its Bundles:
```suggestion
		Bundles:    args[1:],
```

anik120: @joelanford I noticed that's what some of the current interfaces in the `opm index` land does that, i.e bakes in the work to extract and unpack bundles insides their implementations. However, that means that each of those interface implementations instantiate their own `image.Registry`, and @njhale pointed out earlier that that's what makes makes writing e2e tests for these hard.  

>Otherwise other theoretical clients of this package would have to do all this work themselves.

What's the con of telling the client "unpack the bundles yourself they way you desire, and then give me the unpacked bundles"? 

I kind of see the point that that separation of concern from `AddToConfig` and similar interfaces makes them easily testable.

cmd/opm/root/add.go
anik120: >Using defer in a for loop means that a bunch of deferred functions will be added to the defer stack and they only run when this entire function exits.

Isn't that what we want? 

1. Function begins 
2. Extract all the bundles into tmp folders
3. Call AddToConfig(request{bundles: tmpFolders})
4. Clean up all the bundles

Is the concern here that the defer stack will morph into something we cannot predict? If that's the case we'll have to say we cannot depend of defer, and instead re-design AddToConfig to only ever add one bundle at a time (which imho I was happy to avoid by leaning on the defer stack)

joelanford: I see. I had in my mind that we would internally just work on one image at a time. So something like:
1. Load Declcfg, convert to model
2. For each image
   - Pull/extract it
   - Add it to to the model
   - Cleanup the extracted dir
3. Convert model to declcfg, write declcfg

joelanford: My big concern is just the high-water mark of disk space used.

cmd/opm/root/add.go
joelanford: If that's the only problem, then I'd suggest making the image.Registry a field on the `AddConfigRequest` struct.

joelanford: > What's the con of telling the client "unpack the bundles yourself they way you desire, and then give me the unpacked bundles"?

Mainly it's that that callers of this function probably don't really care about the details of image unpacking. They're thinking "here's my image, add it to my declarative config".

I think having the image registry have an automatic default, but making it overridable is the way to go. We could use the functional options pattern to make this clean: https://medium.com/@dreissenzahn/functional-options-in-go-35279f535c6

pkg/registry/registry_to_model.go
joelanford: But `replaces` and `skips` define the graph. So if they're both in there, we effectively have built the graph, right?

cmd/opm/root/add.go
anik120: How does it tie in with what we've been talking about in terms of not baking in external tools to deal with images? 
If we change `BundleAdder` to accept an image instead of a directory, does it make it harder for us to port to `opm index add <bundle-directory> <configs-directory>`?

cmd/opm/root/add.go
anik120: I see your point. But here's the counter argument: 

> we would internally just work on one image at a time

that's not the contract you're signing when you're passing multiple bundle images in the args. You're saying "adding all of these bundles at once". If you are about to add 100s of bundles, it's better to chop that job up into pieces. 

The reason I say this is because this flow: 

```
1. Load Declcfg, convert to model
2. For each image
      Pull/extract it
      Add it to to the model
      Cleanup the extracted dir
6. Convert model to declcfg, write declcfg
```
translates to 
Load declcfg, convert to model 
For each image, AddBundleToModel (instead of AddToConifg) 
And that doesn't seem like the right way to go since that assumes that any and all consumers of BundleAdder must know what the model is, or at the very least must deal with the model before passing it on to BundleAdder, whereas the model is supposed to be an internal model?

internal/model/model_test.go
```diff
@@ -555,3 +555,98 @@ func makePackageChannelBundle() (*Package, *Channel) {
...
+				assert.Equal(t, len(s.model), existingPkgCount+1)
+			}
+			if s.numBundlesIncrease {
+				assert.Equal(t, getNoOfBundles(m, s.pkgBundleAddedTo), existingBundleCount+1)
+			}
+		})
+	}
+}
+
+func getNoOfBundles(m Model, pkg string) int {
```
dinhxuanvu, line 645: `countBundles` would be better.

pkg/action/add_bundle_to_configs.go
```diff
@@ -0,0 +1,83 @@
...
+	Logger *logrus.Entry
+}
+
+func NewBundleAdder(logger *logrus.Entry) BundleAdder {
+	return BundleAdder{
+		Logger: logger,
+	}
+}
+
+func (b BundleAdder) AddToConfig(request AddConfigRequest) error {
```
dinhxuanvu, line 31: A general comment on code formatting for all functions in this PR. You should add an empty line after each block of `{}` to make the code a bit more spacing-out so it will be easier to read and review. Line after line of code is too dense.

pkg/registry/registry_to_model.go
dinhxuanvu: `replaces` and `skips` define the update graph in `replaces` mode for sure. We can't simply ignore them because users include them to build the update graph unless they are using `semver` and `skippatch` mode.

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,232 @@
...
+		Description: desc,
+		Icon:        mIcon,
+		Channels:    make(map[string]*model.Channel),
+	}
+	mb, err := registryBundleToModelBundle(b)
+	mb.Package = pkg
+	if err != nil {
+		return nil, err
+	}
+	for _, ch := range extractChannels(b.Annotations.Channels) {
```
dinhxuanvu, line 42: One thing we need to discuss here is that even though default channel is not required in the annotations. However, users do use it to update the default channel if they include it. Do we still want to reserve that behavior or we are saying we will ignore the default channel info in annotations from now on and if they want to change default channel, they will need to edit the JSON files manually?

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	for p := range requiredGVKs {
+		out = append(out, property.MustBuildGVKRequired(p.Group, p.Version, p.Kind))
+	}
+
+	return out, nil
+}
+
+func convertToModelRelatedImages(csv *ClusterServiceVersion) ([]model.RelatedImage, error) {
```
dinhxuanvu, line 197: This func is the same with what Joe has on his [PR](https://github.com/operator-framework/operator-registry/pull/601/files#diff-485c5555a31cbf01e06186590ac6cf4b5acead26777114cdffeb884ed54ba376R128). It seems it would help with reducing the duplicate code, we should move this func to a command place somewhere to be reused for both PRs.

dinhxuanvu (commented):
Looking good so far. A few comments.

cmd/opm/registry/internal/mode/source_mode.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"os"
+
+	"github.com/h2non/filetype"
+	"github.com/h2non/filetype/matchers"
+)
+
+type Mode string
+
+const (
```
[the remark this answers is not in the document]
anik120: This is outdated and isn't part of any commits now. Strange that it's showing up for you (and concerning :P )

pkg/action/add_bundle_to_configs.go
anik120: Do you mean empty line at line 30 (which is there right now) or at line 32? 
In a previous PR when I introduced an empty line in similar lines like 32, I've been told the preferred approach for this repo is to not include an empty line. 

I prefer the empty line at 32 too, but as long as we can agree with one convention and stick to it I'm fine with either 🤷🏽‍♂️

pkg/registry/registry_to_model.go
anik120: "We" as in opm ins't building the graph, opm is simply translating the intent of the user of opm (communicated via the bundle manifests) onto the configs. Essentially the user is building the graph, we're merely transplanting the very thing the user communicated in the bundle in a very non-opinionated manner onto the configs.

pkg/action/add_bundle_to_configs.go
dinhxuanvu: No. I mean after a block of code `{}`, NOT right after the func signature. For example, at like 36, there should be an empty line afterwards. You can look at [this code](https://github.com/cuelang/cue/blob/master/cuego/cuego.go#L131) for example on how they have empty lines in between block of code to improve readability.

cmd/opm/registry/internal/mode/source_mode.go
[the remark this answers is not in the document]
dinhxuanvu: Probably an old comment that is leftover from previous review.

pkg/registry/registry_to_model.go
anik120: I think the decision was to go with the following: 
1) When the first bundle is added to the package, if default channel isn't specified, pick one and assign it as default (similar to what happens in the db now) 
2) If you want to update the default channel, the workflow of expressing that intent with specifying the new default channel in a bundle you're adding is outdated, and we'll ignore all further metadata about default channels starting from the second bundle in the package. If you want to edit the default channel, you must manually edit it in the configs.

dinhxuanvu: 1. The current behavior is we pick the first channel in lexicographically order. So do we want to reserve that behavior?
2. Sure. That's fine to me but then that's a change from existing `index add` command. So make sure we document it somewhere.

pkg/registry/registry_to_model.go
anik120: We've discussed this, and since we're dealing with two different types of bundles here (api bundles vs registry bundles) I don't think there's much value in trying to find a common place to put just this function while the rest of the things are unique. Not yet at least.

cmd/opm/root/add.go
anik120: @joelanford changed the design of `BundleAdder` to unpack and remove one bundle at a time.

pkg/registry/registry_to_model.go
anik120: 1. Yes we do. 
2. Sure thing.

pkg/registry/registry_to_model.go
dinhxuanvu: In term of func signature (input, output), they are pretty much identical. Yours accepts `*ClusterServiceVersion` while Joe's accepts a CSV JSON string (which is converted to CSV object later) and both output `([]model.RelatedImage, error)`. They can be just one func with small adjustment on the input side. Either way, this is not a blocker and both of your PRs are in parallel pending so we can take care of this at another time.

pkg/action/add_bundle_to_configs.go
anik120: Done!

pkg/registry/registry_to_model.go
dinhxuanvu: I agreed. We are moving away from the `mode` style and this should be documented somewhere.

internal/model/model.go
```diff
@@ -293,3 +293,29 @@ func (m Model) Normalize() {
...
+func (m Model) AddBundle(b Bundle) {
+	if _, present := m[b.Package.Name]; !present {
+		m[b.Package.Name] = b.Package
+	}
+	p := m[b.Package.Name]
+	b.Package = p
+
+	if ch, ok := p.Channels[b.Channel.Name]; ok {
+		b.Channel = ch
+		ch.Bundles[b.Name] = &b
```
joelanford, line 306: Do we need to check if the bundle is already present in the channel?

pkg/action/add_bundle_to_configs.go
```diff
@@ -0,0 +1,88 @@
...
+		return fmt.Errorf("error loading directory %q: %v", request.ConfigsDir, err)
+	}
+
+	model, err := declcfg.ConvertToModel(*decCfg)
+	if err != nil {
+		return fmt.Errorf("error converting configs to internal model:%v", err)
+	}
+
+	for _, bundle := range request.Bundles {
+		b, err := bundle.ExtractBundle(context.TODO())
```
joelanford, line 44: We should plumb the context in from the cobra command's `cmd.Context()` instead of using `context.TODO()`. That would mean adding a context parameter to this `AddToConfig` function.

pkg/action/add_bundle_to_configs.go
```diff
@@ -0,0 +1,88 @@
...
+		return fmt.Errorf("error writing new configs into %q:%v", tmpDir, err)
+	} else {
+		if err := os.RemoveAll(request.ConfigsDir); err != nil { // this is required because WriteDir only writes to an empty directory
+			return fmt.Errorf("could not remove existing output dir %q: %v", request.ConfigsDir, err)
+		}
+		if err := os.Mkdir(request.ConfigsDir, os.ModePerm); err != nil {
+			return fmt.Errorf("could not create new directory %q: %v", request.ConfigsDir, err)
+		}
+		b.Logger.Infof("rewriting contents of existing configs directory %q with contents from %q", request.ConfigsDir, tmpDir)
+		err = copy.Copy(tmpDir, request.ConfigsDir)
```
joelanford, line 73: Can this just be `os.Rename()`?

pkg/action/add_bundle_to_configs.go
```diff
@@ -0,0 +1,88 @@
...
+
+	b.Logger.Infof("writing modified temp directory to %q", tmpDir)
+	err = declcfg.WriteDir(newDecCfg, tmpDir)
+
+	if err != nil {
+		return fmt.Errorf("error writing new configs into %q:%v", tmpDir, err)
+	} else {
+		if err := os.RemoveAll(request.ConfigsDir); err != nil { // this is required because WriteDir only writes to an empty directory
+			return fmt.Errorf("could not remove existing output dir %q: %v", request.ConfigsDir, err)
+		}
```
joelanford, line 68: What if the input directory is under version control (e.g. has a `.git` directory or similar)?

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,235 @@
...
+			var v property.GVK
+			if err := json.Unmarshal(json.RawMessage(p.Value), &v); err != nil {
+				return nil, property.ParseError{Idx: i, Typ: p.Type, Err: err}
+			}
+			k := property.GVKRequired{Group: v.Group, Kind: v.Kind, Version: v.Version}
+			requiredGVKs[k] = struct{}{}
+		case property.TypePackage:
+			out = append(out, property.Property{
+				Type:  property.TypePackageRequired,
+				Value: json.RawMessage(p.Value),
```
joelanford, line 153: This is buggy (see the fix in #625)

pkg/registry/registry_to_model.go
```diff
@@ -0,0 +1,235 @@
...
+				return nil, property.ParseError{Idx: i, Typ: p.Type, Err: err}
+			}
+			k := property.GVK{Group: v.Group, Kind: v.Kind, Version: v.Version}
+			providedGVKs[k] = struct{}{}
+		case property.TypePackage:
+			var v property.Package
+			if err := json.Unmarshal(json.RawMessage(p.Value), &v); err != nil {
+				return nil, property.ParseError{Idx: i, Typ: p.Type, Err: err}
+			}
+			out = append(out, property.MustBuildPackageRequired(v.PackageName, v.Version))
```
joelanford, line 132: This seems incorrect. We should just be converting the `olm.package` properties directly (not changing them to `PackageRequired` properties). Also, shouldn't we be setting `foundPackageProperty = true` here?

This may have been a screw up on my part since, since I know you cribbed from me.

pkg/registry/registry_to_model_test.go
```diff
@@ -0,0 +1,71 @@
...
+		Replaces: "etcdoperator.v0.9.0",
+		Skips:    []string{"etcdoperator.v0.9.1"},
+		Properties: []property.Property{
+			property.MustBuildChannel("alpha", "etcdoperator.v0.9.0"),
+			property.MustBuildChannel("stable", "etcdoperator.v0.9.0"),
+			property.MustBuildPackage("etcd", "0.9.2"),
+			property.MustBuildSkips("etcdoperator.v0.9.1"),
+			//TODO(anik120): check if etcdclusters. is supposed to be prefixed
+			property.MustBuildGVKRequired("etcd.database.coreos.com", "v1beta2", "EtcdCluster"),
+			property.MustBuildGVKRequired("testapi.coreos.com", "v1", "testapi"),
```
joelanford, line 40: Not sure if it matters in this test, but kinds are always uppercased.
```suggestion
			property.MustBuildGVKRequired("testapi.coreos.com", "v1", "Testapi"),
```

joelanford:
/lgtm

I think we can resolve my concerns in a follow-on PR.

dinhxuanvu (approved):
/lgtm

PR #583, opened 2021-02-17
fix: copy out only the selected file(s) from index images via a tar reader

exdx:

Pruning an index via `opm index prune` can fail when permissions on the filesystem in the container are read-only. Since the command copies all files out from the container it is prone to failure. Fora example, the red hat operators catalog is based on the UBI image, which contains some directories with elevated permissions (`/root/*` for example). Extract the entire contents of this container filesystem via `docker cp` will fail. 

The extract database function relies on shelling out to external tools (docker or podman) and does not use the existing containerd unpacking implementation (since that does not support building images). These tools cannot dynamically change the filesystem permissions of files as they are copied out of the container filesystem, based on observed behavior. The solution is to instead copy the selected file(s) to stdout as a tar file and then read -> write them to the local filesystem with the permissions necessary to ensure the copying behavior succeeds always. 

Fixes pruning the red hat operators catalog when running opm as a non-root user. Copying only the database file out from the catalog is simpler as well, since we knows its location ahead of time (via the label). 

**Reviewer Checklist**

### DIFF
--- a/pkg/containertools/exporter.go
+++ b/pkg/containertools/exporter.go
@@ -10,40 +10,38 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// Exporter can read a tar archive being passed along from a writer and write it to a local destination.
-type Exporter struct {
+// exporter can read a tar archive being passed along from a writer and write it to a local destination.
+type exporter struct {
 	dest   string
 	writer io.WriteCloser
 	reader *tar.Reader
 	log    *logrus.Entry
 }
 
-func NewExporter(dest string, logger *logrus.Entry) (*Exporter, error) {
-	// use stdout as pipew
-	piper, pipew := io.Pipe()
+func newExporter(dest string, logger *logrus.Entry, reader *tar.Reader, writer io.WriteCloser) (*exporter, error) {
 	d, err := filepath.Abs(dest)
 	if err != nil {
 		return nil, err
 	}
 
-	return &Exporter{
+	return &exporter{
 		dest:   d,
-		writer: pipew,
-		reader: tar.NewReader(piper),
+		writer: writer,
+		reader: reader,
 		log:    logger,
 	}, nil
 }
 
-func (e *Exporter) Writer() io.WriteCloser {
+func (e *exporter) Writer() io.WriteCloser {
 	return e.writer
 }
 
-func (e *Exporter) Close() error {
+func (e *exporter) Close() error {
 	return e.writer.Close()
 }
 
 // Run reads the tar stream from reader as it comes from stdout and writes the file/dir to the destination specified.
-func (e *Exporter) Run() error {
+func (e *exporter) Run() error {
 	for {
 		header, err := e.reader.Next()
 		if err == io.EOF {
--- a/pkg/containertools/exporter.go
+++ b/pkg/containertools/exporter.go
@@ -1,0 +1,96 @@
+package containertools
+
+import (
+	"archive/tar"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Exporter can read a tar archive being passed along from a writer and write it to a local destination.
+type Exporter struct {
+	dest   string
+	writer io.WriteCloser
+	reader *tar.Reader
+	log    *logrus.Entry
+}
+
+func NewExporter(dest string, logger *logrus.Entry) (*Exporter, error) {
+	// use stdout as pipew
+	piper, pipew := io.Pipe()
+	d, err := filepath.Abs(dest)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Exporter{
+		dest:   d,
+		writer: pipew,
+		reader: tar.NewReader(piper),
+		log:    logger,
+	}, nil
+}
+
+func (e *Exporter) Writer() io.WriteCloser {
+	return e.writer
+}
+
+func (e *Exporter) Close() error {
+	return e.writer.Close()
+}
+
+// Run reads the tar stream from reader as it comes from stdout and writes the file/dir to the destination specified.
+func (e *Exporter) Run() error {
+	for {
+		header, err := e.reader.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+
+		// determine proper file path info
+		finfo := header.FileInfo()
+		fileName := header.Name
+		absFileName := filepath.Join(e.dest, fileName)
+
+		// if a dir, create it, then go to next segment
+		if finfo.Mode().IsDir() {
+			if err := os.MkdirAll(absFileName, os.ModePerm); err != nil {
+				e.log.Debugf("creating %s dir", absFileName)
+				return err
+			}
+			continue
+		}
+
+		// create new file with custom file permissions
+		// TODO sync with existing containerd permissions
+		file, err := os.OpenFile(
+			absFileName,
+			os.O_RDWR|os.O_CREATE|os.O_TRUNC,
+			os.ModePerm,
+		)
+		if err != nil {
+			return err
+		}
+
+		e.log.Debugf("writing %s to disk", absFileName)
+		n, err := io.Copy(file, e.reader)
+		if err != nil {
+			return err
+		}
+		if closeErr := file.Close(); closeErr != nil {
+			return closeErr
+		}
+		if n != finfo.Size() {
+			return fmt.Errorf("unpacking to disk: wrote %d, want %d", n, finfo.Size())
+		}
+	}
+
+	e.log.Debugf("extracted contents from container filesystem")
+	return nil
+}
--- a/pkg/containertools/exporter_test.go
+++ b/pkg/containertools/exporter_test.go
@@ -2,8 +2,11 @@ package containertools
 
 import (
 	"archive/tar"
+	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
+	"reflect"
 	"sync"
 	"testing"
 
@@ -17,7 +20,7 @@ var files = map[string]string{
 }
 
 // 1. create a tar archive in memory
-// 2. setup Exporter with Writer pipe
+// 2. setup exporter with Writer pipe
 // 4. read tar archive contents to a temporary directory on disk
 // 5. Ensure file contents are there
 // 6. Ensure permissions are set as expected
@@ -29,7 +32,9 @@ func TestExporter_Run(t *testing.T) {
 	}
 	defer os.RemoveAll(temp)
 
-	exporter, err := NewExporter(temp, log)
+	piper, pipew := io.Pipe()
+	reader := tar.NewReader(piper)
+	exporter, err := newExporter(temp, log, reader, pipew)
 	if err != nil || exporter == nil {
 		t.Error(err)
 	}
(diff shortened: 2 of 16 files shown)


njhale (changes requested):
Good work!

We're going to want some test cases to ensure this fixes the issue.

There's [another bug](https://bugzilla.redhat.com/show_bug.cgi?id=1934931) with the same root cause, but different symptoms. Let's test that too.

I think we could also run into this situation with a home-grown index image has a database file with elevated privileges, so we'll also need to change the permissions of the files _as we copy them_. Docker `cp`'s observed behavior doesn't enable this directly, but we may be able to work around that limitation.  If we take advantage of the fact that it can be configured to write a tar file to stdout, it should be possible to write the files with any permissions we want. (see https://docs.docker.com/engine/reference/commandline/container_cp/#extended-description and https://golang.org/pkg/archive/tar/#Reader.Next)

exdx:
> Good work!
> 
> We're going to want some test cases to ensure this fixes the issue.
> 
> There's [another bug](https://bugzilla.redhat.com/show_bug.cgi?id=1934931) with the same root cause, but different symptoms. Let's test that too.
> 
> I think we could also run into this situation where a home-grown index image has a database file with elevated privileges, so we'll also need to change the permissions of the files _as we copy them_. Docker `cp`'s observed behavior doesn't enable this directly, but we may be able to work around that limitation. If we take advantage of the fact that it can be configured to write a tar file to stdout, it should be possible to write the files with any permissions we want. (see https://docs.docker.com/engine/reference/commandline/container_cp/#extended-description and https://golang.org/pkg/archive/tar/#Reader.Next)

This approach makes sense to me. Instead of unpacking the container filesystem directly (or individually by file), write the tar file (containing all the parent layers, ie the entire image contents) to stdout and then read it in memory to then write the archive out on disk (with the permission settings we need). 

I wonder if the containerd unpacking implementation you did recently would eliminate the need for this, I need to double check.  

Also podman supports more advanced forms of copying that enable direct use of linux primitives like mounting and unmounting filesystems https://manpages.debian.org/unstable/podman/podman-cp.1.en.html#OPTIONS. I'm not sure we necessarily want to support a different implementation for podman however.

pkg/image/containerdregistry/registry.go
```diff
@@ -91,7 +91,7 @@ func (r *Registry) Pull(ctx context.Context, ref image.Reference) error {
 
 // Unpack writes the unpackaged content of an image to a directory.
 // If the referenced image does not exist in the registry, an error is returned.
-func (r *Registry) Unpack(ctx context.Context, ref image.Reference, dir string) error {
+func (r *Registry) Unpack(ctx context.Context, ref image.Reference, from, to string) error {
```
njhale, line 94: I just thought of an alternative to changing the interface here: Create a new type that implements `image.Reference` but maybe also implements some other interface that we can make an assertion on; e.g.

```
type RootedReference interface {
    Reference
    RootDir() string
}

type SimpleRootedReference struct {
    SimpleReference
    Root
}

func (r SimpleRootedReference) RootDir() string {
    return r.Root
}
```

Here's a playground link that makes sure this can be used the way I think it can: https://play.golang.org/p/tlF1e_joEHo

pkg/image/registry.go
```diff
@@ -14,9 +14,9 @@ type Registry interface {
 	// If the referenced image does not exist in the registry, an error is returned.
 	// Push(ctx context.Context, ref string) error
 
-	// Unpack writes the unpackaged content of an image to a directory.
+	// Unpack recursively copies the filesystem contents of an image from the specified location to a local directory.
 	// If the referenced image does not exist in the registry, an error is returned.
-	Unpack(ctx context.Context, ref Reference, dir string) error
+	Unpack(ctx context.Context, ref Reference, from, to string) error
```
njhale, line 19: nit: I wonder if `to` should become `fs.FS`, or if that's too prescriptive.

pkg/image/containerdregistry/registry.go
```diff
@@ -100,13 +100,13 @@ func (r *Registry) Unpack(ctx context.Context, ref image.Reference, dir string)
 		return err
 	}
 
-	if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+	if err := os.MkdirAll(to, os.ModePerm); err != nil {
 		return err
 	}
 
 	for _, layer := range manifest.Layers {
 		r.log.Debugf("unpacking layer: %v", layer)
```
njhale, line 108: Looks like `from` isn't being respected yet?

njhale (commented):
Really good work so far.

pkg/containertools/exporter.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Exporter can read a tar archive being passed along from a writer and write it to a local destination.
+type Exporter struct {
```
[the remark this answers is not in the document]
exdx: SGTM

pkg/containertools/exporter.go
```diff
@@ -0,0 +1,96 @@
...
+	dest   string
```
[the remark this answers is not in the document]
exdx: yes, I was looking into that. I can try it out, really all the exporter needs is a filepath to write to, but we can make it a filesystem instead.

pkg/containertools/exporter_test.go
```diff
@@ -0,0 +1,96 @@
...
+		info, err := entry.Info()
+		if err != nil {
+			t.Error(err)
+		}
+		// the file mode is changed from 0777 to 0755 due to the default umask 022
+		if info.Mode() != os.FileMode(0755) {
+			t.Errorf("unexpected file mode %s, expected %s", info.Mode(), os.FileMode(0755))
+		}
+	}
+}
```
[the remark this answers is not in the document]
exdx: How would we go about that, take the file from the stream and compare it to the on-disk file, and make sure they are equal? Does that give us much more than checking the number of bytes written?

pkg/containertools/exporter.go
```diff
@@ -0,0 +1,96 @@
...
+	writer io.WriteCloser
+	reader *tar.Reader
+	log    *logrus.Entry
+}
+
+func NewExporter(dest string, logger *logrus.Entry) (*Exporter, error) {
```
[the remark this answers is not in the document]
exdx: That could work

pkg/containertools/runner.go
```diff
@@ -137,18 +138,49 @@ func (r *ContainerCommandRunner) Unpack(image, src, dst string) error {
...
+		exporter.Close()
 		return fmt.Errorf("error copying container directory %s: %v", string(out), err)
 	}
 
+	if expError != nil {
+		exporter.Close()
+		return expError
+	}
+
+	if err := exporter.Close(); err != nil {
```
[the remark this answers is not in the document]
exdx: so calling `Close()` sends the EOF signal on the writer side of the pipe, which causes the tar reader to stop reading, which results in the goroutine exiting. 

`wg.Wait()` without the `Close()` first causes the test case to hang, as the read and write pipes don't close properly, from what I've observed. Maybe it's because I'm using a map with dummy files as a data structure versus a proper tar archive though.

pkg/image/containerdregistry/registry.go
exdx: Yes, in the containerd case, since you're unpacking by layer after looking at the manifest file, I haven't added the `from` since in this case unpacking the entire filesystem makes sense (and the edge cases are resolved). I'm not sure we want to change that behavior in PR necessarily, since the goal is to fix the issue in unpacking when opm shells out to external tools

pkg/image/registry.go
exdx: All we really need is a filepath to write to, but I can see if fs.FS is a good option if it brings some additional benefits.

pkg/image/containerdregistry/registry.go
exdx: I think I'm following, but we end up mixing the image reference with the filepath on the system in the same type right? It's kind of confusing. 

In a previous review you mentioned this interface being underspecified, I agreed with that. I think we should change it, if we are ok with making a change to a public interface, which I believe we are.

pkg/containertools/exporter.go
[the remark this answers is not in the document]
njhale: Cool, thank you!

pkg/image/containerdregistry/registry.go
njhale: > It's kind of confusing.

I could see that view point. My head was automatically phrasing it as "a higher fidelity reference" type.

> In a previous review you mentioned this interface being underspecified

Yeah, I was just trying to come up with a way to work with the existing interfaces in a way that could make sense.

> I think we should change it, if we are ok with making a change to a public interface, which I believe we are.

I think the risk is low for this particular interface, although I would like to get one more set of eyes on this to see what they think. I don't think this is worth agonizing over at this point though.

pkg/image/containerdregistry/registry.go
njhale: I think this **needs** to respect the interface's contract and only copy the files/directories on the path specified with the `from` argument. It would be very odd if this implementation ignores the `from` argument.

The `unpackLayer` method can filter out the paths that get copied, so this may just amount to adding another filter there.

pkg/image/registry.go
njhale: it's probably a discussion for a later date.

pkg/image/containerdregistry/registry.go
njhale: Spoke with @benluddy on this and got some positive feedback for the interface change.

pkg/containertools/exporter.go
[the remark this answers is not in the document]
njhale: outcome: `fs.FS` doesn't define write methods, so we'd need to define our own extension interface anyway, so I'm not going to push this ask in a bugfix -- we can refactor at some point down the road.

pkg/image/containerdregistry/registry.go
njhale: looks like this was fixed! nice

pkg/image/containerdregistry/registry.go
```diff
@@ -261,3 +265,32 @@ func dropXattrs(h *tar.Header) (bool, error) {
...
 }
+
+// only unpack specified files (if not root)
+func fromFilter(from string) archive.Filter {
+	if from == image.Root {
+		// no-op, unpacks everything
+		return func(h *tar.Header) (bool, error) {
+			return true, nil
+		}
+	}
```
njhale, line 276: nice optimization!

pkg/image/containerdregistry/registry.go
```diff
@@ -207,8 +207,12 @@ func (r *Registry) unpackLayer(ctx context.Context, layer ocispec.Descriptor, di
 		return err
 	}
 
-	filters := filterList{adjustPerms, dropXattrs}
-	_, err = archive.Apply(ctx, dir, decompressed, archive.WithFilter(filters.and))
+	// create file filter when unpacking, if the from argument is not root
+	// return a noOp if unpacking the entire filesystem
+	fileFilter := fromFilter(from)
+
+	filters := filterList{adjustPerms, dropXattrs, fileFilter}
```
njhale, line 214:
```suggestion
	filters := filterList{
	    adjustPerms, // Prevent file write issues by tweaking copied file permissions (chmod 777)
	    dropXattrs, // Prevent file attribute permission issues by eliding all xattributes
	    fromFilter(from), // Copy only the given file or directory tree
	}
```

pkg/image/containerdregistry/registry.go
```diff
@@ -261,3 +265,32 @@ func dropXattrs(h *tar.Header) (bool, error) {
...
+			// only unpack this file
+			if from != h.FileInfo().Name() {
+				return false, nil
+			}
+			return true, nil
+		}
+		// we have a directory
+		// unpack the directory and all files under this directory
+		// TODO ensure directory is an exact match and only files under the heirarchy are pulled
+		if from == h.FileInfo().Name() || strings.Contains(from, h.FileInfo().Name()) {
```
njhale, line 291: I don't think this will pull the hierarchy correctly

https://play.golang.org/p/MB2vgqucZhG

If I'm correct, then I think the simplest fix would be to use `strings.HasPrefix` instead: https://pkg.go.dev/strings#HasPrefix

/hold

njhale:
superseded by #775

PR #587, opened 2021-02-19
Deprecate initializer and registry-server

cdjohnson:
Per a discussion on the olm-dev triage call on Feb 4th, it was agreed that both the `initializer` and `registry-server` binaries should be deprecated in favor of `opm` equivalent commands.

The `registry-server` is redundant and is no longer used in mainstream catalog index images, AFAIK.

The `initializer` approach doesn't use `bundle images` which is not strategic, which also requires the sqlite database to encode the entire CSV and Manifests and doesn't work well with the upcoming declarative index feature.

**Reviewer Checklist**


dinhxuanvu (approved):
/lgtm

PR #588, opened 2021-02-22
.github,test: Re-enable Podman e2e tests

timflannagan:
Update the test/e2e/opm_test.go and avoid returning early so the Podman-related e2e tests do not get run.

Update the test Github action workflow and ensure we get a new Podman dependency for both of the minikube and kind e2e tests.

Previously, we were seeing problems when attempting to update/upgrade our existing dependencies from the latest ubuntu image, which contains a default Podman installation. This resulted in failed action runs due to older dependencies being packaged in that image.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,14 +24,17 @@ jobs:
           file: coverage.out
           fail_ci_if_error: false
   e2e-minikube:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - run: |
-          sudo apt-get -y update
-          sudo apt-get -y upgrade
-          sudo apt-get -y install conntrack
-          podman version
+          sudo apt-get update \
+            && sudo apt-get remove buildah podman -y \
+            && sudo apt-get autoremove -y \
+            && sudo apt-get upgrade \
+            && sudo apt-get install podman dbus-x11 conntrack -y \
+            && podman version
+      - run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -24,14 +24,17 @@ jobs:
           file: coverage.out
           fail_ci_if_error: false
   e2e-minikube:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
dinhxuanvu, line 27: I thought 20.04 is the latest. So is this change needed? Or you want to pin a version for a reason?

dinhxuanvu (approved):
/approve

.github/workflows/test.yml
timflannagan: I was tempted to add a review comment note for this change as it's technically a no-op (latest is the 20.04 LTS release) but I'm typically a fan of using explicit versions rather than relying on potentially moving latest tags. I don't have a strong opinion either way so I'm fine to revert.

dinhxuanvu: I'm fine with it given it is a LTS version. However, we do need to manually update this if a newer version requires in the future. I don't think it is gonna be an issue.

test/e2e/opm_test.go
```diff
@@ -493,8 +493,6 @@ var _ = Describe("opm", func() {
 	})
 
 	Context("using podman", func() {
-		//FIXME skip podman tests due to failed to add podman to systemd sandbox cgroup
-		return
```
njhale, line 497: interesting -- this isn't how we should be skipping tests in the first place (see the ginkgo `Skip` function).

PR #590, opened 2021-02-22
Update the vendor Makefile target and update the vendored dependencies 

timflannagan:

**Reviewer Checklist**


go.mod
```diff
@@ -55,7 +55,6 @@ require (
 	k8s.io/api v0.20.0
 	k8s.io/apiextensions-apiserver v0.20.0
 	k8s.io/apimachinery v0.20.0
-	k8s.io/apiserver v0.20.0
```
timflannagan, line 58: I was looking through the codebase and noticed that the apiserver dependency was unused. It looks like running `go mod tidy` locally backed up this assumption.

Makefile
```diff
@@ -66,7 +66,9 @@ image-upstream:
 
 .PHONY: vendor
 vendor:
+	$(GO) mod tidy
```
dinhxuanvu, line 69: Interesting. This is how it is on OLM repo but apparently it is not over here.

dinhxuanvu: 👍

go.mod
dinhxuanvu: We used to vendor more k8s stuff in this repo and had a bit of circular deps between OLM and registry. We decoupled them and resolved those issues. So this is probably leftover that is not cleaned-up.

dinhxuanvu (commented):
/approve

go.mod
timflannagan: Yeah, that's understandable. I was looking through the codebase and noticed that, and I have Go 1.16 locally which was able to spot that in the go.mod. This could also just be an artifact from the Go 1.15 bump from last week.

timflannagan: Actually, this kind of scenario is exactly why I added a `vendor` check in the metering-related repositories. I can add something similar as an action or update the prow configuration.

timflannagan: Nice, it looks like #596 is already producing CI feedback that prompted this PR. It looks like there may be some race conditions in the bundle unpacking test that are popping up this week. We need to merge this change first before attempting to merge that vendor action check.

joelanford (approved):
/lgtm

PR #591, opened 2021-02-22
cmd,pkg: Deflate unnecessary conditionals

timflannagan:
Remove unneeded if/else conditionals where the if/else both specify
return statements.

Deflate if/else conditionals where in loops using the `continue`
keyword.

Switch the if/else logic to avoid nested conditionals.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/utils_test.go
+++ b/pkg/lib/bundle/utils_test.go
@@ -70,8 +70,7 @@ func clearDir(dir string) {
 	for _, item := range items {
 		if item.IsDir() {
 			continue
-		} else {
-			os.Remove(filepath.Join(dir, item.Name()))
 		}
+		os.Remove(filepath.Join(dir, item.Name()))
 	}
 }
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -150,46 +150,45 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 				}
 				return err
 			}
-			if overwritten != "" {
-				// get all bundle paths for that package - we will re-add these to regenerate the graph
-				bundles, err := querier.GetBundlesForPackage(ctx, img.Bundle.Package)
-				if err != nil {
-					return err
-				}
-				type unpackedImage struct {
-					to      image.Reference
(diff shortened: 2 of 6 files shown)


pkg/lib/bundle/utils_test.go
```diff
@@ -70,8 +70,7 @@ func clearDir(dir string) {
 	for _, item := range items {
 		if item.IsDir() {
 			continue
-		} else {
-			os.Remove(filepath.Join(dir, item.Name()))
 		}
+		os.Remove(filepath.Join(dir, item.Name()))
```
timflannagan, line 74: There's already a continue in that first conditional check, so we don't need to wrap file removal logic and can remove that branch entirely.

joelanford (approved):
/lgtm

pkg/lib/registry/registry.go
```diff
@@ -150,46 +150,45 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
...
-					unpack := <-unpacked
-					if unpack.err != nil {
-						return unpack.err
-					}
-					overwriteImageMap[img.Bundle.Package][unpack.to] = unpack.from
-					if _, ok := unpackedImageMap[unpack.to]; ok {
-						delete(unpackedImageMap, unpack.to)
+			if overwritten == "" {
+				return fmt.Errorf("index add --overwrite-latest is only supported when using bundle images")
+			}
```
timflannagan, line 155: Just swapped the if/else conditional checks so we can avoid an unnecessary nested block of code.

Previously, this block looked like this:

```go
if overwritten != "" {
...
} else {
    return fmt.Errorf("index add --overwrite-latest is only supported when using bundle images")
}
```

We can instead just handle that case where the `overwritten` variable is an empty string and exit early, which allows us to deflate the logic for handling the non-empty scenario and improve readability.

dinhxuanvu (approved):
/lgtm

timflannagan:
It looks like codecov isn't overrideable, which is unfortunate. We already have the required labels here and I don't want to have the bot continue to retest while I try and properly update the codecov. Going to force this in.

PR #596, opened 2021-02-23
.github: Add a sanity check for Go packages

timflannagan:
Add a vendor check to the testing actions that runs `$(make) vendor` and
verifies that the vendor directory is up-to-date and no `git diff` has
been produced.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,18 @@ on:
       - '**'
       - '!doc/**'
 jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Install goimports
+        run: go install golang.org/x/tools/cmd/goimports@latest
+      - name: Run sanity checks
+        run: make vendor && make lint && git diff --exit-code
+
   build:
     runs-on: ubuntu-latest
     steps:
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,10 @@ vendor:
 	$(GO) mod vendor
 	$(GO) mod verify
 
+.PHONY: lint
+lint:
+	find . -name '*.go' -not -path "./vendor/*" | xargs goimports -w
+
 .PHONY: codegen
 codegen:
 	protoc -I pkg/api/ --go_out=pkg/api pkg/api/*.proto
--- a/pkg/api/grpc_health_v1/health.pb.go
+++ b/pkg/api/grpc_health_v1/health.pb.go
@@ -7,11 +7,12 @@
 package grpc_health_v1
 
 import (
+	reflect "reflect"
+	sync "sync"
+
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
-	reflect "reflect"
-	sync "sync"
 )
 
 const (
--- a/pkg/api/grpc_health_v1/health_grpc.pb.go
+++ b/pkg/api/grpc_health_v1/health_grpc.pb.go
@@ -4,6 +4,7 @@ package grpc_health_v1
 
 import (
 	context "context"
+
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
--- a/pkg/api/registry.pb.go
+++ b/pkg/api/registry.pb.go
@@ -7,11 +7,12 @@
 package api
 
 import (
+	reflect "reflect"
+	sync "sync"
+
 	proto "github.com/golang/protobuf/proto"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
-	reflect "reflect"
-	sync "sync"
 )
 
 const (
--- a/pkg/api/registry_grpc.pb.go
+++ b/pkg/api/registry_grpc.pb.go
@@ -4,6 +4,7 @@ package api
 
 import (
 	context "context"
+
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
--- a/pkg/image/containerdregistry/resolver.go
+++ b/pkg/image/containerdregistry/resolver.go
@@ -28,7 +28,7 @@ func NewResolver(configDir string, insecure bool, roots *x509.CertPool) (remotes
 		ExpectContinueTimeout: 5 * time.Second,
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: false,
-			RootCAs: roots,
+			RootCAs:            roots,
 		},
 	}
 
@@ -60,7 +60,7 @@ func NewResolver(configDir string, insecure bool, roots *x509.CertPool) (remotes
 	}
 
 	opts := docker.ResolverOptions{
-		Hosts: docker.ConfigureDefaultRegistries(regopts...),
+		Hosts:   docker.ConfigureDefaultRegistries(regopts...),
 		Headers: headers,
 	}
 
--- a/pkg/image/registry.go
+++ b/pkg/image/registry.go
@@ -29,4 +29,3 @@ type Registry interface {
 	// If it exists, it's used as the base image.
 	// Pack(ctx context.Context, ref Reference, from io.Reader) (next string, err error)
 }
-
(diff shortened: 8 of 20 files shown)


.github/workflows/test.yml
```diff
@@ -8,6 +8,12 @@ on:
       - '**'
       - '!doc/**'
 jobs:
+  vendor:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: make vendor && git diff --exit-code
```
joelanford, line 15: WDYT of doing a bunch of tidying here? `go mod tidy`, `go fmt ./...` and `go mod vendor` and then running `git diff`?

timflannagan: So I updated the vendor target in the root Makefile in #590, [which prompted this PR](https://github.com/operator-framework/operator-registry/pull/590#discussion_r581070447) - but adding a linting checking the form of `go fmt ...` sounds like a step in the right direction. I don't have a strong opinion either way about whether the linting and vendor checks should be separate or their own combined test.

joelanford: In SDK, we group all of this under a [`make test-sanity`](https://github.com/operator-framework/operator-sdk/blob/34a55dd5ff27d9d87f96846eb5d17828d0e7b450/Makefile#L119-L126) target and then run that in [CI](https://github.com/operator-framework/operator-sdk/blob/34a55dd5ff27d9d87f96846eb5d17828d0e7b450/.github/workflows/test-sanity.yml#L34), which does what I'm describing. Basically all of the non-unit and non-e2e tests.

We found it nice having that in one make target so that its easy for developers to run it locally before pushing.

This is totally a nit though. I'm always good with incremental improvements.

timflannagan: This kind of check may make more sense as a prow test rather than a github action too.

timflannagan: Ah, gotcha yeah I'm fine with that rationale, although I typically prefer that something like `make test-sanity` just act as a driver and call individual make targets, like `make generate` and `make vendor` and verify the diff that way, but this is something that's pretty straight-forward to implement.

joelanford: I'm not too fussed with the details. The one bit of feedback we got from the community awhile back in SDK is that a sprawling Makefile made it hard to understand what contributors should run to verify their changes. _Maybe_ docs would help with that, but I feel like contributor docs have a way of getting out-of-date quickly.

On the other hand, it's nice having more focused make targets if you know you don't need to keep re-running tests irrelevant to your change.

I'm good either way.

Makefile
```diff
@@ -70,6 +70,10 @@ vendor:
 	$(GO) mod vendor
 	$(GO) mod verify
 
+.PHONY: fmt
+fmt:
+	$(GO) fmt $(go list ./... | grep -v /vendor/)
```
njhale, line 75: we've historically used `goimports` in OLM and operator-registry.

.github/workflows/test.yml
```diff
@@ -8,6 +8,14 @@ on:
       - '**'
       - '!doc/**'
 jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run sanity checks
+        run: |
+          make vendor && make fmt && git diff --exit-code
```
njhale, line 17: It doesn't need to be in this patch, but a `go vet` could be nice too.

Makefile
timflannagan: That's fine with me - I mainly used `gofmt` due to familiarity. The only potential problem is that I've seen `goimports` take upwards of 20 minutes for larger codebases, but it's possible I needed to further refine what paths I was passing to it.

timflannagan: Updated - had to wrangle a bit with how `go get ...` manipulates the go.mod and go.sum and disable module awareness for that single command. I can add a comment for why that's needed if that behavior is confusing and needs more clarification.

.github/workflows/test.yml
```diff
@@ -8,6 +8,18 @@ on:
       - '**'
       - '!doc/**'
 jobs:
+  sanity:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install goimports
+        run:  GO111MODULE=off go get golang.org/x/tools/cmd/goimports
```
timflannagan, line 16: This is needed due to `go get ...` modifying the `[go.mod,go.sum]` files when attempting to grab this dependency: https://github.com/golang/go/issues/24250

I might've been able to get away with just restoring the changes made to those files, but seems unnecessarily complex.

timflannagan, line 16: I guess I could also just do something like this if we think that's clearer:

```yaml
- name: Install goimports
  run: go get ...
  env:
    GO111MODULE: off
...
```

timflannagan:
/retitle .github: Add a sanity check for Go packages

.github/workflows/test.yml
timflannagan: Sounds reasonable to me - added a TODO comment.

.github/workflows/test.yml
joelanford: Not positive, but I think this gets easier with Go 1.16. Pretty sure you can do `go install <cmd>@<version>` without modifying anything in the project.

timflannagan: Yeah, I saw a note about that in the 1.16 release notes, but I was under the impression that was only for the build/test commands. I ran this locally and was able to get `go install ...` working without having to mess around with that `GO111MODULE` environment variable. We'll see if that holds up in the next CI run.

.github/workflows/test.yml
```diff
@@ -8,6 +8,18 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '~1.16'
+      - name: Install goimports
+        run: go install golang.org/x/tools/cmd/goimports@latest
+      - name: Run sanity checks
+        run: make vendor && make lint && git diff --exit-code
```
timflannagan, line 21: I can also just split this out to run the linting check and remove that `lint` make target:

```yaml
goimports -w $(find . -name '*.go' -not -path "./vendor/"
```

And then have the `git diff ...` check in another step. Open to suggestions on how to best organize this, or whether we're fine with throwing this under a single step.

joelanford: Just my personal preference, but I like to have make targets that encapsulate the github jobs so that things are eay to reproduce locally with as little hassle as possible, including getting and using the correct version of whatever tool the Makefile depends on.

Not something I would hold this up for at all, but in SDK we ended up with [this script](https://github.com/operator-framework/operator-sdk/blob/master/tools/scripts/fetch) and then make targets would call it like this: https://github.com/operator-framework/operator-sdk/blob/4e16f401032ca61f1e57203840dbc303de365f42/Makefile#L126

It was nice because:
1. If we ever needed to switch to a different CI runner (which happened three times between Travis/Openshift CI and then to GitHub actions), it was pretty simple.
1. Anyone calling the make target is using the version specified, rather than whatever is on their path.
2. No one has to worry about figuring out how to get the various tools in the first place.

Whenever we had a new tool dependency, it was pretty easy to tack it onto the end of that case block in the fetch script.

Makefile
```diff
@@ -70,6 +70,10 @@ vendor:
 	$(GO) mod vendor
 	$(GO) mod verify
 
+.PHONY: lint
+lint:
+	find . -name '*.go' -not -path "./vendor/*" | xargs goimports -w
```
joelanford, line 75: I thought goimports ignored `./vendor` with `goimports -w ./...`, but that info may be out of date by a few Go versions.

timflannagan: That's entirely possible - I've been using gofmt in the majority of Go-based projects, but Nick had said that goimports had been used throughout the OLM projects in the past so this is all a bit new to me. I can track this down further to see if we can avoid having to specify any explicitly excluded directories.

.github/workflows/test.yml
timflannagan: Yeah, absolutely that makes a lot of sense. I'll dive into those links some more - I think I would prefer getting something merged sooner-than-later just because this PR also changes files after running `goimports` locally, so it seems like some editors aren't fully set up to format on save. Going forward, and there's already a TODO in this PR about running `go vet ...` as well, we should probably aim towards emulating some of that SDK workflow.

Makefile
timflannagan: Ah, so this is what's documented on [pkg.go.dev](https://pkg.go.dev/golang.org/x/tools/cmd/goimports?utm_source=godoc):

```
To exclude directories in your $GOPATH from being scanned for Go files, goimports respects a configuration file at $GOPATH/src/.goimportsignore which may contain blank lines, comment lines (beginning with '#'), or lines naming a directory relative to the configuration file to ignore when scanning. No globbing or regex patterns are allowed. Use the "-v" verbose flag to verify it's working and see what goimports is doing.
```

I think it's fine to keep using that find command to ignore the vendor directory, and in the future if we need more granularity/control over what files are excluded from goimports, add that `.goimportsignore` file to the root repository.

PR #598, opened 2021-02-23
test/e2e: Fix output log message when Podman is not found

timflannagan:
These are minor cosmetic changes that I noted when attempting to run the e2e suite locally.

Switch from `.Log(...)` to `.Logf(...)` when passing an error argument.

Use the `%v` format specifier instead of `%s` when interacting with
`error` typed arguments.

**Reviewer Checklist**

### DIFF
--- a/test/e2e/opm_test.go
+++ b/test/e2e/opm_test.go
@@ -485,7 +485,9 @@ var _ = Describe("opm", func() {
 	}
 
 	Context("using docker", func() {
-		if err := exec.Command("docker").Run(); err != nil {
+		cmd := exec.Command("docker")
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
 			GinkgoT().Logf("container tool docker not found - skipping docker-based opm e2e tests: %v", err)
 			return
 		}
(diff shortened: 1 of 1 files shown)


timflannagan:
I don't think we care about the actual `podman info` that gets returned so we may just want to redirect stderr instead.

/hold

test/e2e/opm_test.go
```diff
@@ -485,16 +485,28 @@ var _ = Describe("opm", func() {
 	}
 
 	Context("using docker", func() {
+		cmd := exec.Command("docker")
+		out, err := cmd.CombinedOutput()
+		if err != nil {
+			GinkgoT().Logf("container tool docker not found - skipping docker-based opm e2e tests: (%s) %v", string(out), err)
+			return
+		}
```
timflannagan, line 493: Hmm yeah, I guess this technically problematic as we're running its CombinedOutput and then Run methods and the command structure doesn't allow being re-used. I'm just going to redirect the stderr pipe to os.Stderr, which would result in the following messages being produced:

```go
time="2021-02-23T12:55:04-05:00" level=error msg="cannot mkdir /run/user/1000/snap.go/libpod: mkdir /run/user/1000/snap.go/libpod: no such file or directory"
container tool podman not found - skipping podman-based opm e2e tests: exit status 1
```

dinhxuanvu (approved):
/lgtm

PR #599, opened 2021-02-24
Stream bundles to the ListBundles response.

benluddy:
The existing implementation of the ListBundles endpoint collects all
bundles first, then writes them to the response. It now writes bundles
to the response as they are read out of the database in order to
reduce the memory usage high watermark.

### DIFF
--- a/pkg/registry/interface.go
+++ b/pkg/registry/interface.go
@@ -17,6 +17,12 @@ type Load interface {
 	ClearNonHeadBundles() error
 }
 
+type BundleCursor interface {
+	Next(context.Context) *api.Bundle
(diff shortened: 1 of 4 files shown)


pkg/registry/interface.go
```diff
@@ -17,6 +17,12 @@ type Load interface {
 	ClearNonHeadBundles() error
 }
 
+type BundleCursor interface {
+	Next(context.Context) *api.Bundle
+	Close()
+	Err() error
+}
+
```
njhale, line 25: is the idea here that we need more client-side control than we get from `<-chan *api.Bundle`?

PR #601, opened 2021-03-03
Implement registry server for declarative config

joelanford:
This PR builds on #617 by adding support for serving the GRPC API from a declarative config format.

A new `opm serve` command will read the config directory, build an internal model of all packages/channels/bundles, and then serve the existing GRPC query interface from that model. 

See https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md

**Reviewer Checklist**


cmd/opm/registry/serve.go
```diff
@@ -41,7 +49,10 @@ func newRegistryServeCmd() *cobra.Command {
 	}
 
 	rootCmd.Flags().Bool("debug", false, "enable debug logging")
-	rootCmd.Flags().StringP("database", "d", "bundles.db", "relative path to sqlite db")
+	rootCmd.Flags().StringP(databaseFlag, "d", "bundles.db", "relative path to sqlite db")
+	// TODO(joelanford): Is there a default configs folder, or will every index image need to specify the config
+	//   folder in the entrypoint?
+	rootCmd.Flags().StringP(configsFlag, "c", "", "path to declarative index configs")
```
njhale, line 55: I think a much better UX would be that the `registry` command takes a path as a required operand. That path can either be to a single database file, or to a config directory. Having multiple options when there can only be one source for serve is confusing IMO.

pkg/declcfg/declcfg.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		if next == nil {
+			return nil, fmt.Errorf("discontinuous or non-linear upgrade graph")
+		}
+		out = append(out, next)
+	}
+	return out, nil
+}
+
+func verifyReplaces(ch model.Channel) error {
```
njhale, line 213: isn't this verification useful for all potential models (not just ones generated from declarative config?)

pkg/model/model.go
```diff
@@ -0,0 +1,148 @@
...
+func (b Bundle) Provides(group, version, kind string) bool {
+	for _, gvk := range b.ProvidedAPIs {
+		if group == gvk.Group && version == gvk.Version && kind == gvk.Kind {
+			return true
+		}
+	}
+	return false
+}
+
+func (b Bundle) ConvertToAPI() *api.Bundle {
```
njhale, line 133: Should the ownership of the IR->$format belong to the specific backend? e.g. in the `api` package?

cmd/opm/registry/serve.go
joelanford: Updated to use a positional argument, deprecate the `--database` flag, and use file type detection to load the appropriate registry source.

pkg/declcfg/declcfg.go
joelanford: Agreed. Latest commit adds (mostly) comprehensive validation to the model package.

pkg/model/model.go
joelanford: Done. I moved the conversion functions related to the `api` types to the `api` package. 

With that change, I was forced to move the `model.Querier` struct to another package to avoid an import cycle between `model` and `api`.  For now its in `registry`, but I'm not sure if that's the right place either.

cmd/opm/registry/serve.go
```diff
@@ -3,33 +3,46 @@ package registry
...
+
 func newRegistryServeCmd() *cobra.Command {
 	rootCmd := &cobra.Command{
-		Use:   "serve",
-		Short: "serve an operator-registry database",
-		Long:  `serve an operator-registry database that is queriable using grpc`,
-
+		Use:   "serve <source_path>",
+		Short: "serve an operator-registry source",
+		Long:  `serve an operator-registry source that is queriable using grpc`,
```
timflannagan, line 44:
```suggestion
		Long:  `serve an operator-registry source that is queryable using grpc`,
```

cmd/opm/registry/serve.go
```diff
@@ -3,33 +3,46 @@ package registry
...
 	"github.com/operator-framework/operator-registry/pkg/lib/graceful"
 	"github.com/operator-framework/operator-registry/pkg/lib/log"
 	"github.com/operator-framework/operator-registry/pkg/lib/tmp"
+	"github.com/operator-framework/operator-registry/pkg/registry"
 	"github.com/operator-framework/operator-registry/pkg/server"
 	"github.com/operator-framework/operator-registry/pkg/sqlite"
 )
 
+const (
+	databaseFlag = "database"
```
timflannagan, line 34: Real minor nit: Does this need to be a global variable for the registry package? It looks like most of the existing code attempts to get the value of this flag through `cmd.Flags().Get...`.

cmd/opm/registry/serve.go
```diff
@@ -66,44 +79,76 @@ func serveFunc(cmd *cobra.Command, args []string) error {
...
+			cfg *declcfg.DeclarativeConfig
+			err error
+		)
+		if s, sErr := os.Stat(source); sErr != nil {
+			return err
+		} else if s.IsDir() {
+			cfg, err = declcfg.LoadDir(source)
+		} else {
+			cfg, err = declcfg.LoadFile(source)
+		}
```
timflannagan, line 110: Another minor nit: this likely can be reduced to an if/else given you return early in while stat-ing the source directory.

cmd/opm/registry/serve.go
joelanford: Yeah I should probably either use hardcoded `"database"` everywhere or change other `cmd.Flags().Get` calls to use a const. Whenever I see a string used this way in more than one place, my default is to change it to a const, to avoid typo mistakes and make use of IDE auto-completion.

As an aside, using `cmd.Flags().Get()` is much more cumbersome than just declaring a flags struct and populating fields in the struct, like this:

```go
type serve struct {
	debug bool
	database string
	port string
	...
}

func (s serve) RunE(cmd *cobra.Command, args []string) error {
	// populate from serveFunc, replace `cmd.Flags().Get("<field>")` with s.<field>
}

func newRegistryServeCmd() *cobra.Command {
	s := serve{}
	rootCmd := &cobra.Command{
		...
		RunE: s.RunE,
	}
	cmd.Flags().BoolVar(&s.debug, "debug", ...)
	cmd.Flags().StringVarP(&s.database, "database", ...)
	cmd.Flags().StringVarP(&s.port, "port", ...)
	...
}
```

cmd/opm/registry/serve.go
joelanford: Now I'm re-reading that section of code and I'm trying to figure out if I mistakenly used `err` when I should have used `sErr` if `os.Stat` returns an error.

joelanford: ^ Fixed

cmd/opm/registry/serve.go
joelanford: For now, reverted to hard-coded strings to be consistent with how other flags work.

In a separate PR, I may see how hard it would be to migrate the existing cobra usage over to structs with flag fields to avoid this issue entirely.

pkg/api/properties.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+type properties struct {
+	providedPackage providedPackage
+	providedGVKs    []*GroupVersionKind
+	requiredGVKs    []*GroupVersionKind
+	skipRange       string
+}
+
+func parseProperties(props []model.Property) (*properties, error) {
```
joelanford, line 40: Some of this code is more or less duplicated in the `pkg/api` package. It initially seemed like a good idea to scope the property parsing to the component that needs it and then ONLY parse the properties required by the component.

But some properties have two or more components that are interested in them, so it seems reasonable to centralize property definitions and parsing code as well. Thoughts?

pkg/declcfg/helpers_test.go
```diff
@@ -0,0 +1,263 @@
...
+					`{"kind": "CustomResourceDefinition", "apiVersion": "apiextensions.k8s.io/v1"}`,
+				},
+			}
+			ch.Bundles[bName] = bundle
+		}
+	}
+	return pkg
+}
+
+func buildBobaFettPkgModel() *model.Package {
```
dinhxuanvu, line 197: Where is buildMandalorianPkgModel()?

pkg/declcfg/bundle.go
```diff
@@ -0,0 +1,30 @@
+package declcfg
+
+import (
+	"github.com/operator-framework/operator-registry/pkg/property"
+)
+
+type bundle struct {
```
dinhxuanvu, line 7: I'm gonna be honest here by saying that I don't see the value of this Bundle API given it is pretty much identical to Bundle on registry package. The diff I can see here is the registry.Bundle has more fields and this bundle has `RelatedImages`. I understand this is helpful for the validation but I wonder if we reinvent the wheel here because I can see we potentially add more fields and it will become exactly like registry.Bundle.

test/e2e/testdata/configs/etcd.json
```diff
@@ -0,0 +1,248 @@
+{
```
dinhxuanvu, line 1: Is this config valid or invalid? This json format looks outdated (pre-your-enhancement-PR).

pkg/declcfg/bundle.go
dinhxuanvu: I recall our conversation about moving this to api repo. I think that makes more sense.

test/e2e/testdata/configs/etcd.json
joelanford: Yeah that's a leftover unused file. I'll delete it.

pkg/declcfg/bundle.go
joelanford: I disagree about re-using `registry.Bundle` in place of the declarative config `bundle` struct. `declcfg.bundle` is the Go specification of the `olm.bundle` schema. It is used to (un)marshal declarative config JSON. That should be its singular purpose.

Looking at `registry.Bundle`, I get the impression that its primary purpose is the representation of the _contents_ of the bundle image, used when we unpack the bundle image and need to understand what it contains.

IMO using one struct to represent both declarative config's `olm.bundle` schema AND the contents of bundle described by that schema will cause unnecessary complexity. For instance, if we need to add a new field to `registry.Bundle` to capture some detail about the contents of the bundle, we also then need to understand how that might impact that struct's use as the specification for the `olm.bundle` spec.

joelanford: I theory, I would agree that `operator-framework/api` is a logical place for `pkg/declcfg`. I think my suggestion for now would be to keep it all encapsulated in `operator-registry` so that we can iterate on it (if necessary) without involving another repository.

We can always move it `operator-framework/api` once it stabilizes. To prevent others from making use of it for now, we could put it in an `internal` tree to make it available only to other `operator-registry` packages.

Related side question: does `operator-registry` make any Go API guarantees? Given that there are v1.x tags, I would assume that we're not allowed to remove or break any exported APIs?

dinhxuanvu: > I disagree about re-using `registry.Bundle` in place of the declarative config `bundle` struct. `declcfg.bundle` is the Go specification of the `olm.bundle` schema. It is used to (un)marshal declarative config JSON. That should be its singular purpose.
> 
> Looking at `registry.Bundle`, I get the impression that its primary purpose is the representation of the _contents_ of the bundle image, used when we unpack the bundle image and need to understand what it contains.
> 
> IMO using one struct to represent both declarative config's `olm.bundle` schema AND the contents of bundle described by that schema will cause unnecessary complexity. For instance, if we need to add a new field to `registry.Bundle` to capture some detail about the contents of the bundle, we also then need to understand how that might impact that struct's use as the specification for the `olm.bundle` spec.

1. I understand your intention to use this `declcfg.bundle` as the representation of `olm.bundle` JSON struct in the config and you are correct regarding `registry.bundle` is the representation of the bundle and its content. However, you blur the line between the two when you include `CsvJSON` and `Objects` in the `declcfg.bundle`. Those two fields are for bundle content so you are duplicating the `registry.bundle` purpose. I can see the comment you have on those two fields about preserving the compatibility with gRPC which actually colliding with another API we have also named Bundle (`api.Bundle` which is for gRPC API). As you can see, we end up with 3 different `something.Bundle` that represent essentially the same thing. I'm not saying this `declcfg.bundle` is a bad thing given we can't really change `api.Bundle` because that would break bundle gRPC API and `registry.Bundle` is kind of a mess due to legacy reason (we don't want to remove fields and keep adding more as we introduced `property` and `dependency` for resolution). So I can see this API perhaps is a necessary step to move forward with config work without dealing with existing API. However, it seems to become a troubling pattern that we have 3 APIs with the same name floating around in the same repo that represents the same thing perhaps in slightly different shapes or forms. Imagining someone looking our registry codebase and wonder why there are 3 different bundles here and frankly I wonder if they even notice these are 3 different APIs or they just assume it is the same thing.
2. We can decide where to keep this bundle API at a later time. I don't feel strongly about where it should belong though as I mentioned above, perhaps it should be in api for the sake of not having multiple bundle APIs in the same repo. Plus, it depends on who we think who is dictating the config format. I often think personally api repo is where OLM and registry have in common and we have everything there so they can share and reuse. In term of config, OLM doesn't use it directly so the config API is actually being dictated by registry which means it can stay here. There is nothing wrong with folks using API from registry repo directly.
3. I don't know if we guarantee the backward compatibility for any API registry except the gRPC API for bundle. The rest is pretty much for internal use. No one is using the APIs in registry directly as far as I know.

joelanford: 1. Pretty much agree with everything you said. Perhaps we could beef up the GoDoc comments for each `Bundle` type to try to clarify where/why they're used?
2. I think we need to consider `operator-sdk` in the list of clients of `operator-framework/api`. I haven't thought about it _too_ much, but I could imagine `operator-sdk` needing to manipulate declarative configs at some point. Either way, my vote would be to leave it here for now as an API that can only be used internally by `operator-registry` components.

    Based on this, I'm thinking I'll move `declcfg` and `model` into an `internal` package so that other projects will be prevented from importing it. Then if/when there's a need for others to consume it, we can have this conversation again. And by then we'll have a concrete motivation to help us decide. WDYT?
3. This point is going off on a tangent, but I know [SDK imports `operator-registry`](https://github.com/operator-framework/operator-sdk/blob/b3547bcacaca9e10c7225ced6ebd6210067d7fc6/go.mod#L16), so there's at least one consumer of "internal" APIs. We should move everything that's actually internal to `internal` directory trees to prevent other projects from using them.

cmd/opm/registry/serve.go
```diff
@@ -66,92 +78,122 @@ func serveFunc(cmd *cobra.Command, args []string) error {
...
+
+	if s.skipMigrate {
+		return dbStore, nil
+	}
+
+	m, err := sqlite.ToModel(ctx, dbStore)
+	if err != nil {
+		return nil, err
+	}
+	return registry.NewQuerier(m), nil
```
joelanford, line 193: Do we want to serve sqlite databases from the model, or should we just return the `dbStore` directly (which would mean continuing to serve sqlite from the sqlite querier)?

There's some risk involved in serving sqlite from the model because that's a change from current code paths that has not been thoroughly tested.

I suppose if we started getting bugs against this, it would be simple to revert back to serving directly from sqlite.

joelanford: Plan discussed offline: by default keep serving sqlite databases as we always have.

If we want to load the sqlite db into memory and serve it from the model, we can do that really easily in a follow-up PR (probably gated by a feature flag).

pkg/declcfg/model_to_declcfg.go
```diff
@@ -0,0 +1,81 @@
...
+			b, ok := bundles[chb.Name]
+			if !ok {
+				b = &Bundle{
+					Schema:        schemaBundle,
+					Name:          chb.Name,
+					Package:       chb.Package.Name,
+					Image:         chb.Image,
+					RelatedImages: modelRelatedImagesToRelatedImages(chb.RelatedImages),
+					CsvJSON:       chb.CsvJSON,
+					Objects:       chb.Objects,
```
anik120, line 56: If these are not set, for eg if CsvJSON and Objects for `chb` is not set during creating a bundle blob, won't this lead to a nil pointer dereferencing?

joelanford: This particular spot is okay as long as `chb` is non-nil. Which will always be the case unless there's a bug in our code that converts any of the representations to the model (or someone else is importing and using `model.Model` directly).

If `chb.CsvJSON` and/or `chb.Objects` is nil (definitely possible and happens frequently), then `declcfg` code should handle that. I'll check to see if the `Load` and `Write` function tests include cases for this (I think they do, but can't hurt to look again).

anik120: @joelanford here's an annotations.yaml file for the bundle quay.io/operatorhubio/etcd:v0.6.1

```
annotations:
  operators.operatorframework.io.bundle.channel.default.v1: ""
  operators.operatorframework.io.bundle.channels.v1: alpha
  operators.operatorframework.io.bundle.manifests.v1: manifests/
  operators.operatorframework.io.bundle.mediatype.v1: registry+v1
  operators.operatorframework.io.bundle.metadata.v1: metadata/
  operators.operatorframework.io.bundle.package.v1: etcd
```

The default channel is not set here, which means when we're creating the model from that bundle, the `model.Package.DefaultChannel` is nil. 
Since we're accessing `mpkg.DefaultChannel.Name` here directly, it'll run into a nil pointer dereferencing error

joelanford: If the model, has been [validated](https://github.com/operator-framework/operator-registry/pull/601/files#diff-79ca10368c383e5ed5c7d8885fe1f4c2758b5ae4afd3feaf629e3a348b69b054R57-R59), `mpkg.DefaultChannel` won't be nil.

And (at least for operator-registry's uses of the model package) the conversion functions run the validations.
- [Declarative config to model](https://github.com/operator-framework/operator-registry/pull/601/files#diff-345bb90148fa16097184a5f10d48e80bf0cc6f8bfa5b6c7054b41070a50235aeR84-R86)
- [Sqlite to model](https://github.com/operator-framework/operator-registry/pull/601/files#diff-806b9455cc4618f4bd3e89d89ab1cb18499ec2e67f21b2914f8e2b69b057dc7eR28-R30)

Nonetheless, you're right that this code could result in a panic if used in other places or ways.  I think there are two possible solutions:
1. Add a `nil` check here for `mpkg.DefaultChannel` and if `nil`, use an empty string.
2. Run the full model validation at the beginning of the function, which would guarantee that `mpkg.DefaultChannel` would be non-nil.

With option 1, we're _kind of_ saying, that it's okay for this field to be unset, and as far as I know, we currently enforce that every package has a default channel (correct me if I'm wrong).

With option 2, we might be enforcing validations that are too strict.

As an aside, I think we should relax the model validations down to enforce only what the GRPC API requires.

pkg/registry/query.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+	if len(entries) == 0 {
+		return nil, fmt.Errorf("no channel entries found that replace %s", name)
+	}
+	return entries, nil
+}
+
+// TODO(joelanford): What if multiple bundles replace this one?
+//    Don't worry about this. Not used anymore.
```
benluddy, line 138: This one appears to still used by catalog-operator:

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L1011
https://github.com/operator-framework/operator-lifecycle-manager/blob/d9ff24161ab4f7d4d323afb662a56f453791329c/pkg/controller/registry/resolver/querier.go#L143
https://github.com/operator-framework/operator-registry/blob/a1fc9d0f1d2e04c14229801605ed727cd80e956e/pkg/client/client.go#L74

joelanford: I'll double check, but I think there's some dead code in OLM, and this might be one of those cases.

joelanford: Oh now I see you already did the leg work. Thanks!

In that case, I'll cross-reference with the sqlite implementation of this function and see if its deterministic (and if so, I'll use the same algorithm).

joelanford: The sqlite implementation looks non-deterministic to me since there's no ORDER BY in the query.

The model implementation in this PR is ALSO non-deterministic because it iterates over a map and chooses the first bundle it finds that meets the criteria. Since map iteration is non-deterministic, the returned bundle is non-deterministic.

I'll remove this TODO comment, and add some comments in the code that call this out.

joelanford: Done!

pkg/declcfg/helpers_test.go
joelanford: It doesn't exist. Did I miss something?

pkg/declcfg/model_to_declcfg.go
joelanford: Fixed the `nil` pointer access issue with `model.Package.DefaultChannel.Name`.

pkg/property/property.go
```diff
@@ -0,0 +1,322 @@
...
+		if _, ok := props[k]; ok {
+			continue
+		}
+		props[k] = p
+		out = append(out, p)
+	}
+	return out
+}
+
+func ValidateBackCompat(in Properties) error {
```
dinhxuanvu, line 233: Can we comment on what are the criteria we are using here for backward compatibility?

pkg/model/model.go
```diff
@@ -0,0 +1,279 @@
...
+type Channel struct {
+	Package *Package
+	Name    string
+	Bundles map[string]*Bundle
+}
+
+// TODO(joelanford): This function determines the channel head by finding the bundle that has 0
+//   incoming edges, based on replaces and skips. It also expects to find exactly one such bundle.
+//   Is this the correct algorithm?
+func (c Channel) Head() (*Bundle, error) {
```
dinhxuanvu, line 130: Answer to your question in the `TODO`. We currently use this ([link](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/registry/resolver/resolver.go#L591)) to sort bundles in a channel (which effectively finds the head of channel). You should consult with that code and perhaps we should align them.

joelanford: I think that's doing the same thing my code is doing. It's just that my code is not concerned with figuring out the sorted order of the entire channel.

dinhxuanvu: Then I think we should be fine with this.

pkg/property/property.go
joelanford: This is a bit of a relic function from 10s of commits ago and as you can see it's much shorter now. I'll revisit this to see if its necessary to keep the last remaining check separated out in this function.

pkg/model/model.go
joelanford: @ecordell Did mention that there are index maintainer use cases for potential multi-head channels, so I'll leave this comment in place, but update it to link to the SAT solver implementation and mention Evan's potential use case.

dinhxuanvu (commented):
This PR needs to rebase with Joe's base PR because I still see code from the query which should be on Joe's serve PR only.
Plus, we should review the part that belongs to Joe's base PR on that PR, not here to avoid polluting the review process on the `add` part.

joelanford:
@dinhxuanvu rebased!

joelanford:
No idea what's going on here: https://github.com/operator-framework/operator-registry/pull/601/checks?check_run_id=2246459783#step:5:161

I can't get this to happen locally. I'm using go 1.16.2 and I'm running the same steps from the sanity test workflow.
```
$ go install golang.org/x/tools/cmd/goimports@latest
go: downloading golang.org/x/tools v0.1.0
go: downloading golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4

$ make vendor && make lint && git diff --exit-code
GOFLAGS="-mod=vendor" go mod tidy
GOFLAGS="-mod=vendor" go mod vendor
GOFLAGS="-mod=vendor" go mod verify
all modules verified
find . -name '*.go' -not -path "./vendor/*" | xargs goimports -w
```

pkg/api/api_to_model.go
```diff
@@ -0,0 +1,165 @@
+package api
```
joelanford, line 1: Code that converts the GRPC API and sqlite DBs into the model isn't strictly required for `opm serve`. However it's used in `pkg/server/server_test.go` to test that both of the following scenarios result in the equivalent GRPC responses:
1. sqlite -> GRPC
2. sqlite -> model -> GRPC

pkg/registry/query.go
```diff
@@ -0,0 +1,322 @@
...
+			continue
+		}
+		if entry.ChannelName == pkg.DefaultChannel.Name {
+			return q.GetBundle(ctx, entry.PackageName, entry.ChannelName, entry.BundleName)
+		}
+	}
+	return nil, fmt.Errorf("no entry found that provides group:%q version:%q kind:%q", group, version, kind)
+}
+
+func doesModelBundleProvide(b model.Bundle, group, version, kind string) (bool, error) {
```
dinhxuanvu, line 254: I wonder if it is beneficial to convert this func and `bundleReplaces` to public funcs on `model` packages.
For example:
`func(b model.Bundle) Replaces(name string) bool {}`
Just a thought.

pkg/api/model_to_api.go
```diff
@@ -0,0 +1,116 @@
...
+		skipRange = string(props.SkipRanges[0])
+	}
+	return &Bundle{
+		CsvName:      b.Name,
+		PackageName:  b.Package.Name,
+		ChannelName:  b.Channel.Name,
+		BundlePath:   b.Image,
+		ProvidedApis: gvksProvidedtoAPIGVKs(props.GVKs),
+		RequiredApis: gvksRequirestoAPIGVKs(props.GVKsRequired),
+		Version:      props.Packages[0].Version,
```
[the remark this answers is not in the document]
dinhxuanvu: I do notice that the `parseProperties` func does validate this scenario so it won't happen.

dinhxuanvu (commented):
Looking good. Just a few comments.

pkg/api/model_to_api.go
```diff
@@ -0,0 +1,116 @@
...
+		Dependencies: convertModelPropertiesToAPIDependencies(b.Properties),
+		Properties:   convertModelPropertiesToAPIProperties(b.Properties),
+		Replaces:     b.Replaces,
+		Skips:        b.Skips,
+		CsvJson:      b.CsvJSON,
+		Object:       b.Objects,
+	}, nil
+}
+
+func parseProperties(in []property.Property) (*property.Properties, error) {
```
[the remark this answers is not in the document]
joelanford: The main reason for the difference is because `api.Bundle` has a `SkipRange` field that allows either 0 or 1 skip ranges. When converting to the `api.Bundle`, we have to verify that no more than one skip range is defined.

The model itself doesn't have this constraint built-in, which is why the declcfg_to_model function does not include this check.

pkg/registry/query.go
joelanford: At one point, I think I had a version of the model package that had these kind of functions. Let me see if I can work out why I changed it.

joelanford: Ahh. I think I did it this way because I wanted to keep property the model out of the business of parsing and interpreting properties, which would be necessary to answer these types of questions.

joelanford: We _could_ put a `DoesReplace` function in model since `model.Bundle` has a single `Replaces` field in the struct. But the provided GVKs are all stored in opaque properties in the model.

My instinct is to leave the model pretty minimal. If there's a future where multiple packages have to implement these same sort of functions against the model, we could revisit.

dinhxuanvu: I see. I agree that this func `doesModelBundleProvide` isn't necessarily beneficial to have on the model because it does require some conversion. The replace one is more straightforward to have on model. Either way, it is not a big deal. I just thought about reorganizing the codebase a bit but it is not a concern at the moment. Move on.

pkg/api/model_to_api.go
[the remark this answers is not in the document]
dinhxuanvu: I'm a bit confused. What do you mean by `allows either 0 or 1 skip ranges`? skipRange is just a string in `api.Bundle`. You mean with `model.Bundle`, they can have multiple `skipRange` properties?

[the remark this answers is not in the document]
joelanford: Correct.

pkg/sqlite/conversion.go
```diff
@@ -0,0 +1,131 @@
+package sqlite
```
anik120, line 1: Are we converting the sqllite db to model before serving the db anymore now that we've decided serving with configs is a separate command altogether? 

I couldn't find a place where we're doing that in this PR, maybe I'm missing something?

pkg/registry/query.go
```diff
@@ -0,0 +1,322 @@
...
+
+var _ GRPCQuery = &Querier{}
+
+func NewQuerier(packages model.Model) *Querier {
+	return &Querier{
+		pkgs: packages,
+	}
+}
+
+func (q Querier) ListPackages(_ context.Context) ([]string, error) {
```
anik120, line 24: Are we testing these new implementations against the existing tests somewhere?

anik120 (commented):
Looking really good so far @joelanford, had a couple of questions.

pkg/sqlite/conversion.go
joelanford: You're correct. The `opm registry serve` command is completely unchanged, so sqlite DBs will be served directly (without converting to the model).

The reason this file is present in this PR is to enable the test in `./pkg/server/server_test.go`. In that test, we serve the DB directly, and we convert the DB to the model and serve that via the model-based GRPC server. This helps us guarantee that the sqlite GRPC server and the model-based GRPC server function the same against the same underlying data.

pkg/registry/query.go
joelanford: Yes, `pkg/server/server_test.go`.

pkg/api/model_to_api.go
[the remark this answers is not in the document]
dinhxuanvu: Given this restriction on skipRange is universal, then I don't see the reason why we don't impose this restriction on `declcfg` side as well.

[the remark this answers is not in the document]
joelanford: > this restriction on skipRange is universal

I think that's the current state of the world with the current resolver and the current GRPC API. But that might not always necessarily be the case when we've updated the resolver and gotten rid of the GRPC server.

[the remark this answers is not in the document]
dinhxuanvu: Sure. I can see that you want to model side to be more generic and not discriminating against this. I can see a future where the api.Bundle no longer needs a specific field for `skipRange` (or no longer exists in the new non-grpc API we use) but I don't see a future where we allow multiple `skipRange, right? However, the fact that we use the model to validate config. If the config is considered valid even with multiple `skipRange`, then it will fail the serve part. That is not a good UX.

pkg/registry/query.go
anik120: Okay I see it (for some reason github decided to contract that file while I was reviewing it so it escaped my eye 🤦🏽 )

joelanford: Yeah GitHub has some weird behaviors like that. Super annoying.

anik120:
/lgtm 
if @dinhxuanvu's remaining concern is addressed it looks good to me.

pkg/api/model_to_api.go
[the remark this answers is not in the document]
joelanford: Chatted offline with @dinhxuanvu. We'll handle this concern in #618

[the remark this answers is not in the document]
dinhxuanvu: Joe and I agree that we should have this "validation" on the model side so the validate and serve can be in sync in term of pass/fail on the same scenario which in case is multiple skipRange. We will keep things as they are in Joe's PR and will address this on my PR (which should be merged after Joe's).

dinhxuanvu (approved):
/lgtm

PR #605, opened 2021-03-14
Update api version, vendor & enhance testing

### LINKED ISSUE #568: Detect common invalid annotations:  e.g. olm.skiprange
It's a very common problem to accidentally use the annotation `olm.skiprange` instead of `olm.skipRange`.  It can take days or weeks for a new developer to not understand why their operators will never upgrade to a new channel.

It would be very helpful for new developers if `opm <index|registry> add` would simply detect some of the most common errors like this.

jchunkins:

The original purpose for this PR was to update the `api` project to v0.6.1 to pick up validator changes, but I decided to use this opportunity to add further testing enhancements and improve documentation. I wanted to run the e2e tests locally without needing to setup a TLS enabled registry, and discovered some code paths which are capable of handling "skip TLS" but did not expose the feature. I've executed unit tests and e2e tests using my local development machine (a Mac) and all run cleanly (with exception of one e2e test which requires a linux executable to function correctly, so it should work in the GitHub Actions environment).

- update api project to v0.7.1 to pick up validator (and other) changes
- update code paths which support skip tls but did not expose it
- add test case and test data for annotation case sensitive validation
- update test case to handle multiple errors
- add .gitignore for artifacts generated from e2e tests
- use mkcerts for better crossplatform SSL cert generation
- create dedicated script for starting regsitry with & without SSL
- implement e2e ctx (i.e. context) package
- Provision impl for kind servers
- Provision impl for kubeconfig (i.e. all other server types)
- adjust makefile to account for `kind` build tag
- allow e2e target to use focus flag
- allow e2e target to use skip tls flag for local testing
- adjust test.yml git action workflow to use start_registry script
- adjust test.yml git action to remove kind (since its now dynamic)
- add e2e documentation for all scenarios
- update vendor

Implements https://github.com/operator-framework/operator-registry/issues/568

Closes #568

Signed-off-by: John Hunkins <jhunkins@us.ibm.com>

See Description

**Reviewer Checklist**


jchunkins:
@dinhxuanvu any idea on how to restart e2e-minikube and e2e-kind ?

timflannagan:
You need to manually kick those failed jobs. I'm not entirely sure prow has the ability to requeue GitHub actions, but I requeued them in the meantime. When you click on one of the failed checks, there's an option to re-test - It's possible that requires some level of permissions in the organization.

jchunkins:
@timflannagan Thanks for the assist. Since I am not in the org, I don't see the UI controls to allow restarting GitHub Actions. Maybe someday I'll get added. At any rate, thanks again for your help.

timflannagan:
Hmm, it looks like that recently added bundle unpack command has been producing more flakes recently. Going to retest the jobs again, but we may need to start diving into those errors.

jchunkins:
I edited the make file on my local system and added `--untilItFails` like so:
```
$(GO) run github.com/onsi/ginkgo/ginkgo --v --untilItFails --randomizeAllSpecs --randomizeSuites --race $(if $(TEST),-focus '$(TEST)') $(TAGS) ./test/e2e -- $(if $(SKIPTLS),-skip-tls true) 
```

And then executed the test:
```
KUBECONFIG="/tmp/kindconfig" DOCKER_REGISTRY_HOST=localhost:5000 make build e2e TEST='fails to unpack'
```

And it repeatedly runs without failing. Must be some combination of running with other tests that makes it fail?

timflannagan:
Same, I wasn't able to reproduce it consistently when I was attempting to dive into those errors locally a couple of weeks back. It seems like it pops up often enough in the e2e-[minikube,kind] tests.

@njhale Any idea why that bundle unpack command occasionally fails the e2e suite? It seems like we're expecting that command will purposefully fail when feeding it an invalid bundle, but the error output is nil.

jchunkins:
@timflannagan @njhale I am not sure if the unpack flake is due to TLS in some oddball circumstances? Or maybe it only happens on linux? Since @timflannagan saw the issue previously it seems unlikely to be related to my changes in this PR.

While I am thinking about TLS, I have included some changes relating to `skip-tls` handling in some of the code paths of this PR as an enhancement for testing. I think it is worthwhile discussing the actual argument to enable/disable TLS when running the actual commands. Note that at [this line](https://github.com/operator-framework/operator-registry/blob/master/cmd/opm/index/cmd.go#L29) the parent command (in this case it would be the root command) gets its flag setup for `skip-tls`. 

This effectively makes the `skip-tls` argument apply to the root and all of its children, which as far as I can tell is the intended behavior. It just seems odd to do it in the index command.... if it is truly an argument that applies to all sub commands (i.e. global) then why not just define it in `cmd/opm/root/cmd.go`? This would make it much more obvious. If it is supposed to be global then there is no need to define the same variable in other places like [add.go](https://github.com/operator-framework/operator-registry/blob/master/cmd/opm/registry/add.go#L32) and [unpack.go](https://github.com/operator-framework/operator-registry/blob/master/cmd/opm/alpha/bundle/unpack.go#L31)

To see all of the places where `skip-tls` shows up, use this query https://github.com/operator-framework/operator-registry/search?q=%22skip-tls%22

Is there anyone else I should tag on this PR to discuss this? I can always open up a separate PR to handle this sort of change if we think its worthwhile.

jchunkins:
Regarding the flakes in `fails to unpack`, I tried to see if this was a linux vs. Mac difference and setup a ubuntu 20.04 system and used `-untilItFails` like I indicated earlier and it also runs without errors continuously. Getting the feeling that maybe running other tests at the same time somehow causes the environment to become dirty in some way, leading to failure depending on what test might proceed `fails to unpack`. Seems possible given that `--randomizeAllSpecs` and `--randomizeSuites` is used. Obviously this is just a total guess on my part.

timflannagan:
Yeah, agreed and we just merged a fix that disabled those problematic/flaky tests, with the intention of re-enabling in the next couple of weeks. It looks like GitHub actions will run out-of-date workflows/tests, so we may need to open/close this PR and re-apply any labels if the flakes are too overwhelming, or force an empty commit.

timflannagan:
But just to clarify, those bundle unpack errors are known and tracked, and we won't block merging PR content with e2e failures that fail solely for something that's difficult to reproduce locally.

jchunkins:
OK... I updated `github.com/operator-framework/api` v0.7.1 since that was now available and rebased / squashed commits. Hopefully the tests run cleanly and this can get reviewed / merged.

jchunkins:
Excellent, all tests pass now. Now we just need review labels.

jchunkins:
@hasbro17 or @anik120 Can either of you review this PR please?

docs/contributors/e2e_tests.md
```diff
@@ -0,0 +1,57 @@
+# Execute end-to-end tests for local development
```
joelanford, line 1: It's super helpful to have this documented, but its even better if the e2e suite setup could do these steps automatically. Is that something you considered?

For example: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/test/e2e/ctx/provisioner_kind.go

docs/contributors/e2e_tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+To execute the end-to-end (e2e) test suite on your local development system:
+
+1. install `kind` (see https://kind.sigs.k8s.io/docs/user/quick-start/#installation)
```
joelanford, line 5: Ideally, the codebase would specify _which_ version of kind and enforce that version is used in the tests.

joelanford (commented):
Mostly a bunch of non-blocking nits, though I'm interested in the choice to match error strings in their entirety.

Also, is it necessary to include the api and k8s bump in the same PR as the other changes? It seems like it could be a good idea to make that dependency bump a standalone PR.

.gitignore
```diff
@@ -462,3 +462,10 @@ pkg/apprclient/openapi/git_push.sh
 !test/**testdata/**
 
 /coverage.out
+
+# Test artifacts generated by e2e
+test/e2e/bundle.Dockerfile
+test/e2e/downloaded/**
+test/e2e/opm-*
+test/e2e/bundle_tmp*
+test/e2e/index_tmp*
```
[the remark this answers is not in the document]
jchunkins: In general I think the tests do clean themselves up assuming they are allowed to run to completion. If you're debugging and stop execution, those artifacts get left around. Then they show up as possible items to commit, and the only way to prevent accidental checkin is to add them to .gitignore.

docs/contributors/e2e_tests.md
```diff
@@ -0,0 +1,57 @@
...
+
+   ```bash
+   GOFLAGS="-mod=vendor" go run github.com/onsi/ginkgo/ginkgo --v --randomizeAllSpecs --randomizeSuites --race -dryRun -focus 'builds and manipulates bundle and index images' -tags "json1" ./test/e2e
+   ```
+
+## Known Limitations
+
+Currently the test case `Launch bundle` in test/e2e/bundle_image_test.go assumes that the `opm` executable used in the test is compiled for linux.
+If you run this test on a darwin environment, the kube job will not succeed unless you manually cross compile for linux and include
+the binary at `bin/opm`.
```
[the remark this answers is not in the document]
jchunkins: Issue filed as https://github.com/operator-framework/operator-registry/issues/620

docs/contributors/e2e_tests.md
jchunkins: When I update the code in my next push this will be resolved since I have implemented `ctx` similar to what is done in OLM.

docs/contributors/e2e_tests.md
jchunkins: When I update the code in my next push this will be resolved since I have implemented `ctx` similar to what is done in OLM.

pkg/lib/bundle/validate_test.go
```diff
@@ -173,9 +184,10 @@ func TestValidateBundleContent(t *testing.T) {
 			require.True(t, isValidationErr)
 		}
 		require.Len(t, validationError.Errors, tt.numErrors, table[i].description)
-		if len(validationError.Errors) > 0 {
-			e := validationError.Errors[0]
-			require.Contains(t, e.Error(), tt.errString)
```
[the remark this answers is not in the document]
jchunkins: That's fair. I'd considered this and decided to go ahead with this change. When I updated the code to account for array of string I wanted to use `require.ElementsMatch` to eliminate the potential for different ordering, which meant checking for full strings. 

I'll just have to make some changes to accomplish the same thing with require.Contains instead in my next set of changes.

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/onsi/gomega v1.10.2
 	github.com/opencontainers/image-spec v1.0.2-0.20190823105129-775207bd45b6
 	github.com/opencontainers/runc v0.1.1 // indirect
-	github.com/operator-framework/api v0.5.0
+	github.com/operator-framework/api v0.7.1
```
timflannagan, line 35: Can we move this API bump to a separate PR? I don't see anything in this PR that requires the v0.7.x tag, but it's possible I'm missing something.

Makefile
```diff
@@ -98,4 +98,4 @@ clean:
 
 .PHONY: e2e
 e2e:
-	$(GO) run github.com/onsi/ginkgo/ginkgo --v --randomizeAllSpecs --randomizeSuites --race $(TAGS) ./test/e2e
+	$(GO) run github.com/onsi/ginkgo/ginkgo --v --randomizeAllSpecs --randomizeSuites --race $(if $(TEST),-focus '$(TEST)') $(TAGS) ./test/e2e -- $(if $(SKIPTLS),-skip-tls true)
```
timflannagan, line 101: Looks at some of the other review comments, and the e2e_tests.md in this PR, it looks like we're getting to the point where we might want to use a helper bash script that this `make e2e` target will call, that will do the heavy-lifting for us, like checking if the `KUBECONFIG` or enforcing the version of kind, etc. We can evaluate that outside of this PR though.

pkg/lib/bundle/validate_test.go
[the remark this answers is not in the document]
timflannagan: Looks like Joe already got this, but continuing to using `require.Contains(...)`, or some equivalent gets a +1 from me. I've run into some considerable problems in other projects maintaining error validation checks in tests when ensuring that error(s) match exactly what we're defined in the test tables and it's typically sufficient to enforce a subset of the error matches what we expect.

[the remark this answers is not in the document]
jchunkins: I just pushed a change to address this. It uses a "contains" approach without using `require.Contains` and reverts the error messages back to what they were.

go.mod
jchunkins: The api bump is needed because the original purpose of the PR was to bring in validator changes, and then morphed from there as I did testing.

Makefile
jchunkins: Please see the latest changes that I pushed. I've implemented Joe's request for bringing forward the "provisioning" code that was in OLM.

go.mod
timflannagan: Ah, gotcha that makes more sense. I think I would prefer if we still moved that bump to another PR, as we need to do it regardless of the changes in this PR, and have this PR depend on that PR, but this isn't a blocking concern for me. @joelanford Any strong opinion on whether we should land the API bump first, rather than throw it into this PR along with the other changes?

Makefile
timflannagan: Oh man, that's awesome - I was under the impression we were only going to tackle documenting running e2e tests locally as that wasn't immediately clear to me as well when I started diving into the registry codebase recently, but attempting to automate that as well is super nice.

go.mod
joelanford: Yeah I agree it would be good to separate the concerns of the PRs.

@jchunkins is there any coupling between the test changes and the API changes? If the test changes help validate the API changes, can the test changes go in first without bumping API?

Somewhat related, though not directly: I'm not sure how operator-registry has managed kubernetes version bumps in the past with respect to releases and branching, but IMO once a Kubernetes version bump lands, we _must_ not cut a patch release on an existing minor release. Put another way, every new Kubernetes version should automatically get at least a new minor release. I don't see any operator-registry-based version branching in the repo, so I'm concerned we're not currently doing this.

jchunkins: Summary of the `API bump and its associated code`:

- I added a new test case "condition" to an existing test which exercises the validator changes coming from API v0.7.1. See this section https://github.com/operator-framework/operator-registry/pull/605/files#diff-fab19c28689357747c4449f5f7d8b9bd400867d1d096c65d656b07d6a5b4001fR166-R176

- Also, these items were added to make the test case work:

   pkg/lib/bundle/testdata/validate/invalid_manifests_bundle/invalid_annotation_name/etcdbackups.etcd.database.coreos.com.crd.yaml
   pkg/lib/bundle/testdata/validate/invalid_manifests_bundle/invalid_annotation_name/etcdclusters.etcd.database.coreos.com.crd.yaml
   pkg/lib/bundle/testdata/validate/invalid_manifests_bundle/invalid_annotation_name/etcdoperator.v0.9.4.clusterserviceversion.yaml
   pkg/lib/bundle/testdata/validate/invalid_manifests_bundle/invalid_annotation_name/etcdrestores.etcd.database.coreos.com.crd.yaml

`Everything else` is related to enhancements for running tests (i.e. TLS handling, kind & other server provisioning, etc.)

So I think you're asking me to strip out `API bump and associated code` into its own PR and get that merged first, then `Everything else` in this PR.

timflannagan: Sorry for the slow reply. We just need to bump the API version listed in the root go.mod and re-generate the vendor tree (via `make vendor` locally) in a separate PR. If you want, I can throw up a PR that does that, as we realize we need that bump sooner rather than later before the openshift CI gets out of control in a couple of days. Else feel free to open that PR and I can that pushed through with the requisite labels.

jchunkins: Thanks Tim. I created PR: https://github.com/operator-framework/operator-registry/pull/624

timflannagan: Nice, that merged so we'll need to rebase this PR. Thanks for throwing that up 👍

jchunkins:
Looks like all tests are green (other than `codecov/patch`) so I think this PR is good to go. Just need LGTM & Approve labels

dinhxuanvu (approved):
/approve

dinhxuanvu:
@jchunkins The PR looks good so we will add lgtm label next week and merge the PR. We are holding it until next week so we can focus on merging other feature PRs first. The CI can be flaky under load of too many PRs trying to get in this week so we try to be careful here. I'll make sure it will merge next week. Thanks.

dinhxuanvu (approved):
/lgtm

jchunkins:
@dinhxuanvu How critical is the codecov/patch job? It would seem that prevents the merging.

timflannagan:
@jchunkins We've been running into problems with codecov blocking PRs with insignificant diffs (e.g. <0.1%) which seems rather unreasonable, and looking at the status of the checks here, the codecov/patch is reporting a 0.00% of diff hit so I'm not too sure what's happening there. I'd like to bump the delta for failed codecov patches before force-merging (there's no way to override that codecov check(s) when I lasted checked) if possible, but if we can't tackle that in the next couple of days then I think it's reasonable to force this in.

timflannagan:
Going to hold so the bot stops going crazy. I'm in the process of getting marking those codecov checks as optional, but I need to get those changes merged and wait until that new configuration has been rolled out. Hopefully, we can get tide unblocked by tomorrow and get this merged.

/hold

jchunkins:
Sorry for not getting to this until now. I pushed a change with a rather stupid test which bumped up the code coverage by an infinitely small amount, so maybe that would make the coverage tool happy.

jchunkins:
Of course that now makes the LGTM label go away, so that'll need to get reapplied.

jchunkins:
Guess I don't understand what I should do to fix the coverage problem. I'll just wait til your change is complete.

timflannagan:
Oh man that frustrating that tide isn't merging that after the change had landed. I'm going to throw the hold back so the bot doesn't go crazy before now and tomorrow.

/hold

timflannagan:
Looks like the codecov/project is pointing to the wrong commit? Going to try and close and then re-open and see if it cycles to the head commit in master.

timflannagan:
> Guess I don't understand what I should do to fix the coverage problem. I'll just wait til your change is complete.

@jchunkins Looks like the checks are being populated with a `Required` status now, but there's still an issue with our tide configuration somewhere where it's still not merging a failed, yet optional check. When I dived into that codecov/project, I tried to re-test using a more up-to-date master commit but it looks like I'm either getting rate limited or simply don't have the requisite permissions for that action. Can you rebase this PR against master, and I'll re-apply the lgtm?

jchunkins:
@timflannagan rebased and pushed

jchunkins:
@timflannagan can you poke the git actions for e2e-kind and e2e-minikube it appears Podman installs are not happening via apt-get. Hopefully whatever was wrong with the repos is fixed or its a transient issue 🤷

timflannagan:
@jchunkins I'm working on fixing that in the background right now as it's affecting other PRs too. Ever since I joined the team, that podman dependency has been real problematic, so hopefully it's as simple as reverting the changes (i.e. workarounds) that were made over the last month or so.

timflannagan:
@jchunkins Alright those fixes landed in #635 - I re-spun the e2e-* jobs but they're still running tests against a stale master commit, which makes sense as you're also introducing changes to those same workflows that were touched. Can you rebase again :laughing:?

timflannagan:
As an aside, this is one of the obvious pain points of migrating from largely prow-based tests to GH actions, but hopefully, we're almost there with merging this.

.github/workflows/test.yml
```diff
@@ -57,26 +47,7 @@ jobs:
...
             && podman version
       - run: |
-          curl -sLo kind "$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/latest | jq -r '[.assets[] | select(.name == "kind-linux-amd64")] | first | .browser_download_url')"
-          chmod +x kind
-          sudo mv kind /bin/
-      - run: |
-          kind create cluster
-          kind export kubeconfig
-          sudo chown -R "$USER" "$HOME/.kube"
-          sudo usermod -aG docker "$USER"
```
timflannagan, line 67: Was this a mistake? I didn't see anything in the scripts/start_registry bash script, but I could be looking at an out-of-date diff?

jchunkins: I've done a lot of force pushing in this branch. But the GitHub PR viewer shows me the script and I see it in Visual Studio Git Graph, as well on my local system, so I think the file is still present.

timflannagan: Apologies, my comment wasn't super clear. I meant removing the process of curling the kind binary down and creating a kind cluster appears to be removed in that diff.

timflannagan: And looking at that script you introduced, it looks like we still want to curl/create that kind cluster, but want to replace the job that's responsible for creating and wiring up the certificates needed to properly authenticate to the local docker container registry, with that script.

jchunkins:
OK merged again... we'll see how it goes

.github/workflows/test.yml
jchunkins: Actually the kind cluster is running directly within the test binary now, so no need to pull it down separately. See https://github.com/operator-framework/operator-registry/pull/605/files#diff-ace2659936556766681fb6406432bca037081e3c00cf76b735ff153865b167a1R70

jchunkins:
@timflannagan Thank you very much for your help getting this across the finish line.

PR #610, opened 2021-03-25
Fix nil function in buildContext return values

komish:

The change sets a return value for `cleanup` (type `func()`) to a no-op at the top of the function logic. This helps prevent a situation where returning before setting that function in `buildContext()` can cause a `<nil>` to be returned for that value which throws an error at runtime when called.

**Concrete Context**

During an `opm index prune` -

In pkg/lib/indexer/indexer.go:

```go
func (i ImageIndexer) PruneFromIndex(request PruneFromIndexRequest) error {
    buildDir, outDockerfile, cleanup, err := buildContext(request.Generate, request.OutDockerfile)
    defer cleanup()
    if err != nil {
        return err
    }

    // ...
```

If the user has not passed `--generate` to `opm index prune` AND they cannot write to the underlying directory, `buildContext(...)` runs into an error trying to `TempDir(...)`:

```go
func buildContext(generate bool, requestedDockerfile string) (buildDir, outDockerfile string, cleanup func(), err error) {
    if generate {
        ...
    }

    // set a temp directory for building the new image
    buildDir, err = ioutil.TempDir(".", tmpBuildDirPrefix)     // error is here
    if err != nil {
        return
    }
    cleanup = func() {
        os.RemoveAll(buildDir)
    }
```

Up until that point `cleanup` hasn't been defined, so it's `<nil>`.  Then, `buildContext` completes and the next line of `PruneFromIndex` calls `defer cleanup()` (i.e. `defer <nil>()`) which causes the panic.

https://bugzilla.redhat.com/show_bug.cgi?id=1943284 - While needing write permissions is fair, other commands return a permission denied (bubbling up from `ioutil.TempDir`). I think that was probably the intention for this command as well.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/indexer/indexer.go
+++ b/pkg/lib/indexer/indexer.go
@@ -394,6 +394,9 @@ func copyDatabaseTo(databaseFile, targetDir string) (string, error) {
 }
 
 func buildContext(generate bool, requestedDockerfile string) (buildDir, outDockerfile string, cleanup func(), err error) {
+	// set cleanup to a no-op until explicitly set
+	cleanup = func() {}
+
 	if generate {
 		buildDir = "./"
 		if len(requestedDockerfile) == 0 {
--- a/pkg/lib/indexer/indexer_test.go
+++ b/pkg/lib/indexer/indexer_test.go
@@ -29,7 +29,7 @@ func TestGetBundlesToExport(t *testing.T) {
 		t.Fatalf("creating querier: %s", err)
 	}
 
-	bundleMap, err := getBundlesToExport(dbQuerier, []string {"etcd"})
+	bundleMap, err := getBundlesToExport(dbQuerier, []string{"etcd"})
 	if err != nil {
 		t.Fatalf("exporting bundles from db: %s", err)
 	}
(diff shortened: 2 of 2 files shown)


komish:
I don't believe similar PR https://github.com/operator-framework/operator-registry/pull/583 addresses this issue (though it addresses another that I ran into as well, so that's awesome).

tonyskapunk:
fwiw, I tested your change @komish and is showing a more clear error message:

```
$ REGISTRY_AUTH_FILE=/path/to/pull-secret.txt \
    ./bin/opm index prune \
    --from-index registry.redhat.io/redhat/redhat-operator-index:v${VER} \
    --packages performance-addon-operator,sriov-network-operator,ptp-operator,kubevirt-hyperconverged \
    --tag ${LOCAL_REGISTRY}/olm/redhat-operator-index:v${VER}

INFO[0000] pruning the index                             packages="[performance-addon-operator sriov-network-operator ptp-operator kubevirt-hyperconverged]"
Error: mkdir index_build_tmp647216914: permission denied

$ echo $?
1
```

Thanks!

ylamgarchal:
LGTM, thanks komish !
Any chances to create these temporary files in /tmp for example ?

komish:
> LGTM, thanks komish !
> Any chances to create these temporary files in /tmp for example ?

@ylamgarchal I think there are some places where you can specify the destination dir (see: `opm index export` flags, which has `--download-folder`) that might be somewhat reusable, but I don't think this does exactly what you're asking for in this case.

Either way, might make sense to request the option as a new issue as there are probably multiple throughout the code base where tmp location to store files could be used.

Maintainers can probably comment further.

joelanford:
/lgtm

Would it be possible to add a unit test to cover this case to prevent regressions?

komish:
@joelanford sure thing. I'm targeting `pkg/lib/indexer/indexer_test.go` and I assume it's ok if I add some additional tests for the `buildContext()` function? If not, I can always roll the changes back and just include a basic test for the `nil` function being returned.

I'll push a commit shortly.

komish:
> /lgtm
> 
> Would it be possible to add a unit test to cover this case to prevent regressions?

Done. Added some basic testing for `buildContext`. I didn't have a reliable way to determine expected values for some of the things that are generated (e.g. `ioutil.TempFiles`) so I tried to account for that in some way by optionally skipping a test if the resulting value would have been generated.

In hindsight, I probably should have added these as a separate PR to prevent the need for re-reviewing. My apologies.

ylamgarchal:
> > LGTM, thanks komish !
> > Any chances to create these temporary files in /tmp for example ?
> 
> @ylamgarchal I think there are some places where you can specify the destination dir (see: `opm index export` flags, which has `--download-folder`) that might be somewhat reusable, but I don't think this does exactly what you're asking for in this case.
> 
> Either way, might make sense to request the option as a new issue as there are probably multiple throughout the code base where tmp location to store files could be used.
> 
> Maintainers can probably comment further.

Hi komish, this temporary directory is used "internally" by the function and cleaned before the command ends. I don't think it makes sense to have such option explicitly from the command line. Maybe a solution would be from the ioutil.TempDir documentation:

> If dir is the empty string, TempDir uses the default directory for temporary files (see os.TempDir).

wdyt

komish:
Just noticing this still hasn't merge. Is there anything needed on my end with regards to this PR? @joelanford @timflannagan

timflannagan:
@komish Neither Joe or I have approval permissions for this repository - you can find the list of people that do in the root OWNERS file. In the meantime, maybe @kevinrizza or @njhale might be able to take a look when they have a chance?

komish:
> @komish Neither Joe or I have approval permissions for this repository - you can find the list of people that do in the root OWNERS file. In the meantime, maybe @kevinrizza or @njhale might be able to take a look when they have a chance?

Understood, my apologies - pinged you both because of the listed Assignees. Thanks for looping in the right folks!

timflannagan:
(Hopefully removing some of the expired context's)

/test all

dinhxuanvu (approved):
/lgtm

PR #612, opened 2021-03-26
(feat) substitutes-for logic

gallettilance:

Implements [substitutes-for enhancement](https://github.com/operator-framework/enhancements/blob/master/enhancements/substitutes-for.md)

**Implementation Check List:**
- [x] Add substitutes-for field to DB (Note: this Isn't required from an implementation perspective but is good to have for visual inspection of the DB)
- [x] Add substitutes for field to CSV type
- [x] Make susbtituted bundle a skips of the substitutes-for bundle
- [x] Bundles in other channels that replace substituted bundles are patched to replace the latest substitutes-for bundle instead
- [x] Bundles in other channels that skip substituted bundles are patched to also skip the substitutes-for bundles
- [x] Handle substitutions of substitutions
- [x] Handle skipRange (*on the assumption that buildID is used to update the version AND buildIDs are not used in the skipRange field)
- [x] Handle adding of a new bundle that replaces a bundle that is substituted for by switching out the replacement to the latest substitution
- [x] Works with batch add (in case the index needs to be rebuilt or rebuilt bundles need to be batch added) 
- [x] Can handle multiple bundles substituting for the same bundle
- [x] Works with non heads of channels
- [x] Test Coverage

**Reviewer Checklist**


bundles/prometheus.0.15.0.substitutesfor/manifests/prometheusoperator.0.15.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,265 @@
+#! parse-kind: ClusterServiceVersion
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: prometheusoperator.0.15.0+1-freshmaker-rebuild
```
gallettilance, line 5: New CSV name

bundles/prometheus.0.15.0.substitutesfor/manifests/prometheusoperator.0.15.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,265 @@
...
+  namespace: placeholder
+  annotations:
+    tectonic-visibility: ocs
+    alm-examples: '[{"apiVersion":"monitoring.coreos.com/v1","kind":"Prometheus","metadata":{"name":"example","labels":{"prometheus":"k8s"}},"spec":{"replicas":2,"version":"v1.7.0","serviceAccountName":"prometheus-k8s","serviceMonitorSelector":{"matchExpressions":[{"key":"k8s-app","operator":"Exists"}]},"ruleSelector":{"matchLabels":{"role":"prometheus-rulefiles","prometheus":"k8s"}},"resources":{"requests":{"memory":"400Mi"}},"alerting":{"alertmanagers":[{"namespace":"monitoring","name":"alertmanager-main","port":"web"}]}}},{"apiVersion":"monitoring.coreos.com/v1","kind":"ServiceMonitor","metadata":{"name":"example","labels":{"k8s-app":"prometheus"}},"spec":{"selector":{"matchLabels":{"k8s-app":"prometheus","prometheus":"k8s"}},"namespaceSelector":{"matchNames":["monitoring"]},"endpoints":[{"port":"web","interval":"30s"}]}},{"apiVersion":"monitoring.coreos.com/v1","kind":"Alertmanager","metadata":{"name":"alertmanager-main"},"spec":{"replicas":3}}]'
+spec:
+  replaces: prometheusoperator.0.14.0
+  substitutes-for: prometheusoperator.0.15.0
```
gallettilance, line 12: This is the substitutes-for field

bundles/prometheus.0.15.0.substitutesfor/manifests/prometheusoperator.0.15.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,265 @@
...
+                  name: http
+                resources:
+                  limits:
+                    cpu: 200m
+                    memory: 100Mi
+                  requests:
+                    cpu: 100m
+                    memory: 50Mi
+  maturity: alpha
+  version: 0.15.0+1
```
gallettilance, line 173: this is the updated version

pkg/registry/interface.go
```diff
@@ -10,7 +10,7 @@ type Load interface {
 	AddOperatorBundle(bundle *Bundle) error
 	AddBundleSemver(graph *Package, bundle *Bundle) error
 	AddPackageChannels(manifest PackageManifest) error
-	AddBundlePackageChannels(manifest PackageManifest, bundle *Bundle) error
+	// AddBundlePackageChannels(manifest PackageManifest, bundle *Bundle) error
```
gallettilance, line 13: Note: commented this out because it doesn't seem to be used anywhere

pkg/registry/populator.go
```diff
@@ -349,18 +349,18 @@ func (i *ImageInput) findCSV(manifests string) (*unstructured.Unstructured, erro
...
+// 	if manifest.PackageName == "" {
+// 		return nil
+// 	}
+
+// 	if err := i.loader.AddBundlePackageChannels(manifest, bundle); err != nil {
+// 		return fmt.Errorf("error loading bundle into db: %s", err)
+// 	}
+
+// 	return nil
+// }
```
gallettilance, line 363: Note: commented this out because it doesn't seem to be used anywhere

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,78 @@ func TestSemverPackageManifest(t *testing.T) {
...
+	// Check channel head is unchanged for preview channel
+	head, err := query.GetCurrentCSVNameForChannel(context.Background(), "prometheus", "preview")
+	require.NoError(t, err)
+	require.Equal(t, "prometheusoperator.0.22.2", head)
+
+	// Check channel head is changed for stable channel
+	head, err = query.GetCurrentCSVNameForChannel(context.Background(), "prometheus", "stable")
+	require.NoError(t, err)
+	require.Equal(t, "prometheusoperator.0.15.0+1-freshmaker-rebuild", head)
+}
```
gallettilance, line 1611: Note: still need more tests for more complex cases and more thorough graph checking

pkg/sqlite/load.go
```diff
@@ -757,29 +790,29 @@ func (s *sqlLoader) AddBundleSemver(graph *registry.Package, bundle *registry.Bu
...
-		return err
-	}
+// 	if err := s.addPackageChannels(tx, manifest); err != nil {
+// 		return err
+// 	}
 
-	return tx.Commit()
-}
+// 	return tx.Commit()
+// }
```
gallettilance, line 815: Note: commented this out because it doesn't seem to be used anywhere

pkg/registry/interface.go
camilamacedo86: could we remove instead, so?

pkg/registry/populator.go
camilamacedo86: same

timflannagan:
@gallettilance We can ignore the ci/prow/images test failures for now as the ci-operator is expecting to build the registry image using a Dockerfile that doesn't exist in this repository.

njhale:
> @gallettilance We can ignore the ci/prow/images test failures for now as the ci-operator is expecting to build the registry image using a Dockerfile that doesn't exist in this repository.

doesn't this also impact the signal for `ci/prow/e2e-aws`?

pkg/sqlite/load.go
njhale: If it's unused, let's delete it entirely.

timflannagan:
> doesn't this also impact the signal for ci/prow/e2e-aws?

Yeah, sorry I should've been more explicit - the images test is an implicit dependency for running any of the e2e tests, so we can also ignore those tests as well.

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,425 @@ func TestSemverPackageManifest(t *testing.T) {
...
+					"quay.io/test/prometheus.0.14.0",
+				},
+				channels: []string{
+					"stable",
+					"preview",
+					"preview",
+					"stable",
+					"stable",
+					"preview",
+					"preview",
```
njhale, line 1583: I think the expected graph would be a lot easier to understand if we organized replacement chains by channel; e.g.

```go
channels: map[string][]Replaces{ // channel name/replacement tuples in channel
    "stable": [{ next: "prometheusoperator.0.15.0+1-freshmaker-rebuild", prev: "..."}, ...],
    "preview": [{ next: "prometheusoperator.0.22.2", prev: "..."}, ...],
}
```

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,425 @@ func TestSemverPackageManifest(t *testing.T) {
...
+					"prometheus.0.14.0",
+					"prometheus.0.15.0",
+				},
+				rebuilds: []string{
+					"prometheus.0.14.0.substitutesfor",
+				},
+			},
+			expected: expected{
+				bundles: []string{
+					"quay.io/test/prometheus.0.22.2",
```
njhale, line 1606: generating these fixtures at runtime could be preferable. not sure where others stand on this.

pkg/sqlite/load.go
```diff
@@ -67,7 +67,7 @@ func (s *sqlLoader) AddOperatorBundle(bundle *registry.Bundle) error {
 }
 
 func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error {
-	addBundle, err := tx.Prepare("insert into operatorbundle(name, csv, bundle, bundlepath, version, skiprange, replaces, skips) values(?, ?, ?, ?, ?, ?, ?, ?)")
+	addBundle, err := tx.Prepare("insert into operatorbundle(name, csv, bundle, bundlepath, version, skiprange, replaces, skips, substitutesfor) values(?, ?, ?, ?, ?, ?, ?, ?, ?)")
```
njhale, line 70: is there a reason why we shouldn't use `properties` table w/ a `substitutesFor` property and forgo the database schema changes?

pkg/sqlite/load.go
```diff
@@ -104,8 +110,85 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
...
+			skips = append(skips, replaces)
+			// update the replaces to a newer substitution
+			replaces = substitutesFor
+			// try to get the substitution of the substitution
+			substitutesFor, err = s.getBundleThatSubstitutesFor(tx, replaces)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
njhale, line 155: IIUC, it seems like altering the graph here makes `addOperatorBundle` non-commutative; i.e. order of an addition matters -- which is something we can't have.

I think we want to do this _after_ all of the bundles have been added to the database, so we have complete substitutes-for information.

bundles/prometheus.0.14.0.substitutesfor/manifests/alertmanager.crd.yaml
```diff
@@ -0,0 +1,2398 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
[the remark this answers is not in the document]
exdx: Also setting the `apiVersion` to `v1` for any CRDs would be good, since `v1beta1` will be deprecated in the next kube bump

bundles/prometheus.0.15.0.substitutesfor/manifests/servicemonitor.crd.yaml
```diff
@@ -0,0 +1,224 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
exdx, line 1: nit: same note about CRD versioning ie using `v1` instead of `v1beta1`. I can see that since real CRDs and operators are being used in the tests this can be annoying, but its a case of doing this now or in a few months

bundles/prometheus.0.14.0.substitutesfor/manifests/alertmanager.crd.yaml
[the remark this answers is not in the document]
exdx: I can see the appeal in using production quality bundles for the tests but agree its not strictly necessary if we are just interested in testing the substitutes-for functionality

pkg/registry/populator.go
exdx: Let's remove dead code?

bundles/prometheus.0.14.0.substitutesfor/manifests/prometheusoperator.0.14.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,241 @@
...
+                  name: http
+                resources:
+                  limits:
+                    cpu: 200m
+                    memory: 100Mi
+                  requests:
+                    cpu: 100m
+                    memory: 50Mi
+  maturity: alpha
+  version: 0.14.0+1
```
gallettilance, line 158: @njhale

bundles/prometheus.0.14.0.substitutesfor/manifests/prometheusoperator.0.14.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,241 @@
+#! parse-kind: ClusterServiceVersion
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: prometheusoperator.0.14.0+1-rebuild
```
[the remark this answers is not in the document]
gallettilance: https://github.com/operator-framework/operator-registry/pull/612/files/91dcef308d7265847c6b9a756d24b118eacd6619#r603535195

bundles/prometheus.0.14.0.substitutesfor/manifests/alertmanager.crd.yaml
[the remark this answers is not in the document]
gallettilance: Yeah I really want the most realistic test possible here in order to be as confident as possible wrt production impact.

@exdx
```
Also setting the apiVersion to v1 for any CRDs would be good, since v1beta1 will be deprecated in the next kube bump
```
should I change all the crd versions everywhere then?

pkg/registry/csv.go
```diff
@@ -47,6 +47,9 @@ const (
 
 	// The yaml attribute that specifies the skipRange of the ClusterServiceVersion
 	skipRangeAnnotationKey = "olm.skipRange"
+
+	// The yaml attibure that specifies the optional substitutesfor of the ClusterServiceVersion
+	substitutesFor = "substitutes-for"
```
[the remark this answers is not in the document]
gallettilance: sounds good to me!

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,425 @@ func TestSemverPackageManifest(t *testing.T) {
...
+			bundles, err := query.ListBundles(context.Background())
+			require.NoError(t, err)
+			var bundlePaths []string
+			var channels []string
+			for _, bundle := range bundles {
+				bundlePaths = append(bundlePaths, bundle.BundlePath)
+				channels = append(channels, bundle.ChannelName)
+			}
+			require.ElementsMatch(t, tt.expected.bundles, bundlePaths)
+			require.ElementsMatch(t, tt.expected.channels, channels)
```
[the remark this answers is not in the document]
gallettilance: yeah the test just checks that the channel heads are correct. I probably need to check the rest of the channel is what I expect

pkg/registry/populator_test.go
gallettilance: what do you mean?

pkg/sqlite/load.go
gallettilance: no reason - what is better in your opinion wrt declaritive config?

bundles/prometheus.0.14.0.substitutesfor/manifests/alertmanager.crd.yaml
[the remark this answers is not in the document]
exdx: I would recommend it, yes. You will have to mess with the YAML CRD definitions a bit though for kube to accept them. You have a schema, but the v1 wants the schema to be on a per-version level versus at the top level. 

BUT at the same time, since these fixtures are never going to be applied on-cluster, idk if it's even worth the effort now that I'm thinking about it. Maybe not worth the effort unless others feel strongly.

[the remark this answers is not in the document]
exdx: We may remove the ability to even add a v1beta1 CRD to a bundle though , so perhaps it does make sense actually

pkg/sqlite/load.go
gallettilance: all the tests have shown me that it's still commutative. Let me know if you see something I don't

pkg/registry/interface.go
gallettilance: sounds good - will do! Just thought to flag it here for visibility

bundles/prometheus.0.14.0.substitutesfor/manifests/alertmanager.crd.yaml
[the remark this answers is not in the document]
njhale: > Yeah I really want the most realistic test possible here in order to be as confident as possible wrt production impact.

Ultimately, we're not changing anything related to CRD parsing though. Supporting enormous test fixtures is a burden on the team and, honestly,  it's hard to grok.

bundles/prometheus.0.14.0.substitutesfor/manifests/prometheusoperator.0.14.0.clusterserviceversion.yaml
njhale: the prerelease suffix `-rebuild` in the name is not here in the version

pkg/sqlite/load.go
njhale: this seems to be exactly the kind of the thing properties were made for.

pkg/sqlite/load.go
njhale: the tests were a little difficult for me to grok. 

stepping through the code a second time, it looks like this works out for `replaces`.

bundles/prometheus.0.14.0.substitutesfor/manifests/prometheusoperator.0.14.0.clusterserviceversion.yaml
gallettilance: freshmaker will / should only use buildIDs to update the version it's rebuilding.

pkg/sqlite/migrations/011_substitutes_for.go
```diff
@@ -0,0 +1,55 @@
...
+		sql := `
+		ALTER TABLE operatorbundle 
+		ADD COLUMN substitutesfor TEXT;
+		`
+		_, err := tx.ExecContext(ctx, sql)
+		return err
+	},
+	Down: func(ctx context.Context, tx *sql.Tx) error {
+		foreignKeyOff := `PRAGMA foreign_keys = 0`
+		createTempTable := `CREATE TABLE operatorbundle_backup (name TEXT, csv TEXT, bundle TEXT, bundlepath TEXT, version TEXT, skiprange TEXT, replaces TEXT, skips TEXT)`
```
ecordell, line 27: Is this dropping the foreign keys from the table entirely?

pkg/registry/populator_test.go
ecordell: I think nick's point is that the "substitutes-for" variants are 1 line different from the other manifests, we could create these fixtures by loading one of the existing fixtures and mutating it.

pkg/sqlite/load.go
```diff
@@ -104,8 +110,85 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 	if err != nil {
 		return err
 	}
+	substitutesFor, err := bundle.SubstitutesFor()
+	if err != nil {
+		return err
+	}
+
+	// Update the replaces that reference the substituted-for bundle
+	if substitutesFor != "" {
```
ecordell, line 119: I think the substitutes-for processing generally needs to happen after all bundles have been added, right?

If I add a bundle after this `add` has run that would have been substituted by this bundle, it will be missed. There's an existing division in the codebase around this: 1. load all bundles, 2. wire them up.

ecordell (changes requested):

pkg/sqlite/migrations/011_substitutes_for.go
gallettilance: oh wow - yes it is... good catch!

pkg/sqlite/load.go
gallettilance: I can add a test for that but we should be ok in that respect because the graph gets rebuilt everytime from information stored during 1.

bundles/prometheus.0.15.0.substitutesfor/manifests/prometheusoperator.0.15.0.clusterserviceversion.yaml
njhale: related: does OLM generate any resources with tighter [name constraints](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) using CSV names?

pkg/registry/populator.go
```diff
@@ -373,18 +373,24 @@ type bundleVersion struct {
 // compare returns a value less than one if the receiver arg is less smaller the given version, greater than one if it is larger, and zero if they are equal.
 // This comparison follows typical semver precedence rules, with one addition: whenever two versions are equal with the exception of their build-ids, the build-ids are compared using prerelease precedence rules. Further, versions with no build-id are always less than versions with build-ids; e.g. 1.0.0 < 1.0.0+1.
 func (b bundleVersion) compare(v bundleVersion) (int, error) {
-	if c := b.version.Compare(v.version); c != 0 {
+	return BuildIdCompare(b.version, v.version)
+}
+
+// BuildIdCompare compares two versions and returns negative one if the first arg is less than the second arg, positive one if it is larger, and zero if they are equal.
+// This comparison follows typical semver precedence rules, with one addition: whenever two versions are equal with the exception of their build-ids, the build-ids are compared using prerelease precedence rules. Further, versions with no build-id are always less than versions with build-ids; e.g. 1.0.0 < 1.0.0+1.
+func BuildIdCompare(b semver.Version, v semver.Version) (int, error) {
```
njhale, line 381: If we're going to expose this comparison and use it across multiple packages, we should give it a package of its own.

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,759 @@ func TestSemverPackageManifest(t *testing.T) {
...
+					"stable",
+					"preview",
+					"preview",
+				},
+				substitutions: map[string]string{
+					"prometheusoperator.0.22.2":                      "",
+					"prometheusoperator.0.14.0":                      "",
+					"prometheusoperator.0.15.0":                      "",
+					"prometheusoperator.0.15.0+1-freshmaker-rebuild": "prometheusoperator.0.15.0",
+				},
```
njhale, line 1593: Do you actually need any of this when you have already have `whatReplaces`, `channelHeads`, and `defaultChannel` representing the effective graph already?

pkg/sqlite/load.go
njhale: To clarify -- I think we should use properties here. I believe this PR can be simplified quite a bit.

I would like to hear some other voices though (cc @benluddy @ecordell @joelanford)

joelanford: My preference would be to reduce the number of properties that directly translate to upgrade edges. From a user perspective, if I see a declarative config that uses 3 or 4 different properties (channel, skips, skipRange, substitutesFor), it becomes harder and harder to build a mental model of that and figure out when I should use one property over another.

joelanford: See https://github.com/operator-framework/operator-registry/pull/617#discussion_r605121223

pkg/registry/populator_test.go
njhale: Yes, in that vein. I see two appealing benefits:

1. Reduce the footprint of our test fixtures
2. Make it clear to the test reader how the fixture has been perturbed

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,759 @@ func TestSemverPackageManifest(t *testing.T) {
...
+				whatReplaces: map[string]replaces{
+					"prometheusoperator.0.22.2": {
+						"preview": "",
+					},
```
njhale, line 1597: The fact that a "blank replaces" is needed to indicate channel membership makes me think we want to change around the relationship here. Is there a practical reason this is preferred over [the example in the previous comment](https://github.com/operator-framework/operator-registry/pull/612/files#r603506828)?

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,759 @@ func TestSemverPackageManifest(t *testing.T) {
...
+			for _, bundle := range bundles {
+				bundlePaths = append(bundlePaths, bundle.BundlePath)
+				channels = append(channels, bundle.ChannelName)
+				bundleThatReplaces, _ := query.GetBundleThatReplaces(context.Background(), bundle.CsvName, bundle.PackageName, bundle.ChannelName)
+				if bundleThatReplaces != nil {
+					require.Equal(t, tt.expected.whatReplaces[bundle.CsvName][bundle.ChannelName], bundleThatReplaces.CsvName)
+				} else {
+					require.Equal(t, tt.expected.whatReplaces[bundle.CsvName][bundle.ChannelName], "")
+				}
+				substitution, err := query.GetBundleSubstitution(context.Background(), bundle.CsvName)
```
njhale, line 2272: Is this method only used in this test?

gallettilance: moved it to the tests only

pkg/registry/populator_test.go
gallettilance: I'm not sure I understand. Are you just saying that checking channel heads and channel membership is redundant after this check?

pkg/registry/populator_test.go
gallettilance: Makes sense! Let's see what other reviewers think.

pkg/registry/populator.go
```diff
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"github.com/operator-framework/operator-registry/pkg/image"
+	csemver "github.com/operator-framework/operator-registry/pkg/lib/semver"
```
njhale, line 19: nit: `libsemver`

pkg/lib/semver/semver.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+
+	"github.com/blang/semver"
+)
+
+// BuildIdCompare compares two versions and returns negative one if the first arg is less than the second arg, positive one if it is larger, and zero if they are equal.
+// This comparison follows typical semver precedence rules, with one addition: whenever two versions are equal with the exception of their build-ids, the build-ids are compared using prerelease precedence rules. Further, versions with no build-id are always less than versions with build-ids; e.g. 1.0.0 < 1.0.0+1.
+func BuildIdCompare(b semver.Version, v semver.Version) (int, error) {
```
njhale, line 11: nit: would be nice to have a few tests directly confirming that this works as expected.

pkg/registry/populator_test.go
njhale: I'm saying that `<bundle> -> []<channel>:<replaces>` is hard to grok and looks odd when you have a single node graph for a channel. I'm suggesting that `channel -> []<bundle>:<replacement>` conveys the same information without looking odd for that edge case (also, I think it gives a cleaner view of the actual path an upgrade could take).

pkg/lib/semver/semver.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return nil, err
+		}
+		pre = append(pre, p)
+	}
+
+	var major uint64
+	if len(pre) > 0 {
+		// Adjust for the case where we compare a build-id prerelease analog to a version without a build-id.
+		// Without this `0.0.0+1` and `0.0.0` would become `0.0.0-1` and `0.0.0`, where the rules of prerelease comparison would
```
ecordell, line 42: we only need to call this comparison if the original version `A.Compare(B) == 0`, so this should never happen, right?

pkg/registry/csv.go
```diff
@@ -341,3 +344,26 @@ func (csv *ClusterServiceVersion) GetOperatorImages() (map[string]struct{}, erro
 
 	return images, nil
 }
+
+// GetSubstitutesFor returns the name of the ClusterServiceVersion object that
+// is substituted by this ClusterServiceVersion object.
+//
+// If not defined, the function returns an empty string.
+func (csv *ClusterServiceVersion) GetSubstitutesFor() (string, error) {
```
ecordell, line 352: This won't find substitutes-for if it's in a property, right?

gallettilance: No this change is purely on the CSV. We'll have to make a change when csvless bundles arrive to make substitutes for a field in the properties.yaml file.

pkg/lib/semver/semver.go
gallettilance: cc @njhale

pkg/registry/csv.go
joelanford: Do we envision substitutes-for being a permanent graph-altering property? (I hope not)

If we can express substitutes-for semantics with existing graph-defining fields (e.g. replaces/skips), then I don't think we want yet another field thrown into the mix.

See my related comment: https://github.com/operator-framework/operator-registry/pull/612#discussion_r605923786

gallettilance: what do you mean by permanent graph altering property? I expect the substitution to take affect until a new substitution or bundle defines a different graph just like replaces / skips works today.

pkg/sqlite/load.go
gallettilance: we're banking on the visualization work being done but this is a good point. What alternative are you suggesting?

bundles/prometheus.0.14.0.substitutesfor/manifests/prometheusoperator.0.14.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,242 @@
...
+                  name: http
+                resources:
+                  limits:
+                    cpu: 200m
+                    memory: 100Mi
+                  requests:
+                    cpu: 100m
+                    memory: 50Mi
+  maturity: alpha
+  version: 0.14.0+0.1234
```
Xia-Zhao-rh, line 159: Is it support the version set to be 0.14.0-rebuild-20210304?

pkg/registry/populator_test.go
```diff
@@ -1534,3 +1534,709 @@ func TestSemverPackageManifest(t *testing.T) {
...
+	type expected struct {
+		err error
+	}
+	tests := []struct {
+		description string
+		args        args
+		expected    expected
+	}{
+		{
+			description: "EnableAlpha/SubstitutesForTrue",
```
ecordell, line 2175: nit: name stuttering, the test will be prefixed with EnableAlpha already

pkg/sqlite/load.go
```diff
@@ -104,8 +107,16 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 	if err != nil {
 		return err
 	}
+	substitutesFor, err := bundle.SubstitutesFor()
+	if err != nil {
+		return err
+	}
+
+	if substitutesFor != "" && !s.enableAlpha {
```
ecordell, line 115: is this the behavior we want? I think a warning might be more appropriate instead of stopping with an error

pkg/registry/populator_test.go
gallettilance: ah right!

pkg/sqlite/load.go
ecordell: discussed offline. there are not a lot of reasonable ways to use a bundle with substitutes-for set without adding it to a catalog that has substitution enabled.

pkg/registry/csv.go
```diff
@@ -58,6 +58,9 @@ const (
 
 	// The yaml attribute that specifies the skipRange of the ClusterServiceVersion
 	skipRangeAnnotationKey = "olm.skipRange"
+
+	// The yaml attibure that specifies the optional substitutesfor of the ClusterServiceVersion
```
ecordell, line 62:
```suggestion
	// The yaml attribute that specifies the optional substitutesfor of the ClusterServiceVersion
```

pkg/registry/csv.go
```diff
@@ -58,6 +58,9 @@ const (
...
+	substitutesForAnnotationKey = "substitutesFor"
```
ecordell, line 63: if this is for an annotation it should be namespaced - we have used `olm` for skiprange

pkg/sqlite/load.go
```diff
@@ -7,15 +7,18 @@ import (
...
+	libsemver "github.com/operator-framework/operator-registry/pkg/lib/semver"
 	"github.com/operator-framework/operator-registry/pkg/registry"
 )
 
 type sqlLoader struct {
-	db       *sql.DB
-	migrator Migrator
+	db          *sql.DB
+	migrator    Migrator
+	enableAlpha *bool
```
ecordell, line 21: by the time we get here I don't think we need a tri-state. let's just make it a `bool`?

pkg/sqlite/load.go
```diff
@@ -40,7 +43,7 @@ func NewSQLLiteLoader(db *sql.DB, opts ...DbOption) (MigratableLoader, error) {
 		return nil, err
 	}
 
-	return &sqlLoader{db: db, migrator: migrator}, nil
+	return &sqlLoader{db: db, migrator: migrator, enableAlpha: &options.EnableAlpha}, nil
```
ecordell, line 46: I think it would keep more in line with the current apis if this were a DbOption you pass in:

`NewSqliteLoader(db, WithEnableAlpha(true))`

see db_options.go for other examples.

pkg/sqlite/load.go
```diff
@@ -130,9 +141,238 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
...
+	}
+	version, err := bundle.Version()
+	if err != nil {
+		return err
+	}
+	substitutesFor, err := bundle.SubstitutesFor()
+	if err != nil {
+		return err
+	}
+	if substitutesFor != "" {
```
ecordell, line 198: this is a major branch, is codecov lying about it not being covered?

pkg/sqlite/load.go
```diff
@@ -130,9 +141,238 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
...
+				currentSubstitutionVersion, err := semver.Parse(version)
+				if err != nil {
+					return err
+				}
+				// Compare versions
+				c, err := libsemver.BuildIdCompare(otherSubstitutionVersion, currentSubstitutionVersion)
+				if err != nil {
+					return err
+				}
+				if c < 0 {
```
ecordell, line 233: same here

pkg/sqlite/load.go
gallettilance: yeah that doesn't look right to me - at least from the tests I've written. What do you think?

ecordell: I agree, just very confusing.

maybe codecov comments don't keep up with force pushes?

Issue #609, opened 2021-03-25
Re-enable the "opm alpha bundle unpack" e2e tests


timflannagan:
The e2e tests around the "opm alpha bundle unpack" command consistently produce failed kind/minikube e2e runs as we expect an error running that command with an invalid bundle, and the `--skip-validation` flag being ommitted, yet no error is produced. This results in the following error message being logged:

```
• Failure [0.880 seconds]
opm alpha bundle
/home/runner/work/operator-registry/operator-registry/test/e2e/opm_bundle_test.go:17
  for an invalid bundle
  /home/runner/work/operator-registry/operator-registry/test/e2e/opm_bundle_test.go:28
    fails to unpack [It]
    /home/runner/work/operator-registry/operator-registry/test/e2e/opm_bundle_test.go:77

    Expected failure, but got no error.

    /home/runner/work/operator-registry/operator-registry/test/e2e/opm_bundle_test.go:90
```

Note: these failures are difficult to reproduce locally, but showing up in github action runs enough to track this to reduce the number of e2e flakes we see.

timflannagan:
/retitle Re-enable the "opm alpha bundle unpack" e2e tests

PR #613, opened 2021-03-26
test/e2e: Disable the bundle unpack opm e2e tests

timflannagan:
Short term fix for #609 

**Reviewer Checklist**


timflannagan:
This doesn't affect any of the OCP e2e so I don't want to take a lease away from an actual tests that needs it.

/override ci/prow/e2e-aws

njhale:
/approve

this is under `alpha` and this looks like it may be an issue with the test itself. let's get it out of the way for now until we get a chance to fix it.

PR #615, opened 2021-03-30
cleanup dead code

gallettilance:

These functions don't seem to be used anywhere

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/interface.go
+++ b/pkg/registry/interface.go
@@ -10,7 +10,6 @@ type Load interface {
 	AddOperatorBundle(bundle *Bundle) error
 	AddBundleSemver(graph *Package, bundle *Bundle) error
 	AddPackageChannels(manifest PackageManifest) error
-	AddBundlePackageChannels(manifest PackageManifest, bundle *Bundle) error
 	RemovePackage(packageName string) error
 	RemoveStrandedBundles() error
 	DeprecateBundle(path string) error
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
This is a public API. Did we put up some sort of deprecation notice before removing it completely from gRPC API?

gallettilance:
@dinhxuanvu who would be using the grpc api to add things to the DB? is that something we even have documented as possible / supported?

dinhxuanvu:
> @dinhxuanvu who would be using the grpc api to add things to the DB? is that something we even have documented as possible / supported?
> 
> I would agree with the deprecation notice for the GET requests - it seems a bit excessive in this case.

I'm not aware of any users who use this particular API endpoint. Still just because we don't know who uses it, it doesn't change the fact it is a public API endpoint that people can use. We don't document every gRPC endpoints that are available but they are present in the code. I'm all for removing code that we no longer use but registry is v1 which means we have to support and deprecate these kind of things properly.

gallettilance:
@dinhxuanvu Ok makes sense - what is the process for sending deprecation notices?

dinhxuanvu:
> @dinhxuanvu Ok makes sense - what is the process for sending deprecation notices?

@anik120 probably knows this part better than me. Maybe you can get in touch with him on this.

anik120:
@gallettilance for "openshift specific" APIs we a) start logging deprecation notices from a particular release b) document them in docs.openshift.com, along with migration strategy etc. A release following that we auto-migrate remaining api clients (whatever that means for your particular api). The following release (2nd releases since deprecation notice was sent out) we remove the api. 

This one however though is probably a great place to start thinking about our upstream deprecation strategy. I'd say we stick to the same (upstream) releases cycle schedule. So we can change this PR to log the deprecation notices if these APIs are called for the 1.17.0 release of operator-registry. We should also document deprecation notice in our release notes for 1.17.0. 
We can then remove the apis in 1.19.0.

PR #617, opened 2021-03-31
feat(declarative config): add base packages

joelanford:
This PR includes the base packages that make up the set of shared dependencies needed to build other declarative-config based `opm` commands, including `opm serve`, `opm add` and `opm validate`.

See the [declarative config enhancement proposal.](https://github.com/operator-framework/enhancements/blob/master/enhancements/declarative-index-config.md)

**Reviewer Checklist**
- [ ] ~Sufficient end-to-end test coverage~
- [ ] ~Docs updated or added to `/docs`~ 


pkg/declcfg/write.go
```diff
@@ -0,0 +1,162 @@
...
+		return fmt.Errorf("parse properties: %v", err)
+	}
+	if len(props.BundleObjects) != len(b.Objects) {
+		return fmt.Errorf("expected %d properties of type %q, found %d", len(b.Objects), property.TypeBundleObject, len(props.BundleObjects))
+	}
+	for i, p := range props.BundleObjects {
+		if p.IsRef() {
+			objPath := filepath.Join(baseDir, p.GetRef())
+			objDir := filepath.Dir(objPath)
+			if err := w.MkdirAll(objDir, 0755); err != nil {
```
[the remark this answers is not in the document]
joelanford: I suppose `umask` still applies, so I can't think of a reason not to use 0777.

pkg/declcfg/write.go
```diff
@@ -0,0 +1,162 @@
...
+
+	for _, p := range cfg.Packages {
+		fcfg := DeclarativeConfig{
+			Packages: []Package{p},
+			Bundles:  bundlesByPackage[p.Name],
+			Others:   othersByPackage[p.Name],
+		}
+		pkgDir := filepath.Join(rootDir, p.Name)
+		if err := w.MkdirAll(pkgDir, 0755); err != nil {
+			return fmt.Errorf("mkdir %q: %v", pkgDir, err)
```
estroz, line 75: Nit: `MkdirAll()` already returns an error with message prefix `mkdir <dir>:`.
```suggestion
			return err
```

pkg/model/model.go
```diff
@@ -0,0 +1,295 @@
...
+type Channel struct {
+	Package *Package
+	Name    string
+	Bundles map[string]*Bundle
+}
+
+// TODO(joelanford): This function determines the channel head by finding the bundle that has 0
+//   incoming edges, based on replaces and skips. It also expects to find exactly one such bundle.
+//   Is this the correct algorithm?
+func (c Channel) Head() (*Bundle, error) {
```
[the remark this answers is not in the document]
joelanford: It seems like adding `substitutes-for` support in declarative config is unnecessary complexity. In declarative config, the upgrade graph can be edited at any point in a process that produces or updates the final index image.

In declarative-config world, inserting a new bundle that substitutes for another bundle is fairly easy:
1. Add a new `olm.bundle` blobs for (using that EP as the example) `1.0.1-patched`, with channel properties that include `replaces: 1.0.0` for each relevant channel.
2. Find all `olm.bundle` blobs that contain a channel property with `replaces: 1.0.0` in the relevant channels, and update those properties to `replaces: 1.0.1-patched`

No changes to the bundle images necessary.

pkg/model/model.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	return result.ErrorOrNil()
+}
+
+type Icon struct {
+	Data      []byte
+	MediaType string
+}
+
+func (i *Icon) Validate() error {
```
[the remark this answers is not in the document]
joelanford: The benefit of putting the validation on the model is that any representation that can be converted to the model can benefit from the exact same validation.

We could extract these somewhat peripheral validations into separate functions on the model that aren't always called when we're validating the core invariants.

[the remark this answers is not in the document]
estroz: Simple validation to make sure an icon is actually an image and has the expected media type seems fine, otherwise the icon could be filled with junk that would not work with downstream consumers. This validation will always run on a package, unlike a validation command which may or may not have been run by the package maintainer.

pkg/model/model.go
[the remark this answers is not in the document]
estroz: Wouldn't the substitution problem go away with declarative configs because the update graph is explicitly defined? I think this is addressed in the [alternatives section](https://github.com/operator-framework/enhancements/blob/master/enhancements/substitutes-for.md#declarative-config).

pkg/model/model.go
```diff
@@ -0,0 +1,295 @@
...
+	for i, skip := range b.Skips {
+		if skip == "" {
+			result = multierror.Append(result, fmt.Errorf("skip[%d] is empty", i))
+		}
+	}
+	// TODO(joelanford): Validate related images? It looks like some
+	//   CSVs in production databases use incorrect fields ([name,value]
+	//   instead of [name,image]), which results in empty image values.
+	//   Example is in redhat-operators: 3scale-operator.v0.5.5
+	//for i, relatedImage := range b.RelatedImages {
```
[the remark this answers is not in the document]
joelanford: Possibly? But related images is a first-class field with a well-defined API. If anything, perhaps this check is redundant because other components are catching it, but clearly that isn't the case because existing CSVs in production catalogs were failing this until I commented it out.

pkg/declcfg/declcfg_to_model.go
```diff
@@ -0,0 +1,108 @@
...
+func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
+	mpkgs := model.Model{}
+	defaultChannels := map[string]string{}
+	for _, p := range cfg.Packages {
+		mpkg := &model.Package{
+			Name:        p.Name,
+			Description: p.Description,
+			Channels:    map[string]*model.Channel{},
+		}
+		if p.Icon != nil {
```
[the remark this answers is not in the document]
joelanford: I don't think I understand your suggestion and how it relates to this particular piece of code.

In general the answer to your question is:
1. We persist icons in `olm.package` blobs in an optional first class field called `icon`. This field contains a base64-encoded `data` value and a mediatype `value` that specifies the format of the data.
2. The icon is in the `olm.package` schema (rather than `olm.bundle`) to reduce unnecessary duplication of the same large icon data blob in all of the `olm.bundle` blobs.

If that's what you're looking for, it seems more applicable as GoDoc comments on the `declcfg.Package` or `declcfg.Icon` types. WDYT?

dinhxuanvu (commented):
Looking good. I feel like I'm ready to lgtm on this PR.

pkg/model/model.go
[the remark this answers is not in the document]
joelanford: I agree, Kevin. So it sounds like we're saying:
1. substitutes-for is a way to tell tooling how to build upgrade edges in an index.
2. the upgrade edge primitives that exist today in declarative config are `olm.skips` and `olm.channel` (with replaces) properties.
3. Index maintainers that wish to consider `substitutes-for` for edge generation in declarative configs will use a tool that considers `substitutes-for` values when manipulating `olm.skips` and `olm.channel` properties in `olm.bundle` blobs.

Does that sound right?

anik120:
@joelanford is it possible to configure deepcopy-gen with this package? 
I couldn't find any other place in this repo that uses deepcopy-gen, so it'd need to be plugged in all the way up to the Makefile.

joelanford:
@anik120 Why does the model need deepcopy functions?

anik120:
@joelanford the IR has a `Channel.Bundles` field deals with `*Bundles`, looks like there'll be many instances where we'll have to copy structs around. Might be a good idea to get ahead and include generated `DeepCopy()` to prevent code duplication.

pkg/model/model.go
```diff
@@ -0,0 +1,295 @@
...
+	return result.ErrorOrNil()
+}
+
+type Bundle struct {
+	Package       *Package
+	Channel       *Channel
+	Name          string
+	Image         string
+	Replaces      string
+	Skips         []string
```
anik120, line 208: I seem to be missing something, did we decided on not storing the skipRange for the bundle?

joelanford: There's a property for skipRange.

pkg/declcfg/load_test.go
```diff
@@ -0,0 +1,124 @@
...
+		t.Run(s.name, func(t *testing.T) {
+			f, err := os.Open(s.file)
+			require.NoError(t, err)
+
+			cfg, err := readJSON(f)
+			s.assertion(t, err)
+			if err == nil {
+				require.NotNil(t, cfg)
+				assert.Equal(t, len(cfg.Packages), s.expectNumPackages, "unexpected package count")
+				assert.Equal(t, len(cfg.Bundles), s.expectNumBundles, "unexpected bundle count")
```
anik120, line 75: What do you think about a  `func (d *DeclarativeConfig) equal (cfg *DeclarativeConfig) bool`? 

These tests are checking if the number of packages/bundles etc are equal to some expected value, but not sure they're proving that the config directory is being converted to the `*DeclarativeConfig` we expect? 
`equal` could help us write down what an expected config is supposed to look like and then check the loaded config against that.

joelanford: Yeah that's a good idea :+1:

pkg/declcfg/write.go
```diff
@@ -0,0 +1,161 @@
...
+func writeToFS(cfg DeclarativeConfig, w fsWriter, rootDir string) error {
+	bundlesByPackage := map[string][]Bundle{}
+	for _, b := range cfg.Bundles {
+		bundlesByPackage[b.Package] = append(bundlesByPackage[b.Package], b)
+	}
+	othersByPackage := map[string][]Meta{}
+	for _, o := range cfg.Others {
+		pkgName := o.Package
+		if pkgName == "" {
+			pkgName = globalName
```
ecordell, line 57: I can see this being useful, but I'm also wondering if we just want to split all blobs into groups: entries that pass the validation for their schema and entries that don't. I imagine an error message saying "package X isn't valid: missing name" is more useful than seeing the `__global` folder with invalid data in it?

pkg/property/property.go
```diff
@@ -0,0 +1,315 @@
...
+	TypePackageRequired = "olm.package.required"
+	TypeChannel         = "olm.channel"
+	TypeGVK             = "olm.gvk"
+	TypeGVKRequired     = "olm.gvk.required"
+	TypeSkips           = "olm.skips"
+	TypeSkipRange       = "olm.skipRange"
+	TypeBundleObject    = "olm.bundle.object"
+)
+
+func Parse(in []Property) (*Properties, error) {
```
ecordell, line 151: This may be a timing issue again, but I would expect this to use the pluggable property validation that vu added in `api`

ecordell (commented):
If we want to go ahead and ship this as I think we would be fine, but I would prefer to keep everything more or less as a private (go) api for now.

I left some more detailed comments in line, but in general I think it would make sense to have:

1. A public api for declarative config, that handles the Metaschema and the pluggable validation from `api` (i.e. given a list of json, is each blob valid)
1.  A private "consumer" of the declarative config, that has a specific set of expected schemas / properties and knows how to serve the GRPC api from it that we use for backwards-compatibility.

Right now I think this mixes some of those two concerns in places. Which I think is fine as a stepping stone, but I'm worried about making that stage of the migration a public interface.

Am I making sense or being too idealistic about this?

pkg/declcfg/load.go
```diff
@@ -0,0 +1,132 @@
...
+		var in Meta
+		if err := json.Unmarshal(*doc, &in); err != nil {
+			// Ignore JSON blobs if they are not parsable as meta objects.
+			continue
+		}
+
+		switch in.Schema {
+		case schemaPackage:
+			var p Package
+			if err := json.Unmarshal(*doc, &p); err != nil {
```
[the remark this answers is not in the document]
joelanford: At some point we have to actually parse the package and bundle blobs to build the model used to serve the GRPC API. Are you suggesting that we should basically have:
```
type DeclarativeConfig struct {
    Blobs []Meta
}
```

And then only parse further when we're converting to the model?

pkg/declcfg/properties.go
```diff
@@ -0,0 +1,32 @@
...
+func parseProperties(props []property.Property) (*property.Properties, error) {
+	out, err := property.Parse(props)
+	if err != nil {
+		return nil, err
+	}
+
+	channels := map[string]struct{}{}
+	for _, ch := range out.Channels {
+		if _, ok := channels[ch.Name]; ok {
+			return nil, propertyDuplicateError{typ: property.TypeChannel, key: ch.Name}
```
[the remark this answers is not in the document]
joelanford: I agree this likely doesn't belong in `pkg/declcfg`, 

However, I think this is something we would need to check when converting to the model because the GRPC API has a built-in assumption that a bundle can only replace one other bundle in a given channel, and the model currently encodes that assumption in its `Bundle` type, which has fields for `Channel` and `Replaces`.

pkg/declcfg/declcfg.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/operator-framework/operator-registry/pkg/property"
+)
+
+const (
+	schemaPackage = "olm.package"
+	schemaBundle  = "olm.bundle"
+)
+
+type DeclarativeConfig struct {
```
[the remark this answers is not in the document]
njhale: I think having an internal package for this is a good idea so long as we can guarantee we won't need to import this anywhere else in operator-framework; i.e. we always use `opm` to interpret declarative config.

pkg/declcfg/write.go
joelanford: This piece of code is not concerned with validation at all. It is just splitting blobs into package-specific vs non-package specific so that it can write blobs out in places convenient for humans to later find them.

I do think it could be interesting to add more logic to this to further split out blobs into valid/invalid, but I think that's orthogonal. For example, you might want `__global-invalid.json` and `<package>/<package>-invalid.json`

pkg/declcfg/properties.go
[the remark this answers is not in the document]
joelanford: I just checked. This code only runs when converting to the model, not during `declcfg.LoadDir`. The only validations that occur at parse time are:
1. If it isn't JSON, ignore it.
1. If it doesn't parse as a `Meta`, ignore it.
1. If it has an empty "schema" field, ignore it.
1. If "schema" == "olm.package", does it the object unmarshal into a `Package` object successfully?
1. If "schema" == "olm.bundle", does it the object unmarshal into a `Bundle` object successfully?
1. If "schema" is anything else, just dump it into the `Others` bucket.

pkg/declcfg/helpers_test.go
```diff
@@ -0,0 +1,275 @@
...
+
+func newTestBundle(packageName, version string, opts ...bundleOpt) Bundle {
+	csvJson := fmt.Sprintf(`{"kind": "ClusterServiceVersion", "apiVersion": "operators.coreos.com/v1alpha1", "metadata":{"name":%q}}`, testBundleName(packageName, version))
+	b := Bundle{
+		Schema:  schemaBundle,
+		Name:    testBundleName(packageName, version),
+		Package: packageName,
+		Image:   testBundleImage(packageName, version),
+		Properties: []property.Property{
+			property.MustBuildPackage(packageName, version),
```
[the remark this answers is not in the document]
joelanford: That would result in callers needing to do:
```
property.Must(property.BuildPackage(packageName, version))
```
which is more stutter.

A generic function already exists: `MustBuild(interface{})`, which calls `Build(interface{})`.

The idea of these property-specific functions is that they should never error/panic since they use structs we've already defined (and thus we know don't error during `json.Marshal`), so with that in mind, its not necessary to expose property-specific error functions that return `*Property, error`.

I wouldn't be opposed to just dropping the `Must` prefix from the property-specific builders though.

pkg/declcfg/helpers_test.go
```diff
@@ -0,0 +1,275 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+
+	"github.com/operator-framework/operator-registry/pkg/model"
+	"github.com/operator-framework/operator-registry/pkg/property"
+)
+
+func buildValidDeclarativeConfig(includeUnrecognized bool) DeclarativeConfig {
```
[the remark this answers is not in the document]
joelanford: IIRC, I was paying special attention to code coverage when building out this function. It seems like `testing/quick` is more like fuzzing than ensuring specific cases are covered?

pkg/declcfg/declcfg.go
```diff
@@ -0,0 +1,82 @@
...
+	Packages []Package
+	Bundles  []Bundle
+	Others   []Meta
+}
+
+type Package struct {
+	Schema         string `json:"schema"`
```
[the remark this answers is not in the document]
joelanford: I considered embedding `Meta` in the other blob structs, but decided against it (so far) because:
1. It didn't seem like it added much value.
2. The Meta type has custom marshal/unmarshal functions that I didn't want interfering with these structs. There's probably a way to make this a little prettier, but it didn't seem worth it to spend time on that.

pkg/declcfg/load_test.go
joelanford: Done!

dinhxuanvu:
So I just want to put a comment here saying that we should move forward with what we already have here. Regarding the cuelang PR on the api repo, I don't think it is gonna make it even though it is ready. It is just not enough time unfortunately. The current validation model in this PR is solid and should work well. This validation model is stricter than the cuelang validation so if we relax the restriction in validation in the future releases, that should be fine.

dinhxuanvu (approved):
There seems to be a few issues with the validation based on validation testing but those are not blockers. Will address on subsequent PRs.

internal/declcfg/declcfg_to_model.go
```diff
@@ -0,0 +1,113 @@
...
+				Skips:         skipsToStrings(props.Skips),
+				Properties:    b.Properties,
+				RelatedImages: relatedImagesToModelRelatedImages(b.RelatedImages),
+				CsvJSON:       b.CsvJSON,
+				Objects:       b.Objects,
+			}
+		}
+	}
+
+	for _, mpkg := range mpkgs {
```
dinhxuanvu, line 76: Will need to revisit this part of code as this will add an empty channel entry to the channel list which will error out later because we don't allow an invalid/empty channel. This happens when `defauftChannel` is missing from package blob. The error will complain about an empty channel which is not entirely accurate when it should be missing default channel. Add this comment here for a record and will address this later.

internal/declcfg/declcfg_to_model.go
```diff
@@ -0,0 +1,113 @@
...
+		if !ok {
+			return nil, fmt.Errorf("unknown package %q for bundle %q", b.Package, b.Name)
+		}
+
+		props, err := parseProperties(b.Properties)
+		if err != nil {
+			return nil, fmt.Errorf("parse properties for bundle %q: %v", b.Name, err)
+		}
+
+		if b.Package != props.Packages[0].PackageName {
```
dinhxuanvu, line 44: It will result an `index out of range` error when there is not `package` property specified in the `properties` list. Will fix this later.

internal/declcfg/declcfg_to_model.go
```diff
@@ -0,0 +1,113 @@
...
+			return nil, fmt.Errorf("package %q does not match %q property %q", b.Package, property.TypePackage, props.Packages[0].PackageName)
+		}
+
+		for _, bundleChannel := range props.Channels {
```
dinhxuanvu, line 48: This doesn't account for the case where channel property is missing and this bundle is orphaned. Will fix later.

internal/model/model.go
```diff
@@ -0,0 +1,295 @@
...
+	RelatedImages []RelatedImage
+
+	// These fields are present so that we can continue serving
+	// the GRPC API the way packageserver expects us to in a
+	// backwards-compatible way.
+	Objects []string
+	CsvJSON string
+}
+
+func (b *Bundle) Validate() error {
```
dinhxuanvu, line 219: I think we are missing the `Image` validation here. If the bundle image is not specified in the bundle blob, then we can't serve this particular bundle given there is no bundle image to pull from. Will fixed later.

joelanford: I'm pretty sure I ran across at least one bundle one of the existing catalogs that did not have an image, so I removed the image validation.

internal/declcfg/declcfg_to_model.go
joelanford: The parseProperties function just above checks for exactly one package property. We can extract the contents of that function here if that makes this more legible. We'll have to check to see if that function is used anywhere else.

internal/model/model.go
dinhxuanvu: I think that catalog must be built with packagemanifests format. In that case, then the content of bundle AKA manifests will be stored in `[]Objects` blob so they can be installed. Without a bundle image nor the content blob, the bundle is not installable.

PR #611, opened 2021-03-25
`opm registry add`: add `--ca-file` to add a cert to the pool used by the command

estroz:
**Description of the change:**
- cmd/opm/registry/add.go: add `--ca-file`, which is mutually exclusive with `--skip-tls=true`.

**Motivation for the change:** `registry add` should be configurable with a cert file, since `registry.AddToRegistryRequest` is.

/kind feature

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

PR #618, opened 2021-03-31
feat(cmd): Add validate subcommand to opm cmd to validate config files

dinhxuanvu:
Validate declarative config (JSON) files in a given directory. The validation
is using the validation library from declcfg, model and property packages.

Allow users to validate their config files.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/root/cmd.go
+++ b/cmd/opm/root/cmd.go
@@ -25,7 +25,7 @@ func NewCmd() *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd(), validate.NewConfigValidateCmd())
+	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd(), validate.NewCmd())
 	index.AddCommand(cmd)
 	version.AddCommand(cmd)
 
--- a/cmd/opm/root/cmd.go
+++ b/cmd/opm/root/cmd.go
@@ -8,6 +8,7 @@ import (
 	"github.com/operator-framework/operator-registry/cmd/opm/index"
 	"github.com/operator-framework/operator-registry/cmd/opm/registry"
 	"github.com/operator-framework/operator-registry/cmd/opm/serve"
+	"github.com/operator-framework/operator-registry/cmd/opm/validate"
 	"github.com/operator-framework/operator-registry/cmd/opm/version"
 )
 
@@ -24,7 +25,7 @@ func NewCmd() *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd())
+	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd(), validate.NewConfigValidateCmd())
 	index.AddCommand(cmd)
 	version.AddCommand(cmd)
 
--- a/cmd/opm/validate/validate.go
+++ b/cmd/opm/validate/validate.go
@@ -1,0 +1,50 @@
+package validate
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/pkg/lib/config"
+)
+
+func NewConfigValidateCmd() *cobra.Command {
+	validate := &cobra.Command{
+		Use:   "validate <directory>",
+		Short: "Validate the declarative index config",
+		Long:  "Validate the declarative config JSON file(s) in a given directory",
+		Args: func(cmd *cobra.Command, args []string) error {
+			return cobra.ExactArgs(1)(cmd, args)
+		},
+		RunE: configValidate,
+	}
+
+	validate.Flags().BoolP("debug", "d", false, "enable debug log output")
+	return validate
+}
+
+func configValidate(cmd *cobra.Command, args []string) error {
+	debug, err := cmd.Flags().GetBool("debug")
+	if err != nil {
+		return err
+	}
+
+	logger := logrus.WithField("cmd", "validate")
+	if debug {
+		logger.Logger.SetLevel(logrus.DebugLevel)
+	}
+
+	if _, err := os.Stat(args[0]); os.IsNotExist(err) {
+		logger.Error(err.Error())
+	}
+
+	err = config.ValidateConfig(args[0])
+	if err != nil {
+		logger.Error(err.Error())
+		return fmt.Errorf("failed to validate config: %s", err)
+	}
+
+	return nil
+}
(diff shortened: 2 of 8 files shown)


pkg/lib/config/validate.go
```diff
@@ -0,0 +1,26 @@
...
+// directory: the directory where declarative config file(s) exist
+// Outputs:
+// error: ValidattionError which contains a list of errors
+func ValidateConfig(directory string) error {
+	// Load config files
+	cfg, err := declcfg.LoadDir(directory)
+	if err != nil {
+		return err
+	}
+	// Validate the config using model validation
```
anik120, line 20: I'm guessing this is not ready to be reviewed yet?

dinhxuanvu: I just rebased and that's pretty much it. The convert func will do validation in the background so that's all we need for validation at this point.

timflannagan: Just driving by as I was curious - can we reflect that we're relying on that ConvertToModel function to implicitly validate as it's not super clear with either the current comment or the current ValidateConfig function comments.

dinhxuanvu: Sure. I will comment on that to make it more transparent.

cmd/opm/validate/validate.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func NewConfigValidateCmd() *cobra.Command {
+	validate := &cobra.Command{
+		Use:   "validate <directory>",
+		Short: "Validate the declarative index config",
+		Long:  "Validate the declarative config JSON file(s) in a given directory",
+		Args: func(cmd *cobra.Command, args []string) error {
+			return cobra.ExactArgs(1)(cmd, args)
+		},
```
timflannagan, line 20: Do we need an inline function for this? It seems unneeded and we can just use `Args: cobra.ExactArgs(1)` or `Args: require(cobra.ExactArgs(1)`.

cmd/opm/validate/validate.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	logger := logrus.WithField("cmd", "validate")
+	if debug {
+		logger.Logger.SetLevel(logrus.DebugLevel)
+	}
+
+	if _, err := os.Stat(args[0]); os.IsNotExist(err) {
+		logger.Error(err.Error())
```
timflannagan, line 40: I'd need to double-check but if we run into an error while stat-ing that declarative config directory, this will just pass through to that ValidateConfig function, no? Do we need to make this a `logger.Fatal(...)` or do we need to also return an error after logging that same error?

timflannagan: Also, I think we should use a variable that holds the value of args[0] instead of passing args[0] to both the os.Stat and config.ValidateConfig functions.

dinhxuanvu: The load function should catch this but I think we should just return here given there is no point to continue when the directory doesn't exist.

timflannagan: Another minor nit, but if we're just going to return an error, let's do the upfront checks first, like getting the `debug` flag or stat-ing the declarative config directory first, before setting up any loggers.

cmd/opm/validate/validate.go
```diff
@@ -0,0 +1,50 @@
...
+
+	validate.Flags().BoolP("debug", "d", false, "enable debug log output")
+	return validate
+}
+
+func configValidate(cmd *cobra.Command, args []string) error {
+	debug, err := cmd.Flags().GetBool("debug")
+	if err != nil {
+		return err
+	}
```
timflannagan, line 32: I realize we're doing almost no logging for this run command, so if we don't run into a stat error, then we're essentially running `opm validate <directory>` and no output is being returned.

pkg/lib/config/validate.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+// ValidateConfig takes a directory containing the declarative config file(s)
+// 1. Validate if declarative config file(s) are valid based on specified schema
+// 2. Validate the `replaces` chains of the upgrade graph
+// Inputs:
+// directory: the directory where declarative config file(s) exist
+// Outputs:
+// error: a wrapped error that contains a list of error strings
+func ValidateConfig(directory string) error {
```
timflannagan, line 14: Do we need to pass a logger here as well? This reads like it's the entry point to the validation work from the command driver.

dinhxuanvu: The reason, I don't add a logger here because the validation funcs being used after this don't have any logger input. We can add logger here for feature use but then we can do that in the future.

cmd/opm/validate/validate.go
dinhxuanvu: Correct. There is no logger hookup on any of validation funcs that we are using here. Frankly, there isn't much to log because there is one-time atomic run kind of a command. There ain't a lot of operations to log. Most of the time, something happens meaning there are errors and should return back.

pkg/lib/config/validate.go
timflannagan: Gotcha, that makes sense, and just wanted to consider the UX of no logging for this subcommand - I didn't dive too deep past this new command and validate package so thanks for the additional context.

timflannagan:
LGTM but holding for someone that's more up-to-date on this effort.

/lgtm
/hold

dinhxuanvu:
Folks, please feel free to lgtm and approve but keep the hold in for now. We try to line up the 3 config PRs for merging today and this one should go last to avoid unnecessary rebase. Thanks.

cmd/opm/root/cmd.go
```diff
@@ -24,7 +25,7 @@ func NewCmd() *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd())
+	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd(), validate.NewConfigValidateCmd())
```
anik120, line 28:
```suggestion
	cmd.AddCommand(registry.NewOpmRegistryCmd(), alpha.NewCmd(), serve.NewCmd(), newAddCmd(), validate.NewCmd())
```
Non blocking nit suggestion

cmd/opm/validate/validate.go
```diff
@@ -0,0 +1,48 @@
...
+
+	directory := args[0]
```
anik120, line 31: We should probably have some check above in `NewValidateCmd()` (eg pass `Args:  cobra.ExactArgs(1)` to &cobra.Command{}) to make sure we're not panicking here.

timflannagan: We had `Args: cobra.ExactArgs(1)` before - not sure if that was removed in the last force push? But yeah, 100% agree we need to add that, or check the length of the arguments list if we're blindingly indexing the first element like this.

timflannagan:
Hmm, it looks like it's possible to override but you may need higher permissions: https://github.com/operator-framework/operator-registry/pull/611#issuecomment-809692942?

dinhxuanvu:
The PR passed all tests so manually merge this to bypass the codecov.

PR #621, opened 2021-04-06
Accept any property type in olm.properties annotation

njhale:

Allow bundle authors to provide properties of any type via their CSV's `olm.properties` annotation.

Delivers [OLM-2133](https://issues.redhat.com/browse/OLM-2133).

This is **stopgap** for bundle authors and will be superseded by [OLM-2130](https://issues.redhat.com/browse/OLM-2130).

As of today, only label-type bundle properties are accepted in the CSV properties annotation when added to an index. Accepting arbitrary types during index addition will allow bundles that specify OpenShift compatibility to be included in an index.


pkg/sqlite/load.go
```diff
@@ -893,6 +893,7 @@ func (s *sqlLoader) addPackageProperty(tx *sql.Tx, bundleName, pkg, version, bun
 }
 
 func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) error {
+	// FIXME: bundle.cache() seems like the proper place for this to be generated. Let's investigate what that looks like at some point.
```
njhale, line 896: Is there a reason this approach wasn't taken the first time around? It seems much nicer to encapsulate this logic for all storage implementations than to rewrite it for each.

pkg/sqlite/load.go
```diff
@@ -921,36 +928,61 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+		case registry.PackageType:
+			if err := json.Unmarshal(prop.Value, &registry.PackageProperty{}); err != nil {
+				return err
 			}
+		case registry.GVKType:
+			if err := json.Unmarshal(prop.Value, &registry.GVKProperty{}); err != nil {
+				return err
+			}
+		case registry.DeprecatedType:
+			// deprecated has no value
```
njhale, line 977: 🙃

pkg/sqlite/load.go
```diff
@@ -921,36 +928,61 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+	}
+
+	var props []registry.Property
+	if csv.GetAnnotations() != nil {
+		v, ok := csv.GetAnnotations()[registry.PropertyKey]
+		if ok {
+			if err := json.Unmarshal([]byte(v), &props); err != nil {
+				return err
+			}
 		}
```
timflannagan, line 953: Not blocking, but I think I'd like to see this reversed and instead check if `!ok`:

```go
if !ok {
    continue
}

if err := json.Unmarshal(...);
...
```

timflannagan:
Also, I'm seeing the following failure in the unit test logs:

```
time="2021-04-09T21:34:25Z" level=debug msg="could not decode file contents for ../../manifests/strimzi-kafka-operator/strimzi-kafka-operator.package.yaml" error="error unmarshaling JSON: while decoding JSON: Object 'Kind' is missing in '{\"channels\":[{\"currentCSV\":\"strimzi-cluster-operator.v0.11.1\",\"name\":\"stable\"},{\"currentCSV\":\"strimzi-cluster-operator.v0.12.1\",\"name\":\"beta\"},{\"currentCSV\":\"strimzi-cluster-operator.v0.12.2\",\"name\":\"alpha\"}],\"defaultChannel\":\"stable\",\"packageName\":\"strimzi-kafka-operator\"}'"
```

pkg/sqlite/load.go
njhale: @timflannagan looks like this isn't in a `for` body, so `continue` won't do anything.

timflannagan: Hmm, I was looking at a bad diff - I did a hard reload in my browser, and I see a different diff now 😵‍💫

timflannagan:
We'll either need to address the codecov difference or force-merge, but having that bot attempt to constantly respin tests isn't going to make a difference. Feel free to remove the hold if we decide to avoid force merging and address the -0.06% difference.

/hold

pkg/sqlite/load.go
```diff
@@ -921,36 +928,61 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
-		if err := s.addProperty(tx, registry.GVKType, string(value), bundle.Name, bundleVersion, bundle.BundleImage); err != nil {
-			return err
+		properties[propstring{Type: registry.GVKType, Value: string(value)}] = struct{}{}
+	}
+
+	// Add properties from annotations
+	csv, err := bundle.ClusterServiceVersion()
+	if err != nil {
+		// FIXME: Returning nil here is in line with the original implementation, but that was probably wrong. We should probably just bubble-up the error.
+		return nil
```
dinhxuanvu, line 938: As the FIXME indicates, we should return the error to avoid a hidden error that potentially drives people crazy during debugging.

pkg/sqlite/load.go
```diff
@@ -921,36 +928,61 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+	}
+
+	if csv == nil {
+		// FIXME: Currently, a CSV is requirement of bundle addition. Should this return an error?
+		return nil
```
dinhxuanvu, line 943: I agreed with the FIXME, CSV is required so if it is missing, an error is appropriate.

dinhxuanvu (commented):
lgtm to me. Just a few comments. Nothing major.

pkg/sqlite/load.go
njhale: IIRC, the reason it wasn't changed is b/c we're not sure if folks are depending on this _not failing_.

PR #623, opened 2021-04-06
.github: Separate the jobs in the test workflow into their own workflows

timflannagan:
Update the test.yaml workflow and migrate some of the jobs to their own
workflow configurations. This is a workaround to GH actions not
supporting re-running individual job actions when they fail.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,16 @@
+name: build
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    paths:
+      - '**'
+      - '!doc/**'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,23 @@
+name: lint
```
joelanford, line 1: nit: align name with job name `lint` vs. `sanity`

Only because the other workflows have aligned names.

timflannagan: Done updated the filename as well.

timflannagan:
Adding the lgtm label as it looks like tide didn't add it after Joe's comment earlier :shrug:

timflannagan:
Looks like some jobs are stuck, or the controller isn't posting status updates.

/override ci/prow/images

PR #631, opened 2021-04-12
Add ppc64le binary to the published OPM binaries

aemperador:
Add PPC64LE to published OPM binaries

We want to use OPM as a pre-built binary for PPC64LE enabled operators to remove the complexity of building the binary from our build processes.

**Reviewer Checklist**


aemperador:
@estroz @timflannagan unsure how to re-run tests, but it looks like the podman install failed

timflannagan:
You may need to either rebase this PR to pick up any changes made to the CI environment or open/close the PR to recycle the checks.

aemperador:
@ankitathomas could you please tell me how to proceed with this PR? it appears to be stuck waiting for a build, but I cannot trigger it myself

aemperador:
will re-open

aemperador:
> You may need to either rebase this PR to pick up any changes made to the CI environment or open/close the PR to recycle the checks.

@timflannagan I pulled in master earlier today. do I need to do anything else?

timflannagan:
I haven't poked around the PR contents (my brain is fried for the day) but it looks like there are a couple of merge commits here, and we typically don't utilize merge commits. By default, our automatic merging tool squashes PR commit(s) into a single commit, so the merge commits amended to that squashed commit might be less clean than desired is the only thing I can think of outside of the DCO check, which appears to be fine so we're good there.

aemperador:
@timflannagan please let me know who can review. we need this to unblock @mayurwaghmode

timflannagan:
Apologies - what's the use case here? You want to be able to build opm during CI environments? I was under the impression that since this repository had integrated goreleaser, we were producing multi-arch (including ppc) binaries and those binaries are attached to specific minor version releases.

timflannagan:
Hmm, interesting it looks like we _don't_ produce individual multi-arch binaries as we don't use goreleaser during the release workflow, which is triggered when new tags are created. We do however produce multi-arch opm container images, but I don't think that's as useful here. @joelanford Any idea on why we don't use goreleaser to also build and attach multi-arch binaries to the releases?

joelanford:
We didn't switch over to GoReleaser for the GitHub Actions release builds because we currently use a macOS github actions build host.

In my original GoReleaser PR, it actually _did_ use GoReleaser for the release binaries, but we had to revert that piece of the PR because it depended on an unlicensed macOS cross-compiler running on a Linux host (Apple typically doesn't permit using XCode and related libraries to cross-compile from non-mac to mac).

We could consider using GoReleaser for all Linux and Windows builds and stick with the separate macOS runner for mac builds. We could have the GoReleaser job publish a draft release, and then have the mac job upload the additional binary to that release.

In the meantime, the workaround would be to use (or extract the binary from) the `quay.io/operator-framework/opm` container image, which has a ppc64le variant.

aemperador:
@mayurwaghmode please see above. will that solve your dependency issue?

@joelanford when do you think the change could be made to GoReleaser to also publish the ppc64le binary?

joelanford:
@aemperador this is not a priority for any of the core maintainers at this time, so there's no estimate on when these changes will happen.

If this is a high priority for you and your team, we would accept (and be very appreciative of) contributions and could help guide the implementation and provide review.

The original PR is here https://github.com/operator-framework/operator-registry/pull/710. The first commit has goreleaser publishing the binaries. The second commit is where we reverted that due to Apple's licensing restrictions. We've since made further refinements, but it should help anyone wanting to work on this.

aemperador:
@joelanford thank you for your help. I will defer to @mayurwaghmode to see if the quay image satisfies his requirement, if not I'll prioritize GoReleaseer.

aemperador:
will update GoReleaser instead. Closing.

PR #637, opened 2021-04-15
go.*,vendor: Bump the k8s dependencies to v1.20.6

timflannagan:
The v1.20.6 contains the github.com/gogo/protobuf version bump:

https://github.com/kubernetes/kubernetes/blob/v1.20.6/go.mod#L53.

Fixes CVE-2021-3121.

**Reviewer Checklist**


timflannagan:
Re-adding the lgtm - had to deal with a merge conflict with a recently merged PR to master.

timflannagan:
It looks like I don't have the requisite permissions to really inspect that codecov/project check, like re-spin it or change the commit it's checking against. It seems like that `pkg/registry/query.go` keeps coming up in the report however, regardless of changes made to that file, the past couple of pull requests to that may warrant further investigation. We may need to bite the bullet and just bump the code coverage for the entire project, as it looks like 50% is currently the default threshold.

PR #641, opened 2021-04-20
Support YAML formatting for declarative config

joelanford:
This PR updates declarative config to support YAML formatting. It:
- Updates readers to handle either YAML or JSON
- Adds a (currently unused) WriteYAML function, which will be used by follow-on PRs.

YAML is more forgiving (e.g. no commas to get right or wrong), and the YAML spec (and YAML tooling) supports multi-document files.

**Reviewer Checklist**


internal/declcfg/write.go
```diff
@@ -131,10 +133,8 @@ func writeFile(cfg DeclarativeConfig, w fsWriter, filename string) error {
 	return nil
 }
 
-func writeJSON(cfg DeclarativeConfig, w io.Writer) error {
-	enc := json.NewEncoder(w)
-	enc.SetIndent("", "    ")
-	enc.SetEscapeHTML(false)
+func writeYAML(cfg DeclarativeConfig, w io.Writer) error {
```
anik120, line 136: Should there be a flag to allow for writing either in JSON or YAML?

anik120 (dismissed):

dinhxuanvu:
I can see the reason to allow YAML or JSON as inputs but honestly I feel JSON is more readable than YAML especially for long and nested structure for k8s objects.

joelanford:
> I can see the reason to allow YAML or JSON as inputs but honestly I feel JSON is more readable than YAML especially for long and nested structure for k8s objects.

@dinhxuanvu I see your point, but the problem is more with editing. I have not found any IDE or tooling that knows how to handle JSON streams, which are effectively what our JSON file format is. Everything gets super confused and spits out warnings for JSON beyond the first object in the file.

YAML handles this seamlessly in its spec by defining `---\n` as the separator.

internal/declcfg/write.go
joelanford: Possibly, but based on some discussion around all the edge cases (how to handle referenced relative files safely) related to writing, I'm no longer convinced we should write declarative configs ever (except for migrating a DB to declarative config, which does not have the same edge case concerns)

anik120: Referenced relative files are supposed to be stop gap measure for backward compatibly serving CSV jsons to packageserver. I don't think we should upend the goal of writing configs for our users because of a temporary inconvenience for us.

joelanford: > Referenced relative files are supposed to be stop gap measure for backward compatibly serving CSV jsons to packageserver.

This is just one particular use case of file references that we have for ourselves. This will always be a problem for us regardless of that use case unless we explicitly forbid any form of file referencing in our declarative config spec.

njhale:
> > I can see the reason to allow YAML or JSON as inputs but honestly I feel JSON is more readable than YAML especially for long and nested structure for k8s objects.
> 
> @dinhxuanvu I see your point, but the problem is more with editing. I have not found any IDE or tooling that knows how to handle JSON streams, which are effectively what our JSON file format is. Everything gets super confused and spits out warnings for JSON beyond the first object in the file.
> 
> YAML handles this seamlessly in its spec by defining `---\n` as the separator.

@joelanford `jq` has the `--slurp` option, which seems to work the way we want: https://jqplay.org/s/i_IexjC6hy

or am I missing something?

joelanford:
Yeah good point. I should have called `jq` out as an exception. I was mainly talking about IDEs and editors. I'm sure plenty of CLI tools handle JSON streams.

internal/declcfg/load.go
```diff
@@ -71,32 +71,32 @@ func extractCSV(objs []string) string {
...
 }
 
-func readJSON(r io.Reader) (*DeclarativeConfig, error) {
+func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	cfg := &DeclarativeConfig{}
-	dec := json.NewDecoder(r)
-	for dec.More() {
-		doc := &json.RawMessage{}
-		if err := dec.Decode(doc); err != nil {
+	dec := yaml.NewYAMLOrJSONDecoder(r, 16)
```
timflannagan, line 76: Any specific reason we're using a buffer size of 16 here, instead of something like 100?

joelanford: I'll set this to 4096. That looks like a typical value used in `kubernetes/kubernetes`.

```suggestion
	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
```

internal/declcfg/testdata/invalid/not-json.txt
```diff
@@ -1 +0,0 @@
-This is not json.not-json.txt
```
njhale, line 1: is this file now empty? did we mean to remove it entirely?

njhale (commented):
If we're saying we're going to support both, I think we need some test of equivalent JSON/YAML that verifies format equality of the parser. This sounds like another good opportunity for a generative/property based test.

internal/declcfg/helpers_test.go
```diff
@@ -40,14 +40,14 @@ func buildValidDeclarativeConfig(includeUnrecognized bool) DeclarativeConfig {
...
 				"package": "anakin",
-				"myField": "foobar"
+				"schema": "custom.3"
 			}`)},
 			{Schema: "custom.3", Package: "boba-fett", Blob: json.RawMessage(`{
-				"schema": "custom.3",
+				"myField": "foobar",
 				"package": "boba-fett",
-				"myField": "foobar"
+				"schema": "custom.3"
```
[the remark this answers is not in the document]
joelanford: No, this order changed because the yaml encoder does not honor struct field order. It looks like it may just be alphanumeric order.

internal/declcfg/load.go
```diff
@@ -71,32 +71,32 @@ func extractCSV(objs []string) string {
...
 }
 
-func readJSON(r io.Reader) (*DeclarativeConfig, error) {
+func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	cfg := &DeclarativeConfig{}
-	dec := json.NewDecoder(r)
-	for dec.More() {
-		doc := &json.RawMessage{}
-		if err := dec.Decode(doc); err != nil {
+	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
```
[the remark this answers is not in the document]
joelanford: I'm not sure I'm following. Are you referencing the `4096` buffer size? If so, that is just the number of bytes to read to determine whether the reader is JSON or YAML. As far as I can tell, it doesn't impact the ability of the decoder to find each item in the stream.

internal/declcfg/testdata/invalid/not-json.txt
joelanford: I thought I renamed it. I'll double check this.

joelanford: Yeah, this file is removed (actually renamed to `not-yaml-or-json.txt`)

internal/declcfg/helpers_test.go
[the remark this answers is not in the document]
njhale: Well that seems like a bug in the encoder, no?

internal/declcfg/load.go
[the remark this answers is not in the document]
njhale: My bad -- I misinterpreted the go docs as "number of bytes to read to determine if it's a stream". A closer read confirms your understanding.

internal/declcfg/helpers_test.go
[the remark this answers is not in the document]
joelanford: YAML _is_ field order agnostic. This change is just necessary because the roundtrip loses the original ordering. There's no bug here unless we want declarative config blobs to roundtrip map field orders. 

I think this happens because the underlying library is sigs.k8s.io/yaml, which encodes `value -> JSON -> YAML` and decodes `YAML -> JSON -> value`.

I think this is due to the JSON <-> YAML step, which uses what amounts to a `map[string]interface{}` for JSON/YAML maps, and Golang maps are not ordered.

internal/declcfg/write.go
joelanford: @anik120 I reverted `WriteDir` to encode files as JSON, so this PR should now be a no-op for the `alpha add` subcommand. I'm thinking we can merge this to unblock the `init` and `unpack` commands supporting both JSON and YAML output, and we can revisit `add` and the issues with `declcfg.WriteDir` in a follow-up.

internal/declcfg/write.go
```diff
@@ -131,28 +135,104 @@ func writeFile(cfg DeclarativeConfig, w fsWriter, filename string) error {
...
+}
 
+func newYAMLEncoder(w io.Writer) *yamlEncoder {
+	return &yamlEncoder{w}
+}
+
+func (e *yamlEncoder) Encode(v interface{}) error {
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	enc.SetEscapeHTML(false)
```
anik120, line 161: Should this be allowed as a method of `yamlEncoder`? To maintain parity with the json encoder like so https://github.com/operator-framework/operator-registry/pull/641/files#diff-5e384dd70c8fb37678149443bbdc0d411af527a1839dc91a1a40cdac04fca4d5R140-R141

anik120 (commented):
Looking good @joelanford 
Just had one question.

internal/declcfg/write.go
joelanford: That's not unreasonable. It's not _really_ necessary since we just hardcode it right now, but for the sake of symmetry, I like the suggestion.

internal/declcfg/load.go
```diff
@@ -71,32 +71,32 @@ func extractCSV(objs []string) string {
...
 
-func readJSON(r io.Reader) (*DeclarativeConfig, error) {
+func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	cfg := &DeclarativeConfig{}
-	dec := json.NewDecoder(r)
-	for dec.More() {
-		doc := &json.RawMessage{}
-		if err := dec.Decode(doc); err != nil {
-			return cfg, nil
+	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
```
dinhxuanvu, line 76: One quick comment here as a note that this `NewYAMLOrJSONDecoder` doesn't validate the JSON blob if it is a string in the value of a key-value. Given most of the value is `json.RawMessage`, this shouldn't be an issue. Plus, there is a test to check for this in the codebase. Consider it handled.

dinhxuanvu (approved):
/lgtm

PR #642, opened 2021-04-20
Bug 1951387: (bug) add after deprecation

gallettilance:

Could not add after deprecation because the graph is rebuilt on every add based on the information stored in the operatorbundle table. The table contains replaces refs to bundles that have been truncated causing add to fail.

Solution: empty the replaces field in the operatorbundle table for deprecated bundles

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/populator_test.go
+++ b/pkg/registry/populator_test.go
@@ -911,6 +911,140 @@ func TestDeprecateBundle(t *testing.T) {
 	}
 }
 
+func TestAddAfterDeprecate(t *testing.T) {
+	type args struct {
+		firstBundles      []string
+		deprecatedBundles []string
+		secondBundles     []string
+	}
+	type pkgChannel map[string][]string
+	type expected struct {
+		err                  error
+		remainingBundles     []string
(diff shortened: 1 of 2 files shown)


dinhxuanvu:
I don't think this change is enough as we made some changes to only use `channel_entry` table to the source of truth to form the upgrade graph. This PR only updates the `operatorbundle` table and that may not be enough. Bring @benluddy in for further review.

gallettilance:
the channel_entry table is taken care of through the deletion of the replacements bundles which cascades

dinhxuanvu (approved):
/lgtm

dinhxuanvu:
> I don't think this change is enough as we made some changes to only use `channel_entry` table to the source of truth to form the upgrade graph. This PR only updates the `operatorbundle` table and that may not be enough. Bring @benluddy in for further review.

We had a discussion further on Slack and it was determined that the `replaces` info between `channel_entry` and `operatorbundle` are kept in sync via FK constraint and additional query. So this is no longer a concern.

PR #644, opened 2021-04-22
Test for bundlepaths before deprecate

ankitathomas:
Allow deprecatetruncate to handle multiple bundles on the same upgrade chain. Only the most latest bundle in the chain will be marked as deprecated, the remaining will be truncated with the rest of the upgrade chain older than the deprecated bundle.

### DIFF
--- a/pkg/sqlite/deprecate_test.go
+++ b/pkg/sqlite/deprecate_test.go
@@ -1,0 +1,329 @@
+package sqlite
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/blang/semver/v4"
+	"github.com/operator-framework/api/pkg/lib/version"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-registry/pkg/api"
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/registry"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+type testBundle struct {
+	version        string
+	pkg            string
+	channels       []string
+	defaultChannel string
+	skips          []string
+	replaces       string
+}
+
+func newRegistryBundle(t *testing.T, b testBundle) *registry.Bundle {
+	name := b.version
+	csvYAML, err := yaml.Marshal(&v1alpha1.ClusterServiceVersion{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       v1alpha1.ClusterServiceVersionKind,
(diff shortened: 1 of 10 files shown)


pkg/sqlite/deprecate_test.go
```diff
@@ -0,0 +1,330 @@
...
+			querier := NewSQLLiteQuerierFromDb(db)
+
+			bundleLoader := registry.BundleGraphLoader{}
+			for _, b := range tt.initial {
+				graph, err := graphLoader.Generate(b.pkg)
+				if err != nil {
+					require.EqualError(t, err, registry.ErrPackageNotInDatabase.Error(), "Expected empty error or package not in database")
+				}
+				bndl := newRegistryBundle(t, b)
+				graph, err = bundleLoader.AddBundleToGraph(bndl, graph, &registry.AnnotationsFile{Annotations: *bndl.Annotations}, false)
```
hasbro17, line 317: Missing error check?
```suggestion
				graph, err = bundleLoader.AddBundleToGraph(bndl, graph, &registry.AnnotationsFile{Annotations: *bndl.Annotations}, false)
				require.NoError(t, err)
```

hasbro17:
SGTM overall.

pkg/sqlite/deprecate_test.go
```diff
@@ -0,0 +1,330 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/blang/semver/v4"
+	"github.com/operator-framework/api/pkg/lib/version"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-registry/pkg/api"
+	image2 "github.com/operator-framework/operator-registry/pkg/image"
```
timflannagan, line 17: Minor nit, but any reason why we need to rename this import declaration and can't just use the regular path?

pkg/sqlite/deprecate_test.go
```diff
@@ -0,0 +1,330 @@
...
+				graph, err = bundleLoader.AddBundleToGraph(bndl, graph, &registry.AnnotationsFile{Annotations: *bndl.Annotations}, false)
+				require.NoError(t, store.AddBundleSemver(graph, bndl))
+			}
+			err = NewSQLDeprecatorForBundles(store, querier, tt.deprecate).Deprecate()
+			if tt.expectedError != nil {
+				require.EqualError(t, err, tt.expectedError.Error())
+				return
+			}
+			require.NoError(t, err)
+			result, err := querier.ListBundles(context.TODO())
```
timflannagan, line 326: Do we need to also check that this error is not nil?

pkg/sqlite/deprecate_test.go
timflannagan: Yeah, agreed we should catch this error as well as the store.AddBundleSemver.

pkg/sqlite/deprecate_test.go
ankitathomas: That's fair, I can take out that check

hasbro17 (commented):
/lgtm

PR #647, opened 2021-04-30
Improve load func error handling to expose more information

dinhxuanvu:
Many of sqlite query returns errors due to constrain issues. Those
errors are often vague and not helpful for debugging reason.

Those sqlite errors should be wrapped around with other information
to expose the root cause and provide further data for debugging.

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -117,7 +117,7 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 	}
 
 	if _, err := addBundle.Exec(csvName, csvBytes, bundleBytes, bundleImage, version, skiprange, replaces, strings.Join(skips, ","), substitutesFor); err != nil {
-		return err
+		return fmt.Errorf("failed to add bundle %q: %s", csvName, err.Error())
 	}
 
 	imgs, err := bundle.Images()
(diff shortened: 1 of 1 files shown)


pkg/sqlite/load.go
```diff
@@ -117,7 +117,7 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 	}
 
 	if _, err := addBundle.Exec(csvName, csvBytes, bundleBytes, bundleImage, version, skiprange, replaces, strings.Join(skips, ","), substitutesFor); err != nil {
-		return err
+		return fmt.Errorf("Failed to add bundle %q: %s", csvName, err.Error())
```
timflannagan, line 120: Go recommends not capitalizing error message strings: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

timflannagan:
This looks reasonable to me besides the error capitalization issue I linked to and the codecov lines outlined by the bot that don't have any coverage.

pkg/sqlite/load.go
```diff
@@ -575,20 +575,20 @@ func (s *sqlLoader) addPackageChannels(tx *sql.Tx, manifest registry.PackageMani
 	var errs []error
 
 	if _, err := addPackage.Exec(manifest.PackageName); err != nil {
-		errs = append(errs, err)
+		errs = append(errs, fmt.Errorf("failed to add package %q: %s", manifest.PackageName, err.Error()))
 		return utilerrors.NewAggregate(errs)
```
timflannagan, line 579: Minor nit as you're only changing the error message and the implementation in this PR, but this seems strange to define a list of errors, run the Exec method and if the error isn't nil, append to an empty list and return that list. Do we want to instead just return an error here using the formatted error message you currently have and define the `errs` variable further down?

PR #649, opened 2021-04-30
scripts: Update the GOFLAGS when installing the mkcert executable

timflannagan:
Update the scripts/start_registry.sh bash script and update the GOFLAGS
argument to "-mod=mod" as we're cloning the mkcert source code to a
tmpdir that doesn't contain a vendor directory and that's problematic
when GOFLAGS="-mod=vendor" locally.


joelanford (approved):
/lgtm

PR #656, opened 2021-05-06
check presence of all bundles before deprecate

ankitathomas:
Allow deprecatetruncate to handle multiple bundles on the same upgrade chain. Only the most latest bundle in the chain will be marked as deprecated, the remaining will be truncated with the rest of the upgrade chain older than the deprecated bundle.

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -342,7 +342,21 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
 		return fmt.Errorf("unable to migrate database: %s", err)
 	}
 
-	deprecator := sqlite.NewSQLDeprecatorForBundles(dbLoader, request.Bundles)
+	// Check if all bundlepaths are valid
+	var toDeprecate []string
+
+	dbQuerier := sqlite.NewSQLLiteQuerierFromDb(db)
+
+	toDeprecate, _, err = checkForBundlePaths(dbQuerier, request.Bundles)
+	if err != nil {
+		if !request.Permissive {
+			r.Logger.WithError(err).Error("permissive mode disabled")
+			return err
+		}
+		r.Logger.WithError(err).Warn("permissive mode enabled")
+	}
+
+	deprecator := sqlite.NewSQLDeprecatorForBundles(dbLoader, toDeprecate)
 	if err := deprecator.Deprecate(); err != nil {
 		r.Logger.Debugf("unable to deprecate bundles from database: %s", err)
 		if !request.Permissive {
@@ -353,4 +367,38 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
 	}
 
 	return nil
+}
+
+// checkForBundlePaths verifies presence of a list of bundle paths in the registry.
+func checkForBundlePaths(querier registry.GRPCQuery, bundlePaths []string) ([]string, []string, error) {
+	if len(bundlePaths) == 0 {
+		return bundlePaths, nil, nil
+	}
+
+	registryBundles, err := querier.ListBundles(context.TODO())
+	if err != nil {
+		return bundlePaths, nil, err
+	}
+
+	if len(registryBundles) == 0 {
+		return nil, bundlePaths, nil
+	}
+
+	registryBundlePaths := map[string]struct{}{}
+	for _, b := range registryBundles {
+		registryBundlePaths[b.BundlePath] = struct{}{}
+	}
+
+	var found, missing []string
+	for _, b := range bundlePaths {
+		if _, ok := registryBundlePaths[b]; ok {
+			found = append(found, b)
+			continue
+		}
+		missing = append(missing, b)
+	}
+	if len(missing) > 0 {
+		return found, missing, fmt.Errorf("target bundlepaths for deprecation missing from registry: %v", missing)
+	}
+	return found, missing, nil
 }
--- a/pkg/lib/registry/registry_test.go
+++ b/pkg/lib/registry/registry_test.go
@@ -1,0 +1,243 @@
+package registry
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"testing"
+	"testing/fstest"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-registry/internal/model"
+	"github.com/operator-framework/operator-registry/internal/property"
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+func fakeBundlePathFromName(name string) string {
+	return fmt.Sprintf("%s-path", name)
+}
+
+func newQuerier(bundles []*model.Bundle) *registry.Querier {
+	pkgs := map[string]*model.Package{}
+	channels := map[string]map[string]*model.Channel{}
+
+	for _, b := range bundles {
+		if len(b.Image) == 0 {
+			b.Image = fakeBundlePathFromName(b.Name)
+		}
+		channelName := b.Channel.Name
+		packageName := b.Package.Name
+		if _, ok := pkgs[packageName]; !ok {
+			pkgs[packageName] = &model.Package{
+				Name: packageName,
+			}
+			channels[packageName] = map[string]*model.Channel{
+				channelName: {
+					Package: pkgs[packageName],
+					Name:    channelName,
+					Bundles: map[string]*model.Bundle{b.Name: b},
+				},
+			}
+			pkgs[packageName].Channels = channels[packageName]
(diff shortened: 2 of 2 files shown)


pkg/sqlite/deprecate.go
```diff
@@ -35,10 +38,35 @@ func (d *BundleDeprecator) Deprecate() error {
...
+	type nameVersion struct {
+		name    string
+		version string
+	}
+	bundleInfo := make(map[string]nameVersion)
+	// Check if all bundlepaths are valid
 	for _, bundlePath := range d.bundles {
-		if err := d.store.DeprecateBundle(bundlePath); err != nil {
-			if !errors.Is(err, registry.ErrBundleImageNotInDatabase) && !errors.Is(err, registry.ErrRemovingDefaultChannelDuringDeprecation) {
+		name, version, err := d.querier.GetBundleNameAndVersionForImage(context.TODO(), bundlePath)
```
estroz, line 48: Is `bundlePath` supposed to be a globally unique image tag, or can it be a directory path?

ankitathomas: It's a globally unique tag.

pkg/lib/registry/registry.go
```diff
@@ -342,9 +342,45 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+	dbQuerier := sqlite.NewSQLLiteQuerierFromDb(db)
+	registryBundles, err := dbQuerier.ListBundles(context.TODO())
+	if err != nil {
+		if !request.Permissive {
+			return fmt.Errorf("cannot verify bundle presence in registry: %v", err)
+		}
+		toDeprecate = request.Bundles
+	}
+
+	if registryBundles != nil {
```
timflannagan, line 357: Is this the right conditional check? Should we be checking the length of this array instead?

pkg/lib/registry/registry.go
```diff
@@ -342,9 +342,45 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+	// Check if all bundlepaths are valid
+	var toDeprecate []string
+
+	dbQuerier := sqlite.NewSQLLiteQuerierFromDb(db)
+	registryBundles, err := dbQuerier.ListBundles(context.TODO())
+	if err != nil {
+		if !request.Permissive {
+			return fmt.Errorf("cannot verify bundle presence in registry: %v", err)
+		}
+		toDeprecate = request.Bundles
```
timflannagan, line 354: Is there any reason we wouldn't want to log that we ran into an error listing the bundles here, even at the debug level? I'm trying to wrap my head around the current permissive workflow, and whether we'd want to ignore if we weren't able to get the list of bundles in the database.

pkg/lib/registry/registry.go
timflannagan: I almost wonder whether this logic would be served being contained in a helper function that we can instead call in this method. It would be easier to add unit tests to hit this case so we don't regress in future releases.

ankitathomas: It's a edge case we'd almost never run into, but if you try to deprecate a bundle from an empty registry, it wouldn't complain about missing bundles if we look for length.

pkg/lib/registry/registry.go
```diff
@@ -342,9 +342,45 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+		for _, bundlePath := range request.Bundles {
+			if _, ok := validPath[bundlePath]; !ok {
+				missingPaths = append(missingPaths, bundlePath)
+				continue
+			}
+			toDeprecate = append(toDeprecate, bundlePath)
+		}
+
+		if len(missingPaths) > 0 {
+			if !request.Permissive {
```
njhale, line 374: IIRC, permissive should decide to return an error at the top-level command, not inform whether an error is returned. I think the proper thing to do in this case is to aggregate all errors and return them at the end of the method.

pkg/lib/registry/registry.go
```diff
@@ -354,3 +368,39 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+			if found == nil {
+				found = make([]string, 0)
+			}
+			found = append(found, b)
+			continue
+		}
+		if missing == nil {
+			missing = make([]string, 0)
+		}
+		missing = append(missing, b)
```
timflannagan, line 403: Am I missing something here? It looks like missing was never declared before referencing it here.

pkg/lib/registry/registry.go
```diff
@@ -354,3 +368,39 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+	}
+	return
```
timflannagan, line 405: nothing blocking, but this return is redundent.

pkg/lib/registry/registry.go
ankitathomas: missing was declared in the function signature, it looked easier to understand CheckForBundlePaths that way

pkg/lib/registry/registry.go
ankitathomas: it's autopopulated with the args from the signature but the return itself still needs to be here

pkg/lib/registry/registry.go
ankitathomas: It actually doesn't make much sense to try and remove bundles when you can't even locate them, permissive mode shouldn't even be used here. I changed to fail when those checks fail

pkg/lib/registry/registry.go
joelanford: Could we avoid using named return values in function signatures? 

IMO, unless the function is very short  (like <5 lines), it harms readability.

https://tour.golang.org/basics/7

pkg/lib/registry/registry.go
```diff
@@ -354,3 +373,40 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+		if _, ok := registryBundlePaths[b]; ok {
+			if found == nil {
+				found = make([]string, 0)
+			}
+			found = append(found, b)
+			continue
+		}
+		if missing == nil {
+			missing = make([]string, 0)
+		}
```
njhale, line 408: this isn't necessary -- `append` works with `nil` field values so long as the field is declared as a slice type.

pkg/lib/registry/registry.go
```diff
@@ -354,3 +373,40 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+	for _, b := range registryBundles {
+		registryBundlePaths[b.BundlePath] = struct{}{}
+	}
+
+	var found, missing []string
+	for _, b := range bundlePaths {
+		if _, ok := registryBundlePaths[b]; ok {
+			if found == nil {
+				found = make([]string, 0)
+			}
```
njhale, line 402: this isn't necessary -- `append` works with `nil` field values so long as the field is declared as a slice type.

pkg/lib/registry/registry.go
```diff
@@ -354,3 +373,40 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
...
+	registryBundles, err := querier.ListBundles(context.TODO())
+	if err != nil {
+		return bundlePaths, nil, err
+	}
+
+	if len(registryBundles) == 0 {
+		return nil, bundlePaths, nil
+	}
+
+	registryBundlePaths := make(map[string]struct{})
```
njhale, line 392:
```suggestion
	registryBundlePaths := map[string]struct{}{}
```

pkg/lib/registry/registry.go
```diff
@@ -354,3 +373,40 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
 
 	return nil
 }
+
+// CheckForBundlePaths verifies presence of a list of bundle paths in the registry.
+func CheckForBundlePaths(querier registry.Query, bundlePaths []string) ([]string, []string, error) {
```
njhale, line 378: any reason we want this to be exported?

pkg/lib/registry/registry_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+			require.NoError(t, store.Migrate(context.TODO()))
+
+			for _, bundle := range tt.bundles {
+				require.NoError(t, store.AddOperatorBundle(bundle))
+			}
+
+			for _, pkg := range tt.pkgs {
+				require.NoError(t, store.AddPackageChannels(pkg))
+			}
```
njhale, line 166: Is it feasible for us to mock the querier instead of building a real database?

bparees:
@ankitathomas @njhale who needs to approve this?

njhale:
> who needs to approve this?

meant to approve this earlier.

/approve

PR #657, opened 2021-05-07
declcfg: allow properties to have unset or null values

joelanford:
The olm.deprecated property is an example of a property
that has no value. Its presence conveys its meaning.

**Reviewer Checklist**

### DIFF
--- a/internal/property/property.go
+++ b/internal/property/property.go
@@ -21,12 +21,11 @@ func (p Property) Validate() error {
 	if len(p.Type) == 0 {
 		return errors.New("type must be set")
 	}
-	if len(p.Value) == 0 {
-		return errors.New("value must be set")
-	}
-	var raw json.RawMessage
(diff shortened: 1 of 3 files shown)


internal/property/property.go
```diff
@@ -202,8 +201,10 @@ func Parse(in []Property) (*Properties, error) {
 			out.BundleObjects = append(out.BundleObjects, p)
 		default:
 			var p json.RawMessage
-			if err := json.Unmarshal(prop.Value, &p); err != nil {
-				return nil, ParseError{Idx: i, Typ: prop.Type, Err: err}
+			if len(prop.Value) > 0 {
```
benluddy, line 204: Can `value` be `""`?

joelanford:
Just force pushed a fix for the coverage drop.

joelanford:
Closing because the `olm.deprecated` property actually _does_ have a value. It is an empty struct, which marshals to `{}`.

PR #659, opened 2021-05-10
Elide channels starting with deprecated bundles

njhale:

Elide channels that start with deprecated bundles.

When a bundle is deprecated, authors don't expect channels starting with that bundle to be visible to cluster admins (since they can't install anything from such channels anyway).

**Downstream bug report:** [BZ 1959957](https://bugzilla.redhat.com/show_bug.cgi?id=1959957)

**Notes:**

~~This change brings into question the usefulness of retaining deprecated bundles at all. It's probably more straightforward to treat a deprecated bundle in a similar way to bundles referenced by `skips`; i.e. use for outbound edges only and don't return it in `ListBundles`.~~

~~I'm going to implement the above for comparison -- holding while I do that.~~

~~/hold~~

I was unable to get consensus on this, so I'm opting to get this patch in now and pick up the conversation afterwards.

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -576,9 +576,16 @@ func (s *sqlLoader) addPackageChannels(tx *sql.Tx, manifest registry.PackageMani
 		return fmt.Errorf("failed to add package %q: %s", manifest.PackageName, err.Error())
 	}
 
-	var errs []error
-	hasDefault := false
+	var (
+		errs       []error
+		channels   []registry.PackageChannel
+		hasDefault bool
+	)
 	for _, c := range manifest.Channels {
+		if deprecated, err := s.deprecated(tx, c.CurrentCSVName); err != nil || deprecated {
+			// Elide channels that start with a deprecated bundle
+			continue
+		}
 		if _, err := addChannel.Exec(c.Name, manifest.PackageName, c.CurrentCSVName); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/sqlite/load.go
```diff
@@ -1427,16 +1419,20 @@ func (s *sqlLoader) DeprecateBundle(path string) error {
...
-		err = s.rmChannelEntry(tx, bundle)
-		if err != nil {
+		if err := s.rmChannelEntry(tx, bundle); err != nil {
 			return err
 		}
-		err := s.rmBundle(tx, bundle)
-		if err != nil {
+		if err := s.rmBundle(tx, bundle); err != nil {
 			return err
 		}
```
joelanford, line 1427: I don't think you're changing this behavior in this PR but...

It looks like we include `skips` in the tail bundles without verifying that they are all non-channel-heads. Isn't it possible to have:
```
package foo, default channel: alpha

channel: alpha, head: v2
channel: beta, head: v4

v4 replaces: v3
v3 replaces: v1, skips: v2
v2 replaces: v1
```

If I'm understanding correctly, when we deprecate v3, we would remove v1 (fine) and v2 (not okay because it is the head of the alpha channel, which is the default channel)

njhale: good catch.

It's probably simplest to add a queue here and perform a breadth-first search for channel heads.

PR #660, opened 2021-05-11
bump containerd to 1.5.0

joelanford:
This commit bumps the containerd/containerd dependency to 1.5.0.

I left a breadcrumb comment in the go.mod in case this introduces a regression we don't immediately detect.


joelanford:
/hold
so I can have a few folks test this on possibly affected registry proxies.

PR #662, opened 2021-05-12
Consistently use latest go1.16 in all workflows

joelanford:
Explicitly configure all workflows to use go1.16

Workflows that don't specify a version seem to default to go1.15


ecordell (approved):
/approve

PR #671, opened 2021-06-03
Explicit bundle properties

njhale:

Defines a new configuration file that can be provided by bundle authors to explicitly declare properties.

The file itself can have any name, but must _both_ exist in the top-level `metadata` directory of the bundle _and_ consist of YAML with a top-level `properties` element containing an array of type/value tuples.

e.g.

```yaml
properties:
- type: olm.label
  value: "foo"
```

All properties sourced from this file, CSV annotations, and those derived from bundle manifests are exposed on the bundle (after being deduplicated).

See the [EP](https://github.com/operator-framework/enhancements/blob/master/enhancements/properties.md) for motivation and details.

**Reviewer Checklist**

### DIFF
--- a/internal/property/property.go
+++ b/internal/property/property.go
@@ -30,6 +30,10 @@ func (p Property) Validate() error {
 	return nil
 }
 
+func (p Property) String() string {
+	return fmt.Sprintf("type: %q, value: %q", p.Type, p.Value)
+}
+
 type Package struct {
 	PackageName string `json:"packageName"`
 	Version     string `json:"version"`
--- a/pkg/registry/decode.go
+++ b/pkg/registry/decode.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/yaml"
@@ -41,4 +42,16 @@ func DecodePackageManifest(reader io.Reader) (manifest *PackageManifest, err err
 
 	manifest = obj
 	return
+}
+
+func decodeFileFS(root fs.FS, path string, into interface{}) error {
+	fileReader, err := root.Open(path)
+	if err != nil {
+		return fmt.Errorf("unable to read file %s: %s", path, err)
+	}
+	defer fileReader.Close()
+
+	decoder := yaml.NewYAMLOrJSONDecoder(fileReader, 30)
+
+	return decoder.Decode(into)
 }
--- a/pkg/registry/parse.go
+++ b/pkg/registry/parse.go
@@ -1,0 +1,252 @@
+package registry
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+)
+
+type bundleParser struct {
+	log *logrus.Entry
+}
+
+func newBundleParser(log *logrus.Entry) *bundleParser {
+	return &bundleParser{
+		log: log,
+	}
+}
+
+// Parse parses the given FS into a Bundle.
+func (b *bundleParser) Parse(root fs.FS) (*Bundle, error) {
+	if root == nil {
+		return nil, fmt.Errorf("filesystem is nil")
+	}
+
+	bundle := &Bundle{}
+	manifests, err := fs.Sub(root, "manifests")
+	if err != nil {
+		return nil, fmt.Errorf("error opening manifests directory: %s", err)
+	}
+	if err := b.addManifests(manifests, bundle); err != nil {
+		return nil, err
+	}
+
+	metadata, err := fs.Sub(root, "metadata")
+	if err != nil {
+		return nil, fmt.Errorf("error opening metadata directory: %s", err)
+	}
+	if err := b.addMetadata(metadata, bundle); err != nil {
+		return nil, err
+	}
+
+	derived, err := b.derivedProperties(bundle)
+	if err != nil {
+		return nil, fmt.Errorf("failed to derive properties: %s", err)
+	}
+
+	bundle.Properties = propertySet(append(bundle.Properties, derived...))
+
+	return bundle, nil
+}
+
+// addManifests adds the result of parsing the manifests directory to a bundle.
+func (b *bundleParser) addManifests(manifests fs.FS, bundle *Bundle) error {
+	files, err := fs.ReadDir(manifests, ".")
+	if err != nil {
+		return err
+	}
+
+	var csvFound bool
+	for _, f := range files {
+		if f.IsDir() {
+			continue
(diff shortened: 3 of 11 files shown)


exdx (commented):
Nice! Left some feedback

pkg/registry/parse.go
```diff
@@ -0,0 +1,303 @@
...
+		visited  = map[string]struct{}{}
+	)
+	for _, p := range properties {
+		if _, ok := visited[p.String()]; ok {
+			continue
+		}
+		visited[p.String()] = struct{}{}
+
+		switch p.Type {
+		case PackageType:
```
ecordell, line 254: is there a way / is it reasonable to not have to worry about this class of problem?

could we ensure properties are sorted by their source (properties.yaml -> csv) and then leave it up to the consumer / resolver to deal with multiple properties (or external validation tools to warn that it's not good to have multiple package properties?)

pkg/registry/parse.go
```diff
@@ -0,0 +1,303 @@
...
+// Parse parses the given FS into a Bundle.
+func (b *bundleParser) Parse(root fs.FS) (*Bundle, error) {
+	if root == nil {
+		return nil, fmt.Errorf("filesystem is nil")
+	}
+
+	bundle := &Bundle{}
+	manifests, err := fs.Sub(root, "manifests")
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
njhale: sgtm, I'll add it.

pkg/registry/parse.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		obj := &unstructured.Unstructured{}
+		if err = decodeFileFS(manifests, name, obj); err != nil {
+			b.log.Warnf("failed to decode: %s", err)
+			continue
+		}
+
+		// Only include the first CSV we find in the
+		if obj.GetKind() == "ClusterServiceVersion" {
```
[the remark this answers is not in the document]
njhale: Yup -- I have an idea of where to find it. Nice find!

pkg/registry/parse.go
```diff
@@ -0,0 +1,303 @@
...
+			}
+			csvFound = true
+		}
+
+		if obj.Object != nil {
+			bundle.Add(obj)
+		}
+	}
+
+	if bundle.Size() == 0 {
```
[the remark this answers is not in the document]
njhale: don't we need to actually parse the directory before we know if the bundle has manifests or not?

I may not be understanding what you mean.

Also, the CSV check in the loop above is to ensure we only add the first CSV we encounter to the bundle.

pkg/registry/parse.go
```diff
@@ -0,0 +1,303 @@
...
+	var (
+		af *AnnotationsFile
+		df *DependenciesFile
+		pf *PropertiesFile
+	)
+	for _, f := range files {
+		name := f.Name()
+		if af == nil {
+			decoded := AnnotationsFile{}
+			if err = decodeFileFS(metadata, name, &decoded); err == nil {
```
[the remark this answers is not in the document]
njhale: Will every file we look at that doesn't decode into the specific file we're looking for log an error? If so, that just seems like extra noise for users.

pkg/registry/parse.go
njhale: That sounds reasonable.

I'll drop the special case for package properties here, as well as in the expanded properties below.

pkg/registry/parse.go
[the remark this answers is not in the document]
exdx: ah, that makes sense

pkg/registry/parse.go
```diff
@@ -0,0 +1,253 @@
...
+			continue
+		}
+
+		obj := &unstructured.Unstructured{}
+		if err = decodeFileFS(manifests, name, obj); err != nil {
+			b.log.Warnf("failed to decode: %s", err)
+			continue
+		}
+
+		// Only include the first CSV we find in the
```
exdx, line 82: is this behavior documented anywhere (that we only care about the first CSV in the bundle that we find?) I'm not sure how `ReadDir` lists files, if it's alphabetical/lexicographic or random. Can users with more than 1 CSV in their bundle know how this behaves deterministically? 

I thought this would be an error condition, if anything other than exactly 1 CSV is present in the bundle.

pkg/registry/parse.go
```diff
@@ -0,0 +1,253 @@
...
+	if csv == nil {
+		return fmt.Errorf("no csv in bundle")
+	}
+
+	bundle.Name = csv.GetName()
+	if err := bundle.AllProvidedAPIsInBundle(); err != nil {
+		return fmt.Errorf("error checking provided apis in bundle %s: %s", bundle.Name, err)
+	}
+
+	return err
```
exdx, line 112: nit: should we return `nil` at the end here?

internal/property/property.go
```diff
@@ -30,6 +30,10 @@ func (p Property) Validate() error {
 	return nil
 }
 
+func (p Property) String() string {
+	return fmt.Sprintf("type: %s, value: %s", p.Type, p.Value)
```
joelanford, line 34: In case either is an empty string
```suggestion
	return fmt.Sprintf("type: %q, value: %q", p.Type, p.Value)
```

pkg/registry/parse.go
njhale: that's probably more idiomatic than what's currently here. I'll change that.

pkg/registry/parse.go
njhale: > I'm not sure how ReadDir lists files, if it's alphabetical/lexicographic or random.

Very good points!

I don't think we can assume anything about the order files are returned by implementations of `fs.FS`.

I know that 1 CSV per bundle is an invariant for bundles, although I'm unsure where -- if at all -- this is documented.

I think it behooves us to have some deterministic behavior, but I don't want to stoop to some arbitrary heuristic to do so.

I'll dig up what the behavior is for the old parse logic and see if we want to replicate that (it might return an error)

estroz: IIRC the first CSV found by `filepath.Walk()` was used, and finding another CSV afterwards resulted in an error.

pkg/registry/decode.go
```diff
@@ -42,3 +43,19 @@ func DecodePackageManifest(reader io.Reader) (manifest *PackageManifest, err err
 	manifest = obj
 	return
 }
+
+func decodeFileFS(root fs.FS, path string, into interface{}) error {
+	if into == nil {
```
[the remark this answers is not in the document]
njhale: Ah, nice catch (and refresher).

IIUC all of the refactoring options -- for a generic decode function at least -- would put the onus on the caller to make sure they value is non-nil. This makes me think that removing this check isn't a huge loss anyway.

This is a private helper function, so we can always change the signature later too.

pkg/registry/parse.go
njhale: After reviewing the code again, it _does_ seem like the behavior -- pick the first CSV and ignore the rest -- matches what's currently in `master`.

See https://github.com/operator-framework/operator-registry/blob/f1629cd112099303272c245f1f9de3f8ab6fdcf2/pkg/registry/imageinput.go#L90 for more details.

PR #673, opened 2021-06-04
feat(render): render sqlite database files

estroz:
**Description of the change:** render sqlite db files directly

**Motivation for the change:** follow up on https://github.com/operator-framework/operator-registry/pull/661#discussion_r639255740

**Reviewer Checklist**

### DIFF
--- a/internal/action/render.go
+++ b/internal/action/render.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/h2non/filetype"
+	"github.com/h2non/filetype/matchers"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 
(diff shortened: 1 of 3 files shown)


internal/action/render.go
```diff
@@ -140,6 +150,35 @@ func (r Render) imageToDeclcfg(ctx context.Context, imageRef string) (*declcfg.D
...
+	}()
+
+	// Just read enough bytes to detect if ref is an sqlite3 db
+	// so the whole file is not loaded into memory.
+	var buf [4]byte
+	if _, err := f.Read(buf[:]); err != nil {
+		return fmt.Errorf("error reading first 4 bytes of ref %q: %v", ref, err)
+	}
+
+	typ, err := filetype.Match(buf[:])
```
joelanford, line 172: Use `filetype.MatchFile()` here instead?

joelanford (changes requested):
Just one suggestion. Otherwise looks great! Thanks!

internal/action/render.go
estroz: :+1: `MatchFile()` only [loads 8kB](https://github.com/h2non/filetype/blob/master/match.go#L56) into memory which is fine.

joelanford (approved):
/lgtm

joelanford:
/hold

This needs to be rebased and tweaked to account for changes in #687 (we removed the testdata sqlite db file; its being generated as part of the test now)

dinhxuanvu (approved):
/lgtm

PR #674, opened 2021-06-07
Remove `opm alpha add`, `declcfg.WriteDir`, and related now-unused code/tests

joelanford:
This PR removes the `opm alpha add` command, the `declcfg.WriteDir` method, and any code and tests that are used only by their code paths.

* `opm alpha add`

	We decided to go with the `init` and `render` workflow of generating DC blobs, but not directly adding them to an index. This path is more in line with our vision for a declarative index in that `opm` no longer directly manipulates the index.

* `declcfg.WriteDir`

	The `WriteDir` function implemented an opinionated filesystem structure for a declarative config that is generally in-line with the OLM community's thoughts on a good structure (i.e. a package-based structure where each package gets a subdirectory inside the index root directory).

	However, it suffered from being _too_ opinionated and not handling potential future extensions well. For example, DC does not prohibit extension schemas and properties from including relative file paths in the DC config. Since extension schemas and properties are opaque to `opm`, the `WriteDir` command would have no way to know if relative file references exist, and therefore could not move referenced files with the DC blob they are referenced from, thus breaking these references.

**Reviewer Checklist**


joelanford:
The `go-apidiff` tests are failing on recently added `pkg/action` types that we're used only for `opm alpha add` codepaths. They _are_ included in `v1.17.0`, but they are not likely to be used because they are specific to an alpha subcommand.

Note: In the future if there's a need for alpha-level public/exported code, I think we should put it in an `alpha`-rooted package to make clear to API consumers that it is subject to change despite the v1 repo status.

estroz:
>However, it suffered from being too opinionated and not handling potential future extensions well.

I still see value in `WriteDir` now even if it does not capture "the future" adequately. There needs to be some option to split by a demarcation, in this case package, for big DC's, which can't be easily opened by text editors for modification if in one big file. You can't natively split using unix pipes/redirects without a tool like `jq`, and I can see lots of users writing their own "split by package" `jq` scripts. I know `WriteDir` is not hooked into `opm alpha render` yet, but probably should be.

>For example, DC does not prohibit extension schemas and properties from including relative file paths in the DC config. Since extension schemas and properties are opaque to opm, the WriteDir command would have no way to know if relative file references exist, and therefore could not move referenced files with the DC blob they are referenced from, thus breaking these references.

I am pretty sure this problem still exists when rendering a DC without `WriteDir`. The invocation
```sh
opm alpha render ./configs > ./configs/index.json
```
may still work, but that is a special case since
```sh
opm alpha render ./configs-1 ./configs-2 my/catalog/image:v1.0.0 > ./configs-1/index.json
```
will break relative paths in `./configs-2` and `my/catalog/image:v1.0.0`.

joelanford:
If I'm reading the codecov diff stats correctly, I think the overall coverage percentage dropped slightly because the removed tests covered a larger relative percentage than the rest of the repo. Hence, I don't think there's a reasonable way to improve coverage. We would have to add unrelated tests.

estroz:
>but at least in that case it isn't a bug with opm. It's a bug in the script that calls opm. We should have good docs about render outlining this problem.

I can get behind this argument. It would be awesome to have a doc with some very common operations implemented in `jq`/`yq`, and a few complex scripts. Especially one script that
1. Takes a (potentially large) DC output from `render`.
2. Divides it into files by package name.
3. Updates file references by either copying files or changing the referenced path.
4. Builds a new index image.

/lgtm

joelanford:
Something like this perhaps :)

https://github.com/joelanford/declcfg/blob/3f38d2aa1f30a8222717605874f53c715ab8f99e/lib/funcs.sh#L119-L135

PR #679, opened 2021-06-07
Add declarative config `.indexignore` support

joelanford:
**Description**:
This commit adds support for .indexignore files (exact same functionality as .gitignore files) inside of declarative config directories.

A typical .indexignore file for a repo might look like this:

```gitignore
# Ignore everything
*

# Except, do NOT ignore json and yaml files
!*.json
!*.yaml

# However, DO ignore referenced olm.bundle.object
# files (overrides json and yaml exlusions)
objects/*.json
objects/*.yaml
```

**Motivation**:    
Prior to this commit, files present in the DC directory that did not successfully parse as declarative config files were silently ignored. In general, this is fine. However simple syntax mistakes in JSON or YAML would cause a file a user expected to be DC to be ignored with no warning.

Since users now have a way to communicate intent for what files should and should not be processed as DC, errors can now be emitted when intended DC files fail to parse.

**Reviewer Checklist**


dinhxuanvu (approved):
/approve

PR #680, opened 2021-06-14
declcfg: change LoadDir to LoadFS

joelanford:
Refactor `declcfg.LoadDir()` to `declcfg.LoadFS()`

Decouple loading declarative config directories from an underlying filesystem implementation. Now arbitrary DC can be loaded from an arbitrary `fs.FS`.

This approach is actually very similar to the now-deleted internal custom `fsWalker` interface, but now its using the interface from Go's stdlib, which should make it extensible to many other implementations.

**Reviewer Checklist**


internal/action/render.go
```diff
@@ -51,7 +51,7 @@ func (r Render) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 		)
 		// TODO(joelanford): Add support for detecting and rendering sqlite files.
 		if stat, serr := os.Stat(ref); serr == nil && stat.IsDir() {
-			cfg, err = declcfg.LoadDir(ref)
+			cfg, err = declcfg.LoadFS(os.DirFS(ref))
```
njhale, line 54: I think it makes sense -- from a testing perspective at least -- to DI this as part of the `Render` struct.

internal/declcfg/load.go
```diff
@@ -17,28 +17,26 @@ import (
 	"github.com/operator-framework/operator-registry/internal/property"
 )
 
-func LoadDir(configDir string) (*DeclarativeConfig, error) {
-	w := &dirWalker{}
-	return loadFS(configDir, w)
-}
-
-func loadFS(root string, w fsWalker) (*DeclarativeConfig, error) {
+func LoadFS(configFS fs.FS) (*DeclarativeConfig, error) {
```
njhale, line 20: nit: we can still call the fs `root`?

internal/declcfg/load_test.go
```diff
@@ -265,9 +274,534 @@ func TestLoadDir(t *testing.T) {
...
+	}
+
+	indexIgnore = &fstest.MapFile{
+		Data: []byte(`*
+!*.json
+!*.yaml
+
+*.clusterserviceversion.yaml`),
+	}
+	cockroachdb = &fstest.MapFile{
```
njhale, line 308: nit: in the future, it might be nice to handcraft smaller datasets for these tests. Sharing large datasets between testcases often leads to issues later on.

njhale (commented):
Looking great! Thanks for doing this.

/approve

internal/declcfg/load.go
joelanford: I changed the name because I thought `configFS` was clearer than `root`, but thinking about this a second time, I agree `root` (+ some GoDoc comment text) is better.

internal/action/render.go
joelanford: Totally agree. I somewhat blindly just replaced `LoadDir()` calls with `LoadFS(os.DirFS())` calls. I'll comb through again to see if there are any other places we can inject from higher levels.

internal/declcfg/load_test.go
joelanford: Yeah agreed. At the time, context was "let's make this look like the EP/existing tests to make sure I don't inadvertently miss something"

Now that we've got good test coverage, we could probably (at some point) refactor/minify test data and ensure coverage stays the same.

internal/action/render.go
joelanford: Hmm. Upon closer inspection, in the case of `render` it isn't trivial to inject references as filesystems since `render` supports image references, DC directories, and ([probably soon](https://github.com/operator-framework/operator-registry/pull/673)) sqlite DB files.

IMO implementing a containerd fs.FS implementation to read an image reference makes a lot of sense, but I think that's best left to a separate PR.

PR #683, opened 2021-06-16
Use fedora as base for image builder

kevinrizza:
Remove dependencies on Openshift and DockerHub for image build steps

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+FROM quay.io/fedora/fedora:34-x86_64 AS builder
 
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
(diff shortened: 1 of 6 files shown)


upstream-builder.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16-alpine as builder
+FROM quay.io/fedora/fedora:34-x86_64 as builder
```
njhale, line 1: Was there a reason to change this from `alpine`? 

I suspect this image build is also broken now due to the `apk` calls below.

njhale, line 1: Can we rename this `Dockerfile` and drop the original?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+FROM quay.io/fedora/fedora:34-x86_64 AS builder
```
njhale, line 2: There are still a lot of openshift specifics in this file. I'm wondering if we can drop it entirely.

njhale (changes requested):

joelanford:
@kevinrizza can this PR be closed out now that we're building via goreleaser and distroless?

PR #685, opened 2021-06-21
Add an option to store Gzip-encoded bundle data in ConfigMaps

zcahana:

This PR adds an option to compress/encode bundle data stored in a ConfigMap, so that large bundles have better chance fitting within the 1048576 bytes limit for ConfigMaps.

It's basically a stop-gap solution to operator-framework/operator-lifecycle-manager#1523, until the design proposed in operator-framework/enhancements#40 is finalized and implemented.

Some key aspects of this implementation:
- The compression option is opt-in, and is activated by a `-z`/`--gzip` flag to `opm alpha bundle extract` command.
- When enabled, the manifests content is compressed using *Gzip* and then encoded using *Base64*.
- *Base64* isn't strictly required here, but results with some nicer looking/better debuggable ConfigMaps without garbled binary data, in the expense of a slightly larger storage size.
- ConfigMaps with compressed data are discriminated using a `olm.contentEncoding: gzip+base64` annotation.
- This PR should be complemented with a PR in OLM to set the `--gzip` flag while [invoking the bundle unpacker job](https://github.com/operator-framework/operator-lifecycle-manager/blob/74881ec69721176a69d509409f348179eb9014d6/pkg/controller/bundle/bundle_unpacker.go#L105).

So far I've been manually testing this with the [kubevirt/hyperconverged-cluster-operator](https://github.com/kubevirt/hyperconverged-cluster-operator/tree/main/deploy/olm-catalog/community-kubevirt-hyperconverged/1.5.0) bundle.
With compression enabled, this bundle takes 180KB, compared to 1004KB uncompressed (and 152KB compressed w/o base64 encoding).
I'll be extending this PR with unit+e2e tests, given the approach taken here is acceptable by the project maintainers.

Thanks!

There's a growing number of operators who are hitting the 1048576 bytes barrier, including kubevirt/hyperconverged-cluster-operator, cert-manager, Strimzi, OCS, ...

**Reviewer Checklist**


kevinrizza:
IMO This looks more or less exactly how I was expecting an implementation would look. I think we had some thoughts on whether it was better to only compress the CRDs + other raw kube manifests and leave the CSVs as is because we know that a large portion of the filesize of CSVs comes from already base64 encoded image metadata -- it would probably be interesting to see if we could run this against a few other operator bundles to see what kinds of reductions we get. But I imagine even just the naive implementation that compresses all of the contents will get very good results given the bulk of the size comes from the CRD structural schemas.

pkg/lib/encoding/encoding.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return ungzipped, nil
+}
+
+func gzipEncode(data []byte) ([]byte, error) {
+	buf := &bytes.Buffer{}
+
+	writer := gzip.NewWriter(buf)
+	_, err := writer.Write(data)
```
exdx, line 47: I suppose we don't check the number of bytes written after compression because don't have a way of knowing how big the compressed file should be? I suppose lossy conversion is not a concern in this particular case?

pkg/lib/encoding/encoding.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		writer.Close()
```
exdx, line 49: could we `defer writer.Close()` instead?

joelanford (commented):
> Base64 isn't strictly required here, but results with some nicer looking/better debuggable ConfigMaps without garbled binary data, in the expense of a slightly larger storage size.

Can the ConfigMap `binaryData` field be used to solve this problem?

pkg/lib/encoding/encoding.go
```diff
@@ -0,0 +1,95 @@
...
+	zBuf := bytes.NewBuffer(data)
+
+	reader, err := gzip.NewReader(zBuf)
+	if err != nil {
+		return nil, err
+	}
+	defer reader.Close()
+
+	buf := &bytes.Buffer{}
+	_, err = io.Copy(buf, reader)
```
exdx, line 68: here I think maybe we should check to see the number of bytes read matches what was passed-in?

exdx (commented):
Really impressive PR! Just have some nits and general observations. 

I think this would definitely be valuable in OLM.

pkg/configmap/configmap_writer.go
```diff
@@ -218,3 +239,14 @@ func LaunchBundleImage(kubeclient kubernetes.Interface, bundleImage, initImage,
 
 	return newConfigMap, launchedJob, nil
 }
+
+func setGzipEncodingAnnotation(cm *corev1.ConfigMap) {
+	annotations := cm.GetAnnotations()
```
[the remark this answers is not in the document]
zcahana: good point, will add that.

pkg/configmap/configmap.go
```diff
@@ -50,17 +53,32 @@ func (l *BundleLoader) Load(cm *corev1.ConfigMap) (bundle *api.Bundle, err error
...
 }
 
-func loadBundle(entry *logrus.Entry, data map[string]string) (bundle *api.Bundle, skipped map[string]string, err error) {
+func loadBundle(entry *logrus.Entry, cm *corev1.ConfigMap) (bundle *api.Bundle, skipped map[string]string, err error) {
 	bundle = &api.Bundle{Object: []string{}}
 	skipped = map[string]string{}
 
+	gzipped := hasGzipEncodingAnnotation(cm)
+	if gzipped {
+		entry.Info("Decoding gzip-encoded bundle data")
```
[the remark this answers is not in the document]
zcahana: Does this also apply to the catalog-operator? Since this code is actually invoked from there during steps generation. If so, this could definitely be reduced to debug level.

cmd/opm/alpha/bundle/extract.go
```diff
@@ -55,9 +56,13 @@ func runExtractCmd(cmd *cobra.Command, args []string) error {
...
 		return err
 	}
+	gzip, err := cmd.Flags().GetBool("gzip")
+	if err != nil {
+		return err
+	}
 
 	loader := configmap.NewConfigMapLoaderForDirectory(configmapName, namespace, manifestsDir, kubeconfig)
-	if err := loader.Populate(datalimit); err != nil {
+	if err := loader.Populate(datalimit, gzip); err != nil {
```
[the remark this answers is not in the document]
zcahana: I've been debating this with myself too, and didn't reach a clear conclusion :-) Eventually I chose to pass this parameter together with `datalimit`, since they're somehow related in dealing with bundle size. That was my logic at least. Anyway, no problem to change to a loader attribute.

pkg/lib/encoding/encoding.go
zcahana: > I suppose we don't check the number of bytes written after compression because don't have a way of knowing how big the compressed file should be?

Right, I see nothing useful we could do with this number rather than the spamming the already-noisy logs :-)

> I suppose lossy conversion is not a concern in this particular case?
gzip is loseless - this is an absolute requirement for our compression use here...

pkg/lib/encoding/encoding.go
zcahana: We can't in this case, since we must `Close()` (--> flush) before accessing the underlying `buf.Bytes()`.
Originally, I actually wrote this with `defer writer.Close()`, and discovered this thanks to the unit tests.

pkg/lib/encoding/encoding.go
zcahana: Don't think we'd have what to compare it to - the number of bytes returned from `io.Copy()` would be the number of uncompressed bytes, whereas this function receives only the compressed bytes.

zcahana:
Thanks all for this thorough review. I'll work on the suggested changes as well as unit/e2e test.

pkg/configmap/configmap_writer.go
```diff
@@ -116,9 +107,35 @@ func (c *ConfigMapWriter) Populate(maxDataSizeLimit uint64) error {
...
-			} else {
-				configMapPopulate.Data[validConfigMapKey] = string(content)
+
+				// annotations aren't accounted for the ConfigMap data size
+				// limit, and rather have their own limit of 262144 bytes.
+				continue
 			}
+
+			if gzip {
+				content, err = encoding.EncodeGzipBase64(content)
```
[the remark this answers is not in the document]
zcahana: Probably so. This might actually turn both more efficient and more compact. Will try to redesign.

zcahana:
> Base64 isn't strictly required here, but results with some nicer looking/better debuggable ConfigMaps without garbled binary data, in the expense of a slightly larger storage size.

> Can the ConfigMap `binaryData` field be used to solve this problem?

~No~, yes! I've only now realized that base64 encoding is applied to anything passed into `binaryData`, so we're basically doing double base64 encoding here (and wastefully increasing data size). I'll certainly look into removing our own base64 encoding.

joelanford:
@zcahana As I understand it, `binaryData` fields are actually stored as raw binary (it's a [`map[string][]byte`](https://github.com/kubernetes/api/blob/0f1d05d0f33885faa7bc4a760cee2c01641bb897/core/v1/types.go#L5977-L5985)). It's just rendered as base64 when encoded as YAML or JSON.

zcahana:
>  As I understand it, `binaryData` fields are actually stored as raw binary (it's a `map[string][]byte`). It's just rendered as base64 when encoded as YAML or JSON.

Best of both worlds :-)

zcahana:
> IMO This looks more or less exactly how I was expecting an implementation would look. I think we had some thoughts on whether it was better to only compress the CRDs + other raw kube manifests and leave the CSVs as is because we know that a large portion of the filesize of CSVs comes from already base64 encoded image metadata -- it would probably be interesting to see if we could run this against a few other operator bundles to see what kinds of reductions we get. But I imagine even just the naive implementation that compresses all of the contents will get very good results given the bulk of the size comes from the CRD structural schemas.

@kevinrizza here's a quick look at the top-5 CSVs by size, as well as their compressed sizes:
```bash
$ ll ./redis-operator/0.5.0/redis-operator.v0.5.0.clusterserviceversion.yaml*
-rw-r--r-- 1 zvic zvic 872918 Jun 21 23:15 ./redis-operator/0.5.0/redis-operator.v0.5.0.clusterserviceversion.yaml
-rw-r--r-- 1 zvic zvic 352004 Jun 21 22:37 ./redis-operator/0.5.0/redis-operator.v0.5.0.clusterserviceversion.yaml.gz

$ ll ./ember-csi-community-operator/0.9.4/ember-csi-community-operator.v0.9.4.clusterserviceversion.yaml*
-rw-r--r-- 1 zvic zvic 576990 Jun 21 23:15 ./ember-csi-community-operator/0.9.4/ember-csi-community-operator.v0.9.4.clusterserviceversion.yaml
-rw-r--r-- 1 zvic zvic  51783 Jun 21 22:37 ./ember-csi-community-operator/0.9.4/ember-csi-community-operator.v0.9.4.clusterserviceversion.yaml.gz

$ ll ./hedvig-operator/1.0.1/manifests/hedvig-operator.v1.0.1.clusterserviceversion.yaml*
-rwxr-xr-x 1 zvic zvic 254115 Jun 21 23:15 ./hedvig-operator/1.0.1/manifests/hedvig-operator.v1.0.1.clusterserviceversion.yaml*
-rwxr-xr-x 1 zvic zvic 182983 Jun 21 22:37 ./hedvig-operator/1.0.1/manifests/hedvig-operator.v1.0.1.clusterserviceversion.yaml.gz*

$ ll ./jaeger/1.20.0/jaeger.v1.20.0.clusterserviceversion.yaml*
-rw-r--r-- 1 zvic zvic 231809 Jun 21 23:15 ./jaeger/1.20.0/jaeger.v1.20.0.clusterserviceversion.yaml
-rw-r--r-- 1 zvic zvic  68659 Jun 21 22:37 ./jaeger/1.20.0/jaeger.v1.20.0.clusterserviceversion.yaml.gz

$ ll ./cockroachdb/2.0.9/manifests/cockroachdb.v2.0.9.clusterserviceversion.yaml*
-rw-r--r-- 1 zvic zvic 186843 Jun 21 23:15 ./cockroachdb/2.0.9/manifests/cockroachdb.v2.0.9.clusterserviceversion.yaml
-rw-r--r-- 1 zvic zvic 131128 Jun 21 22:37 ./cockroachdb/2.0.9/manifests/cockroachdb.v2.0.9.clusterserviceversion.yaml.gz
```

As you see, the compression ratios vary substantially. Lower ratios indeed correspond with CSVs with base64 encoded data inline. Still, some CSVs benefit nicely from compression (particularly, `redis-operator.v0.5.0` stepping away from the red line, and `ember-csi-community-operator.v0.9.4` getting a x10 reduction in size), so I think it worthwhile to keep compressing CSVs too.

pkg/lib/encoding/encoding.go
zcahana: Actually, it turns out that Gzip stores the length and checksum of the uncompressed data, and the [gzip.Reader](https://golang.org/pkg/compress/gzip/#Reader) verifies that the number of uncompressed bytes matches that. So in case of any mismatch, an `ErrChecksum` will be propagated from the `gzip.Reader` up to the `io.Copy`.

pkg/configmap/configmap.go
```diff
@@ -50,17 +53,32 @@ func (l *BundleLoader) Load(cm *corev1.ConfigMap) (bundle *api.Bundle, err error
...
 			"key": name,
 		})
 
+		var reader io.Reader
+		if gzipped {
+			decoded, err :=  encoding.DecodeGzipBase64([]byte(content))
+			if err != nil {
+				return nil, nil, fmt.Errorf("error decoding gzip-encoded bundle data: %v", err)
+			}
+			reader = bytes.NewReader(decoded)
```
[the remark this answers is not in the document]
zcahana: The `reader` is only used by the `registry.DecodeUnstructured(reader)` below, but the actual decompressed `content []byte` is still needed further down (`bundle.Object = append(bundle.Object, content)`). But since we've given up on our own base64 encoding, this clearly uses less memory now.

pkg/configmap/configmap_writer.go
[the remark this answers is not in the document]
zcahana: After spending some time on this, I don't think it make sense anymore to redesign around `io.Writer`/`io.Reader`:
1. Most of the redundant (intermediate) `[]byte`'s have been eliminated with the removal of the base64 encoding/decoding.
2. Some of the code paths on `configmap_writer.go`/`configmap.go` still need the actual `[]byte`. Trying to optimize the other code paths that *don't* need it actually results with more complex/less cohesive code.

zcahana:
@all,

I've worked through all review comments and added unit tests.
As for e2e tests, I've taken a quick look through the existing e2e infra both here and at the OLM repo, and it seems to me that it would be easier to implement it in there, not to mention we'd be able to test the actual e2e flow (bundle image > 1mb, opm bundle extract, ConfigMap, InstallPlan, ...). What do you think?

exdx:
Hi @zcahana,

w.r.t. e2e testing, we think its best to have a specific operator-registry e2e test where we we spin up a cluster, run the unpack job, and check that the content made it to the configmap in the right format. There is an e2e test in operator-registry that runs in-cluster already, so it should be ok to use the existing infrastructure for the test. 

Although the OLM e2e suite is more robust, we don't need this test to integrate into the OLM APIs itself. The test in OLM is already there: do an install and make sure that the unpacking still works after adding the compression flag. This can be turned on in a subsequent PR (after vendoring in this change) and if CI is green then we can consider it working on the OLM side as well.

zcahana:
> As I understand it, binaryData fields are actually stored as raw binary (it's a map[string][]byte). It's just rendered as base64 when encoded as YAML or JSON.

@joelanford I've been playing with the base64 encoding/decoding back and forth and eventually realized we must base64-decode what we get back when reading the binaryData, regardless if we base64-encoded it when writing, or didn't.
As for writing, we can either pass the raw bytes in, or base64-encode first and then pass it in - the result is the same, as well as the size limit (which always apply to the un-encoded raw bytes, even if we pass in base64 encoded data).

Bottom line, we're back to gzip+base64 encoding/decoding, but at least this time I've rewritten it with better memory efficiency.

zcahana:
> w.r.t. e2e testing, we think its best to have a specific operator-registry e2e test where we we spin up a cluster, run the unpack job, and check that the content made it to the configmap in the right format. There is an e2e test in operator-registry that runs in-cluster already, so it should be ok to use the existing infrastructure for the test.

@exdx thanks for the tip about the e2e test. I've indeed extended the existing one to launch the extract job, and then read the configmap and verify we get the expected objects. The test runs twice: once for a small bundle, without applying compression, and once for a large (> 1MiB) bundle), with compression applied.

test/e2e/bundle_image_test.go
```diff
@@ -3,6 +3,8 @@ package e2e_test
 import (
 	"context"
 	"fmt"
+	"github.com/onsi/ginkgo/extensions/table"
```
exdx, line 6: nit: looks like the ordering of the new dependencies is causing goimports to fail so CI is red
```
find . -name '*.go' -not -path "./vendor/*" | xargs goimports -w
diff --git a/test/e2e/bundle_image_test.go b/test/e2e/bundle_image_test.go
index 637795d..5615346 100644
--- a/test/e2e/bundle_image_test.go
+++ b/test/e2e/bundle_image_test.go
```

exdx (commented):
nice work! I'm ready to lgtm once CI is green

test/e2e/bundle_image_test.go
exdx: Also, squashing all the commits for the final PR would be great as well

zcahana: Fixed both.

kevinrizza:
/approve

Looks like this still has valid failing CI because a public api function signature was modified:

https://github.com/operator-framework/operator-registry/pull/685/checks?check_run_id=2944060605#step:4:34

I'm not sure I see a way around that for this change, so I think it's reasonable to override this failure

exdx:
```
 LaunchBundleImage: changed from func(k8s.io/client-go/kubernetes.Interface, string, string, string) (*k8s.io/api/core/v1.ConfigMap, *k8s.io/api/batch/v1.Job, error) to func(k8s.io/client-go/kubernetes.Interface, string, string, string, bool) (*k8s.io/api/core/v1.ConfigMap, *k8s.io/api/batch/v1.Job, error)
  ```

I believe this will have to be updated in OLM with the next registry bump

Looks like the e2e test is spinning, maybe a retry will help it pass. 

/lgtm

**Comment by zcahana** [comment]
> I believe this will have to be updated in OLM with the next registry bump

I don't see `LaunchBundleImage` used anywhere in OLM (anymore?), looks like it has [its own code](https://github.com/operator-framework/operator-lifecycle-manager/blob/a4d4ab60bc28850e42b99a14a60a2019e5d68ab2/pkg/controller/bundle/bundle_unpacker.go#L85) to spin up the extract job.

> Looks like the e2e test is spinning, maybe a retry will help it pass.

Since this includes a new/modified e2e test, let's let it timeout to ensure it's not related.

**Comment by zcahana** [comment]
> Since this includes a new/modified e2e test, let's let it timeout to ensure it's not related.

Indeed the failure was in the new e2e test. According to the test logs the ConfigMap was empty. I suspect that the extract job wasn't given a proper chance to complete (only 15s) before moving on to reading the ConfigMap. I increased that timeout to 120 seconds and also added a log message if we hit that timeout.

**Comment by timflannagan** [comment]
Looks like there are some failures in the e2e test:

```
[Fail] Launch bundle Deploy bundle job should populate specified configmap [It] Small bundle, uncompressed 
/home/runner/work/operator-registry/operator-registry/test/e2e/bundle_image_test.go:198

[Fail] Launch bundle Deploy bundle job should populate specified configmap [It] Large bundle, compressed 
/home/runner/work/operator-registry/operator-registry/test/e2e/bundle_image_test.go:112
```

I haven't dived into those errors, but placing a hold on this PR in the meantime so the bot doesn't go crazy attempting to retest non-prow-based jobs.

/hold

**Comment by zcahana** [comment]
The current failures are still due to a timeout (120 sec) in waiting for the extract job to complete.
Unfortunately I couldn't reproduce in my local env - tried both minikube and kind, with/without TLS, whole test suite or just the single test - everything passes time after time. 

I can try to increase the timeout even further, but it's hard to believe the test is running that slow on the CI.
It would be best to take a look at the bundle job's logs, but I'm not sure if it's collected somewhere by the e2e running here. Alternatively I can add test code that dumps it to the test logs, and remove it when the issue is resolved.

**Comment by timflannagan** [comment]
@zcahana We have a couple of testing flakes that are difficult to reproduce locally but will pop up frequently while in CI - it's possible there's some sort of ginko parallelization magic going on that causes these flakes, but it's difficult to track down. It's possible that the testing environment that GH actions spins up for us has limited resources, which may help explain why more frequent tests are flaking due to exceeding the configured test timeout period, but I haven't dived too much into that angle either. Increasing the test timeout to find a good medium seems like a sufficient workaround in the short term though.

**Comment by joelanford** [comment]
@zcahana 

> I've been playing with the base64 encoding/decoding back and forth and eventually realized we must base64-decode what we get back when reading the binaryData, regardless if we base64-encoded it when writing, or didn't.

I'm not sure I'm following the reasoning here, though its entirely possible I'm overlooking something. I checked out the PR and removed the `base64.NewEncoder` and `base64.NewDecoder` (and related) lines in `pkg/lib/encoding`, ran the unit tests, and everything still passed.

I would have expected the `TestLoadWriteRead/BundleCompressed` test in `pkg/configmap/configmap_test.go` to fail if base64 decoding the data from the configmap is always required.

**Comment by zcahana** [comment]
> I'm not sure I'm following the reasoning here, though its entirely possible I'm overlooking something. I checked out the PR and removed the base64.NewEncoder and base64.NewDecoder (and related) lines in pkg/lib/encoding, ran the unit tests, and everything still passed.
>
> I would have expected the TestLoadWriteRead/BundleCompressed test in pkg/configmap/configmap_test.go to fail if base64 decoding the data from the configmap is always required.

@joelanford The unit test still passes since the fake client just returns whatever was written into the in-memory ConfigMap.binaryData as-is, encoded or not. In a real cluster (and indeed I discovered this only after starting to actually test the code using a real cluster), the ConfigMap binaryData field *always* returns base64-encoded when read from the api-server. This happens whether I base64-encoded the data put in binaryData (as I do now), or used the "raw" gzipped bytes (as I initially did)

**Comment by joelanford** [comment]
> The unit test still passes since the fake client just returns whatever was written into the in-memory ConfigMap.binaryData as-is, encoded or not. In a real cluster (and indeed I discovered this only after starting to actually test the code using a real cluster), the ConfigMap binaryData field always returns base64-encoded when read from the api-server. This happens whether I base64-encoded the data put in binaryData (as I do now), or used the "raw" gzipped bytes (as I initially did)

Oh wow, that is extremely surprising (to me at least). I wonder if this is documented somewhere... :thinking:

**Comment by zcahana** [comment]
Guys, can someone please kick off the CI? The two last commits increase the job timeout as well as dump job logs in case of timeout error. Let's see where we are with this.

**Comment by dinhxuanvu** [comment]
@zcahana Would you mind rebasing your PR against master? We recently fixed a kind test on CI so your PR won't pass CI until it is rebased. Thanks.

**Comment by zcahana** [comment]
> Would you mind rebasing your PR against latest master? We recently fixed a kind test on CI so your PR won't pass CI until it is rebased. Thanks.

Will do, thanks.
Based on the [recently added debug logging](https://github.com/operator-framework/operator-registry/pull/685/checks?check_run_id=2974753814#step:6:409), it appears that there's no schedulable node hence the bundle extract job runs into timeout (10m) and the test fails:

```
INFO: Displaying job status for deploy-bundle-image-zclds
&JobStatus{Conditions:[]JobCondition{},StartTime:2021-07-02 18:16:48 +0000 UTC,CompletionTime:<nil>,Active:1,Succeeded:0,Failed:0,}
INFO: Found 1 pods matching job selector job-name=deploy-bundle-image-zclds
INFO: Displaying pod status for deploy-bundle-image-zclds-rtw2s
&PodStatus{Phase:Pending,Conditions:[]PodCondition{PodCondition{Type:PodScheduled,Status:False,LastProbeTime:0001-01-01 00:00:00 +0000 UTC,LastTransitionTime:2021-07-02 18:16:48 +0000 UTC,Reason:Unschedulable,Message:0/1 nodes are available: 1 node(s) had taint {node.kubernetes.io/not-ready: }, that the pod didn't tolerate.,},},Message:,Reason:,HostIP:,PodIP:,StartTime:<nil>,ContainerStatuses:[]ContainerStatus{},QOSClass:BestEffort,InitContainerStatuses:[]ContainerStatus{},NominatedNodeName:,PodIPs:[]PodIP{},EphemeralContainerStatuses:[]ContainerStatus{},}
```

zcahana:
@kevinrizza @joelanford @exdx @timflannagan @dinhxuanvu 

Thank you all for the thorough review and assistance throughout this PR.
To close the loop, we'll need to set the new `--gzip` flag in the bundle extract job spawned by OLM.
Do you have an estimation for when there will be an `operator-registry` release including this patch?

PR #687, opened 2021-06-23
Query related_image table when rendering DC from sqlite

joelanford:
When rendering declarative config from an sqlite database, query the `related_image` table to find any related images we haven't already found from the CSVs in the `operatorbundle` table.

Typically, the `operatorbundle` table contains CSVs only for channel heads. Therefore querying _just_ the CSVs for related images will typically result in finding related images for channel heads.

To find all related images, the `related_image` table should also be queried.

Why not use _just_ the `related_image` table? Because the CSV's `relatedImages` field has an image name and image reference, whereas the `related_image` table contains just the image reference. The declarative config's `olm.bundle` `relatedImages` field allows a `name` to be set, so we use the `related_image` table to supplement what we find the the CSVs.  

**Reviewer Checklist**


internal/action/render.go
```diff
@@ -165,9 +166,55 @@ func sqliteToDeclcfg(ctx context.Context, dbFile string) (*declcfg.DeclarativeCo
...
 	return &cfg, nil
 }
 
+func populateDBRelatedImages(ctx context.Context, cfg *declcfg.DeclarativeConfig, db *sql.DB) error {
+	rows, err := db.QueryContext(ctx, "SELECT image, operatorbundle_name FROM related_image")
+	if err != nil {
+		return err
+	}
+	images := map[string]sets.String{}
+	for rows.Next() {
```
timflannagan, line 182: I was blank on whether you need to defer the rows.Close() here. Looking at the godocs, it looks like it should be safe assuming rows.Next() eventually returns false, as the rows will get closed automatically.

joelanford: Good catch. Can't hurt to do it explicitly.

joelanford: Done!

PR #690, opened 2021-06-24
WIP declcfg: allow multiple replaces values per bundle per channel

joelanford:

Relax the restriction that a bundle can only replace one other bundle per channel. Now a bundle can replace any number of other bundles in a channel provided that the replaced bundle has no more than one replacer in the channel that is not skipped by any other bundle in the channel.

The end result is that the ListBundles response may now contain multiple copies of a bundle in the same channel where only the `api.Bundle.Replaces` value differs.

It also opens up more possibilities, because a skipped bundle is removed as an upgrade candidate throughout the entire channel graph, regardless of whether the bundle that skips it is a candidate for a particular update. By changing `skips` to `replaces`, operator teams can have more fine-grained control of upgrade candidate edges.

It also can greatly simplify upgrade graphs because it is possible to simplify replaces/skips graphs down to just replaces graphs.  

Consider what a typical semver-skiprange graph looks like today: 
[](https://mermaid-js.github.io/mermaid-live-editor/edit/##eyJjb2RlIjoiZ3JhcGggUkxcblxuICAxLjAuMSAtLSByZXBsYWNlcyAtLT4gMS4wLjBcbiAgMS4wLjIgLS0gcmVwbGFjZXMgLS0-IDEuMC4wXG4gIDEuMC4yIC0tIHNraXBzIC0tPiAxLjAuMVxuICAxLjAuMyAtLSByZXBsYWNlcyAtLT4gMS4wLjBcbiAgMS4wLjMgLS0gc2tpcHMgLS0-IDEuMC4xXG4gIDEuMC4zIC0tIHNraXBzIC0tPiAxLjAuMlxuICAxLjAuNCAtLSByZXBsYWNlcyAtLT4gMS4wLjBcbiAgMS4wLjQgLS0gc2tpcHMgLS0-IDEuMC4xXG4gIDEuMC40IC0tIHNraXBzIC0tPiAxLjAuMiBcbiAgMS4wLjQgLS0gc2tpcHMgLS0-IDEuMC4zIiwibWVybWFpZCI6IntcbiAgXCJ0aGVtZVwiOiBcImRlZmF1bHRcIlxufSIsInVwZGF0ZUVkaXRvciI6ZmFsc2UsImF1dG9TeW5jIjp0cnVlLCJ1cGRhdGVEaWFncmFtIjpmYWxzZX0)

With this change, the same graph could be represented with just replaces:
[](https://mermaid-js.github.io/mermaid-live-editor/edit/##eyJjb2RlIjoiZ3JhcGggUkxcbiAgMS4wLjQgLS0gcmVwbGFjZXMgLS0-IDEuMC4wXG4gIDEuMC40IC0tIHJlcGxhY2VzIC0tPiAxLjAuMVxuICAxLjAuNCAtLSByZXBsYWNlcyAtLT4gMS4wLjJcbiAgMS4wLjQgLS0gcmVwbGFjZXMgLS0-IDEuMC4iLCJtZXJtYWlkIjoie1xuICBcInRoZW1lXCI6IFwiZGVmYXVsdFwiXG59IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ)

As mentioned in the description, the following graph is still not allowed because A is replaced by both B and C, and neither B nor C are skipped in the channel. In this situation the OLM resolver in unable to determine whether to upgrade to B or C, so `opm alpha validate` is updated to emit an error for this case:
[](https://mermaid-js.github.io/mermaid-live-editor/edit/##eyJjb2RlIjoiZ3JhcGggUkxcbiAgQyAtLSByZXBsYWNlcyAtLT4gQVxuICBDIC0tIHJlcGxhY2VzIC0tPiBCXG4gIEIgLS0gcmVwbGFjZXMgLS0-IEEiLCJtZXJtYWlkIjoie1xuICBcInRoZW1lXCI6IFwiZGVmYXVsdFwiXG59IiwidXBkYXRlRWRpdG9yIjp0cnVlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6dHJ1ZX0)

In OLM, all skipped bundles in a channel are eliminated as possible upgrade candidates. Operator authors can change `skips` bundles to `replaces` bundles, which effectively signals to the OLM resolver that the these bundles can still be candidates for upgrades in other areas of the graph, if so desired.

**Reviewer Checklist**


joelanford:
@benluddy pointed out to me that there are several reasons this won't work.

The primary problem is that for fresh installs if the channel head does not meet existing cluster constraints, OLM needs the rest of the channel to be sorted so that it can choose the next closest bundle from the head.

The unique ID the resolver uses for bundles in a channel is `catalog` + `channel` + `csvName`. This PR would cause multiple bundles to have the same ID in a channel because the `replaces` field is the only field that would differ. That is currently a case in which the resolver will emit an error.

And even if we updated OLM to make this situation a non-error, we would still have the fundamental problem of having multiple possible choices that are not ordered relative to each other.

PR #691, opened 2021-06-24
declcfg: validation should allow non-existent replaces value

joelanford:
Remove the validation check that ensures all bundle replaces values are present in the channel.

It is not necessary for a catalog to contain all referenced bundles. An old catalog may contain a bundle that has been replaced by a bundle that exists in a new catalog. The OLM resolver already handles this case.

Note that it is still a requirement that within a given DC catalog, there must be exactly one channel head per channel (where channel head is defined as a node with 0 incoming edges)

**Reviewer Checklist**


estroz (approved):
Is it worth noting this characteristic in the DC enhancement or some other docs?

/lgtm

PR #694, opened 2021-07-01
Update case to match resource.

Jamstah:
https://github.com/openshift/api/blob/a99ffa1cac6709edf8f502b16890b16f9a557e00/console/v1/types_console_yaml_sample.go

Signed-off-by: James Hewitt <james.hewitt@uk.ibm.com>

Update string to match the real world

Kind is not recognised because the case does not match the resource.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/supported_resources.go
+++ b/pkg/lib/bundle/supported_resources.go
@@ -16,7 +16,7 @@ const (
 	PodDisruptionBudgetKind   = "PodDisruptionBudget"
 	PriorityClassKind         = "PriorityClass"
 	VerticalPodAutoscalerKind = "VerticalPodAutoscaler"
-	ConsoleYamlSampleKind     = "ConsoleYamlSample"
+	ConsoleYAMLSampleKind     = "ConsoleYAMLSample"
 	ConsoleQuickStartKind     = "ConsoleQuickStart"
 	ConsoleCLIDownloadKind    = "ConsoleCLIDownload"
 	ConsoleLinkKind           = "ConsoleLink"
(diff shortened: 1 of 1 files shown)


Jamstah:
I switched this to change just the value and not the constant, although I expect the current constant isn't being used (as its wrong anyway), so it might be better to make the breaking API change instead.

Jamstah:
Hah, changing the value is considered incompatible too, so I'm going to go ahead and say "we should make an incompatible change here".

pkg/lib/bundle/supported_resources.go
```diff
@@ -16,7 +16,7 @@ const (
 	PodDisruptionBudgetKind   = "PodDisruptionBudget"
 	PriorityClassKind         = "PriorityClass"
 	VerticalPodAutoscalerKind = "VerticalPodAutoscaler"
-	ConsoleYamlSampleKind     = "ConsoleYamlSample"
+	ConsoleYAMLSampleKind     = "ConsoleYAMLSampleKind"
```
dinhxuanvu, line 19: Is the `Kind` part expected on the value side?

dinhxuanvu:
Can you please rebase this PR against latest master? We recently fixed a kind test on CI. Thanks.

pkg/lib/bundle/supported_resources.go
Jamstah: No! Thank you. Copy paste bug from the repeated tests. I've been upgrading my go to be able to run the e2e locally.

Jamstah:
> Can you please rebase this PR against latest master? We recently fixed a kind test on CI. Thanks.

Done.

Jamstah:
/assign @dinhxuanvu

Assigning because you already seem to be looking at it

timflannagan:
Related: https://github.com/operator-framework/operator-registry/pull/757.

Jamstah:
@timflannagan rebased, can I have the lgtm back?

dinhxuanvu (approved):
/approve

Jamstah:
@timflannagan I had to rebase this so it lost your LGTM, can you please provide another?

PR #696, opened 2021-07-08
opm alpha diff: generate a diff between two sets of catalog/bundle refs

estroz:
**Description of the change:** Implements the `opm diff` component of [catalog-data-transfer-reduction](https://github.com/operator-framework/enhancements/blob/master/enhancements/catalog-data-transfer-reduction.md).

**Motivation for the change:** generate a diff for disconnected catalogs.

**Reviewer Checklist**

TODO's:
- [x] code comments
- [x] add usage docs
- [x] more tests

Signed-off-by: Eric Stroczynski <ericstroczynski@gmail.com>


internal/declcfg/declcfg_to_model.go
```diff
@@ -66,6 +68,23 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
 				}
 				mpkg.Channels[bundleChannel.Name] = pkgChannel
 			}
+
+			// Parse version from the package property, falling back to the CSV's spec.version field.
```
estroz, line 72: @joelanford opinions on this being a bundle field? Pre-parsing version makes comparison way easier and more efficient.

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	retryInterval = time.Second * 5
+	timeout       = time.Minute * 1
+)
+
+type diff struct {
+	oldRefs []string
+	newRefs []string
+
+	caFile   string
+	pullTool string
```
joelanford, line 28: Any reason we need `pullTool` to be anything other than `none`? Or can we just hardcode use of containerd to pull images?

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	logger *logrus.Entry
+}
+
+func NewCmd() *cobra.Command {
+	a := diff{
+		logger:   logrus.NewEntry(logrus.New()),
+		pullTool: "none",
+	}
+	rootCmd := &cobra.Command{
+		Use: "diff --old { render-target }... --new { render-target }...",
```
joelanford, line 40: Do we need `--old` and `--new` flags, or can we just use positional comma-separated args?

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (a *diff) addFunc(cmd *cobra.Command, args []string) error {
+	skipTLS, err := cmd.PersistentFlags().GetBool("skip-tls")
+	if err != nil {
+		a.logger.Fatal(err)
+	}
+
+	rootCAs, err := certs.RootCAs(a.caFile)
+	if err != nil {
+		return fmt.Errorf("failed to get root CAs: %v", err)
```
joelanford, line 74: Maybe debateable, but IMO this is more of an actual execution error than a usage error. Returning an error results in cobra printing the error and then outputting the usage text for the command. Do you think the usage text is valuable for this error?
```suggestion
		a.logger.Fatalf("failed to get root CAs: %v", err)
```

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	rootCmd.Flags().StringVarP(&a.caFile, "ca-file", "", "", "the root Certificates to use with this command")
+
+	rootCmd.Flags().BoolVar(&a.debug, "debug", false, "enable debug logging")
+	return rootCmd
+}
+
+func (a *diff) addFunc(cmd *cobra.Command, args []string) error {
+	skipTLS, err := cmd.PersistentFlags().GetBool("skip-tls")
+	if err != nil {
+		a.logger.Fatal(err)
```
joelanford, line 69: If this returns an error, that's only because the `skip-tls` flag hasn't been registered as a persistent flag right? In that case, wdyt about panic-ing because this is a programmer error, not a user error?
```suggestion
		panic(err)
```

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	rootCAs, err := certs.RootCAs(a.caFile)
+	if err != nil {
+		return fmt.Errorf("failed to get root CAs: %v", err)
+	}
+	reg, err := imgreg.NewRegistry(imgreg.SkipTLS(skipTLS), imgreg.WithLog(a.logger), imgreg.WithRootCAs(rootCAs))
+	if err != nil {
+		return err
```
joelanford, line 78: wdyt?
```suggestion
		return a.logger.Fatal(err)
```

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+
+	diff := action.Diff{
+		Registry: reg,
+		OldRefs:  a.oldRefs,
+		NewRefs:  a.newRefs,
+		Logger:   a.logger,
+	}
+	cfg, err := diff.Run(ctx)
+	if err != nil {
+		return err
```
joelanford, line 97:
```suggestion
		return a.logger.Fatal(err)
```

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return declcfg.WriteYAML(*cfg, os.Stdout)
```
joelanford, line 100:
```suggestion
	if err := declcfg.WriteYAML(*cfg, os.Stdout); err != nil {
		a.logger.Fatal(err)
	}
	return nil
```

joelanford: Also, would it be worth supporting JSON or YAML output, similar to https://github.com/operator-framework/operator-registry/blob/97579420458d7f0d3dae78848d286df45ce49114/cmd/opm/render/cmd.go#L28-L36 and https://github.com/operator-framework/operator-registry/blob/97579420458d7f0d3dae78848d286df45ce49114/cmd/opm/render/cmd.go#L48-L50

joelanford (commented):
First pass, mostly focused on structural stuff. I'll make another pass of the actual diff logic.

internal/declcfg/declcfg_to_model.go
estroz: Apologies, I meant on the `model.Bundle` type itself. I agree with what you said above.

I could also make the `olm.package`'s Go `Version` field a `semver.Version`.

internal/declcfg/diff.go
```diff
@@ -0,0 +1,393 @@
...
+
+	"github.com/operator-framework/operator-registry/internal/model"
+	"github.com/operator-framework/operator-registry/internal/property"
+)
+
+// Diff returns a Model containing everything in newModel not in oldModel,
+// and all bundles that exist in oldModel but are different in newModel.
+// If oldModel is empty, only channel heads in newModel's packages are
+// added to the output Model. All dependencies not in oldModel are also added.
+func Diff(oldModel, newModel model.Model) (model.Model, error) {
```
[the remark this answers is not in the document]
estroz: As things stand now, this makes sense. However the signature might change (follow-up) to accept an iterator type that can block-load packages from disk to overcome memory limitations. I'm not 100% sure where that iterator's definition will live yet, but it makes sense to place it in `declcfg` because it'll index packages on-disk at parse time.

internal/model/model.go
```diff
@@ -215,6 +216,10 @@ type Bundle struct {
 	// backwards-compatible way.
 	Objects []string
 	CsvJSON string
+
+	// These fields are used to compare bundles in a diff.
+	PropertiesP *property.Properties
+	Version     semver.Version
```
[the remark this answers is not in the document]
estroz: I had that thought originally. You'd need to implement a separate `model.Model` type specifically for diffing along with a `ConvertToDiffModel()` function, which imo is less maintainable. Plus these fields might be useful for other operations in the future like sorting by version.

cmd/opm/alpha/diff/cmd.go
estroz: Nope, this was copied from an earlier version of `render` I think.

cmd/opm/alpha/diff/cmd.go
estroz: Either works. I figured, why not make the args obvious with flags. Also if some wrapper script fails to collect the list of old render targets silently, then you'll get a helpful error rather than a heads-only diff:

```sh
local old_targets=""

# Target-lookup logic that fails silently

# Returns an error
opm alpha diff --old ${old_targets} --new reg/catalog:latest
# Generates a heads-only diff, which is likely not what you want
opm alpha diff ${old_targets} reg/catalog:latest
```

I suppose this leads to a discussion of whether heads-only mode is best exposed as-is, but I think by only using `--old/--new` we keep the number of flags low while being descriptive and exposing both modes intuitively.

cmd/opm/alpha/diff/cmd.go
estroz: SGTM.

cmd/opm/alpha/diff/cmd.go
joelanford: My three main arguments against the flags are:
1) If positional args are good enough for `git diff`, they're probably good enough for `opm diff`
1) In general, non-sqlite based `opm` commands have been shifting to positional args, especially for required arguments, so using positional args would be consistent with the overall CLI design of `opm` going forward.
1) Positional args require more typing.

cmd/opm/alpha/diff/cmd.go
```diff
@@ -0,0 +1,157 @@
...
+
+FROM scratch
+LABEL operators.operatorframework.io.index.configs.v1=/index/index.yaml
+COPY ./my-catalog-index /index
+COPY --from=builder /bin/opm /opm
+COPY --from=builder /bin/grpc_health_probe /bin/grpc_health_probe
+EXPOSE 50051
+ENTRYPOINT ["/opm"]
+CMD ["serve", "/index"]
+EOF
```
joelanford, line 76: This will likely need to change as soon as @ankitathomas's upcoming PR to build DC index images merges.

internal/action/diff.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	newRender := Render{Refs: a.NewRefs, Registry: a.Registry}
+	newCfg, err := newRender.Run(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error rendering new refs: %v", err)
+	}
+	newModel, err := declcfg.ConvertToModel(*newCfg)
+	if err != nil {
+		return nil, fmt.Errorf("error converting new delcarative config to model: %v", err)
```
joelanford, line 49: typo
```suggestion
		return nil, fmt.Errorf("error converting new declarative config to model: %v", err)
```

internal/declcfg/declcfg_to_model.go
joelanford: `Version` on `model.Bundle` seems reasonable, I suppose. I think we would need to add a bundle validation for `model.Bundle.Version == model.Bundle.Properties["olm.package"].Value.Version`

And then always make sure to populate that field when converting everything to the model.

internal/model/model.go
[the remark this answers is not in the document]
joelanford: Would it work if they were unexported? That would at least be a signal that they're derived fields. You'd have to put the function that actually looks at these fields in the model package or have a method on `model.Bundle`, but that doesn't seem too unreasonable.

joelanford (commented):
I didn't realize that you didn't need to make immediate use of `WalkFS`, so feel free to take that back out if you want.

I'm personally fine with it staying in though (I was going to submit a separate PR with just that change).

cmd/opm/alpha/diff/cmd.go
estroz: Sounds good.

internal/declcfg/declcfg_to_model.go
estroz: Mind if I handle this in a follow-up?

internal/model/model.go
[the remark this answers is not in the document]
estroz: Mind if I handle this in a follow-up?

cmd/opm/alpha/diff/cmd.go
estroz: Done.

cmd/opm/alpha/diff/cmd.go
estroz: Updated.

internal/action/diff_test.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+	}
+
+	for _, s := range specs {
+		t.Run(s.name, func(t *testing.T) {
+			actualCfg, actualErr := s.diff.Run(context.Background())
+			s.assertion(t, actualErr)
+			// for _, b := range actualCfg.Bundles {
+			// 	sort.Strings(b.Objects)
+			// }
```
joelanford, line 59: nit: remove extraneous comments.

internal/action/diff_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+//go:embed testdata/index-declcfgs
+var indicesDir embed.FS
+
+func newDiffRegistry() (image.Registry, error) {
+	subDeclcfgImage, err := fs.Sub(indicesDir, "testdata/index-declcfgs")
+	if err != nil {
+		return nil, err
+	}
+	const configsLabel = "operators.operatorframework.io.index.configs.v1"
```
joelanford, line 119: This is now a constant in the `containertools` package.

PR #701, opened 2021-07-08
opm catalog-snapshot: compile and output all installed operators in a cluster

estroz:
**Description of the change:** `opm catalog-snapshot` takes a snapshot of all operators installed in a cluster via their Subscriptions and writes them to a (relatively bare) declarative config.

**Motivation for the change:** useful for creating a diff (#696) between some current and new state.

**Reviewer Checklist**

/kind feature


estroz:
This command is not actually as valuable as I had expected for the listed use case. Closing for now.

PR #707, opened 2021-07-12
declcfg: move dc opm commands out of alpha

ankitathomas:
Move the init, serve, validate and render commands out of alpha


joelanford:
@ankitathomas I think we also need to actually move these to `opm init/render/validate/serve` by moving the setup code from `cmd/opm/alpha/cmd.go` to `cmd/opm/root/cmd.go`

PR #710, opened 2021-07-15
opm: add multi-arch builds with goreleaser

joelanford:
Update the release workflow to do three things:
1. Run on PRs so that we test our image build process (minus actually publishing the images) on every change
3. Build a multi-arch opm container images for all of the linux variants.

**Other notes:**
- Using `gcr.io/distroless/base:debug`, which gives us a very similar base image as alpine + ca-certificates, and enables our piece of the image build to be simply copying files (no `RUN` commands).
- Using a new image name, `quay.io/operator-framework/opm`, which enables existing users to keep using the existing image and migrate at their leisure in case there are substantive differences between alpine and distroless. This (imo) is also just a better name from the perspective of discoverability and obviousness.

**Outputs**
Images (pushed to quay on tags and pushes to master):
- `quay.io/operator-framework/opm:{{ .Tag || "master" }}` (multi-arch manifest list)
- `quay.io/operator-framework/opm:{{ .Tag || "master" }}-amd64`
- `quay.io/operator-framework/opm:{{ .Tag || "master" }}-arm64`
- `quay.io/operator-framework/opm:{{ .Tag || "master" }}-ppc64le`
- `quay.io/operator-framework/opm:{{ .Tag || "master" }}-s390x`

- More repeatable, testable image builds process (image builds happen on every PR), pushes happen on master pushes and tags.
- Support for more platforms
- Better visibility when an image build job fails (it will happen in Github Actions rather than in quay's image builds)

**Follow-ups**
1. Figure out a (legal) way to compile darwin binaries via github actions, ideally via GoReleaser.
1. Uncomment `.github/workflows/goreleaser.yaml` osxcross steps (if permissible)
1. Uncomment `.goreleaser.yaml` darwin builds
1. Remove `.goreleaser.yaml`'s `.release.disable=true` configuration to enable GoReleaser's release steps
1. Delete `.github/workflows/release.yaml`


.github/workflows/release.yaml
```diff
@@ -3,46 +3,43 @@ on:
...
       - run: |
-          echo "asset_path=bin/opm" >> $GITHUB_ENV
-          echo "asset_name=$(go env GOOS)-$(go env GOARCH)-opm$(go env GOEXE)" >> $GITHUB_ENV
-      - run: make ${{ env.asset_path }}
-      - uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
+          git clone https://github.com/tpoechtrager/osxcross /tmp/osxcross
+          cd /tmp/osxcross
+          wget -P tarballs 'https://www.dropbox.com/s/cklcenl4667il6b/MacOSX11.1.sdk.tar.xz'
```
joelanford, line 25: I built and uploaded this file to dropbox. Downloading the SDK from Apple directly requires a logged in Apple developer account, which seems out of the question for a GH action.

timflannagan: I'm just trying to wrap my head around this as I'm not a frequent mac user - is there any reason we couldn't just dedicate a job here to installing this macos sdk that `runs-on: macos-<some tag>`? I poked around the osxcross library real quick, but it looked like there's a script housed in that repository that should generate the sdk: https://github.com/tpoechtrager/osxcross/blob/master/tools/gen_sdk_package.sh. I need to dive deeper into that script and it might be beneficial to do a quick github search to see how others are handling the installment of this sdk.

timflannagan:
These changes seem reasonable and work towards a much needed improvement in the release process. I'm not sure familiar with the gorelease tooling so other reviewers may need to be more involved in reviewing that aspect.

/lgtm

.github/workflows/release.yaml
joelanford: I think it would be possible to run osxcross's `gen_sdk_package.sh` on a macos github host and upload the resulting package via a `actions/upload-artifact@v2` step

Then in the linux job, we could download that artifact instead of doing the wget to dropbox. I _think_ caching would just work the same on the linux host, and we'd probably want to look into whether its possible to also cache the generated package on the macos host.

.github/workflows/release.yaml
```diff
@@ -1,48 +1,94 @@
...
-      matrix:
-        os:
-          - ubuntu-latest
-          - macos-latest
-          - windows-latest
-    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
timflannagan, line 18: Any value in adding a comment here as to why this is necessary? It looks like the [goreleaser workflow example](https://github.com/goreleaser/goreleaser-action#workflow) calls out this is necessary to have changelogs work correctly, but it wasn't immediately obvious to me and I had to go poke around that linked README.md.

.github/workflows/release.yaml
timflannagan: Yeah I think that all sounds reasonable to me. I'm fine with keeping this implementation for now, but I'd like to move away from housing a custom dropbox file as using something more native is likely cleaner and easier to maintain going forward. I'm not super familiar with how caching works at a workflow vs. job level, but I'd assume that all jobs defined in an individual workflow share a local cache despite jobs using different machine os'?

opm.Dockerfile
```diff
@@ -0,0 +1,8 @@
+# NOTE: This Dockerfile is used in conjuction with GoReleaser to
+#   build opm images. See the configurations in .goreleaser.yaml
+#   and .github/workflows/release.yaml.
+
+FROM gcr.io/distroless/base:debug
+COPY ["nsswitch.conf", "/etc/nsswitch.conf"]
+COPY opm /bin/opm
```
timflannagan, line 7: I'm not super familiar with the current role of the upstream-opm-builder.Dockerfile - it looks like we're blindly copying a local version of opm without building it, which would be problematic outside of the changes made here to include it's usage in the release workflow. Is anything/anyone depending on that dockerfile to build opm in an overall multi-stage image?

.github/workflows/release.yaml
```diff
@@ -1,48 +1,94 @@
...
+
+      - name: "Login to Quay"
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
         with:
-          upload_url: ${{ needs.create.outputs.upload_url }}
-          asset_path: ${{ env.asset_path }}
-          asset_name: ${{ env.asset_name }}
-          asset_content_type: application/octet-stream
+          username: operatorlifecyclemanager
```
timflannagan, line 85: Should this key also be stored as a secret?

opm.Dockerfile
estroz: operator-sdk is effectively [dependent](https://github.com/operator-framework/operator-sdk/blob/1be35bedc4356d8a804ba8258e455cb9a3f4e419/internal/olm/operator/registry/index_image.go#L43) on it, but we're flexible.

estroz (commented):
Aside from cred username being public, LGTM

opm.Dockerfile
joelanford: > problematic outside of the changes made here to include it's usage in the release workflow

Agreed. WDYT about moving/renaming this new `opm.Dockerfile` to something like `release/goreleaser.opm.Dockerfile` to make it more obvious what context this is intended to be used in.

.github/workflows/release.yaml
joelanford: If so, we'd probably also want a different user name at this point since this one is now public.

@kevinrizza thoughts?

timflannagan: Yeah I imagine we'd need to change the username.

opm.Dockerfile
timflannagan: Moving this to a release directory and retaining the old Dockerfile sounds reasonable to me - as an aside, I wasn't able to immediately find any documentation around the various Dockerfile(s) that are housed in this repository, which is unfortunate if you don't have much historical context.

.github/workflows/release.yaml
joelanford: Sounds good. I vote to merge this PR without this change and do this work in a follow-up. I wouldn't be surprised to see some other minor issues pop up after merge since there's some logic in the workflow file that is hard to simulate locally.

estroz (approved):
/lgtm

.github/workflows/release.yaml
ecordell: /hold while we find out if this is legal

joelanford: For now, I removed the darwin build targets, disabled GoReleaser's GitHub release steps, and re-introduces our existing github release workflow.

So the current state of this PR is just the addition of GoReleaser multi-arch image builds/pushes.

release/goreleaser.opm.Dockerfile
```diff
@@ -0,0 +1,8 @@
+# NOTE: This Dockerfile is used in conjuction with GoReleaser to
+#   build opm images. See the configurations in .goreleaser.yaml
+#   and .github/workflows/release.yaml.
+
+FROM gcr.io/distroless/base:debug
+COPY ["nsswitch.conf", "/etc/nsswitch.conf"]
+COPY opm /bin/opm
```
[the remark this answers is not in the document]
timflannagan: I meant to double-check that same thing when I was reviewing this PR last week - my limited understanding is that it's namely there for the registry-server and configmap-server container images. @njhale Maybe you'd have an idea?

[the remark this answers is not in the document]
joelanford: Just checked -- it is required:

```console
$ k get pods redhat-operators-p5xm4 -o json | jq '.spec.containers[0].readinessProbe.exec.command[0]'
"grpc_health_probe"
```

.github/workflows/goreleaser.yaml
```diff
@@ -0,0 +1,96 @@
...
+#          key: ${{ runner.os }}-osxcross-${{ hashFiles('/tmp/osxcross/**/*') }}
+#          restore-keys: |
+#            ${{ runner.os }}-osxcross-
+#
+#      - name: "Build osxcross"
+#        if: steps.cache-osxcross.outputs.cache-hit != 'true'
+#        run: |
+#          cd /tmp/osxcross
+#          sudo ./tools/get_dependencies.sh
+#          UNATTENDED=1 ./build.sh
```
njhale, line 47: for cleanliness, could we drop disabled config with an additional commit?

(dogfooding some feedback I've dished out to other contributors in the past)

timflannagan: I'd also prefer this approach :+1:

joelanford: We can, but we'll have to merge this PR to keep the commits intact.

/label tide/merge-method-merge

release/goreleaser.opm.Dockerfile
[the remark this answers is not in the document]
joelanford: Alright, grpc_health_probes are added. The fact that a multi-arch grpc_health_probe image exists now made this _very_ easy.

.github/workflows/goreleaser.yaml
joelanford: Done.

PR #712, opened 2021-07-21
generate DC index images

ankitathomas:
The generate command allows opm to create index images serving from a declarative config directory. This requires a base image with the opm and grpc_health_probe binaries in /bin. The command generates manifests for each platform supported by the base image.

**Reviewer Checklist**


pkg/image/containerdregistry/builder/dockerfile.go
```diff
@@ -0,0 +1,87 @@
...
+			generateList[fmt.Sprintf("%s-%s", config.OS, config.Architecture)] = desc.Digest
+		}
+		return nil, nil
+	}), *i.head)
+
+	if err != nil {
+		return err
+	}
+
+	for p, dgst := range generateList {
```
ankitathomas, line 57: this creates a dockerfile for each manifest found, it does not create a manifest list like output oci-archive

cmd/opm/alpha/generate/generate.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := b.ExposePorts(ctx, matcher, "50051"); err != nil {
+		return err
+	}
+	if err := b.Add(ctx, matcher, args[0], "/database"); err != nil {
+		return err
+	}
+	if err := b.SetEntrypoint(ctx, matcher, []string{"/bin/opm"}); err != nil {
+		return err
+	}
+	if err := b.SetEntrypoint(ctx, matcher, []string{"serve", "/database"}); err != nil {
```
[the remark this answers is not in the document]
joelanford: Also, do we need to call `SetEntrypoint` twice? Can we just do:

```go
b.SetEntrypoint(ctx, matcher, []string{"/bin/opm", "serve", "/configs"})
```

cmd/opm/alpha/generate/generate.go
```diff
@@ -0,0 +1,117 @@
...
+		Short: "Generate an oci-archive index image for a declarative config directory",
+		Args:  cobra.MinimumNArgs(1),
+		RunE: generate,
+	}
+	cmd.Flags().StringP("binary-image", "i", "quay.io/joelanford/opm:a8c80ac9", "container image for on-image `opm` command")
+	cmd.Flags().StringP("tag", "t", "", "custom tag for container image being built")
+	cmd.Flags().Bool("generate", false, "generate a dockerfile")
+	cmd.Flags().StringP("out-dockerfile", "d", "", "if generating the dockerfile, this flag is used to (optionally) specify a dockerfile name")
+	cmd.Flags().StringP("output", "o", "", "name of the output archive")
+	cmd.Flags().StringSlice("labels", nil, "list of label key=value pairs to apply on the index image")
```
joelanford, line 26: What do you think about making the CLI a light wrapper around another action API, similar to  the [render command](https://github.com/operator-framework/operator-registry/blob/9a4d16f18f26bf81d3bdbffdba2773481a5a95d1/cmd/opm/render/cmd.go#L18) and [`action.Render`](https://github.com/operator-framework/operator-registry/blob/9a4d16f18f26bf81d3bdbffdba2773481a5a95d1/internal/action/render.go#L52-L56)?

This way we can make a nice Go API for building OCI archives for declarative config indexes.

pkg/image/containerdregistry/builder/actions.go
```diff
@@ -0,0 +1,111 @@
...
+func (i *imageBuilder) SetLabels(ctx context.Context, platformMatcher platforms.MatchComparer, labels map[string]string) error {
+	ctx = ensureNamespace(ctx)
+
+	var action string
+	if len(labels) > 0 {
+		action := "LABEL"
+		for k, v := range labels {
+			action = fmt.Sprintf("%s %s=%s", action, k, v)
+		}
+	}
```
joelanford, line 20: Nit: a `strings.Builder` would be slightly better here because it minimizes memory copying and allocations. This is likely never a memory intensive operation, so no big deal.

cmd/opm/alpha/generate/generate.go
```diff
@@ -0,0 +1,117 @@
+package generate
```
joelanford, line 1: I think my preference would be to call this "build" or "save" instead of "generate" since that aligns more with the docker/podman CLIs that folks are familiar with. WDYT?

joelanford (commented):
First pass review.

Overall this looks great! I'm having a little trouble putting all of the pieces together. Once I run this command, what are the next steps to actually get the image pushed into a registry or loaded into a local runtime?

cmd/opm/alpha/generate/generate.go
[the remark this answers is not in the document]
joelanford: Alternatively, we could have the second `b.SetEntrypoint` call be `b.SetCmd` instead?

cmd/opm/alpha/generate/generate.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"github.com/containerd/containerd/platforms"
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
+	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry/builder"
+	"github.com/spf13/cobra"
+	"strings"
+)
+
+func NewCmd() *cobra.Command {
```
joelanford, line 14: Looks like this needs to be hooked up to the `opm alpha` cmd?

cmd/opm/alpha/generate/generate.go
```diff
@@ -0,0 +1,117 @@
...
+	cmd := &cobra.Command{
+		Use:   "generate <path to declarative config directory>",
+		Short: "Generate an oci-archive index image for a declarative config directory",
+		Args:  cobra.MinimumNArgs(1),
+		RunE: generate,
+	}
+	cmd.Flags().StringP("binary-image", "i", "quay.io/joelanford/opm:a8c80ac9", "container image for on-image `opm` command")
```
joelanford, line 21: This is fine for now, but we'll want to drop in the official image before this merges.
```suggestion
	cmd.Flags().StringP("binary-image", "i", "quay.io/operator-framework/upstream-opm-builder:latest", "container image for on-image `opm` command")
```

pkg/image/containerdregistry/builder/export.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, nil
+	}), desc); err != nil {
+		return err
+	}
+	err = tarWriter.Close()
+	if err != nil {
+		return err
+	}
+
+	f, err := os.OpenFile(filepath.Join(filepath.Dir(dst), fmt.Sprintf("%s.tar", filepath.Base(dst))), os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0644)
```
joelanford, line 129: I think we should users to give us the full filename they want us to write to.
```suggestion
	f, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0644)
```

joelanford: Going even further (along the lines of [this comment](https://github.com/operator-framework/operator-registry/pull/712/files#discussion_r674297332)), what do you think of passing an `io.Writer` into the image builder and using that?

That way the caller gets to decide where to write the output.

If we had an `action.Build` struct, one of the fields could be `ArchiveWriter io.Writer`.

PR #713, opened 2021-07-22
Add command to generate DC dockerfile

joelanford:
Add subcommand `opm alpha generate dockerfile` to generate a dockerfile for declarative config indexes.

Make it easier for users to onboard onto declarative config

**Example**
```sh
mkdir index
opm render quay.io/openshift-community-operators/catalog:latest -o yaml > index/index.yaml
opm alpha generate dockerfile index
docker build -f index.Dockerfile -t quay.io/<myorg>/catalog:latest-dc .
docker run -it --rm -p 50051:50051 quay.io/<myorg>/catalog:latest-dc

# in another terminal
grpcurl -plaintext localhost:50051 api.Registry/ListBundles | head -10
```

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/generate/cmd.go
+++ b/cmd/opm/alpha/generate/cmd.go
@@ -54,7 +54,7 @@ Dockerfile with the same name already exists, this command will fail.`,
 			if s, err := os.Stat(fromDir); err != nil {
 				return err
 			} else if !s.IsDir() {
-				return fmt.Errorf("provided root path %q is not a directory")
+				return fmt.Errorf("provided root path %q is not a directory", fromDir)
 			}
 
 			f, err := os.OpenFile(dockerfilePath, os.O_CREATE|os.O_WRONLY, 0666)
--- a/cmd/opm/alpha/generate/cmd.go
+++ b/cmd/opm/alpha/generate/cmd.go
@@ -15,7 +15,10 @@ import (
 )
 
 func NewCmd() *cobra.Command {
-	cmd := &cobra.Command{Use: "generate"}
+	cmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate various artifacts for declarative config indexes",
+	}
 	cmd.AddCommand(
 		newDockerfileCmd(),
 	)
--- a/cmd/opm/alpha/generate/cmd.go
+++ b/cmd/opm/alpha/generate/cmd.go
@@ -1,0 +1,104 @@
+package generate
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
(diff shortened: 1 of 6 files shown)


cmd/opm/alpha/generate/cmd.go
```diff
@@ -0,0 +1,104 @@
...
+			dockerfilePath := filepath.Join(dir, fmt.Sprintf("%s.Dockerfile", indexName))
+
+			if err := ensureNotExist(dockerfilePath); err != nil {
+				logrus.Fatal(err)
+			}
+
+			if s, err := os.Stat(fromDir); err != nil {
+				return err
+			} else if !s.IsDir() {
+				return fmt.Errorf("provided root path %q is not a directory")
```
timflannagan, line 57: This appears to be missing the corresponding variable to that format specifier.

timflannagan:
Nice, I think this implementation looks reasonable. I'm going to play around with it locally in the background.

cmd/opm/alpha/generate/cmd.go
joelanford: Good catch. Thanks!

internal/action/generate_dockerfile.go
```diff
@@ -0,0 +1,63 @@
...
+# Set DC-specific label for the location of the DC root directory
+# in the image
+LABEL ` + containertools.ConfigsLocationLabel + `=/configs
+{{- if .ExtraLabels }}
+
+# Set other custom labels
+{{- range $key, $value := .ExtraLabels }}
+LABEL "{{ $key }}"="{{ $value }}"
+{{- end }}
+{{- end }}
```
timflannagan, line 62: Just wanted to clarify the behavior here in case there are multiple key/value pairs that are referenced in the `--extra-labels` flag - if I specify duplicate keys but differing values, it looks like the last value in the chain will always be overriden (e.g. `--extra-labels test=test1,test=test2` would produce a `LABEL test=test2` entry), which seems like the correct behavior. I wasn't sure whether that kind of edge case would belong in the validation logic though.

joelanford: I'll make a note in the help text that the last label wins if there are duplicates.

ankitathomas (approved):
/lgtm

timflannagan:
The only real minor detail I saw locally was the there's no short description for this new generate command:

```
$ opm alpha --help
Run an alpha subcommand

Usage:
  opm alpha [command]

Available Commands:
  bundle      Operator bundle commands
  generate    
  list        List contents of an index
```

I'm not sure there's even a decent description given only the dockerfile subcommand exists today.

joelanford:
I'll add one a short description for generate, even it it doesn't really say much. It just looks weird without out.

PR #706, opened 2021-07-12
Add list subcommands for packages, channels, and bundles

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

**Description of the change:**
Add `opm alpha list packages|channels|bundles` subcommands

Examples:

`opm alpha list packages quay.io/openshift-community-operators/catalog:latest | head -10` 
```
NAME                                       DISPLAY NAME                                       DEFAULT CHANNEL
3scale-community-operator                  3scale API Management                              threescale-2.10
akka-cluster-operator                      Akka Cluster Operator                              alpha
alvearie-imaging-ingestion                 Alvearie Imaging Ingestion Operator                alpha
api-operator                               API Operator for Kubernetes                        2.x-stable
apicast-community-operator                 APIcast                                            stable
apicurio-registry                          Apicurio Registry Operator                         2.x
apicurito                                  Apicurito Operator                                 alpha
aqua                                       Aqua Security Operator                             6.2.0
argocd-operator                            Argo CD                                            alpha
```

`opm alpha list channels quay.io/openshift-community-operators/catalog:latest etcd`
```
PACKAGE  CHANNEL                HEAD
etcd     alpha                  etcdoperator.v0.6.1
etcd     clusterwide-alpha      etcdoperator.v0.9.4-clusterwide
etcd     singlenamespace-alpha  etcdoperator.v0.9.4
```

`opm alpha list bundles quay.io/openshift-community-operators/catalog:latest etcd`
```
PACKAGE  CHANNEL                BUNDLE                           REPLACES                         SKIPS  SKIP RANGE  IMAGE
etcd     alpha                  etcdoperator.v0.6.1                                                                  quay.io/openshift-community-operators/etcd:v0.6.1
etcd     clusterwide-alpha      etcdoperator.v0.9.0                                                                  quay.io/openshift-community-operators/etcd:v0.9.0
etcd     clusterwide-alpha      etcdoperator.v0.9.2-clusterwide  etcdoperator.v0.9.0                                 quay.io/openshift-community-operators/etcd:v0.9.2-clusterwide
etcd     clusterwide-alpha      etcdoperator.v0.9.4-clusterwide  etcdoperator.v0.9.2-clusterwide                     quay.io/openshift-community-operators/etcd:v0.9.4-clusterwide
etcd     singlenamespace-alpha  etcdoperator.v0.9.0                                                                  quay.io/openshift-community-operators/etcd:v0.9.0
[...]

PR #714, opened 2021-07-22
bug(declcfg): load and render both sort properties and bundle objects

estroz:
**Description of the change:** sort both properties and bundle objects.

**Motivation for the change:** I noticed that property and bundle object order can get out of sync depending on when objects are loaded from properties and whether a config was loaded or converted. This follows up on #706.

**Reviewer Checklist**

/cc @joelanford


estroz:
Fixing order issues will be handled elsewise.

PR #718, opened 2021-07-27
Prune operator version

monkey1016:
Introduces a new command, `prune-version` which allows the catalog to be pruned similar to the original `prune` command , but instead keeps specific versions of operators instead of all versions for a given operator. It was added as a hidden command since it is a command that could cause instability if done incorrectly, for example breaking upgrade paths if a specific version required in the upgrade path is not added to the list.

We've run into situations many times where we don't need every version of a given operator, only the latest or the second latest (or maybe even slightly older) operator on a given day due to how approvals for application versions work for the organization. The current `prune` command pulls in every version of the operator which causes us to mirror a large number of images we don't care about. Just three operators would mirror more than 3500 images, even though we only care about one or two versions for each operator.

**Reviewer Checklist**


estroz:
@monkey1016 have you seen #696 and #716? The latter effectively implements the functionality you want in this PR, with channel support as well; does it cover your goals?

monkey1016:
@estroz I have not. I'll take a look at them and cancel this pull request. Thanks!

PR #720, opened 2021-07-28
Aggregate properties/dependencies in-query for ListBundles.

benluddy:
This fixes the result set to contain one row per result Bundle instead
of potentially several, which should allow for gRPC result streaming.

### DIFF
--- a/pkg/sqlite/query.go
+++ b/pkg/sqlite/query.go
@@ -970,6 +970,16 @@ tip (depth) AS (
       INNER JOIN channel_entry AS skipped_entry
         ON skips_entry.skips = skipped_entry.entry_id
     GROUP BY all_entry.operatorbundle_name, all_entry.package_name, all_entry.channel_name
+),
+merged_properties (bundle_name, merged) AS (
+  SELECT operatorbundle_name, json_group_array(json_object('type', properties.type, 'value', properties.value))
+    FROM properties
+    GROUP BY operatorbundle_name
+),
+merged_dependencies (bundle_name, merged) AS (
+  SELECT operatorbundle_name, json_group_array(json_object('type', dependencies.type, 'value', CAST(dependencies.value AS TEXT)))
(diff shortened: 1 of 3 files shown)


pkg/sqlite/query.go
```diff
@@ -970,6 +970,16 @@ tip (depth) AS (
       INNER JOIN channel_entry AS skipped_entry
         ON skips_entry.skips = skipped_entry.entry_id
     GROUP BY all_entry.operatorbundle_name, all_entry.package_name, all_entry.channel_name
+),
+merged_properties (bundle_name, merged) AS (
+  SELECT operatorbundle_name, json_group_array(json_object('type', properties.type, 'value', properties.value))
```
awgreene, line 975: I'n the [existing code](https://github.com/operator-framework/operator-registry/pull/720/files#diff-5d7535c9d61e38ec8f7cd26e7746b174a76b808fb64e098679f584c2a0986f34L1035), properties / dependencies are only merged by cases of matching name + version + path + channel name, it seems as though this change only considers bundle name. Is there a reason to be concerned that this may change existing user behavior?

pkg/sqlite/query.go
```diff
@@ -970,6 +970,16 @@ tip (depth) AS (
...
+    FROM properties
+    GROUP BY operatorbundle_name
+),
+merged_dependencies (bundle_name, merged) AS (
```
awgreene, line 979: Same as above.

awgreene (changes requested):
One question but it looks good overall.

pkg/sqlite/query.go
benluddy: `name` is the full primary key of `operatorbundle`, so there can't be two rows in that table that have the same name but different path/version. Channel is already a nonissue because each result row is pretty much hung off of a single row in `channel_entry`.

PR #723, opened 2021-07-28
Update Registry to stream bundles

awgreene:


awgreene:
/hold
Waiting on #720

awgreene:
/hold

I need to run this locally and ensure that bundles are actually returned as they become available.

pkg/streamer/streamer.go
```diff
@@ -0,0 +1,7 @@
+package streamer
+
+import "github.com/operator-framework/operator-registry/pkg/api"
+
+type Stream interface {
+	Send(*api.Bundle) error
+}
```
benluddy, line 7:
```suggestion
type BundleSender interface {
	Send(*api.Bundle) error
}
```

I'm sure this is debatable, but I would like to see the interface name be oriented to the behavior (sending Bundles somewhere). Stream is similar, but it invites confusion with gRPC "streams" and doesn't specify direction (can I receive from a Stream?).

pkg/streamer/slicestreamer.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+type sliceStreamer struct {
+	Slice []*api.Bundle
+}
+
+func (s *sliceStreamer) Send(b *api.Bundle) error {
+	s.Slice = append(s.Slice, b)
+	return nil
+}
```
benluddy, line 18: Could this avoid the struct completely:

```suggestion
type sliceStreamer []*api.Bundle
```

And let users do things like:

```
DoThing(sliceStreamer(bundles))
```

without needing to have a `New...` function?

benluddy (commented):
Looks good. I have a few comments but they're all style focused.

pkg/registry/query.go
```diff
@@ -30,20 +31,29 @@ func (q Querier) ListPackages(_ context.Context) ([]string, error) {
 }
 
 func (q Querier) ListBundles(_ context.Context) ([]*api.Bundle, error) {
-	var bundles []*api.Bundle
+	sliceStreamer := streamer.NewSliceStreamer()
 
+	err := q.StreamBundles(context.TODO(), sliceStreamer)
```
joelanford, line 36: Assign a name to the `ListBundles()` `context.Context` argument and use that instead of `context.TODO()`

pkg/streamer/streamer.go
```diff
@@ -0,0 +1,7 @@
+package streamer
+
+import "github.com/operator-framework/operator-registry/pkg/api"
+
+type Stream interface {
+	Send(*api.Bundle) error
```
joelanford, line 6: Does this need a `context.Context` parameter?

joelanford: Nevermind, I see this is derived from the generated GRPC server code.

pkg/registry/query.go
```diff
@@ -30,20 +31,29 @@ func (q Querier) ListPackages(_ context.Context) ([]string, error) {
...
 	for _, pkg := range q.pkgs {
 		for _, ch := range pkg.Channels {
 			for _, b := range ch.Bundles {
 				apiBundle, err := api.ConvertModelBundleToAPIBundle(*b)
 				if err != nil {
-					return nil, fmt.Errorf("convert bundle %q: %v", b.Name, err)
+					return fmt.Errorf("convert bundle %q: %v", b.Name, err)
 				}
-				bundles = append(bundles, apiBundle)
+				s.Send(apiBundle)
```
joelanford, line 52: Does this return an error that should be checked?

Also, do we need to check whether the context is `Done()` before each `Send()`? 

It seems like the GRPC server has it's own context that will be plumbed into this function, so I'm not really sure if `s.Send` will handle a context that has been cancelled. It's odd to me that this API does not accept a context directly.

pkg/registry/interface.go
```diff
@@ -21,6 +22,9 @@ type GRPCQuery interface {
 	// List all available package names in the index
 	ListPackages(ctx context.Context) ([]string, error)
 
+	// Streams all available bundles in the index
+	StreamBundles(ctx context.Context, stream streamer.Stream) error
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the PR.

pkg/registry/interface.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/operator-framework/operator-registry/pkg/api"
+	"github.com/operator-framework/operator-registry/pkg/streamer"
```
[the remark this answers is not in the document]
awgreene: Addressed in the latest version of the PR.

pkg/streamer/streamer.go
awgreene: Addressed in the latest version of the PR.

pkg/registry/query.go
awgreene: Addressed in the latest version of the PR.

pkg/registry/query.go
benluddy: +1, the error needs to be checked.

>  do we need to check whether the context is Done() before each Send()?

My (unverified) assumption is that Send returns an error (e.g. ContextCancelled) in these cases and can be treated as an implementation detail.

pkg/streamer/slicestreamer.go
awgreene: Resolved in the latest commit.

pkg/server/server.go
```diff
@@ -33,15 +33,10 @@ func (s *RegistryServer) ListPackages(req *api.ListPackageRequest, stream api.Re
 }
 
 func (s *RegistryServer) ListBundles(req *api.ListBundlesRequest, stream api.Registry_ListBundlesServer) error {
-	bundles, err := s.store.ListBundles(stream.Context())
+	err := s.store.SendBundles(stream.Context(), stream)
```
benluddy, line 36:
```suggestion
	return s.store.SendBundles(stream.Context(), stream)
```

benluddy:
/approve

@joelanford can you be sold on https://github.com/operator-framework/operator-registry/pull/723#discussion_r678619932?

pkg/registry/query.go
joelanford: Yeah I was trying to find the actual implementation of that `Send` call, but I quickly hit a dead end trying to track that down through generated code and interfaces.

I think what I'm slightly worried about is that the context that is passed in isn't necessarily tied at all to the stream's context. 

It looks like `Send()` blocks in various situations, so this is probably not really a solvable problem for us. Even if we check our context's `Done()` before every `Send()`, `Send()` can block forever and never let us find out we later have a cancelled context.

TL;DR: I think its fine to just handle the error and not worry about the context that is passed in.

benluddy: > I think what I'm slightly worried about is that the context that is passed in isn't necessarily tied at all to the stream's context.

That sounds accurate to me, but IMO that's up to the caller, not the Querier implementation. In the RegistryServer case, we know that the contexts should be linked (identical) because we're passing `stream.Context()` to ListBundles, but it definitely looks possible to use this in such a way that you could cancel the ListBundles Context and go on to hang indefinitely on a Sender implementation that does not respect that cancellation.

PR #724, opened 2021-07-28
Set legacy API fields in ListBundles using available properties.

benluddy:
For each bundle in its result, ListBundles has been making three
separate database queries -- for properties, dependencies, and plural
API names -- via GetApisForEntry in order to populate the bundle
fields providedApis and requiredApis. Properties and dependencies are
already in hand and can be reused. Setting plural names on the
elements of providedApis and requiredApis is pointless because plurals
are discarded by the catalog operator at runtime.

### DIFF
--- a/pkg/sqlite/query.go
+++ b/pkg/sqlite/query.go
@@ -1056,22 +1056,12 @@ func (s *SQLQuerier) ListBundles(ctx context.Context) ([]*api.Bundle, error) {
 			out.Skips = strings.Split(skips.String, ",")
 		}
 
-		provided, required, err := s.GetApisForEntry(ctx, entryID.Int64)
-		if err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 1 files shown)


benluddy:
@joelanford: Can we drop https://github.com/operator-framework/operator-registry/blob/fd099516ef679bd14d53212f40b4dae8f61ca8b7/pkg/api/api_to_model.go#L97-L116? Both fields are populated from the `properties` and `dependencies` tables as the source of truth, so there should never be additional information there.

joelanford:
@benluddy Yeah I don't see why we couldn't drop that if those fields are (and have been) populated from the same place. In that case, it seems like we could drop the `requiredGVKs` and `providedGVKs` maps and just directly append those properties to `out`.

ecordell (commented):
/lgtm

PR #725, opened 2021-07-28
Omit manifests from ListBundles result if bundle image is available.

benluddy:
The "object" and "csvJson" fields of bundles returned from ListBundles
is only consumed by the catalog operator in cases where no bundle
image reference is present.

### DIFF
--- a/cmd/opm/registry/serve.go
+++ b/cmd/opm/registry/serve.go
@@ -95,7 +95,7 @@ func serveFunc(cmd *cobra.Command, args []string) error {
 		logger.WithError(err).Warnf("couldn't migrate db")
 	}
 
-	store := sqlite.NewSQLLiteQuerierFromDb(db)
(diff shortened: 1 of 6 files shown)


dinhxuanvu:
This fails the TestRender test (for render command) on unit test.

joelanford:
Do we still need to send manifests on the ListBundles response for packageserver?

benluddy:
> Do we still need to send manifests on the ListBundles response for packageserver?

Only the catalog operator uses ListBundles, and it discards both fields based on the presence/absence of `bundlePath`: https://github.com/operator-framework/operator-lifecycle-manager/blob/7fa42934ea0ce5a5f4c589767ce88b8fac7f6c68/pkg/controller/registry/resolver/operators.go#L293-L300.

benluddy:
/hold

Needs a rebase, plus I'll update the DC-based server to make the behavior consistent with the SQLite-based server.

PR #728, opened 2021-07-29
Set soft_heap_limit for SQLite-based registry servers.

benluddy:
Registry pods receive low request volumes and are not
latency-sensitive, so there's little downside to advising SQLite to
keep its page cache as small as possible.

### DIFF
--- a/cmd/opm/registry/serve.go
+++ b/cmd/opm/registry/serve.go
@@ -90,6 +90,10 @@ func serveFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if _, err := db.ExecContext(context.TODO(), `PRAGMA soft_heap_limit=1`); err != nil {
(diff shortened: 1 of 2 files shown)


dinhxuanvu (approved):
/lgtm

cmd/opm/registry/serve.go
```diff
@@ -90,6 +90,10 @@ func serveFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if _, err := db.ExecContext(context.TODO(), `PRAGMA soft_heap_limit=1`); err != nil {
```
awgreene, line 93: In case anyone is unfamiliar with this setting, https://www.sqlite.org/c3ref/hard_heap_limit64.html (includes info on `soft_heap_limit`)

PR #729, opened 2021-07-30
cmd/opm: Fix broken link to validating bundles

timflannagan:
Update the opm alpha bundle validate command and fix the broken link to
the OLM bundle validation documentation.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/bundle/validate.go
+++ b/cmd/opm/alpha/bundle/validate.go
@@ -25,17 +25,18 @@ func newBundleValidateCmd() *cobra.Command {
 		Use:   "validate",
 		Short: "Validate bundle image",
 		Long: `The "opm alpha bundle validate" command will validate a bundle image
-from a remote source to determine if its format and content information are accurate. 
-Required validators. These validators will run by default on every invocation of the command. 
+from a remote source to determine if its format and content information are accurate.
+Required validators. These validators will run by default on every invocation of the command.
  * CSV validator - validates the CSV name and replaces fields.
- * CRD validator - validates the CRDs OpenAPI V3 schema. 
- * Bundle validator - validates the bundle format and annotations.yaml file as well as the optional dependencies.yaml file. 
+ * CRD validator - validates the CRDs OpenAPI V3 schema.
+ * Bundle validator - validates the bundle format and annotations.yaml file as well as the optional dependencies.yaml file.
 
-Optional validators. These validators are disabled by default and can be enabled via the --optional-validators flag. 
- * Operatorhub validator - performs operatorhub.io validation. To validate a bundle using custom categories use with the OPERATOR_BUNDLE_CATEGORIES environmental variable to point to a json-encoded categories file.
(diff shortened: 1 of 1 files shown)


cmd/opm/alpha/bundle/validate.go
```diff
@@ -25,17 +25,18 @@ func newBundleValidateCmd() *cobra.Command {
...
 
-Optional validators. These validators are disabled by default and can be enabled via the --optional-validators flag. 
- * Operatorhub validator - performs operatorhub.io validation. To validate a bundle using custom categories use with the OPERATOR_BUNDLE_CATEGORIES environmental variable to point to a json-encoded categories file.
- * Bundle objects validator - performs validation on resources like PodDisruptionBudgets and PriorityClasses. 
+Optional validators. These validators are disabled by default and can be enabled via the --optional-validators flag.
+ * Operatorhub validator - performs operatorhub.io validation. To validate a bundle using custom categories use with the $OPERATOR_BUNDLE_CATEGORIES environmental variable to point to a json-encoded categories file.
+ * Bundle objects validator - performs validation on resources like PodDisruptionBudgets and PriorityClasses.
 
-See https://olm.operatorframework.io/docs/tasks/validate-package/#validation for more info.`,
+See https://olm.operatorframework.io/docs/tasks/creating-operator-bundle/#validating-your-bundle for more info.
```
timflannagan, line 38: @joelanford I'm curious whether you have an opinion on this - I think we'd like to rely on `opm validate <declarative config>` in the future and remove this bundle validation command entirely. Is that accurate? Would it be better to instead remove this link entirely, mark it as deprecated, and link out to the operator-sdk validation command instead?

joelanford: > I think we'd like to rely on opm validate <declarative config> in the future and remove this bundle validation command entirely. Is that accurate?

The `opm validate` is mainly geared towards validating the index content, not necessarily the bundle.

While there is some overlap between index and bundle validation, there is also some bundle-specific validations that `opm alpha bundle validate` does.

> Would it be better to instead remove this link entirely, mark it as deprecated, and link out to the operator-sdk validation command instead?

I think this is probably deserving of a larger conversation. SDK looks to have more plans related to pluggable bundle validation, so it seems like they're driving that forward more than we are.
That said, I think there's an argument to be made for putting bundle validation in opm, but I'm not sure I'm the person to answer/decide that. Maybe @dmesser and/or @tlwu2013 have some thoughts?

If bundle validation _is_ going to stay in `opm`, I'd vote to move it into the `opm validate` command and have `opm validate` automatically detect what kind of artifact was passed into it to do the appropriate validation.

timflannagan: @joelanford What's more favorable in the short term then, updating this link so it at least resolves, or removing it entirely? I don't think I have a strong opinion on either front, but having a broken link in a command seems like it's worth fixing anyways.

joelanford: I'd vote to update it to a the new URL, I think, but I also don't have a strong opinion.

PR #730, opened 2021-07-30
.github/workflows: Replace the e2e-kind job name with e2e in the test workflow

timflannagan:
Update the test.yml test workflow and replace the `e2e-kind` job name with
`e2e`. Before, the e2e test workflow had ran e2e tests against both kind
and minikube clusters before the latter job was removed entirely.

**Reviewer Checklist**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,22 +8,25 @@ on:
       - '**'
       - '!doc/**'
 jobs:
-  e2e-kind:
+  e2e:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


timflannagan:
This has been annoying me for a bit - I'm happy with closing this out if we don't want to head in this direction. This is technically a breaking change in the context of tide. I'm also happy to tackle this update on the tide side-of-things as well if we're comfortable with this change.

joelanford:
/lgtm

> This has been annoying me for a bit

Me too 🙏 🪓

timflannagan:
Likely want to merge https://github.com/openshift/release/pull/20871 first before merging this to ensure little downtime.

PR #731, opened 2021-08-02
internal/action: Fix opm render outputting empty related image names

timflannagan:
Update internal/action/render.go and ensure the RelatedImage.Name key is
being populated when converting a sqlite-based index image to a
declarative config one.

When constructing a declarative config representation from a
sqlite-based index image, we extract the related images properties from
the related_images table name. When extracting that data and building up
a list of related images for the rendered declarative config, we neglect
to populate the RelatedImage.Name field when iterating over processed
row results.

This can lead to situations where an individual bundle's relatedImages
properties have an empty "name" key, but a populated image key:

```json
    "relatedImages": [
        {
            "name": "",
            "image": "quay.io/olmqe/cockroachdb-operator:5.0.3-2199"
        },
        {
            "name": "",
            "image": "quay.io/helmoperators/cockroachdb:v5.0.3"
        },
        {
            "name": "",
            "image": "gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0"
        }
```

Reference to BZ tracking this: https://bugzilla.redhat.com/show_bug.cgi?id=1988948.

**Reviewer Checklist**

### DIFF
--- a/internal/action/render.go
+++ b/internal/action/render.go
@@ -111,17 +111,16 @@ func (r Render) renderReference(ctx context.Context, ref string) (*declcfg.Decla
 				return nil, fmt.Errorf("cannot render DC directory: %w", &ErrNotAllowed{})
 			}
 			return declcfg.LoadFS(os.DirFS(ref))
-		} else {
-			// The only supported file type is an sqlite DB file,
-			// since declarative configs will be in a directory.
-			if err := checkDBFile(ref); err != nil {
-				return nil, err
-			}
-			if !r.AllowedRefMask.Allowed(RefSqliteFile) {
-				return nil, fmt.Errorf("cannot render sqlite file: %w", &ErrNotAllowed{})
-			}
-			return sqliteToDeclcfg(ctx, ref)
 		}
+		// The only supported file type is an sqlite DB file,
+		// since declarative configs will be in a directory.
+		if err := checkDBFile(ref); err != nil {
+			return nil, err
+		}
+		if !r.AllowedRefMask.Allowed(RefSqliteFile) {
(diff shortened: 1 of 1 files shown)


internal/action/render.go
```diff
@@ -272,7 +271,7 @@ func populateDBRelatedImages(ctx context.Context, cfg *declcfg.DeclarativeConfig
 			}
 		}
 		for ri := range ris {
-			cfg.Bundles[i].RelatedImages = append(cfg.Bundles[i].RelatedImages, declcfg.RelatedImage{Image: ri})
+			cfg.Bundles[i].RelatedImages = append(cfg.Bundles[i].RelatedImages, declcfg.RelatedImage{Name: b.Name, Image: ri})
```
timflannagan, line 274: This technically will change the ordering of related image contents in the rendered version of the index image as the underlying type is a set representation, and it looks like that ordering is based on the name key when playing around with this locally.

timflannagan:
Hmm a cursory look at the unit test failures almost reads like this was expected behavior?

```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -139,3 +139,3 @@
        	            	     (declcfg.RelatedImage) {
        	            	-     Name: (string) "",
        	            	+     Name: (string) (len=10) "foo.v0.1.0",
        	            	      Image: (string) (len=44) "test.registry/foo-operator/foo-bundle:v0.1.0"
        	            	@@ -297,3 +297,3 @@
        	            	     (declcfg.RelatedImage) {
        	            	-     Name: (string) "",
        	            	+     Name: (string) (len=10) "foo.v0.2.0",
        	            	      Image: (string) (len=44) "test.registry/foo-operator/foo-bundle:v0.2.0"
```

PR #732, opened 2021-08-02
check for pruned bundles on add in replaces mode

ankitathomas:
In replaces mode, opm uses the bundle with the highest semver in each channel to select a channel head. The edges for a package's upgrade graph are added by following the replaces chains from the channel head till the end of the chain. However, this does not process any bundle that has the highest semver bundle as a member along its replaces chain. This can cause bundles to be pruned silently. This PR helps opm raise an error on detecting any pruning of added bundles.

### DIFF
--- a/cmd/opm/index/add.go
+++ b/cmd/opm/index/add.go
@@ -18,7 +18,13 @@ var (
 
 		This command will add the given set of bundle images (specified by the --bundles option) to an index image (provided by the --from-index option).
 
-		If multiple bundles are given with '--mode=replaces' (the default), bundles are added to the index by order of ascending (semver) version unless the update graph specified by replaces requires a different input order; e.g. 1.0.0 replaces 1.0.1 would result in [1.0.1, 1.0.0] instead of the [1.0.0, 1.0.1] normally expected of semver. However, for most cases (e.g. 1.0.1 replaces 1.0.0) the bundle with the highest version is used to set the default channel of the related package.
+		If multiple bundles are given with '--mode=replaces' (the default), bundles are added to the index by order of ascending (semver) version unless the update graph specified by replaces requires a different input order; e.g. 1.0.0 replaces 1.0.1 would result in [1.0.1, 1.0.0] instead of the [1.0.0, 1.0.1] normally expected of semver. However, for most cases (e.g. 1.0.1 replaces 1.0.0) the bundle with the highest version is used to set the default channel of the related package.  
+
+		Caveat: in replaces mode, the head of a channel is always the bundle with the highest semver. Any bundles upgrading from this channel-head will be pruned.  
+		An upgrade graph that looks like:  
+		0.1.1 -> 0.1.2 -> 0.1.2-1  
+		will be pruned on add to:  
+		0.1.1 -> 0.1.2
 	`)
 
 	addExample = templates.Examples(`
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -196,7 +196,16 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 	}
 
 	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap, overwriteImageMap, overwrite)
-	return populator.Populate(mode)
+	if err := populator.Populate(mode); err != nil {
+		return err
+	}
+
+	for _, imgMap := range overwriteImageMap {
+		for to, from := range imgMap {
+			unpackedImageMap[to] = from
+		}
+	}
+	return checkForBundles(ctx, querier.(*sqlite.SQLQuerier), graphLoader, unpackedImageMap)
 }
 
 type DeleteFromRegistryRequest struct {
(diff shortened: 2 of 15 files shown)


pkg/lib/registry/registry.go
```diff
@@ -402,3 +411,80 @@ func checkForBundlePaths(querier registry.GRPCQuery, bundlePaths []string) ([]st
...
+					}
+					var deprecated bool
+					for _, prop := range props {
+						if prop.Type == registry.DeprecatedType {
+							deprecated = true
+							break
+						}
+					}
+					if !deprecated {
+						errs = append(errs, fmt.Errorf("added bundle %s pruned from package %s, channel %s: verify upgrade paths to ensure channel head has the highest semver", bundle.BundlePath, pkg, channel))
```
kevinrizza, line 483: This logic seems reasonable, but should this error message be slightly more generic? This method essentially checks "did all the bundles your command attempted to add actually get added?" -- and I suspect there are some other edge cases aside from the highest semver picking that could be included in that case.

kevinrizza (approved):
/approve

Just one small nit about the error message we're propogating.

pkg/lib/registry/registry.go
ankitathomas: Switching to having a caveat in the command help text and a less specific error. The reason is now listed as a potential cause instead

cmd/opm/index/add.go
```diff
@@ -19,6 +19,12 @@ var (
 		This command will add the given set of bundle images (specified by the --bundles option) to an index image (provided by the --from-index option).
 
 		If multiple bundles are given with '--mode=replaces' (the default), bundles are added to the index by order of ascending (semver) version unless the update graph specified by replaces requires a different input order; e.g. 1.0.0 replaces 1.0.1 would result in [1.0.1, 1.0.0] instead of the [1.0.0, 1.0.1] normally expected of semver. However, for most cases (e.g. 1.0.1 replaces 1.0.0) the bundle with the highest version is used to set the default channel of the related package.
+
+        Caveat: in replaces mode, the head of a channel is always the bundle with the highest semver. Any bundles upgrading from this channel-head will be pruned.
+		An upgrade graph that looks like:
+		0.1.1 -> 0.1.2 -> 0.1.2-1
+		will be pruned on add to:
+		0.1.1 -> 0.1.2
```
timflannagan, line 27: nit: This renders somewhat strangely when I was playing around with this locally:

```bash
$ ./bin/opm index add --help
Add operator bundles to an index.

 This command will add the given set of bundle images (specified by the --bundles option) to an index image (provided by the --from-index option).

 If multiple bundles are given with '--mode=replaces' (the default), bundles are added to the index by order of ascending (semver) version unless the update graph specified by replaces requires a different input order; e.g. 1.0.0 replaces 1.0.1 would result in [1.0.1, 1.0.0] instead of the [1.0.0, 1.0.1] normally expected of semver. However, for most cases (e.g. 1.0.1 replaces 1.0.0) the bundle with the highest version is used to set the default channel of the related package.

    Caveat: in replaces mode, the head of a channel is always the bundle with the highest semver. Any bundles upgrading from this channel-head will be pruned.
  
 An upgrade graph that looks like: 0.1.1 -> 0.1.2 -> 0.1.2-1 will be pruned on add to: 0.1.1 -> 0.1.2
```

timflannagan: (nothing blocking of course.)

pkg/lib/registry/registry.go
```diff
@@ -402,3 +411,80 @@ func checkForBundlePaths(querier registry.GRPCQuery, bundlePaths []string) ([]st
...
+		}
+		for _, c := range b.Bundle.Channels {
+			if _, ok := foundEntries[b.Bundle.Package][c]; !ok {
+				foundEntries[b.Bundle.Package][c] = map[registry.BundleKey]bool{}
+			}
+			foundEntries[b.Bundle.Package][c][key] = false
+		}
+	}
+
+	for pkg, channels := range foundEntries {
```
joelanford, line 452: same comment as above. if there's a reasonable way to decrease the overall complexity of this large function (i.e. fewer conditionals and nested blocks per function) by breaking it into smaller functions, it'll be easier to comprehend and maintain.

joelanford (commented):
I'm not sure I _completely_ follow the logic of how this is checking the bundles in the DB. I naively thought I'd find something like:
```
for _, b := range q.ListBundles() {
    if !existsInAddedBundles(b) {
		// not found error
    }
}
```

But I'm not nearly experienced enough with the sqlite stuff to know whether that's reasonable.

cmd/opm/index/add.go
ankitathomas: I formatted the heredoc wrong, corrected that

pkg/lib/registry/registry_test.go
```diff
@@ -241,3 +246,156 @@ func TestUnpackImage(t *testing.T) {
...
+
+	return db, func() {
+		defer func() {
+			if err := os.Remove(dbName); err != nil {
+				t.Fatal(err)
+			}
+		}()
+		if err := db.Close(); err != nil {
+			t.Fatal(err)
+		}
```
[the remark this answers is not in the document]
ankitathomas: We should still be removing the db even if we fail to close it, the deferred func will run regardless of whether we hit a db close error.

timflannagan:
(respinning everything as I just renamed the e2e-kind workflow to e2e)

dinhxuanvu (approved):
/lgtm

PR #733, opened 2021-08-03
internal/declcfg: Avoid processing the .indexignore file when walking the root fs

timflannagan:
Update internal/declcfg/load.go implementation and avoid processing any files that are named `.indexignore` when walking the declarative config index root filesystem. When validating declarative config directories, the .indexignore file was being processed and validated.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1989398.

### DIFF
--- a/internal/declcfg/load.go
+++ b/internal/declcfg/load.go
@@ -17,6 +17,10 @@ import (
 	"github.com/operator-framework/operator-registry/internal/property"
 )
 
+const (
+	indexIgnoreFilename = ".indexignore"
+)
+
 type WalkFunc func(path string, cfg *DeclarativeConfig, err error) error
 
 // WalkFS walks root using a gitignore-style filename matcher to skip files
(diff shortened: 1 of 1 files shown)


joelanford:
This is _probably_ reasonable, but another solution could be to ignore these files using a pattern _in_ the `.indexignore` file.

The canonical example I use for `.indexignore` is to explicitly ignore everything except non-k8s-object json/yaml files:
```gitignore
**/*
!**/*.yaml
!**/*.json
**/objects/*
**/objects/*
```

But I'm good with just hardcoding `.indexignore` to be ignored since there's no use case to process it as DC.

/lgtm

joelanford:
We need a similar change in `oc adm catalog mirror` [here](https://github.com/openshift/oc/blob/1557476e1d927004f3392179859f027b8477fa5f/pkg/cli/admin/catalog/mirror.go#L475).

timflannagan:
Yeah, I had the same thought locally and it wasn't immediately clear to me what the best UX is here. I'm under the impression that we should avoid validating the .indexignore file, as I don't think we need to be validating individual patterns outlined in that file, so misconfiguring the .indexignore file would fall under user-error. That said, I think you can make the argument that using a more general pattern (i.e. only validate YAML/JSON files and ignore everything else) is valid, but whether it's the best UX is another question.

PR #734, opened 2021-08-03
feat: support removing the default channel head in deprecatetruncate

exdx:

OPM currently does not allow an end user to call deprecatetruncate on a bundle if that bundle is the head of the default channel. This is intended to prevent an ambiguous case – if all the bundles from a channel are removed and as a result the channel itself is removed, then the default channel no longer exists and the package is in an invalid state.

However, if the deprecatetruncate command would result in all channels being removed, there is no ambiguity to deal with. Functionally, that is equivalent to that package becoming unavailable in the index.

This PR adds a flag to the deprecatetruncate command --allow-package-removal that, if set, removes the entire package ala `opm registry rm` so that all package metadata is removed if all of the channels would be removed without that channel head requirement. It does not touch the existing deprecate codepath, and enables other bundles (not in the removed package) to continue to be deprecated normally via the command. 

Support pipeline usecases 

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/index/deprecatetruncate.go
+++ b/cmd/opm/index/deprecatetruncate.go
@@ -25,7 +25,9 @@ var deprecateLong = templates.LongDesc(`
 		Produces the following update graph in quay.io/my/index:v2
 		1.4.0 -- replaces -> 1.3.0 [deprecated]
 		
-	Deprecating a bundle that removes the default channel is not allowed. Changing the default channel prior to deprecation is possible by publishing a new bundle to the index.
+	Deprecating a bundle that removes the default channel is not allowed unless the head(s) of all channels are being deprecated (the package is subsequently removed from the index). 
+    This behavior can be enabled via the allow-package-removal flag. 
+    Changing the default channel prior to deprecation is possible by publishing a new bundle to the index.
 	`)
 
 func newIndexDeprecateTruncateCmd() *cobra.Command {
@@ -60,6 +62,7 @@ func newIndexDeprecateTruncateCmd() *cobra.Command {
 	if err := indexCmd.Flags().MarkHidden("debug"); err != nil {
 		logrus.Panic(err.Error())
 	}
+	indexCmd.Flags().Bool("allow-package-removal", false, "removes the entire package if the heads of all channels in the package are deprecated")
 
 	return indexCmd
 }
@@ -110,6 +113,11 @@ func runIndexDeprecateTruncateCmdFunc(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	allowPackageRemoval, err := cmd.Flags().GetBool("allow-package-removal")
+	if err != nil {
+		return err
+	}
+
 	logger := logrus.WithFields(logrus.Fields{"bundles": bundles})
 
 	logger.Info("deprecating bundles from the index")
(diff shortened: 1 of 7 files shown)


pkg/sqlite/deprecate.go
```diff
@@ -45,3 +48,82 @@ func (d *BundleDeprecator) Deprecate() error {
...
+	// Remove bundles from the removed package from the deprecation request
+	// This enables other bundles to be deprecated via the expected flow
+	// Build a new array with just the outstanding bundles
+	for _, bundlePath := range d.bundles {
+		if contains(bundlePath, removedBundlePaths) {
+			continue
+		}
+		remainingBundlePaths = append(remainingBundlePaths, bundlePath)
+	}
+	d.bundles = remainingBundlePaths
```
gallettilance, line 117: I'm not sure I understand what this does. This is the only function using the bundleDeprecator right?

gallettilance (commented):
@exdx can you add a test to check that all metadata for that package gets cleaned up?

pkg/sqlite/deprecate.go
```diff
@@ -45,3 +48,82 @@ func (d *BundleDeprecator) Deprecate() error {
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("error checking if bundle is default channel head %s: %s", bundle, err))
+		}
+		if found != "" {
+			packages = append(packages, found)
+		}
+	}
+
+	if len(packages) == 0 {
+		return nil
```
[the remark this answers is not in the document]
exdx: This function would return and then the regular `Deprecate()` call would proceed as before. 

Essentially, if we don't have any default channel head bundles, we don't proceed with package removal, but still deprecate those bundles.

pkg/sqlite/deprecate.go
exdx: No, there is also the existing `Deprecate()` logic which I did not change. This line ensures that the `Deprecate()` call will run on a subset of packages after the `MaybeRemoveDeprecatedPackages()` call returns. 

For example, say that there are three bundles to deprecatetruncate A B C. A and B are in one package, A is the head of the default channel, B is the head of the only other channel. This package would be removed entirely from the index, but C would also need to be deprecated via the existing logic (find tail bundles, remove those, add the deprecation property, etc). So what this logic does is take out bundles that were associated with packages that were then removed (A and B), leaving only those that should still be deprecated via the expected flow (C)

pkg/sqlite/deprecate.go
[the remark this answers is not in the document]
gallettilance: ah ok thanks for clarifying

pkg/sqlite/deprecate.go
gallettilance: gotcha thanks!

exdx:
> exdx can you add a test to check that all metadata for that package gets cleaned up?

I'm going to add a test to ensure that the package removal works as intended. Since it relies on the existing `opm registry rm` codepath to remove packages it should be reliable.

pkg/lib/registry/registry.go
```diff
@@ -357,6 +358,23 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
 	}
 
 	deprecator := sqlite.NewSQLDeprecatorForBundles(dbLoader, toDeprecate)
+
+	// Check for deprecation of head of default channel. If deprecation request includes heads of all other channels,
+	// then remove the package entirely. Otherwise, deprecate provided bundles. This enables deprecating an entire package.
+	// By default deprecating the head of default channel is not permitted.
+	if request.AllowPackageRemoval {
```
ankitathomas, line 365: This is more related to package removal by itself, but what would this mean for previously deprecated entries? They persist in the deprecated table as just a bundle name when the package is removed

exdx: That's true -- if I understand correctly this is being tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1982781 which we will address separately. I didn't want to tackle that in this PR.

cmd/opm/index/deprecatetruncate.go
```diff
@@ -25,7 +25,9 @@ var deprecateLong = templates.LongDesc(`
 		Produces the following update graph in quay.io/my/index:v2
 		1.4.0 -- replaces -> 1.3.0 [deprecated]
 		
-	Deprecating a bundle that removes the default channel is not allowed. Changing the default channel prior to deprecation is possible by publishing a new bundle to the index.
+	Deprecating a bundle that removes the default channel is not allowed unless the head(s) of all channels are being deprecated (the package is subsequently removed from the index).
```
camilamacedo86, line 28: How it can work for our automated process that requires deprecating many cases? 

c/c @gallettilance

PR #739, opened 2021-08-05
fix(declcfg): collect informative errors in Load()

estroz:
**Description of the change:** collect informative (name/package and schema type) errors in Load()

**Motivation for the change:** able to fix all load errors in one go.

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/declcfg.go
+++ b/alpha/declcfg/declcfg.go
@@ -65,6 +65,7 @@ type RelatedImage struct {
 
 type Meta struct {
 	Schema  string
+	Name    string
 	Package string
 
 	Blob json.RawMessage
@@ -77,6 +78,7 @@ func (m Meta) MarshalJSON() ([]byte, error) {
 func (m *Meta) UnmarshalJSON(blob []byte) error {
 	type tmp struct {
 		Schema     string              `json:"schema"`
+		Name       string              `json:"name,omitempty"`
 		Package    string              `json:"package,omitempty"`
 		Properties []property.Property `json:"properties,omitempty"`
 	}
@@ -86,6 +88,7 @@ func (m *Meta) UnmarshalJSON(blob []byte) error {
 	}
 	m.Schema = t.Schema
 	m.Package = t.Package
+	m.Name = t.Name
 	m.Blob = blob
 	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/declcfg/load.go
```diff
@@ -135,20 +137,24 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 		case schemaPackage:
 			var p Package
 			if err := json.Unmarshal(doc, &p); err != nil {
-				return nil, fmt.Errorf("parse package: %v", err)
+				errs = append(errs, fmt.Errorf("parse %s %s: %v", in.Schema, in.Name, err))
```
joelanford, line 140: If schema and name are empty strings (which is definitely possible if we're hitting an unmarshal error), this error message will look like `parse    : error`

Maybe something like this. I don't love the square brackets, so if you have a better idea, I'm all ears.
```suggestion
				errs = append(errs, fmt.Errorf("parse [schema:%q name:%q]: %v", in.Schema, in.Name, err))
```

internal/declcfg/declcfg.go
```diff
@@ -86,6 +88,11 @@ func (m *Meta) UnmarshalJSON(blob []byte) error {
 	}
 	m.Schema = t.Schema
 	m.Package = t.Package
+	m.Name = t.Name
+	// Package name is encoded as "name" in the package schema.
+	if m.Schema == schemaPackage {
+		m.Package = t.Name
+	}
```
joelanford, line 95: Not sure I'm convinced this assignement is really necessary. `olm.package` blobs don't have a package field, so IMO it would be fine for this field to just be an empty string for packages.

internal/declcfg/load.go
ankitathomas: It might be helpful to have the processed filenames on the errors in the Walk function, where schema and name might be empty.

https://github.com/operator-framework/operator-registry/blob/78f27b39dc098c2e56681d9a3d8e786aae95d4fe/internal/declcfg/load.go#L39

estroz: At the point these errors are being constructed, at least `Meta.Schema` is known, so the error will always at least look like `parse olm.<package|bundle>  : error`, so `parse %s [name:%q]: %v` would be reasonable. Thoughts?

>It might be helpful to have the processed filenames on the errors in the Walk function

@ankitathomas I'll add a path component to some errors in that function.

joelanford: Good point about schema being known. That actually simplifies this IMO. How about just:

```
return fmt.Errorf("parse %s name=%q: %v", m.Schema, name, err)
```

I think subbing in `"unknown"` when name is empty could be more confusing than just leaving it blank.

awgreene (commented):
Nice work @estroz 
/lgtm

perdasilva:
closing as stale

PR #741, opened 2021-08-05
fix(opm): clarify that bundle declcfgs are not valid refs alone

estroz:
**Description of the change:** clarify that bundle declcfgs are not valid refs alone, and require an `olm.package` declcfg. An error is returned if a bundle is directly specified.

**Motivation for the change:** diff is really intended for indices. There's no way to infer which channel is the default for a package if more than one bundle in the same package is passed to diff either, so we could get non-deterministic behavior unless diff imposes some ordering on channels. Assuming icon doesn't exist (or other future `olm.package` fields) for a package may cause issues when using the diff'ed catalog in-cluster too.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/diff/cmd.go
+++ b/cmd/opm/alpha/diff/cmd.go
@@ -45,7 +45,10 @@ func NewCmd() *cobra.Command {
 Diff a set of old and new catalog references ("refs") to produce a
 declarative config containing only packages channels, and versions not present
 in the old set, and versions that differ between the old and new sets. This is known as "latest" mode.
+
 These references are passed through 'opm render' to produce a single declarative config.
+Bundle image refs are not supported directly;
+a valid "olm.package" declarative config object referring to the bundle's package must exist in all input refs.
 
 This command has special behavior when old-refs are omitted, called "heads-only" mode:
 instead of the output being that of 'opm render refs...'
--- a/cmd/opm/alpha/diff/cmd.go
+++ b/cmd/opm/alpha/diff/cmd.go
@@ -47,8 +47,8 @@ declarative config containing only packages channels, and versions not present
 in the old set, and versions that differ between the old and new sets. This is known as "latest" mode.
 
 These references are passed through 'opm render' to produce a single declarative config.
-Bundle image refs are not supported directly;
-a valid "olm.package" declarative config object referring to the bundle's package must exist in all input refs.
+Bundle image refs are not supported directly; a valid "olm.package" declarative config object
+referring to the bundle's package must exist in all input refs.
 
 This command has special behavior when old-refs are omitted, called "heads-only" mode:
 instead of the output being that of 'opm render refs...'
--- a/internal/action/diff.go
+++ b/internal/action/diff.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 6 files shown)


ankitathomas (approved):
/lgtm

internal/action/diff.go
```diff
@@ -25,12 +26,18 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
...
+	mask := RefDCDir | RefDCImage | RefSqliteFile | RefSqliteImage
+
 	// Heads-only mode does not require an old ref, so there may be nothing to render.
 	var oldModel model.Model
 	if len(a.OldRefs) != 0 {
-		oldRender := Render{Refs: a.OldRefs, Registry: a.Registry}
+		oldRender := Render{Refs: a.OldRefs, Registry: a.Registry, AllowedRefMask: mask}
 		oldCfg, err := oldRender.Run(ctx)
 		if err != nil {
+			if isNotAllowedError(err) {
```
joelanford, line 38: Can you add unit tests to cover both old and new refs being bundles?

internal/action/render.go
```diff
@@ -108,7 +103,7 @@ func (r Render) renderReference(ctx context.Context, ref string) (*declcfg.Decla
 	if stat, serr := os.Stat(ref); serr == nil {
 		if stat.IsDir() {
 			if !r.AllowedRefMask.Allowed(RefDCDir) {
-				return nil, fmt.Errorf("cannot render DC directory: %w", &ErrNotAllowed{})
+				return nil, fmt.Errorf("cannot render DC directory: %w", ErrNotAllowed)
```
awgreene, line 106: Might be worth writing DeclarativeConfig out for users unfamiliar with the type.

internal/action/render.go
```diff
@@ -147,23 +142,23 @@ func (r Render) imageToDeclcfg(ctx context.Context, imageRef string) (*declcfg.D
...
+			return nil, fmt.Errorf("cannot render sqlite image: %w", ErrNotAllowed)
 		}
 		cfg, err = sqliteToDeclcfg(ctx, filepath.Join(tmpDir, dbFile))
 		if err != nil {
 			return nil, err
 		}
 	} else if configsDir, ok := labels[containertools.ConfigsLocationLabel]; ok {
 		if !r.AllowedRefMask.Allowed(RefDCImage) {
-			return nil, fmt.Errorf("cannot render DC image: %w", &ErrNotAllowed{})
+			return nil, fmt.Errorf("cannot render DC image: %w", ErrNotAllowed)
```
awgreene, line 153: Same as above.

awgreene (changes requested):
Nice work @estroz, a few nits.

cmd/opm/alpha/diff/cmd.go
```diff
@@ -45,7 +45,10 @@ func NewCmd() *cobra.Command {
 Diff a set of old and new catalog references ("refs") to produce a
 declarative config containing only packages channels, and versions not present
 in the old set, and versions that differ between the old and new sets. This is known as "latest" mode.
+
 These references are passed through 'opm render' to produce a single declarative config.
+Bundle image refs are not supported directly;
+a valid "olm.package" declarative config object referring to the bundle's package must exist in all input refs.
```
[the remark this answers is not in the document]
estroz: Done.

dinhxuanvu (approved):
/approve

PR #742, opened 2021-08-05
Generate package property during migration when no provided APIs.

benluddy:
All entries beginning with schema version 9 should have at least one
property -- a generated package property. This did not occur for
catalog entries that had no provided APIs.

The existing migration is updated to fix the bug, and a new
migration is added to retroactively generate the package property for
databases that have already applied the existing migration.


kevinrizza:
does editing the existing migration make sense? don't we need to handle dbs already past this version?

benluddy:
That's my question too. Properties are supposed to be opaque to the catalog, so it's also not great to go in and start generating properties after-the-fact as part of a later migration. Information is sort of implicitly lost after version 9, because we can no longer distinguish between the properties that were generated automatically as part of the migration versus a set of properties that the catalog maintainer may have specifically curated.

pkg/sqlite/migrations/013_ensure_package_property.go
```diff
@@ -0,0 +1,34 @@
...
+func init() {
+	registerMigration(EnsurePackagePropertyMigrationKey, ensurePackagePropertyMigration)
+}
+
+var ensurePackagePropertyMigration = &Migration{
+	Id: EnsurePackagePropertyMigrationKey,
+	Up: func(ctx context.Context, tx *sql.Tx) error {
+		if _, err := tx.ExecContext(ctx, `
+INSERT INTO properties(type, value, operatorbundle_name, operatorbundle_version, operatorbundle_path)
+  SELECT DISTINCT :property_type, json_object('packageName', channel_entry.package_name, 'version', operatorbundle.version), operatorbundle.name, operatorbundle.version, operatorbundle.bundlepath
```
ecordell, line 22: do you need a `CAST` here?

benluddy:
With https://github.com/operator-framework/operator-lifecycle-manager/pull/2291 merged, plus the existing package property inference feature, I don't think there is any pressure to make this change.

PR #745, opened 2021-08-09
read stdout for reading container name

### LINKED ISSUE #689: OPM fails to create index with Podman 3.0.1
There was a change introduced in Podman after version` 2.2.1` to display a warning message when the owner of the current cgroup could not be detected. When using OPM `v1.14.3-34` with Podman `3.0.1-7` to add a bundle to an index, OPM fails because it does not expect this warning message. I have traced this problem to runner.go where OPM tries to run `podman create`. When trying to extract the ID from the create output, OPM fails as it does not expect the warning message to be in the output (`id := strings.TrimSuffix(string(out), "\n")`.

I am running rootless in a container with this base image: `registry.redhat.io/rhel8/podman:8.4`. Runnnig `yum update -y` in the Dockerfile gives me Podman `3.0.1-7`.

Here is the line of code where the problem is:
https://github.com/operator-framework/operator-registry/blob/master/pkg/containertools/runner.go#L139. Though the warning may cause problems with other Podman commands that OPM runs.

OPM failure log:
```
...
INFO[0003] running podman cp"
ERRO[0003] time="2021-06-23T09:27:21-04:00" level=warning msg="Failed to detect the owner for the current cgroup: stat /sys/fs/cgroup/systemd/docker/350ea3d90cc8a53bae50e9d52e0a2ba7fb1320b02b5649d9299e032481b38a1d: no such file or directory"
Error: container "time=\"2021-06-23T09" does not exist"
ERRO[0003] permissive mode disabled                      error="error copying container directory time=\"2021-06-23T09:27:21-04:00\" level=warning msg=\"Failed to detect the owner for the current cgroup: stat /sys/fs/cgroup/systemd/docker/350ea3d90cc8a53bae50e9d52e0a2ba7fb1320b02b5649d9299e032481b38a1d: no such file or directory\"\nError: container \"time=\\\"2021-06-23T09\" does not exist\n: exit status 125"
Error: error copying container directory time="2021-06-23T09:27:21-04:00" level=warning msg="Failed to detect the owner for the current cgroup: stat /sys/fs/cgroup/systemd/docker/350ea3d90cc8a53bae50e9d52e0a2ba7fb1320b02b5649d9299e032481b38a1d: no such file or directory"
Error: container "time=\"2021-06-23T09" does not exist
```

ankitathomas:
Read container names from only stdout to avoid podman warnings on stderr.

Closes #689

### DIFF
--- a/pkg/containertools/runner.go
+++ b/pkg/containertools/runner.go
@@ -130,10 +130,13 @@ func (r *ContainerCommandRunner) Unpack(image, src, dst string) error {
 	r.logger.Infof("running %s create", r.containerTool)
 	r.logger.Debugf("%s", command.Args)
 
-	out, err := command.CombinedOutput()
+	out, err := command.Output()
 	if err != nil {
-		r.logger.Errorf(string(out))
-		return fmt.Errorf("error creating container %s: %v", string(out), err)
+		msg := err.Error()
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			msg = fmt.Sprintf("%s: %s", err, exitErr.Stderr)
+		}
+		return fmt.Errorf("error creating container %s: %s", string(out), msg)
 	}
 
 	id := strings.TrimSuffix(string(out), "\n")


pkg/containertools/runner.go
```diff
@@ -130,10 +130,14 @@ func (r *ContainerCommandRunner) Unpack(image, src, dst string) error {
...
-	out, err := command.CombinedOutput()
+	out, err := command.Output()
 	if err != nil {
-		r.logger.Errorf(string(out))
-		return fmt.Errorf("error creating container %s: %v", string(out), err)
+		msg := err.Error()
+		if exitErr, ok := err.(*exec.ExitError); ok {
+			msg = fmt.Sprintf("%s: %s", err, exitErr.Stderr)
+		}
+		r.logger.Errorf(msg)
```
exdx, line 139: nit: would logging the error here and then bubbling it up cause the message to occur twice in a row in the command output?

timflannagan: Yeah, generally speaking, logging at the source and then returning an error can lead to duplicate error logging, so if we can double-check how we're handling logging at the various call sites, and instead bubble the error up that should be sufficient.

dubee (changes requested):
@ankitathomas, can you also change the other occurrences of `CombinedOutput` in the `Unpack` method? It looks like there are two more calls in `Unpack` that combine stdout and stderr.

dubee (commented):
@ankitathomas, any updates on this one

ankitathomas:
> @ankitathomas, can you also change the other occurrences of `CombinedOutput` in the `Unpack` method? It looks like there are two more calls in `Unpack` that combine stdout and stderr.

The other uses of CombinedOutput are only ever used as error messages, unlike the one instance fixed here that was using the CombinedOutput to look for a container name on stdout. The issue should go away with just this one fixed.

PR #746, opened 2021-08-09
fix(deprecate): explicit pkg rm drops latest deprecated

njhale:
- When using opm index|registry rm <pkg>, drop the "latest" related bundles
from the deprecated table. This makes "un-deprecating" a bundle
possible, which is useful if that  bundle was deprecated by
accident
- Add a migration and logic on deprecatetruncate that drops
truncated bundles (i.e. w/o assoc. channel_entry rows) from the deprecated
table, clearing the "deprecation history" so that operations post package
removal don't assume a truncated bundle should be deprecated

See https://bugzilla.redhat.com/show_bug.cgi?id=1982781 for motivation

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -28,7 +28,7 @@ type MigratableLoader interface {
 
 var _ MigratableLoader = &sqlLoader{}
 
-func NewSQLLiteLoader(db *sql.DB, opts ...DbOption) (MigratableLoader, error) {
+func newSQLLoader(db *sql.DB, opts ...DbOption) (*sqlLoader, error) {
 	options := defaultDBOptions()
 	for _, o := range opts {
 		o(options)
@@ -44,6 +44,10 @@ func NewSQLLiteLoader(db *sql.DB, opts ...DbOption) (MigratableLoader, error) {
 	}
 
 	return &sqlLoader{db: db, migrator: migrator, enableAlpha: options.EnableAlpha}, nil
+}
+
+func NewSQLLiteLoader(db *sql.DB, opts ...DbOption) (MigratableLoader, error) {
+	return newSQLLoader(db, opts...)
 }
 
 func (s *sqlLoader) Migrate(ctx context.Context) error {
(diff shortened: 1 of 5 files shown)


njhale:
/hold

until we resolve a conversation with some downstream stakeholders

pkg/sqlite/load.go
```diff
@@ -1536,3 +1547,46 @@ func (s *sqlLoader) deprecated(tx *sql.Tx, name string) (bool, error) {
...
+	// Use that info to relate the package to a set of rows in the deprecated table.
+	_, err = tx.Exec(`DELETE FROM deprecated WHERE deprecated.operatorbundle_name IN (SELECT DISTINCT deprecated.operatorbundle_name FROM (deprecated INNER JOIN channel_entry ON deprecated.operatorbundle_name = channel_entry.operatorbundle_name) WHERE channel_entry.package_name = ?)`, pkg)
+	if err != nil {
+		return err
+	}
+
+	return tx.Commit()
+}
+
+func (d *DeprecationAwareLoader) RemovePackage(pkg string) error {
```
njhale, line 1586: Created a decorator type because we don't want to drop the deprecation history in certain cases; e.g. `--overwrite-latest` needs to remember deprecations but remove everything else in the package.

njhale:
/hold

more discussions

pkg/sqlite/migrations/013_rm_truncated_deprecations.go
```diff
@@ -0,0 +1,31 @@
...
+		// Delete deprecation history for all bundles that no longer exist in the channel_entries table
+		// These bundles have been truncated by more recent deprecations and would only confuse future operations on an index;
+		// e.g. adding a previously truncated bundle to a package removed via `opm index|registry rm` would lead to that bundle
+		// being deprecated
+		_, err := tx.ExecContext(ctx, `DELETE FROM deprecated WHERE deprecated.operatorbundle_name NOT IN (SELECT DISTINCT deprecated.operatorbundle_name FROM (deprecated INNER JOIN channel_entry ON deprecated.operatorbundle_name = channel_entry.operatorbundle_name))`)
+
+		return err
+	},
+	Down: func(ctx context.Context, tx *sql.Tx) error {
+		// No-op
```
kevinrizza, line 28: 👋  bye!

pkg/sqlite/load.go
```diff
@@ -1536,3 +1547,46 @@ func (s *sqlLoader) deprecated(tx *sql.Tx, name string) (bool, error) {
...
 }
+
+// DeprecationAwareLoader understands how bundle deprecations are handled in SQLite and decorates
+// the sqlLoader with proxy methods that handle deprecation related table housekeeping.
+type DeprecationAwareLoader struct {
+	*sqlLoader
+}
+
+// NewDeprecationAwareLoader returns a new DeprecationAwareLoader.
+func NewDeprecationAwareLoader(db *sql.DB, opts ...DbOption) (*DeprecationAwareLoader, error) {
```
kevinrizza, line 1558: 👍

PR #747, opened 2021-08-10
feat: bump k8s dependencies to v0.22.0

exdx:

Bumps k8s dependencies to v0.22.0.

Note that this PR does not actually deprecate any types such as v1beta1 CRDs that are currently handled in the registry. There will need to be a deprecation warning for these types, and that may live in registry or in API. 

New release 

**Reviewer Checklist**


estroz:
Probably worth bumping api in this PR after https://github.com/operator-framework/api/pull/140 is merged to make sure bundle validation still works.

pkg/lib/bundle/validate.go
```diff
@@ -343,6 +343,8 @@ func (i imageValidator) ValidateBundleContent(manifestDir string) error {
 					}
 				}
 			case v1beta1CRDapiVersion:
+				i.logger.Warnf("GroupVersionKind apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition was deprecated in Kubernetes 1.16+, removed in 1.22. Support" +
+					"for apiextensions.k8s.io/v1beta1 CustomResourceDefinitions will be removed in the future in favor of apiextensions.k8s.io/v1 CustomResourceDefinitions.")
```
kevinrizza, line 347: should we also extend this check to the other manifest types that we support that are getting deprecated? webhooks and rbac I think

exdx: Right now for bundle content (in registry), we check
* the CSV
* the CRDs (by APIVersion)
* check to see that other resources are “kubectl-able” on cluster

This validation is pretty light, I think it makes sense to keep it that way and just alert on v1beta1 CRDs.

There is additional validation in API that checks to see if objects are deprecated by looking at a hard-coded list of GVKs: https://github.com/operator-framework/api/blob/0fe04f80e0102dc6973109a221a7babd6bee005e/pkg/validation/internal/removed_apis.go#L33

This validation is not run by default, and only applies to a few validators (OperatorHub and Community ones). We would have to reimplement this in registry, or use the existing one, but we end up much more restrictive than we were previously. That's why myself and @njhale prefer to keep the light validation we have in registry currently.

njhale (approved):
/approve

PR #748, opened 2021-08-13
render: support rendering FBC from bundle directories

joelanford:
This PR adds support to the `opm render` command for bundle directories.

When rendering bundle directories, there is no metadata about the bundle image reference, so the FBC rendered from a bundle directory contains inlined `olm.bundle.object` properties but lacks an `image` value.

However callers of `opm render` with bundle directories will generally know the image that will contain the bundle directory, so they can insert the bundle image reference themselves. To facilitate this image reference injection, this PR adds a `--image-ref-template` flag that supports templating based on package, name, and version of the bundle.

If specified, `render` will populate `olm.bundle` objects' image field with the templated value and use the `olm.csv.metadata` property. If not specified, `render` will leave the image field empty and will instead include all of the bundle manifests in the catalog as `olm.bundle.object` properties.

Some catalog pipelines build the bundle images based on the bundle sources that are included in a source repository. This structure works well in imperative flows where bundle sources alone are sufficient to construct a valid catalog.

However with catalogs themselves now being declarative, a new use case has appeared. Operator authors would like to be able to provide the pipeline with their bundle sources _and_ their catalog metadata. But there is a chicken and egg problem because existing `opm` tooling only supports generating catalog metadata from a bundle image. If a pipeline builds the bundle images and `opm` expects the bundle images to exist in order to generate catalog metadata, it is impossible for an operator author to provide catalog and bundle sources at the same time.

In order to fully support this style of pipeline, a follow-on PR may be necessary to update `opm`'s template to support using bundle directory references and supplying an image reference template.

**Reviewer Checklist**

### DIFF
--- a/alpha/action/render.go
+++ b/alpha/action/render.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"text/template"
 
 	"github.com/h2non/filetype"
 	"github.com/h2non/filetype/matchers"
@@ -39,6 +40,7 @@ const (
 	RefSqliteFile
 	RefDCImage
 	RefDCDir
+	RefBundleDir
 
 	RefAll = 0
 )
@@ -50,10 +52,11 @@ func (r RefType) Allowed(refType RefType) bool {
 var ErrNotAllowed = errors.New("not allowed")
 
 type Render struct {
-	Refs           []string
-	Registry       image.Registry
-	AllowedRefMask RefType
-	Migrate        bool
+	Refs             []string
+	Registry         image.Registry
+	AllowedRefMask   RefType
+	Migrate          bool
+	ImageRefTemplate *template.Template
 
 	skipSqliteDeprecationLog bool
 }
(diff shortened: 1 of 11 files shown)


joelanford:
/hold
We might not need this, at least not urgently.

alpha/action/render.go
```diff
@@ -454,3 +482,150 @@ func combineConfigs(cfgs []declcfg.DeclarativeConfig) *declcfg.DeclarativeConfig
...
+}
+
+type imageReferenceTemplateData struct {
+	Package string
+	Name    string
+	Version string
+}
+
+func (r *Render) renderPackageManifest(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
+	db, err := sql.Open("sqlite3", fmt.Sprintf("file:%s?mode=memory&cache=shared&_foreign_keys=on", ref))
```
grokspawn, line 521: I'm sure I'm missing some context here, but I'm wondering...  why we are reading a yaml file into an sqlite DB? 
For e.g. an example I find for this is something like
```yaml
channels:
- currentCSV: myoperator.v0.1.0
  name: original
- currentCSV: myoperator.v1.9.0
  name: alpha
defaultChannel: alpha
packageName: myop
```

Additional question:  
is doing this going to make fulfilling https://github.com/operator-framework/operator-registry/issues/1105 more difficult?

joelanford: I removed this code from the PR because I don't think we need to support package manifest directories anymore.

But to answer your question, this code would take a package manifests root directory (not just the packages.yml file) and load it into an sqlite DB. The package manifest format was actually declarative in that it put all of the bundle manifests for all of the versions in a particular directory structure, which meant that one could idempotently rebuild the database by tweaking something in the source files and constructing the DB again from scratch.

alpha/action/render_test.go
```diff
@@ -875,6 +987,8 @@ func TestAllowRefMask(t *testing.T) {
 					"test.registry/foo-operator/foo-index-declcfg:v0.2.0",
 					"testdata/foo-index-v0.2.0-declcfg",
 					"test.registry/foo-operator/foo-bundle:v0.2.0",
+					"testdata/foo-bundle-v0.2.0",
+					"testdata/foo-index-v0.2.0-package-manifest",
```
grokspawn, line 991: Need to take 990, 991 out to pass unit, I think.

PR #753, opened 2021-08-18
feat(opm): configure diff to not include dependency bundles with --skip-deps

estroz:
**Description of the change:** configure `opm alpha diff` to not include dependency bundles with `--skip-deps`

**Motivation for the change:** there are situations where dependencies are not desired, ex. they are fulfilled by another catalog and this is known prior to invoking `opm alpha diff`.

**Reviewer Checklist**

/kind feature

### DIFF
--- a/internal/declcfg/diff.go
+++ b/internal/declcfg/diff.go
@@ -3,19 +3,41 @@ package declcfg
 import (
 	"reflect"
 	"sort"
+	"sync"
 
 	"github.com/blang/semver"
 	"github.com/mitchellh/hashstructure/v2"
+	"github.com/sirupsen/logrus"
 
 	"github.com/operator-framework/operator-registry/internal/model"
 	"github.com/operator-framework/operator-registry/internal/property"
 )
 
-// Diff returns a Model containing everything in newModel not in oldModel,
+// DiffGenerator configures how diffs are created via Run().
+type DiffGenerator struct {
+	Logger *logrus.Entry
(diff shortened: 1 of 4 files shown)


internal/declcfg/diff.go
```diff
@@ -71,9 +94,11 @@ func Diff(oldModel, newModel model.Model) (model.Model, error) {
 		}
 	}
 
-	// Add dependencies to outputModel not already present in oldModel.
-	if err := addAllDependencies(newModel, oldModel, outputModel); err != nil {
-		return nil, err
+	if !g.NoDependencies {
```
kevinrizza, line 97: this is a nit, but I don't love double negatives. It would be nice if this was actually:

`if g.IncludeDependencies {`

joelanford: My general approach for struct fields (and especially for `bool`s) is to make their default value meaningful and the default that you actually want.

Since the default value for `bool`s is `false`, it would mean we want the `diff` functionality to default to including dependencies.

estroz: I don't either, although I want the default to be false so dependencies are added by default. May just make this a pointer.

estroz: Changed to `SkipDependencies`.

joelanford (commented):
/approve

dinhxuanvu (approved):
/lgtm

PR #756, opened 2021-08-18
feat(opm): fine-grained dependency selection in diffs

estroz:
**Description of the change:** select the latest bundle that satisfies each dependency (deduplicated).

**Motivation for the change:** currently entire dependency packages are added to `opm alpha diff`'s output. This is not optimal even when not using the resolver because multiple bundles may satisfy the same dependency when the latest bundle (version) will suffice. This implementation does not attempt to resolve dependencies, but pre-groom the set passed to the resolver.

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/diff.go
+++ b/alpha/declcfg/diff.go
@@ -1,6 +1,7 @@
 package declcfg
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
 	"sync"
@@ -167,7 +168,7 @@ func bundlesEqual(b1, b2 *model.Bundle) (bool, error) {
 
 func addAllDependencies(newModel, oldModel, outputModel model.Model) error {
 	// Get every oldModel's bundle's dependencies, and their dependencies, etc. by BFS.
-	allProvidingBundles := []*model.Bundle{}
+	providingBundlesByPackage := map[string][]*model.Bundle{}
 	for curr := getBundles(outputModel); len(curr) != 0; {
 		reqGVKs, reqPkgs, err := findDependencies(curr)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


alpha/declcfg/diff_include.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return graph
+}
+
+// findIntersectingBundles finds the intersecting bundle of start and end in the
+// replaces upgrade graph graph by traversing down to the lowest graph node,
+// then returns every bundle higher than the intersection. It is possible
+// to find no intersection; this should only happen when start and end
+// are not part of the same upgrade graph.
+// Output bundle order is not guaranteed.
```
joelanford, line 24: Non-blocking nit: it might be nice to have a simple ASCII diagram in this comment to help explain what this function is doing.

alpha/declcfg/diff_include.go
```diff
@@ -0,0 +1,78 @@
...
+		oldChain[next.Replaces] = next
+	}
+
+	// Trace the new replaces chain from end until the old chain intersects.
+	var intersection string
+	for next := end; next != nil && next.Replaces != ""; next = ch.Bundles[next.Replaces] {
+		if _, inChain := oldChain[next.Replaces]; inChain {
+			intersection = next.Replaces
+			break
+		}
```
joelanford, line 43: Should this be checking if `next.Name` is in the chain rather than `next.Replaces`? What if `end` is in the chain but is not equal to `start`?

Follow up (having looked at the tests)... if start and end are accidentally reversed by a caller such that oldChain contains end, what happens?

joelanford (commented):
Needs rebase and a couple of minor nits questions, but looks great overall!

/approve

alpha/declcfg/diff_include.go
joelanford: Resolving this comment. I think the tests do a good enough job explaining it.

ankitathomas (commented):
Good work on this!

alpha/declcfg/diff.go
```diff
@@ -183,36 +184,86 @@ func addAllDependencies(newModel, oldModel, outputModel model.Model) error {
...
 			curr = append(curr, providingBundles...)
-			allProvidingBundles = append(allProvidingBundles, providingBundles...)
+
+			oldPkg, oldHasPkg := oldModel[pkg.Name]
+			for _, b := range providingBundles {
+				b := b
+				// If the bundle is not in oldModel, add it to the set.
+				// outputModel is checked below.
+				add := true
+				if oldHasPkg {
```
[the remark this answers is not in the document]
estroz: Checking this here saves me having to write two loops, since all providing bundles need to be added if `!oldHasPkg`.

alpha/declcfg/diff_include.go
estroz: >Should this be checking if next.Name is in the chain rather than next.Replaces? What if end is in the chain but is not equal to start?

Each name will also be in a replaces, so the algorithm is effectively just skipping `next.Name`. 
I'll add a check for `next.Name` as a base case.

>Follow up (having looked at the tests)... if start and end are accidentally reversed by a caller such that oldChain contains end, what happens?

The intersection will always be the same because the algorithm constructs the same DAGs for both, so the set produced by BFS will always be the same too. The case you mention causes an issue when deleting bundles between start and intersection inclusively, since deleting from the true end leaf node (which could be reversed as `start` as input) will remove bundles that should be in the output. However the only `end` ever used in `findIntersectingBundles()` is the channel head, so we can assume that deleting bundles between `start` and the intersection is ok. In general we can assume `start` and `end` are never reversed.

joelanford: > However the only end ever used in findIntersectingBundles() is the channel head, so we can assume that deleting bundles between start and the intersection is ok. In general we can assume start and end are never reversed.

That's fair, I'm just worried about this assumption being implicit and the possibility that we later change the code to violate that assumption. Is there a preventative measure we can take?

For example:

> What if end is in the chain but is not equal to start?

Can we have an explicit check/error (or maybe panic) for this, since we always assume/expect that `end` encompasses `start`?

estroz: Added a check for end being in start's chain: https://github.com/operator-framework/operator-registry/pull/756/files#diff-f61e326af35ec0122edb28f797b31bdfdd84cddb1975b1d2df38a10ce8292379R41-R42.

PR #758, opened 2021-08-19
Fix upstream-opm-builder build error due to rate limiting

kevinrizza:
Dockerhub has some pretty severe rate limiting now, so most of our builds that rely on docker.io are failing now. To temporarily get around this, I have mirrored the base images to quay.io until we can find a more permanent base image replacement.


dinhxuanvu (approved):
/lgtm

PR #759, opened 2021-08-19
move internal to alpha to make subpackages available for importing from other projects

joelanford:

Wholesale move of `internal` to `alpha`

Using `alpha` name to denote that these packages do not yet have stable Go APIs, and added a `doc.go` to explain what being in the `alpha` tree means:

```go
// Package alpha contains subpackages that have unstable Go APIs.
// Subpackages that achieve maturity will be deprecated and moved
// to other modules or packages within operator-registry.
//
// Presence in the alpha package should not be construed as being
// of less quality than non-alpha packages. Subpackages in the
// alpha package have the same standard for testing as all other
// packages in this repository.
package alpha
```

To make file-based config packages available to be imported by other projects.

**Reviewer Checklist**


estroz (approved):
/lgtm

PR #767, opened 2021-08-27
handle deprecating branch points

ankitathomas:
Signed-off-by: Ankita Thomas <ankithom@redhat.com>

`deprecatetruncate` deprecates a bundle path and truncates all bundles that have an upgrade path to the deprecated bundle, either via a skips or replaces entry. 
[](https://mermaid-js.github.io/mermaid-live-editor/edit/##eyJjb2RlIjoiZ3JhcGggTFJcbkFbQTogc3RhYmxlXVxuQidbQzogMS4wLnhdXG5CJydbQjogMS4wLnhdXG5CW0Q6IHN0YWJsZSwxLjAueF1cbkNbOiBzdGFibGUsMS4wLl1cblxuQSAtLT4gQiAtLT4gQ1xuQicnIC0tPiBCJyAtLi0-IEIgXG5cbmNsYXNzIEInJyBkZXByZWNhdGVkXG5jbGFzcyBCJyxDLEIgdHJ1bmNhdGVkXG5cbmNsYXNzRGVmIGRlcHJlY2F0ZWQgZmlsbDojZjlmLHN0cm9rZTojMzMzLHN0cm9rZS13aWR0aDo0cHg7XG5jbGFzc0RlZiB0cnVuY2F0ZWQgZmlsbDojYmJmLHN0cm9rZTojZjY2LHN0cm9rZS13aWR0aDoycHgsY29sb3I6I2ZmZixzdHJva2UtZGFzaGFycmF5OiA1IDU7XG4iLCJtZXJtYWlkIjoie1xuICBcInRoZW1lXCI6IFwiZGVmYXVsdFwiXG59IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ)

This causes the index to become invalid when one or more truncated tail bundles are branch points, i.e they have paths to other non-deprecated bundles  
e.g, v1.1.0 replaces v1.0.1 which gets truncated on deprecating v1.0.3

This PR now removes tail bundles only from the channels the deprecated bundle belongs to. Bundles are truncated only if they no longer belong to any channel.
[](https://mermaid-js.github.io/mermaid-live-editor/edit/##eyJjb2RlIjoiZ3JhcGggTFJcbkFbQTogc3RhYmxlXVxuQidbQzE6IDEuMC54XVxuQicnW0I6IDEuMC54XVxuQltEOiBzdGFibGVdXG5DW0U6IHN0YWJsZV1cblxuQSAtLT4gQiAtLT4gQ1xuQicnIC0tPiBCJyAtLi0-IEIgXG5cbmNsYXNzIEInJyBkZXByZWNhdGVkXG5jbGFzcyBCJywgdHJ1bmNhdGVkXG5cbmNsYXNzRGVmIGRlcHJlY2F0ZWQgZmlsbDojZjlmLHN0cm9rZTojMzMzLHN0cm9rZS13aWR0aDo0cHg7XG5jbGFzc0RlZiB0cnVuY2F0ZWQgZmlsbDojYmJmLHN0cm9rZTojZjY2LHN0cm9rZS13aWR0aDoycHgsY29sb3I6I2ZmZixzdHJva2UtZGFzaGFycmF5OiA1IDU7XG4iLCJtZXJtYWlkIjoie1xuICBcInRoZW1lXCI6IFwiZGVmYXVsdFwiXG59IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ)

It also switches to using the `operatorbundle` table to remove truncated bundles from the `deprecated` table to handle deprecating channel heads (which get removed from the `channel_entry` table to avoid removing deprecated channels showing up in console).

To avoid the multiple channel head problem that comes from regenerating such a graph in `--overwrite-latest`, the PR also modifies `--overwrite-latest` to only readd the bundle that is overwritten - the remaining graph is preserved.

This also stops opm add's graph generation from traversing down a replaces chain past the first deprecated entry.

### DIFF
--- a/alpha/action/render_test.go
+++ b/alpha/action/render_test.go
@@ -792,7 +792,7 @@ func generateSqliteFile(path string, imageMap map[image.Reference]string) error 
 		return err
 	}
 
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, dbQuerier, imageMap, nil, false)
+	populator := registry.NewDirectoryPopulator(loader, graphLoader, dbQuerier, imageMap, nil)
 	if err := populator.Populate(registry.ReplacesMode); err != nil {
 		return err
 	}
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -165,7 +165,7 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 		return err
 
 	}
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap, overwrittenBundles, overwrite)
+	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap, overwrittenBundles)
 
 	if err := populator.Populate(mode); err != nil {
 
@@ -258,7 +258,7 @@ func (r RegistryUpdater) PruneFromRegistry(request PruneFromRegistryRequest) err
 	}
 	defer db.Close()
 
-	dbLoader, err := sqlite.NewSQLLiteLoader(db)
+	dbLoader, err := sqlite.NewDeprecationAwareLoader(db)
 	if err != nil {
 		return err
 	}
@@ -311,7 +311,7 @@ func (r RegistryUpdater) DeprecateFromRegistry(request DeprecateFromRegistryRequ
 	}
 	defer db.Close()
 
-	dbLoader, err := sqlite.NewSQLLiteLoader(db)
+	dbLoader, err := sqlite.NewDeprecationAwareLoader(db)
 	if err != nil {
 		return err
 	}
@@ -460,7 +460,7 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
 * eg:  [1.0.2 (alpha, stable)] <- 1.0.1 (alpha)
 * When 1.0.2 in alpha and stable channels is added replacing 1.0.1, 1.0.1's presence will only be marked as expected on
 * the alpha channel, not on the inherited stable channel.
-*/
+ */
 // expectedGraphBundles returns a set of package-channel-bundle tuples that MUST be present following an add.
 func expectedGraphBundles(imagesToAdd []*registry.Bundle, graphLoader registry.GraphLoader, overwrite bool) (map[string]*registry.Package, error) {
 	expectedBundles := map[string]*registry.Package{}
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -155,9 +155,6 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 			if overwritten == "" {
 				return fmt.Errorf("index add --overwrite-latest is only supported when using bundle images")
 			}
-			if len(overwrittenBundles[img.Bundle.Package]) == 0 {
-				overwrittenBundles[img.Bundle.Package] = []string{}
-			}
 			overwrittenBundles[img.Bundle.Package] = append(overwrittenBundles[img.Bundle.Package], img.Bundle.Name)
 		}
 	}
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -446,6 +446,7 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
 	return false, nil
 }
 
+// expectedGraphBundles returns a set of package-channel-bundle tuples that MUST be present following an add.
 /* opm index add drops bundles that replace a channel head, and since channel head selection heuristics
 * choose the bundle with the greatest semver as the channel head, any bundle that replaces such a bundle
 * will be dropped from the graph following an add.
@@ -457,11 +458,10 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
 *
 * Overwritten bundles will only be verified on the channels of the newly added version.
 * Any inherited channels due to addition of a new bundle on its tail bundles may not be verified
-* eg:  [1.0.2 (alpha, stable)] <- 1.0.1 (alpha)
+* eg:  1.0.1 (alpha) <-[1.0.2 (alpha, stable)]
 * When 1.0.2 in alpha and stable channels is added replacing 1.0.1, 1.0.1's presence will only be marked as expected on
 * the alpha channel, not on the inherited stable channel.
  */
-// expectedGraphBundles returns a set of package-channel-bundle tuples that MUST be present following an add.
 func expectedGraphBundles(imagesToAdd []*registry.Bundle, graphLoader registry.GraphLoader, overwrite bool) (map[string]*registry.Package, error) {
 	expectedBundles := map[string]*registry.Package{}
 	for _, bundle := range imagesToAdd {
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -128,6 +128,8 @@ func unpackImage(ctx context.Context, reg image.Registry, ref image.Reference) (
 
 func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode, overwrite bool) error {
 	unpackedImageMap := make(map[image.Reference]string, 0)
+	overwrittenBundles := map[string][]string{}
+	var imagesToAdd []*registry.Bundle
 	for _, ref := range refs {
 		to, from, cleanup, err := unpackImage(ctx, reg, ref)
 		if err != nil {
@@ -135,16 +137,14 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 		}
 		unpackedImageMap[to] = from
 		defer cleanup()
-	}
 
-	overwriteImageMap := make(map[string]map[image.Reference]string, 0)
-	if overwrite {
-		// find all bundles that are attempting to overwrite
-		for to, from := range unpackedImageMap {
-			img, err := registry.NewImageInput(to, from)
-			if err != nil {
-				return err
-			}
+		img, err := registry.NewImageInput(to, from)
+		if err != nil {
+			return err
+		}
+		imagesToAdd = append(imagesToAdd, img.Bundle)
+
+		if overwrite {
 			overwritten, err := querier.GetBundlePathIfExists(ctx, img.Bundle.Name)
 			if err != nil {
 				if err == registry.ErrBundleImageNotInDatabase {
@@ -155,57 +155,28 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
 			if overwritten == "" {
 				return fmt.Errorf("index add --overwrite-latest is only supported when using bundle images")
 			}
-			// get all bundle paths for that package - we will re-add these to regenerate the graph
-			bundles, err := querier.GetBundlesForPackage(ctx, img.Bundle.Package)
-			if err != nil {
-				return err
+			if len(overwrittenBundles[img.Bundle.Package]) == 0 {
+				overwrittenBundles[img.Bundle.Package] = []string{}
 			}
-			type unpackedImage struct {
-				to      image.Reference
-				from    string
-				cleanup func()
-				err     error
-			}
-			unpacked := make(chan unpackedImage)
-			for bundle := range bundles {
-				// parallelize image pulls
-				go func(bundle registry.BundleKey, img *registry.ImageInput) {
-					if bundle.CsvName != img.Bundle.Name {
-						to, from, cleanup, err := unpackImage(ctx, reg, image.SimpleReference(bundle.BundlePath))
-						unpacked <- unpackedImage{to: to, from: from, cleanup: cleanup, err: err}
-					} else {
-						unpacked <- unpackedImage{to: to, from: from, cleanup: func() { return }, err: nil}
-					}
-				}(bundle, img)
-			}
-			if _, ok := overwriteImageMap[img.Bundle.Package]; !ok {
-				overwriteImageMap[img.Bundle.Package] = make(map[image.Reference]string, 0)
-			}
-			for i := 0; i < len(bundles); i++ {
-				unpack := <-unpacked
-				if unpack.err != nil {
-					return unpack.err
-				}
-				overwriteImageMap[img.Bundle.Package][unpack.to] = unpack.from
-				if _, ok := unpackedImageMap[unpack.to]; ok {
-					delete(unpackedImageMap, unpack.to)
-				}
-				defer unpack.cleanup()
-			}
+			overwrittenBundles[img.Bundle.Package] = append(overwrittenBundles[img.Bundle.Package], img.Bundle.Name)
 		}
 	}
 
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap, overwriteImageMap, overwrite)
-	if err := populator.Populate(mode); err != nil {
+	expectedBundles, err := expectedGraphBundles(imagesToAdd, graphLoader, overwrite)
+	if err != nil {
+
 		return err
-	}
 
-	for _, imgMap := range overwriteImageMap {
-		for to, from := range imgMap {
-			unpackedImageMap[to] = from
-		}
 	}
-	return checkForBundles(ctx, querier.(*sqlite.SQLQuerier), graphLoader, unpackedImageMap)
+	populator := registry.NewDirectoryPopulator(loader, graphLoader, querier, unpackedImageMap, overwrittenBundles, overwrite)
+
+	if err := populator.Populate(mode); err != nil {
+
+		return err
+
+	}
+	return checkForBundles(ctx, querier.(*sqlite.SQLQuerier), graphLoader, expectedBundles)
+
 }
 
 type DeleteFromRegistryRequest struct {
@@ -430,55 +401,10 @@ func checkForBundlePaths(querier registry.GRPCQuery, bundlePaths []string) ([]st
 	return found, missing, nil
 }
 
-// packagesFromUnpackedRefs creates packages from a set of unpacked ref dirs without their upgrade edges.
-func packagesFromUnpackedRefs(bundles map[image.Reference]string) (map[string]registry.Package, error) {
-	graph := map[string]registry.Package{}
-	for to, from := range bundles {
-		b, err := registry.NewImageInput(to, from)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse unpacked bundle image %s: %v", to, err)
-		}
-		v, err := b.Bundle.Version()
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse version for %s (%s): %v", b.Bundle.Name, b.Bundle.BundleImage, err)
-		}
-		key := registry.BundleKey{
-			CsvName:    b.Bundle.Name,
-			Version:    v,
-			BundlePath: b.Bundle.BundleImage,
-		}
-		if _, ok := graph[b.Bundle.Package]; !ok {
-			graph[b.Bundle.Package] = registry.Package{
-				Name:     b.Bundle.Package,
-				Channels: map[string]registry.Channel{},
-			}
-		}
-		for _, c := range b.Bundle.Channels {
-			if _, ok := graph[b.Bundle.Package].Channels[c]; !ok {
-				graph[b.Bundle.Package].Channels[c] = registry.Channel{
-					Nodes: map[registry.BundleKey]map[registry.BundleKey]struct{}{},
-				}
-			}
-			graph[b.Bundle.Package].Channels[c].Nodes[key] = nil
-		}
-	}
-
-	return graph, nil
-}
-
 // replaces mode selects highest version as channel head and
 // prunes any bundles in the upgrade chain after the channel head.
 // check for the presence of all bundles after a replaces-mode add.
-func checkForBundles(ctx context.Context, q *sqlite.SQLQuerier, g registry.GraphLoader, bundles map[image.Reference]string) error {
-	if len(bundles) == 0 {
-		return nil
-	}
-
-	required, err := packagesFromUnpackedRefs(bundles)
-	if err != nil {
-		return err
-	}
-
+func checkForBundles(ctx context.Context, q *sqlite.SQLQuerier, g registry.GraphLoader, required map[string]*registry.Package) error {
 	var errs []error
 	for _, pkg := range required {
 		graph, err := g.Generate(pkg.Name)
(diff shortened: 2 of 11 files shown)


pkg/sqlite/load.go
```diff
@@ -1435,35 +1476,94 @@ func (s *sqlLoader) DeprecateBundle(path string) error {
...
+	// At a branch point, a bundle will have a different channel list to the one it gets replaced by.
+	// Removing the entire tail for deprecating 1.1.3-rc will make the replaces value for 1.1.2 invalid,
+	// so all bundles till the last branch point for the tail bundle set must only be deprecated, not removed.
+	//
+	// multiple bundles replacing a deprecated bundle with a skipped edge requires each branch to be handled separately,
+	// 1.1.2 (alpha,1.1.x) <------------- 1.1.0 (alpha,stable)
+	//   ^   ^-...- 1.1.2-rc1 (alpha) <---/    /
+	//    \-..- 1.1.1 (1.1.x,stable,alpha) <--/
+	// deprecating 1.1.2 will deprecate 1.1.1 and 1.1.0 but truncate 1.1.2-rc-1 in the example above.
+
```
camilamacedo86, line 1491: Question: When we only set deprecate what is the OLM behaviour? Is that only ignored in the upgrade?
 What is the OLM behaviour if the head of the channel is flagged as deprecated? Would I still be able to install and see the bundle and channel?

pkg/sqlite/migrations/013_rm_truncated_deprecations.go
```diff
@@ -16,11 +16,11 @@ var rmTruncatedDeprecationsMigration = &Migration{
 	Id: RmTruncatedDeprecationsMigrationKey,
 	Up: func(ctx context.Context, tx *sql.Tx) error {
 
-		// Delete deprecation history for all bundles that no longer exist in the channel_entries table
+		// Delete deprecation history for all bundles that no longer exist in the operatorbundle table
 		// These bundles have been truncated by more recent deprecations and would only confuse future operations on an index;
 		// e.g. adding a previously truncated bundle to a package removed via `opm index|registry rm` would lead to that bundle
 		// being deprecated
-		_, err := tx.ExecContext(ctx, `DELETE FROM deprecated WHERE deprecated.operatorbundle_name NOT IN (SELECT DISTINCT deprecated.operatorbundle_name FROM (deprecated INNER JOIN channel_entry ON deprecated.operatorbundle_name = channel_entry.operatorbundle_name))`)
+		_, err := tx.ExecContext(ctx, `DELETE FROM deprecated WHERE deprecated.operatorbundle_name NOT IN (SELECT DISTINCT deprecated.operatorbundle_name FROM (deprecated INNER JOIN operatorbundle ON deprecated.operatorbundle_name = operatorbundle.name))`)
```
camilamacedo86, line 23: Seems that would also solve: https://bugzilla.redhat.com/show_bug.cgi?id=2001954

pkg/sqlite/load.go
```diff
@@ -1465,14 +1535,12 @@ func (s *sqlLoader) DeprecateBundle(path string) error {
 	if err != nil {
 		return err
 	}
-
 	// Clean up the deprecated table by dropping all truncated bundles
 	// (see pkg/sqlite/migrations/013_rm_truncated_deprecations.go for more details)
-	_, err = tx.Exec(`DELETE FROM deprecated WHERE deprecated.operatorbundle_name NOT IN (SELECT DISTINCT deprecated.operatorbundle_name FROM (deprecated INNER JOIN channel_entry ON deprecated.operatorbundle_name = channel_entry.operatorbundle_name))`)
+	_, err = tx.Exec(`DELETE FROM deprecated WHERE deprecated.operatorbundle_name NOT IN (SELECT DISTINCT deprecated.operatorbundle_name FROM (deprecated INNER JOIN operatorbundle ON deprecated.operatorbundle_name = operatorbundle.name))`)
```
camilamacedo86, line 1540: seems that would also solved : https://bugzilla.redhat.com/show_bug.cgi?id=2001954

pkg/sqlite/load.go
```diff
@@ -1279,121 +1279,141 @@ func (s *sqlLoader) addBundleProperties(tx *sql.Tx, bundle *registry.Bundle) err
...
+	row := tx.QueryRow(getDefaultChannelHead, head)
+	if row == nil {
+		return nil, fmt.Errorf("could not find default channel head for %s", head)
+	}
+	var defaultChannelHead sql.NullString
+	err = row.Scan(&defaultChannelHead)
+	if err != nil {
+		return nil, fmt.Errorf("error getting default channel head for %s: %v", head, err)
+	}
+	if !defaultChannelHead.Valid || len(defaultChannelHead.String) == 0 {
```
camilamacedo86, line 1361: Would that also solve https://bugzilla.redhat.com/show_bug.cgi?id=2001940?

pkg/lib/registry/registry.go
```diff
@@ -128,23 +128,40 @@ func unpackImage(ctx context.Context, reg image.Registry, ref image.Reference) (
 
 func populate(ctx context.Context, loader registry.Load, graphLoader registry.GraphLoader, querier registry.Query, reg image.Registry, refs []image.Reference, mode registry.Mode, overwrite bool) error {
 	unpackedImageMap := make(map[image.Reference]string, 0)
+	expectedBundles := make(map[string]registry.Package)
+	overwrittenBundles := make(map[string][]string)
```
njhale, line 132: In this case `make` is semantically equivalent to `map[string]registry.Package{}`

pkg/lib/registry/registry.go
```diff
@@ -128,23 +128,40 @@ func unpackImage(ctx context.Context, reg image.Registry, ref image.Reference) (
...
-	if overwrite {
-		// find all bundles that are attempting to overwrite
-		for to, from := range unpackedImageMap {
-			img, err := registry.NewImageInput(to, from)
+		img, err := registry.NewImageInput(to, from)
+		if err != nil {
+			return err
+		}
+		if _, ok := expectedBundles[img.Bundle.Package]; !ok {
+			pkg, err := graphLoader.Generate(img.Bundle.Package)
```
njhale, line 146: does the behavior of graphLoader actually match the real loading behavior? I don't think it's actually used for replaces.

pkg/sqlite/migrations/013_rm_truncated_deprecations.go
```diff
@@ -16,11 +16,11 @@ var rmTruncatedDeprecationsMigration = &Migration{
 	Id: RmTruncatedDeprecationsMigrationKey,
 	Up: func(ctx context.Context, tx *sql.Tx) error {
 
-		// Delete deprecation history for all bundles that no longer exist in the channel_entries table
+		// Delete deprecation history for all bundles that no longer exist in the operatorbundle table
```
kevinrizza, line 19: Why are we modifying an existing migration rather than adding a new migration?

ankitathomas: This change is solely for the purpose of keeping around entries in the deprecated table for deprecated channel heads. (otherwise, we have just an operatorbundle table entry for them). If we're adding another migration for this, it would be too late: the deprecated channel heads would already have been removed from everything but the operatorbundle table

pkg/sqlite/load_test.go
```diff
@@ -590,8 +655,9 @@ func TestGetTailFromBundle(t *testing.T) {
 			},
 			expected: expected{
 				err: nil,
-				tail: []string{
-					"csv-c",
+				tail: map[string]tailBundle{
+					"csv-b": {name: "csv-b", channels: []string{"alpha", "stable"}, replaces: []string{"csv-c"}, replacedBy: []string{"csv-a"}},
+					"csv-c": {name: "csv-c", channels: []string{"alpha", "stable"}, replacedBy: []string{"csv-b"}},
```
joelanford, line 660: (possibly answered somewhere else that i haven't seen yet...)

Why does the tail contain the requested bundle itself now? In the previous commit, it looks like we start with the bundle replaced by the requested bundle.

joelanford (commented):
I've made it through the "handle branchpoint deprecation" commit so far. And so far it looks great! Very thorough tests!

I'll pick up with the other two commits tomorrow.

pkg/lib/registry/registry.go
```diff
@@ -155,57 +155,28 @@ func populate(ctx context.Context, loader registry.Load, graphLoader registry.Gr
...
-				if unpack.err != nil {
-					return unpack.err
-				}
-				overwriteImageMap[img.Bundle.Package][unpack.to] = unpack.from
-				if _, ok := unpackedImageMap[unpack.to]; ok {
-					delete(unpackedImageMap, unpack.to)
-				}
-				defer unpack.cleanup()
+			if len(overwrittenBundles[img.Bundle.Package]) == 0 {
+				overwrittenBundles[img.Bundle.Package] = []string{}
```
joelanford, line 159: This is unnecessary. `append` handles allocating the slice if it is nil.

pkg/lib/registry/registry_test.go
```diff
@@ -565,53 +478,63 @@ func TestCheckForBundles(t *testing.T) {
...
-						if _, ok := overwriteRefs[img.Bundle.Package]; ok {
-							overwriteRefs[img.Bundle.Package] = map[image.Reference]string{}
+
+						// bundles to remove for overwrite. Only one per package is permitted.
+						if step.action == actionOverwrite {
+							img, err := registry.NewImageInput(image.SimpleReference(name), dir)
+							require.NoError(t, err)
+							if _, ok := overwriteRefs[img.Bundle.Package]; ok {
+								overwriteRefs[img.Bundle.Package] = make([]string, 0)
+							}
```
joelanford, line 525: Same comment as above: there's no need to initialize the slice here.

pkg/sqlite/load_test.go
```diff
@@ -668,6 +735,57 @@ func TestGetTailFromBundle(t *testing.T) {
 				tail: nil,
 			},
 		},
+		{
+			/*
+				0.1.2 <- 0.1.1 <- 0.1.0
+						   V (skips)
+				1.1.2 <- 1.1.1 <- 0.1.0
```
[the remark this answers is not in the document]
ankitathomas: thanks! That was a typo

pkg/lib/registry/registry.go
```diff
@@ -523,3 +449,59 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
...
+		}
+		for _, c := range bundle.Channels {
+			if _, ok := pkg.Channels[c]; !ok {
+				pkg.Channels[c] = registry.Channel{
+					Nodes: map[registry.BundleKey]map[registry.BundleKey]struct{}{},
+				}
+			}
+			// This can miss out on some channels, when a new bundle has channels that the one it replaces does not.
+			// eg: When bundle A in channel A replaces bundle B in channel B is added, bundle B is also added to channel A
+			// but it is only expected to be in channel B, presence in channel A will be ignored
```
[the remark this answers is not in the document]
ankitathomas: It will still upgrade to bundle A. We wouldn't verify the bundle's presence on that specific channel though, since channel A would be newly added to bundle B when bundle A is added.

pkg/registry/populator.go
```diff
@@ -20,22 +20,22 @@ type Dependencies struct {
...
-		graphLoader:     graphLoader,
-		querier:         querier,
-		imageDirMap:     imageDirMap,
-		overwriteDirMap: overwriteDirMap,
-		overwrite:       overwrite,
+		loader:            loader,
+		graphLoader:       graphLoader,
+		querier:           querier,
+		imageDirMap:       imageDirMap,
+		overwrittenImages: overwrittenImages,
```
kevinrizza, line 37: nit: do we need a separate `overwrite` variable anymore? Isn't `len(overwrittenImages) == 0` sufficient?

pkg/registry/populator.go
```diff
@@ -145,25 +132,34 @@ func (i *DirectoryPopulator) globalSanityCheck(imagesToAdd []*ImageInput) error
...
 	switch mode {
 	case ReplacesMode:
-		for pkg := range i.overwriteDirMap {
-			// TODO: If this succeeds but the add fails there will be a disconnect between
-			// the registry and the index. Loading the bundles in a single transactions as
-			// described above would allow us to do the removable in that same transaction
-			// and ensure that rollback is possible.
-			if err := i.loader.RemovePackage(pkg); err != nil {
-				return err
+		// TODO: If this succeeds but the add fails there will be a disconnect between
```
kevinrizza, line 143: re this todo:

I think we also already don't really have transactional safety, so this shouldn't be a dealbreaker. most users will pull out the db from an image anyway, so that should be trivial to retry. imo this is something we should avoid implementing given the complexity of threading the transaction through without the right abstractions already in place

njhale (commented):
Wow -- thanks for working on this @ankitathomas!

pkg/registry/interface.go
```diff
@@ -15,6 +15,7 @@ type Load interface {
 	RemoveStrandedBundles() error
 	DeprecateBundle(path string) error
 	ClearNonHeadBundles() error
+	RemoveOverwrittenChannelHead(pkg, bundle string) error
```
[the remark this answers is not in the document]
joelanford: If we later add another implementation of `Load` that is not also a `HeadOverwriter`, will this code still produce expected output?

pkg/sqlite/load_test.go
```diff
@@ -825,3 +820,211 @@ func TestAddBundlePropertiesFromAnnotations(t *testing.T) {
...
+			require.NoError(t, err)
+
+			for _, bundle := range tt.fields.bundles {
+				// Throw away any errors loading bundles (not testing this)
+				store.AddOperatorBundle(bundle)
+			}
+
+			for _, pkg := range tt.fields.pkgs {
+				// Throw away any errors loading packages (not testing this)
+				store.AddPackageChannels(pkg)
```
joelanford, line 1115: Do we expect errors to occur? If not, can we `require.NoError` each of them. If we do expect errors, do we know which ones?

It worries me a little that errors happening here may cause false positives or false negatives if we later make changes in the code under test.

pkg/sqlite/load_test.go
```diff
@@ -825,3 +820,211 @@ func TestAddBundlePropertiesFromAnnotations(t *testing.T) {
...
+			args: args{
+				bundle: "csv-a",
+				pkg:    "pkg-0",
+			},
+			expected: expected{
+				err: nil,
+				bundles: map[string]struct{}{
+					"pkg-0/a/csv-b": {},
+					"pkg-0/b/csv-b": {},
+					"pkg-0/a/csv-c": {},
```
joelanford, line 1090: So `csv-b` and `csv-c` end up in channel A even after bundle `a` is removed. In this state, there would be an error because `csv-b` and `csv-c` are both in channel `a`, but they're in different replaces chains.

Is the assumption that the new overwrite bundle will resolve this situation? Or do we just rebuild the graph from scratch based on highest semver, and this intermediate state just doesn't matter?

pkg/sqlite/load_test.go
```diff
@@ -424,6 +424,42 @@ func TestDeprecationAwareLoader(t *testing.T) {
 				},
 			},
 		},
+		{
+			description: "DeprecateTruncateRemoveDeprecatedChannelHeadOnPackageRemoval",
+			fields: fields{
+				bundles: []*registry.Bundle{
+					withBundleImage("quay.io/my/bundle-a", newBundle(t, "csv-a", "pkg-0", []string{"a"}, newUnstructuredCSV(t, "csv-a", ""))),
+					withBundleImage("quay.io/my/bundle-aa", newBundle(t, "csv-aa", "pkg-0", []string{"a"}, newUnstructuredCSV(t, "csv-aa", "csv-a"))),
+					withBundleImage("quay.io/my/bundle-aaa", newBundle(t, "csv-b", "pkg-0", []string{"b"}, newUnstructuredCSV(t, "csv-b", "csv-a"))),
```
joelanford: Is there a reason the image ref is `aaa` but the name is `b`?
```suggestion
					withBundleImage("quay.io/my/bundle-aaa", newBundle(t, "csv-aaa", "pkg-0", []string{"b"}, newUnstructuredCSV(t, "csv-aaa", "csv-a"))),
```

pkg/lib/registry/registry_test.go
```diff
@@ -669,3 +592,87 @@ func TestDeprecated(t *testing.T) {
...
+					}: nil,
+				},
+			},
+		},
+	}
+	testBundleDifferentChannelPkg := &registry.Package{
+		Name: "testPkg",
+		Channels: map[string]registry.Channel{
+			"alpha": {
+				Nodes: map[registry.BundleKey]map[registry.BundleKey]struct{}{
```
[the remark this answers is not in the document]
ankitathomas: That looks like it just avoids duplicate edges, or possibly it's used for verifying specific replaces edges?

pkg/registry/interface.go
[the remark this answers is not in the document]
ankitathomas: That version of `HeadOverwriter` would have to implement the Add correctly to not need this `RemoveOverwrittenHeads`

pkg/sqlite/load_test.go
ankitathomas: switching it to `require.NoError` seems trivial and the right thing to do

pkg/sqlite/load_test.go
ankitathomas: This is an issue when you have a new bundle overwriting an existing one in a new set of channels but not keeping the same edges as a subset of its new edges. The channel head selection heuristic would drop whichever one of b/c was of a lower semver from the a/b channels, but since we're checking for dropped bundles, we'd complain on either case.

So we'd be adding another hidden requirement to overwrites: new bundles must have edges to their old bundles on at least the channels they previously had those edges on.

If overwritten bundles are mainly for modifying bundle data rather than the graph metadata, this is a non-issue. If they do modify graph metadata, we'd at least warn when bundles get dropped.

pkg/sqlite/load_test.go
ankitathomas: Typo, good catch!

pkg/lib/registry/registry_test.go
```diff
@@ -565,53 +478,63 @@ func TestCheckForBundles(t *testing.T) {
...
-						require.NoError(t, err)
-						return
-					}
-					require.EqualError(t, err, tt.wantErr.Error())
+						// List of package-channel-bundle tuples expected to be in the graph.
+						// Does not check for replaces/skips between the nodes.
+						var pkg *registry.Package
+						var ok bool
+						if pkg, ok = expectedBundles[b.annotations.PackageName]; !ok {
+							if pkg, err = graphLoader.Generate(b.annotations.PackageName); err != nil {
```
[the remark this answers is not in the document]
ankitathomas: That part is actually unnecessary unless we're starting will already populated indices, which we aren't for this test.

pkg/lib/registry/registry.go
```diff
@@ -523,3 +449,59 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
 	}
 	return false, nil
 }
+
+func expectedGraphBundles(imagesToAdd []*registry.Bundle, graphLoader registry.GraphLoader, overwrite bool) (map[string]*registry.Package, error) {
```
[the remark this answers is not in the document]
ankitathomas: Added a description for the function

pkg/sqlite/load_test.go
ankitathomas: It does, it was a convenient way to know where to start channel removal from.

pkg/registry/populator_test.go
```diff
@@ -1204,6 +1302,215 @@ func TestAddAfterDeprecate(t *testing.T) {
...
+			},
+			expected: expected{
+				err: nil,
+				remaining: []string{
+					"quay.io/test/testpkg.v0.1.0/0.1",
+					"quay.io/test/testpkg.v0.1.0/0.2",
+					"quay.io/test/testpkg.v0.1.1/0.1",
+					"quay.io/test/testpkg.v0.2.0/0.2",
+					"quay.io/test/testpkg.v0.2.1/0.2",
+					"quay.io/test/testpkg.v0.2.2/0.2",
```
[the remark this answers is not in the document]
ankitathomas: Added this as a new test case, OverwriteDeprecatedChannelHeadWithReplaces

pkg/sqlite/load.go
ankitathomas: Solved by https://github.com/operator-framework/operator-registry/pull/780/files

pkg/sqlite/load.go
ankitathomas: Solved by https://github.com/operator-framework/operator-registry/pull/780/files

pkg/sqlite/migrations/013_rm_truncated_deprecations.go
ankitathomas: This migration is to ensure that deprecated channel head will not get dropped, which they would with the channel table losing the channel entry https://github.com/operator-framework/operator-registry/blob/29e90dea6a1c2ef12075ca866284cd8f267c2bba/pkg/sqlite/load.go#L1448
The channel head+channel_entry had to be dropped so the deprecated bundle wouldn't show up in console and the graph rebuilding from channel heads means any free entry in the channel entry table would be dropped anyway.

This method did bring up an interesting new bug: removing a channel head after deprecating it causes it to persist in the operatorbundle table even on package removal, preventing the bundle from ever being overwritten/added again

The migration should be safe though, we'd retain more deprecated entries if a migration was re-done, but it shouldn't affect any previously migrated index. Assuming no one was using 'deprecate channel head' as an equivalent for 'delete channel head'

pkg/lib/registry/registry.go
```diff
@@ -523,3 +445,75 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
...
+* will be dropped from the graph following an add.
+* eg: 1.0.1 <- 1.0.1-new
+*
+* 1.0.1-new replaces 1.0.1 but will not be chosen as the channel head because of its non-empty pre-release version.
+* expectedGraphBundles gives a set of bundles (old bundles from the graphLoader and the newly added set of bundles from
+* imagesToAdd) that must be present following an add to ensure no bundle is dropped.
+*
+* Overwritten bundles will only be verified on the channels of the newly added version.
+* Any inherited channels due to addition of a new bundle on its tail bundles may not be verified
+* eg:  [1.0.2 (alpha, stable)] <- 1.0.1 (alpha)
```
njhale, line 460: The direction of the arrow in this example denotes "is replaced by", which is the opposite of the previous example's "replaces". I think we should keep the meaning uniform across examples.

pkg/lib/registry/registry.go
```diff
@@ -523,3 +445,75 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
 	}
 	return false, nil
 }
+
+/* opm index add drops bundles that replace a channel head, and since channel head selection heuristics
```
njhale, line 449: nit: go doc style asks that long function descriptions are put beneath the short function description.

(not super important because private methods aren't included in go docs anyway)

alpha/action/render_test.go
```diff
@@ -792,7 +792,7 @@ func generateSqliteFile(path string, imageMap map[image.Reference]string) error
 		return err
 	}
 
-	populator := registry.NewDirectoryPopulator(loader, graphLoader, dbQuerier, imageMap, nil, false)
+	populator := registry.NewDirectoryPopulator(loader, graphLoader, dbQuerier, imageMap, nil)
```
timflannagan, line 795: Completely unrelated, but this likely signals that we should be accepting some sort of configuration struct if we're accepting this number of argument, namely for readability as it's not immediately clear aside from the variable names what the function signature expects.

pkg/lib/registry/registry.go
```diff
@@ -523,3 +449,59 @@ func isDeprecated(ctx context.Context, q *sqlite.SQLQuerier, bundle registry.Bun
...
+				}
+			}
+		}
+		for _, c := range bundle.Channels {
+			if _, ok := pkg.Channels[c]; !ok {
+				pkg.Channels[c] = registry.Channel{
+					Nodes: map[registry.BundleKey]map[registry.BundleKey]struct{}{},
+				}
+			}
+			// This can miss out on some channels, when a new bundle has channels that the one it replaces does not.
```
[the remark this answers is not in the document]
timflannagan: I'll say this was largely my initial read as well - I'd likely defer to I'm missing some context. Any reasonable clarification that could be done here, or is the value add not worth it?

pkg/registry/populator.go
```diff
@@ -20,22 +20,20 @@ type Dependencies struct {
...
-	graphLoader     GraphLoader
-	querier         Query
-	imageDirMap     map[image.Reference]string
-	overwriteDirMap map[string]map[image.Reference]string
-	overwrite       bool
+	loader            Load
+	graphLoader       GraphLoader
+	querier           Query
+	imageDirMap       map[image.Reference]string
+	overwrittenImages map[string][]string
```
timflannagan, line 27: :+1: I largely think that including the type in the variable naming is an anti-pattern (as shown above).

pkg/registry/populator.go
njhale: agreed -- I think we should drop this comment

pkg/registry/populator_test.go
```diff
@@ -1160,7 +1165,111 @@ func TestDeprecatePackage(t *testing.T) {
 }
 
 func TestAddAfterDeprecate(t *testing.T) {
+	tmpdir, err := os.MkdirTemp(".", "add-after-deprecate-*")
```
njhale, line 1168: nit: omitting the `*` here would achieve the same result

njhale (approved):
/approve

Looks good. Thanks for sticking with this!

PR #768, opened 2021-08-28
Introduce olm.channel schema, remove channel metadata-related bundle properties

joelanford:

- Define `olm.channel` schema to specify channel membership and upgrade edges. For example:

    ```yaml
    ---
    schema: olm.channel
    package: foo
    name: stable
    entries:
      - name: foo.v0.0.1
      - name: foo.v0.0.2
      - name: foo.v0.1.0
        replaces: foo.v0.0.1
        skipRange: <0.1.0
        skips: 
          - foo.v0.0.2
    ```

- Remove properties on the `olm.bundle` blobs that are related to channel metadata. Specifically `olm.channel`, `olm.skips`, and `olm.skipRange` are removed.

- Conversions from the model to the file-based config format use the `olm.channel` schema.

- `opm render` no longer populates channel-related metadata when rendering bundle images.

The current schema overlooked the fact that bundle properties contained channel-level metadata, which goes directly against the design goals of file-based configs.

This PR corrects that issue by implementing a new `olm.channel` schema and removing the ability of `opm render` to pull this metadata from bundles created with imperative sqlite-based workflows in mind.

This is also an improvement on the the existing `olm.bundle` properties because:
1. `olm.skips` and `olm.skipRange` properties apply to all channels. This causes channel upgrade graphs to be defined by channel-specific replaces values from the `olm.channel` property and by the global skips and skipRange properties, which can make it extremely difficult to define usable upgrade graphs when multiple channels are involved.
2. It centralizes all of the channel membership and upgrade edge information in a single place, making it easier for humans to visualize and for computers to quickly parse and generate upgrade graphs.
3. It simplifies implementation of veneer APIs that focus on different upgrade graph definitions. Different channel and upgrade graph definitions can be built without touching any `olm.bundle` blobs. 

**Reviewer Checklist**


alpha/action/testdata/index-declcfgs/exp-latest/index.yaml
```diff
@@ -43,35 +49,14 @@ defaultChannel: stable
 name: baz
 schema: olm.package
 ---
-image: test.registry/baz-operator/baz-bundle:v1.0.1
```
joelanford, line 46: Hmm. Need to investigate why this is removed. Perhaps its because the only reason is showed up in the expected output is because of differences in the channel-related metadata?

alpha/declcfg/declcfg_to_model.go
```diff
@@ -39,67 +86,47 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
...
-					Name:    bundleChannel.Name,
-					Bundles: map[string]*model.Bundle{},
-				}
-				if bundleChannel.Name == defaultChannelName {
-					mpkg.DefaultChannel = pkgChannel
-				}
-				mpkg.Channels[bundleChannel.Name] = pkgChannel
-			}
-
-			// Parse version from the package property, falling back to the CSV's spec.version field.
```
joelanford, line 73: @estroz Was this introduced with the `opm diff` stuff? What was the purpose of falling back to the CSV `spec.version` field? The `olm.package` property is required to have a valid semver version in it for all bundles.

dmesser:
This might be out of scope for this particular change, but one positive aspect of having all this update-graph/channel-related metadata in the bundle is that it made adding a bundle to a catalog extremely straight forward, essentially a single step.

It seems we are moving towards a structure where, due to the separate of concerns, now multiple steps are needed in order to add a bundle to an existing set of bundles in a catalog / package definition: adding the bundle reference and then reference the version in the channel definitions and various graph controls.
I'd like us to keep in mind that the proverbial 80% of the catalog interactions are going to be adding a new version that replaces the last version (or potentially everything else) to the default catalog. That was one of the reasons we built the semver-mode for `opm`.  

What does this process look like in the future. Will it always be two/multiple steps?

joelanford:
> This might be out of scope for this particular change, but one positive aspect of having all this update-graph/channel-related metadata in the bundle is that it made adding a bundle to a catalog extremely straight forward, essentially a single step.

I think what this PR brings into focus is that operator authors are the owners of their index. The number of steps for an operator author hasn't changed.

**Before**
1. Set `replaces`, `skips`, and `skipRange` in their CSV
2. Build the bundle
3. Add the bundle to the index

**After**
1. Build the bundle
2. Add the bundle to the index
3. Set `replaces`, `skips`, and `skipRange` in the index

alpha/action/testdata/index-declcfgs/exp-latest/index.yaml
joelanford: Yep, just confirmed that `image: test.registry/baz-operator/baz-bundle:v1.0.1` is now identical in both `old/index.yaml` and `latest/index.yaml`.

On `master`, this bundle has an extra `olm.skips` property for `baz.v1.0.0` in `old/index.yaml` that is not present in `latest/index.yaml`

Since this skips field was moved to the `olm.channel` in this PR, the bundles are now identical and the diff correctly omits it from the result.

dmesser:
For most users of any pipelines that we currently entertain the process looks like this:

1. Merge a PR to your bundle repo branch (i.e. release a new operator version with updated replaces, skips, etc)

That's it, then a pipeline picks it up, builds the image and adds it to the index. Making it a 3-way process by default would regress on the simplicity part quite a bit. We likely want to have some automation like your channelizer. I think I don't want to create more noise in this PR review about this, let's move the discussion to https://issues.redhat.com/browse/OLM-2107

alpha/declcfg/declcfg_to_model.go
estroz: Oh perfect. In that case fallback is unnecessary.

estroz:
LGTM from `diff`'s perspective.

PR #770, opened 2021-09-01
migrate: add subcommand to migrate sqlite catalogs to file-based catalogs

joelanford:
Add `opm migrate` which will convert an sqlite-based catalog (image or db file) into a file-based catalog.

This command is different from the general `render` command in two important ways:
1. It works only on sqlite-based catalogs (this is necessary because it is unsafe to "migrate" file-based catalogs to file-based catalogs. `migrate` writes files in an opinionated location, which could break any relative file references that exist in the input file-based catalog.
2. It writes a full package-based directory structure in the output directory, such that each package in the catalog gets its own file (`render` concatenates everything together into one output stream) 

Note that this command is immediately deprecated because it works with sqlite catalogs, which are deprecated.

To help users migrate their existing indexes to a file-based catalog.

**Reviewer Checklist**

### DIFF
--- a/alpha/action/migrate_test.go
+++ b/alpha/action/migrate_test.go
@@ -1,0 +1,341 @@
+package action_test
+
+import (
+	"context"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/pkg/containertools"
+	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
+)
+
+func TestMigrate(t *testing.T) {
+	type spec struct {
+		name          string
+		migrate       action.Migrate
+		expectedFiles map[string]string
+		expectErr     error
+	}
+
+	sqliteBundles := map[image.Reference]string{
+		image.SimpleReference("test.registry/foo-operator/foo-bundle:v0.1.0"): "testdata/foo-bundle-v0.1.0",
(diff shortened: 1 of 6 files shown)


joelanford:
/hold

I need to add some test cases.

joelanford:
/hold cancel

Added tests

anik120 (commented):
won't it be better if we kept it under alpha? Gives us all the leverage to remove it when we think it's appropriate.

joelanford:
If we put it under `alpha` that kind of means YMMV with users and we don't have to provide support. And I think we need to provide support for this and give users confidence that migrations from sqlite to FBC works.

In this case, I think the deprecation warning is enough to give us the ability to remove this command at the same time that we remove the rest of the sqlite support.

alpha/action/migrate_test.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	sqliteBundles := map[image.Reference]string{
+		image.SimpleReference("test.registry/foo-operator/foo-bundle:v0.1.0"): "testdata/foo-bundle-v0.1.0",
+		image.SimpleReference("test.registry/foo-operator/foo-bundle:v0.2.0"): "testdata/foo-bundle-v0.2.0",
+		image.SimpleReference("test.registry/bar-operator/bar-bundle:v0.1.0"): "testdata/bar-bundle-v0.1.0",
+		image.SimpleReference("test.registry/bar-operator/bar-bundle:v0.2.0"): "testdata/bar-bundle-v0.2.0",
+	}
+
+	fooCatalog := `---
```
dinhxuanvu, line 35: I know we don't like local test files much but I actually prefer this to be a separate file than a long and hairy embedded text.

joelanford: I agree these interrupt the flow of the test as local variables. Rather than put them in separate testdata files where it's more difficult to tell where (and if) they're used, I made new functions at the bottom of the migrate_test.go file that just return them.

anik120 (approved):
/lgtm

PR #775, opened 2021-09-13
Exdx fix/prune filesystem

njhale:
Rebase and add finishing touches to #583

> **Description of the change:**
> 
> Pruning an index via opm index prune can fail when permissions on the filesystem in the container are read-only. Since the command copies all files out from the container it is prone to failure. Fora example, the red hat operators catalog is based on the UBI image, which contains some directories with elevated permissions (/root/* for example). Extract the entire contents of this container filesystem via docker cp will fail.


njhale:
Note: This makes a breaking API change, which was previously discussed and approved

see https://github.com/operator-framework/operator-registry/pull/583/files#r651838537 for more details

pkg/containertools/runner.go
```diff
@@ -121,34 +125,73 @@ func (r *ContainerCommandRunner) Build(dockerfile, tag string) error {
...
+// Unpack copies a file or directory from a local container image to a directory on the local filesystem.
 func (r *ContainerCommandRunner) Unpack(image, src, dst string) error {
-	args := r.argsForCmd("create", image, "")
-
-	command := exec.Command(r.containerTool.String(), args...)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
 
+	args := r.argsForCmd("create", image, "")
+	command := exec.CommandContext(ctx, r.containerTool.String(), args...)
```
timflannagan, line 134: :+1:

pkg/containertools/runner.go
```diff
@@ -121,34 +125,73 @@ func (r *ContainerCommandRunner) Build(dockerfile, tag string) error {
...
+		}
+	}()
+	defer func() {
+		if err := pipew.Close(); err != nil {
+			r.logger.Warnf("error closing command write pipe: %s", err)
+		}
+	}()
+
+	if err != nil {
+		return fmt.Errorf("error setting unpacking destination: %w", err)
```
timflannagan, line 159: What error is being caught here? The error captured in the defer statements?

pkg/containertools/untarer.go
```diff
@@ -0,0 +1,88 @@
...
+			return err
+		}
+
+		u.log.Debugf("writing %s to disk", abs)
+		n, err := io.Copy(file, reader)
+		if err != nil {
+			return err
+		}
+
+		if err := file.Close(); err != nil {
```
timflannagan, line 77: Does this need to live as a defer statement instead closer to the call site? It looks like we could leak file descriptors when the io.Copy call returns an error.

pkg/image/containerdregistry/registry_test.go
```diff
@@ -0,0 +1,41 @@
...
+			dest:   "my-db.db",
+			header: &tar.Header{Name: "my-db.db"},
+			match:  true,
+		},
+		{
+			dest:   "my-db.db",
+			header: &tar.Header{Name: "rocks-db.db"},
+			match:  false,
+		},
+		// TODO directory test cases
```
timflannagan, line 30: Is this still needed?

pkg/containertools/runner.go
njhale: ah, this is a leftover from refactoring. nice catch!

njhale: fixed

pkg/containertools/untarer.go
njhale: Fair point.

I moved this whole section into a helper method so we can ensure the close happens before we copy the next file (we don't want more than one open at a time).

pkg/image/containerdregistry/registry_test.go
njhale: Added some directory tests to this (and refactored some).

pkg/containertools/untarer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"testing"
+
+	"github.com/sirupsen/logrus"
+)
+
+var files = map[string]string{
+	"index.html": `<body>Hello!</body>`,
+	"lang.json":  `[{"code":"eng","name":"English"}]`,
+	"songs.txt":  `Claire de la lune, The Valkyrie, Swan Lake`,
+}
```
timflannagan, line 21: (nothing blocking obviously) this kind of structure feels like it could be baked into a table driven test instead.

timflannagan:
Throwing a hold just in case we want additional eyes on this PR.

/lgtm
/hold

timflannagan:
Hmm this seems related to the newly introduced prune behavior?

```
• Failure [12.719 seconds]
opm
/home/runner/work/operator-registry/operator-registry/test/e2e/opm_test.go:219
  using docker
  /home/runner/work/operator-registry/operator-registry/test/e2e/opm_test.go:489
    builds and manipulates bundle and index images [It]
    /home/runner/work/operator-registry/operator-registry/test/e2e/opm_test.go:260

    Unexpected error:
        <*fs.PathError | 0xc000b5ec30>: {
            Op: "open",
            Path: "index_tmp_805713973/database/index.db",
            Err: <syscall.Errno>0x2,
        }
        open index_tmp_805713973/database/index.db: no such file or directory
    occurred

    /home/runner/work/operator-registry/operator-registry/test/e2e/opm_test.go:290
```

njhale:
stale

PR #776, opened 2021-09-13
test/e2e: Ignore stdout output when checking if podman/docker are present

timflannagan:
Ignore any stdout output that's returned when attempting to check
whether podman/docker are present when running the e2e suite.

There's little need for printing out the `docker --help` or `podman info` stdout output. That command was previously updated to capture stderr to help identify any issues when running the e2e suite locally:

```log
GOFLAGS="-mod=vendor" go run github.com/onsi/ginkgo/ginkgo --v --randomizeAllSpecs --randomizeSuites --race  -tags=json1,kind ./test/e2e -- 
INFO: Using builder at '/usr/bin/docker'


Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers

Options:
      --config string      Location of client config files (default
                           "/home/runner/.docker")
  -c, --context string     Name of the context to use to connect to the
                           daemon (overrides DOCKER_HOST env var and
                           default context set with "docker context use")
  -D, --debug              Enable debug mode
  -H, --host list          Daemon socket(s) to connect to
  -l, --log-level string   Set the logging level
                           ("debug"|"info"|"warn"|"error"|"fatal")
                           (default "info")
      --tls                Use TLS; implied by --tlsverify
      --tlscacert string   Trust certs signed only by this CA (default
                           "/home/runner/.docker/ca.pem")
      --tlscert string     Path to TLS certificate file (default
                           "/home/runner/.docker/cert.pem")
      --tlskey string      Path to TLS key file (default
                           "/home/runner/.docker/key.pem")
      --tlsverify          Use TLS and verify the remote
  -v, --version            Print version information and quit
...
```

**Reviewer Checklist**

### DIFF
--- a/test/e2e/opm_test.go
+++ b/test/e2e/opm_test.go
@@ -487,7 +487,7 @@ var _ = Describe("opm", func() {
 	}
 
 	Context("using docker", func() {
-		cmd := exec.Command("docker")
+		cmd := exec.Command("docker", "--version")
 		cmd.Stderr = os.Stderr
 		if err := cmd.Run(); err != nil {
 			GinkgoT().Logf("container tool docker not found - skipping docker-based opm e2e tests: %v", err)
@@ -497,7 +497,7 @@ var _ = Describe("opm", func() {
 	})
 
 	Context("using podman", func() {
-		cmd := exec.Command("podman", "info")
+		cmd := exec.Command("podman", "--version")
 		cmd.Stderr = os.Stderr
 		if err := cmd.Run(); err != nil {
 			GinkgoT().Logf("container tool podman not found - skipping podman-based opm e2e tests: %v", err)


timflannagan:
Link to a job that has this output: https://github.com/operator-framework/operator-registry/pull/775/checks?check_run_id=3588728523.

test/e2e/opm_test.go
```diff
@@ -489,6 +489,7 @@ var _ = Describe("opm", func() {
 	Context("using docker", func() {
 		cmd := exec.Command("docker")
 		cmd.Stderr = os.Stderr
+		cmd.Stdout = nil
```
njhale, line 492: Doesn't look like we're reading `os.Stderr` for this command, so we might as well do both since an `error` returned by `cmd.Run()` will indicate a non-zero exit code regardless -- see https://pkg.go.dev/os/exec#Cmd.Run

P.S. `cmd.Stdout = nil` is the same as `cmd.Stdout = os.DevNull` -- see https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/os/exec/exec.go;l=102-103

njhale (changes requested):

test/e2e/opm_test.go
timflannagan: In the case that cmd.Run returns an error, a non-zero exit code will only be logged when handling that error:

```bash
package main

func main() { ... }

func run() error {
	cmd := exec.Command("podman")
	cmd.Stderr = os.Stderr
	if err := cmd.Run(); err != nil {
		return err
	}
	return nil
}
```

```bash
$ go run main.go
Error: missing command 'podman COMMAND'
Try 'podman --help' for more information.
exit status 125
```

When removing that call to `cmd.Stderr = os.Stderr` the following is returned instead, which makes debugging local testing environments more difficult:

```bash
$ go run main.go
exit status 125
```

Weirdly enough, it looks like running the `docker` command logs everything to stderr locally, despite the return code being 0 - podman seems to handle logging errors to stderr correctly.

test/e2e/opm_test.go
```diff
@@ -487,7 +487,7 @@ var _ = Describe("opm", func() {
 	}
 
 	Context("using docker", func() {
-		cmd := exec.Command("docker")
+		cmd := exec.Command("docker", "--version")
```
timflannagan, line 490: Ended up just using `docker --version` and `podman --version` as the check for whether these tools are present locally (e.g. in $PATH) to avoid the skew between the two checks. For whatever reason, running `docker` locally without any arguments returns a zero return code, but logs the output to stderr in the e2e suite logs.

timflannagan:
I'd like to dive into whether stderr output is produced by docker/podman locally like last time.

/hold

PR #779, opened 2021-09-14
add warning log entry for removing non-exist package

### LINKED ISSUE #675: Removing a non-existent operator from an index image does not produce an error or a warning message
When supplying a list of non-existent operator(s) from an index image to `opm index rm`, the command results in a successful run and no logs are produced that warn those supplied operators (s) don't exist.

This is tracking the UX improvement to at least warn users that these operator(s) don't exist.

Mirror of https://bugzilla.redhat.com/show_bug.cgi?id=1913644

akihikokuroda:
Adding a warning log entry when `opm index rm` is issued for a non-existing package

Closes #675 

**Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -942,6 +942,9 @@ func (s *sqlLoader) RemovePackage(packageName string) error {
 		if err != nil {
 			return err
 		}
+		if len(csvNames) == 0 {
+			return fmt.Errorf("no package found for packagename %s", packageName)
+		}
 		for _, csvName := range csvNames {
 			if err := s.rmBundle(tx, csvName); err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


pkg/sqlite/load.go
```diff
@@ -942,6 +943,9 @@ func (s *sqlLoader) RemovePackage(packageName string) error {
 		if err != nil {
 			return err
 		}
+		if len(csvNames) == 0 {
+			logrus.WithFields(logrus.Fields{"pkg": packageName}).Warnf("package doesn't exist")
```
kevinrizza, line 947: Should this be an error instead of a warning?

kevinrizza: We should probably just return here.

timflannagan: Do we need some testing if we're changing behavior here?

akihikokuroda: Thanks for reviewing.  I'll change the warning to the error and return here.  I'll look into the test change, too.

kevinrizza: I think the linter error is just fixed by running `go-fmt`

akihikokuroda:
I changed to return the error to caller so that the caller can handle the error.  I also updated the test.  Thanks!

PR #783, opened 2021-09-18
opm serve: pre-compute and store api.Bundles to disk; requests read from disk 

joelanford:

When serving the registry from the model, pre-compute the `api.Bundle` blobs, save them to disk, and reduce the size of the in-memory model to contain only enough to answer requests by looking up and reading bundles from disk.

This change provides a major memory optimization. Using a real-world file-based catalog (created by migrating a production sqlite-based catalog), the steady state RSS is reduced from 373Mb to 55Mb, an 85% decrease.

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ static: build
 
 .PHONY: unit
 unit:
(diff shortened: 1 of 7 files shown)


anik120 (approved):
/lgtm

Issue #703, opened 2021-07-09
OPM doesn't error/warn when given bad cli arguments


eriknelson:
This is a result of some bad automation on our part, but I would have expected OPM to expliclity error and/or warn that the cli arguments were junk:

```
./opm index add -c docker --bundles quay.io/konveyor/mig-operator-bundle:release-1.3.2,quay.io/konveyor/mig-operator-bundle:release-1.4.5 release-1.4.6,quay.io/konveyor/mig-operator-bundle:release-1.5.0,quay.io/konveyor/mig-operator-bundle:latest --tag quay.io/konveyor/mig-operator-index:latest
```
Notice the bad argument behind `1.4.5 1.4.6`

This causes it to silently only build with 1.3 and 1.4:

```
INFO[0000] building the index                            bundles="[quay.io/konveyor/mig-operator-bundle:release-1.3.2 quay.io/konveyor/mig-operator-bundle:release-1.4.5]"
INFO[0000] running /usr/bin/docker pull quay.io/konveyor/mig-operator-bundle:release-1.3.2  bundles="[quay.io/konveyor/mig-operator-bundle:release-1.3.2 quay.io/konveyor/mig-operator-bundle:release-1.4.5]"
```
No error.

Would have expected a complaint or a warning with this.

timflannagan:
There's a similar gap in the opm UX when attempting to _remove_ a specific, non-existent package from an index and the result is no error or warning message logged. IIRC, there's a BZ on the OLM component that tracks this downstream.

exdx:
Seems like opm thinks those are bad arguments, but ignored them instead of throwing an error.

joelanford:
It looks like `opm` may be interpretting `release-1.4.6,quay.io/konveyor/mig-operator-bundle:release-1.5.0,quay.io/konveyor/mig-operator-bundle:latest` as a separate CLI argument instead of a flag value.

We likely need to add some tighter validation in the `opm` cobra commands that have explicit checks for the number (and possibly shape) of non-flag arguments. (e.g. `cobra.ExactArgs(1)`)

PR #786, opened 2021-09-22
Update opm commands to specifcy if they don't take args

tylerslaton:
Updating the `opm` binary to specify when a command takes arguments or not. This will prevent typos from successfully running should a user accidentally type args into a command that does not accept them.

[Issue 703](https://github.com/operator-framework/operator-registry/issues/703)

**Reviewer Checklist**

Closes 703

### DIFF
--- a/cmd/opm/alpha/bundle/build.go
+++ b/cmd/opm/alpha/bundle/build.go
@@ -42,6 +42,7 @@ func newBundleBuildCmd() *cobra.Command {
 		* All manifests yaml must be in the same directory. 
         `,
 		RunE: buildFunc,
+		Args: cobra.NoArgs,
 	}
 
 	bundleBuildCmd.Flags().StringVarP(&buildDir, "directory", "d", "",
@@ -79,7 +80,7 @@ func newBundleBuildCmd() *cobra.Command {
 	return bundleBuildCmd
 }
 
-func buildFunc(cmd *cobra.Command, args []string) error {
+func buildFunc(cmd *cobra.Command, _ []string) error {
 	return bundle.BuildFunc(
 		buildDir,
 		outputDir,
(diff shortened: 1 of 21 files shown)


cmd/opm/alpha/bundle/build.go
```diff
@@ -42,6 +42,7 @@ func newBundleBuildCmd() *cobra.Command {
 		* All manifests yaml must be in the same directory. 
         `,
 		RunE: buildFunc,
+		Args: cobra.NoArgs,
```
timflannagan, line 45: Unrelated to your changes: that bundleBuildCmd long description looks largely misidented to me if you're looking for additional things to tackle.

timflannagan: Yeah, just pulled these changes down and here's the outupt:

```bash
$ ./bin/opm alpha bundle --help
The "opm alpha bundle build" command will generate operator
        bundle metadata if needed and build bundle image with operator manifest
        and metadata for a specific version.

        For example: The command will generate annotations.yaml metadata plus
        Dockerfile for bundle image and then build a container image from
        provided operator bundle manifests generated metadata
        e.g. "quay.io/example/operator:v0.0.1".

        After the build process is completed, a container image would be built
        locally in docker and available to push to a container registry.

        $ opm alpha bundle build --directory /test/0.1.0/ --tag quay.io/example/operator:v0.1.0 \
		--package test-operator --channels stable,beta --default stable --overwrite

		Note:
		* Bundle image is not runnable.
		* All manifests yaml must be in the same directory.
```

tylerslaton: Good call out, I'll tidy that up while I'm at it.

njhale (approved):
/approve

PR #796, opened 2021-09-29
Use the proper grpc_health_probe command per architectures.

### LINKED ISSUE #795: readiness probe failing on multiarch
Due that the project [grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe) build the container only for `linux/amd64` the readiness probe fail when running on multi-arch and pod goes into `CrashLoopBackOff`.

Working on a fix now.

sxd:

Downloads the proper grpc_health_probe binary for the build
architectures during the release process.
 
**Reviewer Checklist**

Closes #795

### DIFF
--- a/release/goreleaser.opm.Dockerfile
+++ b/release/goreleaser.opm.Dockerfile
@@ -2,7 +2,13 @@
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
 
-FROM --platform=$BUILDPLATFORM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.4 as grpc_health_probe
+FROM alpine as grpc_health_probe
+ARG TARGETARCH
+RUN apk update && \
+	apk add curl && \
+	curl --silent --show-error --location --output /grpc_health_probe \
+	https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.5/grpc_health_probe-linux-$TARGETARCH && \
+	chmod 755 /grpc_health_probe
 FROM gcr.io/distroless/static:debug
 COPY --from=grpc_health_probe /grpc_health_probe /bin/grpc_health_probe
 COPY ["nsswitch.conf", "/etc/nsswitch.conf"]
--- a/.github/workflows/goreleaser.yaml
+++ b/.github/workflows/goreleaser.yaml
@@ -65,6 +65,8 @@ jobs:
           password: ${{ secrets.QUAY_PASSWORD }}
           registry: quay.io
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
 
       - name: "Run GoReleaser"
         run: make release


release/goreleaser.opm.Dockerfile
```diff
@@ -2,7 +2,11 @@
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
 
-FROM --platform=$BUILDPLATFORM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.4 as grpc_health_probe
+FROM ubuntu as grpc_health_probe
+RUN apt update && \
+  apt install -y curl &&
+  curl -o /grpc_health_probe \
+  https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.5/grpc_health_probe-linux-$TARGETARCH
```
joelanford, line 9: Does the following do the right thing?
```suggestion
FROM --platform=$TARGETPLATFORM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.4 as grpc_health_probe

Comment (sxd) [reply release/goreleaser.opm.Dockerfile]: yep but grpc-health-probe isn't multi-arch it is build only for linux/amd64 https://github.com/grpc-ecosystem/grpc-health-probe/blob/master/.github/workflows/release.yml#L29

Comment (joelanford) [reply release/goreleaser.opm.Dockerfile]: You're right. I swear I saw something about it being multi-arch when I was originally putting this together, but apparently not.

Disregard my suggestion!

File: release/goreleaser.opm.Dockerfile [hunk 2]
```diff
@@ -2,7 +2,11 @@
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
 
-FROM --platform=$BUILDPLATFORM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.4 as grpc_health_probe
+FROM ubuntu as grpc_health_probe
+RUN apt update && \
+  apt install -y curl && \
+  curl -o /grpc_health_probe \
+  https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.5/grpc_health_probe-linux-$TARGETARCH
```
joelanford, line 9: Does this need to be `chmod`-ed so that it's executable?
```suggestion
  https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.5/grpc_health_probe-linux-$TARGETARCH && \
  chmod 755 /grpc_health_probe
```

release/goreleaser.opm.Dockerfile
```diff
@@ -2,7 +2,11 @@
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
 
-FROM --platform=$BUILDPLATFORM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.4 as grpc_health_probe
+FROM ubuntu as grpc_health_probe
+RUN apt update && \
+  apt install -y curl && \
+  curl -o /grpc_health_probe \
```
joelanford, line 8:
```suggestion
  curl --silent --show-error --location --output /grpc_health_probe \
```

sxd: Working on this line too, looks like curl has some issues on arm64 =( probably we will have to switch to wget

release/goreleaser.opm.Dockerfile
sxd: That's a good catch! we should make sure!

release/goreleaser.opm.Dockerfile
sxd: what about using -SsL ? what do you think?

sxd:
@joelanford changed a few things since @mnencia suggested not to use ubuntu (because arm64 support was broken) so now we use alpine to download the proper binary

mnencia:
We probably need to specify `use: buildx` in goreleaser docker section.

release/goreleaser.opm.Dockerfile
joelanford: In scripts, I generally try to remember to use long form options to make it easier to grok later.

mnencia: I'm going to push new version with the long options

release/goreleaser.opm.Dockerfile
```diff
@@ -2,7 +2,13 @@
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
 
-FROM --platform=$BUILDPLATFORM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.4 as grpc_health_probe
+FROM alpine as grpc_health_probe
```
joelanford, line 5:
```suggestion
FROM --platform=$TARGETPLATFORM alpine as grpc_health_probe
```

mnencia: it should not be needed, it's the default.

mnencia:
I see a failure on s390x, but I'm afraid that it is an error in the base alpine image.

joelanford:
> I see a failure on s390x, but I'm afraid that it is an error in the base alpine image (it works on all the other architectures).

🤣 ubuntu fails with arm64, alpine fails with s390x.

Maybe UBI?

mnencia:
I've tested the alpine image and it works. I believe that the problem is the missing step to set up QEMU to execute the binaries. I've added it in the CI to see if it fixes the build.

mnencia:
It works. Now I've tried again removing the `docker/setup-buildx-action` that seem to be unneeded.

mnencia:
Everything looks fine. I think the PR is ready for the final review.

joelanford (approved):
/approve

PR #797, opened 2021-09-29
render: populate related images even when CSV has none explicitly defined

joelanford:
When `opm render` renders a bundle, it currently produces zero related images if the CSV does not have an explicit `spec.relatedImages` field.

This is update to fix an edge case not already fixed in #771.

Related images can be derived from a CSV even when the CSV does not explicitly contain a `spec.relatedImages` field.

**Reviewer Checklist**


dinhxuanvu (approved):
/lgtm

PR #803, opened 2021-10-01
registry: vacuum db on commands that remove data

joelanford:
Add [`VACUUM`](https://www.sqlite.org/lang_vacuum.html) call to the following commands:
- `deprecatetruncate`
- `prune`
- `prune-stranded`
- `rm`

Users would likely be surprised that the size of the database file is not reduced after a command the removes data from the database completes successfully.

**Reviewer Checklist**


anik120 (approved):
/lgtm

PR #804, opened 2021-10-04
(opm) Introduce deprecatetruncate for `opm registry`

anik120:

This PR exposes the `deprecatetruncate` functionality, previously
only included under `opm index`, under `opm registry`. In line with
all other commands that operate on a sqlite db, this sub-command is
deprecated too. In line with `opm index deprecatetruncate`, the
`deprecatetruncate` sub-command for `opm registry` is hidden too.

Existing catalog building pipelines that operate on index images are about to be migrated to building plaintext files backed catalogs. Infrastructures are being built for them such that during the migration process, they can continue to use the `opm index` UX (i.e imperative workflow) on plaintext files backed catalogs. This is required to provide the pipelines with sufficient time to move over from existing imperative workflow to declarative workflow while building plaintext backed catalogs. The new infrastructures mentioned previously operates exclusively on sqlite databases (and not index images). As a result all existing functionalities for index images are also needed to be present for sqlite databases. `deprecatetruncate` is only functionality that was missing from `opm registry`.      
 
**Reviewer Checklist**

### DIFF
--- a/cmd/opm/registry/deprecatetruncate.go
+++ b/cmd/opm/registry/deprecatetruncate.go
@@ -1,0 +1,76 @@
+package registry
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/pkg/lib/registry"
+	"github.com/operator-framework/operator-registry/pkg/sqlite"
+)
+
+func newRegistryDeprecateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Hidden: true,
+		Use:    "deprecatetruncate",
+		Short:  "deprecate operator bundle from registry DB",
+		Long: `deprecate operator bundle from registry DB
+
+` + sqlite.DeprecationMessage,
+
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			if debug, _ := cmd.Flags().GetBool("debug"); debug {
+				logrus.SetLevel(logrus.DebugLevel)
+			}
+			return nil
+		},
+
+		RunE: deprecateFunc,
+		Args: cobra.NoArgs,
+	}
+
+	cmd.Flags().Bool("debug", false, "enable debug logging")
+	cmd.Flags().StringP("database", "d", "index.db", "relative path to database file")
+	cmd.Flags().StringSliceP("bundle-images", "b", []string{}, "comma separated list of links to bundle image")
+	cmd.Flags().Bool("permissive", false, "allow registry load errors")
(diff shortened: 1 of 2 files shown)


cmd/opm/registry/deprecatetruncate.go
```diff
@@ -0,0 +1,76 @@
...
+		InputDatabase:       fromFilename,
+		Bundles:             bundleImages,
+		AllowPackageRemoval: allowPackageRemoval,
+	}
+
+	logger := logrus.WithFields(logrus.Fields{"bundles": bundleImages})
+
+	logger.Info("deprecating from registry")
+
+	RegistryDeprecator := registry.NewRegistryDeprecator(logger)
```
joelanford, line 69:
```suggestion
	registryDeprecator := registry.NewRegistryDeprecator(logger)
```

cmd/opm/registry/deprecatetruncate.go
```diff
@@ -0,0 +1,76 @@
...
+
+	err = RegistryDeprecator.DeprecateFromRegistry(request)
+	if err != nil {
+		return err
```
joelanford, line 73: If we return an error here, cobra will print out the command usage. However, that seems unhelpful since by this point, we have successfully parsed that flags and handled the CLI -> library translation.

It seems better to just log the returned error and immediately exit to avoid the extra cobra-level usage logging.

```suggestion
		logger.Fatal(err)
```

anik120: @joelanford that's a good point. Changed this to `logger.Fatal(err)` and then `logrus.Exit(1)` here. PTAL, thanks!

joelanford: No need for the `Exit` call. `Fatal` does that for you.

cmd/opm/registry/deprecatetruncate.go
```diff
@@ -0,0 +1,77 @@
...
+	logger := logrus.WithFields(logrus.Fields{"bundles": bundleImages})
+
+	logger.Info("deprecating from registry")
+
+	registryDeprecator := registry.NewRegistryDeprecator(logger)
+
+	err = registryDeprecator.DeprecateFromRegistry(request)
+	if err != nil {
+		logger.Fatal(err)
+		logrus.Exit(1)
```
joelanford, line 74: nit: this line is unreachable because `logger.Fatal` calls `os.Exit(1)` directly.

joelanford (approved):
One last nit. Otherwise, looks good!

/approve

ankitathomas (approved):
/lgtm

PR #807, opened 2021-10-05
render: include olm.bundle.object properties in rendered bundle images

joelanford:

In `opm render`, include `olm.bundle.object` properties in rendered bundle images.

By default, we need bundles to include `olm.bundle.object` properties so that the GRPC server can serve 
them to OLM's `packageserver` component, which is used by OLM clients (e.g. the kubectl-operator plugin) to provide user-friendly installation workflows and display information about the operators available in catalogs installed in the cluster.

Note: `packageserver` only actually uses the CSVs from the heads of each channel. Catalog maintainers may opt to delete objects from non-channel heads to reduce the size of the catalog. 

**Reviewer Checklist**


jianzhangbjz:
Thanks for your quick action! 
> Catalog maintainers may opt to delete objects from non-channel heads to reduce the size of the catalog.

Yes, this is a potential issue, but why not add a command or integrate some code to address this issue at this time? Or am I missing something? Thanks!

alpha/action/render.go
```diff
@@ -296,22 +296,33 @@ func populateDBRelatedImages(ctx context.Context, cfg *declcfg.DeclarativeConfig
 }
 
 func bundleToDeclcfg(bundle *registry.Bundle) (*declcfg.DeclarativeConfig, error) {
-	bundleProperties, err := registry.PropertiesFromBundle(bundle)
+	objs, props, err := registry.ObjectsAndPropertiesFromBundle(bundle)
```
joelanford, line 299: This is the breaking change that go-apidiff is complaining about. This function was introduced with FBC code, so it hasn't been in the wild long.

I'm perfectly willing to refactor to avoid a breaking change if others think its necessary here.

PR #808, opened 2021-10-05
fix(opm): ignore empty include file args

estroz:
**Description of the change:** ignore empty include file args in `opm alpha diff`

**Motivation for the change:** bugfix

/kind bug

**Reviewer Checklist**


dinhxuanvu (approved):
/approve

PR #819, opened 2021-11-02
feat(diff): change default behavior to prune by omission

estroz:
**Description of the change:** change the behavior of `opm alpha diff` to only include stuff from the include file by default, and optionally act additively with `--include-additive`

**Motivation for the change:** it is counterintuitive to include additively by default

**Reviewer Checklist**


cmd/opm/alpha/diff/cmd.go
```diff
@@ -44,27 +57,28 @@ func NewCmd() *cobra.Command {
...
+
+Each set of refs is passed to 'opm render <refs>' to produce a single, normalized delcarative config.
+
+Depending on what arguments are provided to the command, a particular "mode" is invoked to produce a diff:
+
+- If in heads-only mode (old-refs is not specified), then the heads of channels in new-refs are added to the output.
+- If in latest mode (old-refs is specified), a diff between old-refs and new-refs is added to the output.
+- If --include-file is set, items from that file will be added to the diff:
+	- If --include-additive is false (the default), a diff will be generated only on those objects, depending on the mode.
+	- If --include-additive is true, the diff will contain included objects, plus those added by the mode's invocation.
```
kevinrizza, line 71: is it possible to include some optional help text with example calls?

alpha/declcfg/diff.go
```diff
@@ -48,73 +58,101 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
...
 		}
+
+		return nil
+	}
+
+	headsOnlyMode := len(oldModel) == 0
+	latestMode := !headsOnlyMode
+	isInclude := len(g.Includer.Packages) != 0
+
+	switch {
```
kevinrizza, line 86: Is a switch statement better than a nested if statement here? Worrying about the default case running or not seems like it would be easy to break on further refactor

alpha/action/diff_test.go
```diff
@@ -129,6 +133,7 @@ func TestDiff(t *testing.T) {
 						},
 					},
 				},
+				IncludeAdditively: true,
```
kevinrizza, line 136: Do we want to add some test cases where IncludeAdditively is false for explicit output comparison?

kevinrizza (commented):
Some small nits, lgtm otherwise

alpha/declcfg/diff.go
estroz: An if statement would look more confusing, and a switch statement supports future modes more intuitively.

alpha/action/diff_test.go
estroz: The tests I added elsewhere cover this case. These are mostly to ensure input is parsed and output is rendered correctly with real files.

cmd/opm/alpha/diff/cmd.go
estroz: Done.

dinhxuanvu (approved):
/lgtm

estroz:
Thanks!

PR #821, opened 2021-11-05
goreleaser: use netgo and osusergo to ensure completely static binary builds

joelanford:
Use `netgo` and `osusergo` when building binaries for `opm` image to ensure completely static binary builds.

The current goreleaser build results in an image with a binary that panics when the network is used (e.g. `opm render` with a remote image) 

**Reviewer Checklist**

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -11,7 +11,7 @@ builds:
       - CGO_ENABLED=1
     mod_timestamp: "{{ .CommitTimestamp }}"
     flags: &build-flags
-      - -tags=json1
+      - -tags=json1,netgo,osusergo
     asmflags: &build-asmflags
       - all=-trimpath={{ .Env.PWD }}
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -11,7 +11,7 @@ builds:
       - CGO_ENABLED=1
     mod_timestamp: "{{ .CommitTimestamp }}"
     flags: &build-flags
-      - -tags=json1
+      - -tags=json1,netgo,osusergo
```
timflannagan, line 14: I think if I squint I can see why osusergo is needed but it's not immediately obvious outside of opm's general domain what command is triggering the need for that tag.

/lgtm

joelanford: 🙏 

I'm not _positive_ if `osusergo` is actually necessary, but since we don't have anything that actually tests the results of our binary builds, I figured it was best to err on the side of caution.

PR #823, opened 2021-11-05
feat(opm): merge objects of same unique key in `alpha diff`

estroz:
**Description of the change:** `opm alpha diff` now merges packages, channels, and bundles that have the same unique key with ascending preference.

**Motivation for the change:** Deduplicate data that may cause errors at runtime

Test this change by building `opm` then comparing the output of master and this branch's `opm alpha diff` on some test catalogs I built:
```console
$ ./bin/opm-master alpha diff quay.io/estroz/test-catalog:latest,quay.io/estroz/test-catalog:disjoint -o yaml > nomerge.yaml
...
$ ./bin/opm-merge alpha diff quay.io/estroz/test-catalog:latest,quay.io/estroz/test-catalog:disjoint -o yaml > merge.yaml
FATA[0005] error generating diff: error converting new declarative config to model: package "bar", bundle "bar.v0.2.0" not found in any channel entries
```

**Reviewer Checklist**

### DIFF
--- a/alpha/action/diff.go
+++ b/alpha/action/diff.go
@@ -29,12 +29,17 @@ type Diff struct {
 	// IncludeAdditively catalog objects specified in IncludeConfig.
 	IncludeAdditively bool
 
-	MergeType MergeType
+	Merger declcfg.Merger
 
 	Logger *logrus.Entry
 }
 
 func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
+
+	if a.Merger == nil {
+		a.Merger = &declcfg.PreferLastStrategy{}
+	}
+
 	if err := a.validate(); err != nil {
 		return nil, err
 	}
@@ -53,7 +58,11 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 			}
 			return nil, fmt.Errorf("error rendering old refs: %v", err)
 		}
-		if err := a.MergeType.MergeDC(oldCfg); err != nil {
+
+		if err != nil {
+			return nil, fmt.Errorf("error setting merge type: %v", err)
+		}
+		if err := a.Merger.MergeDC(oldCfg); err != nil {
 			return nil, fmt.Errorf("error merging across old refs: %v", err)
 		}
 		oldModel, err = declcfg.ConvertToModel(*oldCfg)
@@ -70,7 +79,7 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 		}
 		return nil, fmt.Errorf("error rendering new refs: %v", err)
 	}
-	if err := a.MergeType.MergeDC(newCfg); err != nil {
+	if err := a.Merger.MergeDC(newCfg); err != nil {
 		return nil, fmt.Errorf("error merging across new refs: %v", err)
 	}
 	newModel, err := declcfg.ConvertToModel(*newCfg)
--- a/alpha/action/diff.go
+++ b/alpha/action/diff.go
@@ -29,6 +29,8 @@ type Diff struct {
 	// IncludeAdditively catalog objects specified in IncludeConfig.
 	IncludeAdditively bool
 
+	MergeType MergeType
+
 	Logger *logrus.Entry
 }
 
@@ -51,6 +53,9 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 			}
 			return nil, fmt.Errorf("error rendering old refs: %v", err)
 		}
+		if err := a.MergeType.mergeDC(oldCfg); err != nil {
+			return nil, fmt.Errorf("error merging across old refs: %v", err)
+		}
 		oldModel, err = declcfg.ConvertToModel(*oldCfg)
 		if err != nil {
 			return nil, fmt.Errorf("error converting old declarative config to model: %v", err)
@@ -64,6 +69,9 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 			return nil, fmt.Errorf("%w (diff does not permit direct bundle references)", err)
 		}
 		return nil, fmt.Errorf("error rendering new refs: %v", err)
+	}
+	if err := a.MergeType.mergeDC(newCfg); err != nil {
+		return nil, fmt.Errorf("error merging across new refs: %v", err)
 	}
 	newModel, err := declcfg.ConvertToModel(*newCfg)
 	if err != nil {
--- a/alpha/action/diff.go
+++ b/alpha/action/diff.go
@@ -58,11 +58,7 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 			}
 			return nil, fmt.Errorf("error rendering old refs: %v", err)
 		}
-
-		if err != nil {
-			return nil, fmt.Errorf("error setting merge type: %v", err)
-		}
-		if err := a.Merger.MergeDC(oldCfg); err != nil {
+		if err := a.Merger.Merge(oldCfg); err != nil {
 			return nil, fmt.Errorf("error merging across old refs: %v", err)
 		}
 		oldModel, err = declcfg.ConvertToModel(*oldCfg)
@@ -79,7 +75,7 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 		}
 		return nil, fmt.Errorf("error rendering new refs: %v", err)
 	}
-	if err := a.Merger.MergeDC(newCfg); err != nil {
+	if err := a.Merger.Merge(newCfg); err != nil {
 		return nil, fmt.Errorf("error merging across new refs: %v", err)
 	}
 	newModel, err := declcfg.ConvertToModel(*newCfg)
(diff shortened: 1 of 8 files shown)


joelanford:
I'm still concerned that this puts `render` in the opinionated camp. And once something has opinions, its hard to say no to making that thing have variations of those opinions. My concern is that as soon as we say "`opm render` can merge things", we'll get all sorts of requests about _how_ to merge those things:
- Can we prefer values from references listed earlier instead of later?
- Can we "intelligently" merge channel entries? (with variations on what's intelligent for various use cases)
- Can we "intelligently" merge bundle properties? (with variations on what's intelligent for various use cases)
- Can we override some values by merge others?
- etc, etc.

If this is a specific need of the `diff` command, can we hardcode the correct merge logic directly into the `diff` command's logic?

estroz:
All valid concerns. I do actually think, in the long term, `opm` or some library it consumes should have the ability to support the different merge strategies you suggest (and more), while being opinionated by default. I also want to point out that you can effectively get different merge strategies by preprocessing `opm render --merge` input refs yourself, ex. reversing the input ref list(s) to reverse merge preference, or breaking out packages/channels/bundles into separate refs and reordering them. Ultimately I'd like the merge library `opm render` uses to be extensible to several merge strategies, but `opm render` itself has only one or two opinionated ways to merge 2+ refs. `opm render --merge` may be used in the future to handle the multiple-catalog-shared-content issue, which is a real concern when one cluster uses multiple curated but uncoordinated content sources, so it would be better to have a tool for cluster admins that isn't perfect than no tool at all to handle this issue. Perhaps this warrants an EP in itself.

>If this is a specific need of the diff command, can we hardcode the correct merge logic directly into the diff command's logic?

For now this is totally fine with me.

cdjohnson:
Is the result of this consistent with how `opm serve` currently merges the stream?  The result should be identical assuming there is no advanced/intelligent merge policy.

joelanford:
> Is the result of this consistent with how opm serve currently merges the stream? The result should be identical assuming there is no advanced/intelligent merge policy.

`opm serve` doesn't merge anything. It loads it all and then errors out (via an internal `opm validate` call) if it finds any duplicates.

cdjohnson:
@joelanford 
> `opm serve` doesn't merge anything. It loads it all and then errors out (via an internal `opm validate` call) if it finds any duplicates.

It appears to do some sort of merging or deduplication (I didn't do an exhaustive test).  Take a look at this FBC:
`docker.io/cdjohnson/depcatalogs@sha256:e067cbe0738608c19f66b71828d8dfa75a25790178316b73585cc7dc2967e635`

```
# See that testoperatorc is duplicated
cat index/index.yaml | yq -r 'select(.schema == "olm.package").name'
testoperatora
testoperatorb
testoperatorc
testoperatorc

cat index/index.yaml | yq -r 'select(.schema == "olm.bundle").name'
testoperatora.v1.0.0
testoperatorb.v2.0.0
testoperatorc.v3.0.0
testoperatorc.v3.0.0
testoperatorc.v3.0.1
```

`opm validate` passes

When I serve it, it ignores (or merges) the duplicates:
`opm serve index &`

```
grpcurl -plaintext localhost:50051 api.Registry/ListBundles | jq '{name: .csvName, version: .version, replaces: .replaces, skipRange: .skipRange, skips: .skips, channelName: .channelName}' | jq -s -c 'sort_by(.name)[]'
{"name":"testoperatora.v1.0.0","version":"1.0.0","replaces":null,"skipRange":null,"skips":null,"channelName":"v1.0"}
{"name":"testoperatorb.v2.0.0","version":"2.0.0","replaces":null,"skipRange":null,"skips":null,"channelName":"v2.0"}
{"name":"testoperatorc.v3.0.0","version":"3.0.0","replaces":null,"skipRange":null,"skips":null,"channelName":"v3.0"}
{"name":"testoperatorc.v3.0.1","version":"3.0.1","replaces":"testoperatorc.v3.0.0","skipRange":"<3.0.1","skips":null,"channelName":"v3.0"}
```

joelanford:
> It appears to do some sort of merging or deduplication

Looking at the code, I think you're right! But I think this is a regression. At one point, I'm fairly sure  `opm validate` was producing and error if it found multiple bundles with the same name in the same package.

joelanford:
Now taking the time to look back through the history of the FBC implementation, I didn't find any code that actually prevented duplicates. It looks like the code has always just accepted the last duplicate item (i.e. no merging, just throw away the old thing and keep the new thing when I encounter a duplicate)

I still consider this an error though because a declarative API (which FBC is), should disallow such ambiguity. Declaring "I want A to look like X AND I want A to look like Y" just doesn't make sense.

PR to fix this is #824.

alpha/action/merge.go
```diff
@@ -0,0 +1,264 @@
...
+	// PreferLast just uses the last/most recent object.
+	// This is the default merge type.
+	PreferLast MergeType = iota
+	// TwoWay fills in object fields by combining objects in ascending order,
+	// starting with the first/oldest.
+	// Bundle properties are NOT merged because an arbitrary property's key is unknowable.
+	TwoWay
+)
+
+func (mt MergeType) MergeDC(cfg *declcfg.DeclarativeConfig) error {
```
joelanford, line 26: This function worries me because it has to be maintained and updated every time we add/remove a merge implementation.

WDYT about refactoring to a `type Merger interface { Merge(*declcfg.DeclarativeConfig) error }` and then implementing each strategy as a different struct implementation of that interface?

Since (AFAICT) this PR doesn't introduce a way to customize the merge implementation via a flag, that would remove this switch entirely from the codebase. If we ever did add a flag, this switch would have to exist in some form, but it would likely be in the CLI, not in the library.

The other benefit is that any consumers of the library could add their own merge implementation without requiring us to merge anything.

alpha/action/merge.go
```diff
@@ -0,0 +1,264 @@
+package action
```
joelanford, line 1: Do you think it makes sense to try to keep a 1-to-1 mapping between CLI subcommands and actions? Since there isn't an `opm merge` CLI subcommand, I imagined this would be added in the `declcfg` package rather than as a separate action.

joelanford (changes requested):
Sorry for the delay on the review. I left a few comments related to some hopefully minor refactoring.

However, I am still concerned that doing this type of merging runs counter to the declarative nature of FBC.

It looks like we're merging `old` on its own, merging `new` on its own, and then diffing the individually merged models. I question why `old` and `new` are ever in a state where they need to be merged in the first place. Shouldn't they both already be fully valid FBCs?

alpha/action/diff.go
```diff
@@ -53,7 +58,11 @@ func (a Diff) Run(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 			}
 			return nil, fmt.Errorf("error rendering old refs: %v", err)
 		}
-		if err := a.MergeType.MergeDC(oldCfg); err != nil {
+
+		if err != nil {
+			return nil, fmt.Errorf("error setting merge type: %v", err)
+		}
+		if err := a.Merger.MergeDC(oldCfg); err != nil {
```
joelanford, line 65: I'm still confused by what it means to merge a single `*declcfg.DeclarativeConfig`.

Shouldn't we be expecting any given individual FBC to be valid?

 I would have expected to see something like:
```go
// Validate the old config
oldModel, err := declcfg.ToModel(oldCfg)
if err != nil {
	return nil, err
}
if err := oldModel.Validate(); err != nil {
	return nil, err
}

// Validate the new config
newModel, err := declcfg.ToModel(newCfg)
if err != nil {
	return nil, err
}
if err := newModel.Validate(); err != nil {
	return nil, err
}

// Merge them together
if err := a.Merger.Merge(oldCfg, newCfg); err != nil {
	return nil, err
}
```

jpower432: Sorry @joelanford, I think I missed the earlier comment when you mentioned this concern before. I think the use-case for keeping the single argument to the `Merge` function is that the `oldCfg` and the `newCfg` are generated by `Render`. Since `render` isn't handling duplicate objects it would be possible for the FBC created by the groups of `OldRefs` and `NewRefs` could be invalid. Would changing from `Merge` to `Normalize` make more sense here?

joelanford: > Would changing from Merge to Normalize make more sense here?

Perhaps

> `oldCfg` and the `newCfg` are generated by Render.

Makes sense, but what's the case that led to those needing to be normalized? Shouldn't old and new be valid standalone FBCs?

joelanford: What I want to avoid is having our exported libraries normalize the possibility of an FBC being invalid and us providing code that can turn an FBC from invalid to valid.

If there's something happening in `opm alpha diff` that gives rise to the need to normalize some intermediate FBC state, I think all of this is fine, but I'd probably make it internal and unexported.

jpower432: Just adding some info here for historical purposes -
- The merge functionality was public because it is required by `oc-mirror`, but we have agreed to move this functionality directly into the `oc-mirror` codebase since the requirement to merge or normalize invalid FBCs is a unique use-case.

cdjohnson: Can you clarify?  Is the entire merge function moving to `oc-mirror` or just the single-fbc normalization function?

jpower432: Just the single-fbc normalization work proposed on this PR was moved to `oc-mirror`

grokspawn:
opm diff has moved to `oc mirror`.  From the conversation it appears that team has already extracted the salient bits of this PR for that project.  Is there any aspect of this PR which hasn't been moved over and is still needed?

PR #824, opened 2021-11-10
opm validate: fail on duplicate packages, channels, and bundles

joelanford:

When `opm validate` (or any code that requires a file-based catalog to be converted to the registry's internal model) encounters a duplicate blob `olm.package`, `olm.channel`, or `olm.bundle` blob, emit an error.

To prevent ambiguity in the way a file-based catalog is presented to consumers when there are duplicates. FBC is declarative, so it doesn't make sense to attempt to declare the same thing twice.

**Reviewer Checklist**


estroz (commented):
/lgtm

estroz:
This will need backporting correct?

PR #826, opened 2021-11-12
Block cross-package skips/replaces on add

ankitathomas:
add validation to ensure bundles only skip/replace members of the same package

Signed-off-by: Ankita Thomas <ankithom@redhat.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.16
 
 require (
 	github.com/Microsoft/hcsshim v0.8.9 // indirect
-	github.com/blang/semver v3.5.1+incompatible
 	github.com/blang/semver/v4 v4.0.0
 	github.com/bugsnag/bugsnag-go v1.5.3 // indirect
 	github.com/bugsnag/panicwrap v1.2.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/Microsoft/hcsshim v0.8.9 // indirect
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/blang/semver/v4 v4.0.0
 	github.com/bugsnag/bugsnag-go v1.5.3 // indirect
 	github.com/bugsnag/panicwrap v1.2.0 // indirect
 	github.com/containerd/containerd v1.4.11
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -7,7 +7,7 @@ import (
 	"os"
 	"sort"
 
-	"github.com/blang/semver"
+	"github.com/blang/semver/v4"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/yaml"
 
@@ -412,6 +412,7 @@ func DecodeFile(path string, into interface{}) error {
 	return decoder.Decode(into)
 }
 
+// ValidateEdgeBundlePackage ensures that all bundles in the input will only skip or replace bundles in the same package.
 func (i *DirectoryPopulator) ValidateEdgeBundlePackage(images []*ImageInput) error {
 	// track packages for encountered bundles
 	expectedBundlePackages := map[string]string{}
@@ -452,7 +453,6 @@ func (i *DirectoryPopulator) ValidateEdgeBundlePackage(images []*ImageInput) err
 		if err != nil {
 			return fmt.Errorf("unable to verify bundles for package %v", err)
 		}
-		fmt.Println(expectedBundlePackages)
 		for _, b := range entries {
 			if bundlePkg, ok := expectedBundlePackages[b.BundleName]; ok && bundlePkg != b.PackageName {
 				return fmt.Errorf("bundle %s belongs to package %s on index, cannot be added as an edge for package %s", b.BundleName, b.PackageName, bundlePkg)
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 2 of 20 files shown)


pkg/registry/populator.go
```diff
@@ -407,3 +411,54 @@ func DecodeFile(path string, into interface{}) error {
...
+	pkgs, err := i.querier.ListPackages(context.TODO())
+	if err != nil {
+		return fmt.Errorf("unable to verify bundle packages: %v", err)
+	}
+	for _, pkg := range pkgs {
+		entries, err := i.querier.GetChannelEntriesFromPackage(context.TODO(), pkg)
+		if err != nil {
+			return fmt.Errorf("unable to verify bundles for package %v", err)
+		}
+		fmt.Println(expectedBundlePackages)
```
dinhxuanvu, line 455: I assume this is test code so it should be removed.

dinhxuanvu (commented):
Just a few nits. Looking good.

pkg/registry/populator.go
ankitathomas: removed

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/Microsoft/hcsshim v0.8.9 // indirect
 	github.com/blang/semver v3.5.1+incompatible
```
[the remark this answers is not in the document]
ankitathomas: Added a new commit with a general version bump throughout operator-registry.

dinhxuanvu (approved):
Looking good.
/approve

pkg/registry/populator.go
```diff
@@ -407,3 +411,54 @@ func DecodeFile(path string, into interface{}) error {
...
+				return fmt.Errorf("bundle %s must belong to exactly one package, found on: %v", bndl, pkgs)
+			}
+			expectedBundlePackages[bndl] = b.Bundle.Package
+		}
+	}
+	if len(expectedBundlePackages) == 0 {
+		return nil
+	}
+
+	pkgs, err := i.querier.ListPackages(context.TODO())
```
anik120, line 447: Ideally, we'd place this kind of check in the table schema itself(eg put a constraint in the replaces/skips fields in the operatorbundle table to allow bundle names that belong to the same package only). Calling the querier, i.e opening a db connection every single time is an expensive operation. Opening a db connection for every additional step we take while loading a bundle = making the load operation a little bit more inefficient every time. 

Having said that, if this is being considered as a quick dirty fix for the time sqllite operations are deprecated, and we're okay with the load operation being a little bit inefficient, consider this comment withdrawn.

pkg/registry/populator_test.go
```diff
@@ -3068,3 +3071,86 @@ func newUnpackedTestBundle(root, dir, name string, csvSpec json.RawMessage, anno
...
+		require.NoError(t, err)
+
+		b, err := registry.NewBundleFromStrings(name, versionString, pkg, defaultChannel, channels, string(jsonout))
+		require.NoError(t, err)
+		return &registry.ImageInput{Bundle: b}
+	}
+
+	logrus.SetLevel(logrus.DebugLevel)
+	db, _ := CreateTestDb(t)
+	//	defer cleanup()
```
anik120, line 3112: I'm guessing this was meant to be uncommented.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.16
 
 require (
 	github.com/Microsoft/hcsshim v0.8.9 // indirect
-	github.com/blang/semver v3.5.1+incompatible
+	github.com/blang/semver/v4 v4.0.0
```
anik120, line 7: Why was this necessary?

anik120 (changes requested):

pkg/registry/populator.go
ankitathomas: We could add it to schema, but we'd also need to enforce the one package per bundle rule or this still happens. It's not just making sure it's in the right package, but that it isn't in any other package.

go.mod
ankitathomas: The `Version` in `ClusterServiceVersionSpec` requires semver/v4. It didn't make sense to keep both semver versions around, and since operator-lifecycle-manager already uses v4 it seemed like a good choice to keep.

dinhxuanvu (approved):
/approve

pkg/registry/populator.go
anik120: Shouldn't we be doing that anyway then?

go.mod
anik120: Could you link where exactly you need `semver/v4` in the `ClusterServiceVersionSpec` for this PR?

ankitathomas: https://github.com/operator-framework/operator-registry/pull/826/files#diff-bc4c211037518ff105dee0994c7e5fe4ae00bfd9e3230cb17b82605894dfc33aR3083

That requires `OperatorVersion`, which is `semver/v4`

anik120: Okay that makes sense.

pkg/registry/populator.go
ankitathomas: We can have a trigger checking for this but with the way adding bundles in sqlite works, that alone would get us bundles that are partially added (i.e, in `operatorbundle` table alone with some entries into `channel_entry` table). Considering we also nuke the package for reconstructing bundle graphs, having that start and fail half way would be really bad.

So we'd need to do this check before starting the add anyway, since our bundle adds aren't one atomic operation.

kevinrizza: I don't think we should touch the schema or make a db migration unless we absolutely have to

anik120: That's a good point too.

anik120 (approved):
/lgtm

PR #872, opened 2021-11-26
The `--skip-tls` flag shouldn't use the http protocol

omertuc:
This commit makes it so that the `https` scheme is still used even when
the `--skip-tls` flag is specified

When using the `operator-sdk run bundle`'s `--skip-tls` flag which is
described as:

> skip authentication of image registry TLS certificate when pulling a bundle image in-cluster

It tries to access the image registry (given in the <bundle_image>
positional argument) using HTTP rather than HTTPS.

This behavior is unexpected and fails when the image registry only
speaks the HTTPS protocol.

The commit (a16399f05c380ba432691d2e1d99a69aa5a269e2) which seems to have introduced
this behavior doesn't mention this behavior anywhere, so I'm assuming
it's unintentional and therefore a bug that needs fixing - Let me know
if you think otherwise.

**Reviewer Checklist**

### DIFF
--- a/pkg/image/containerdregistry/resolver.go
+++ b/pkg/image/containerdregistry/resolver.go
@@ -27,16 +27,11 @@ func NewResolver(configDir string, insecure bool, roots *x509.CertPool) (remotes
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 5 * time.Second,
 		TLSClientConfig: &tls.Config{
-			InsecureSkipVerify: false,
+			InsecureSkipVerify: insecure,
 			RootCAs:            roots,
 		},
 	}
 
-	if insecure {
-		transport.TLSClientConfig = &tls.Config{
-			InsecureSkipVerify: insecure,
-		}
-	}
 	headers := http.Header{}
 	headers.Set("User-Agent", "opm/alpha")
 
@@ -54,9 +49,6 @@ func NewResolver(configDir string, insecure bool, roots *x509.CertPool) (remotes
 			docker.WithAuthCreds(credential(cfg)),
 		)),
 		docker.WithClient(client),
-	}
-	if insecure {
-		regopts = append(regopts, docker.WithPlainHTTP(docker.MatchAllHosts))
 	}
 
 	opts := docker.ResolverOptions{


pkg/image/containerdregistry/resolver.go
```diff
@@ -55,9 +50,6 @@ func NewResolver(configDir string, insecure bool, roots *x509.CertPool) (remotes
 		)),
 		docker.WithClient(client),
 	}
-	if insecure {
```
perdasilva, line 58: Question: Have you checked your changes still work with http endpoints?
I'm not sure what these `regopts` do exactly. Should we maybe still keep this but only add if if the url scheme is http?
@njhale any thoughts here?

omertuc: I have not checked that, good point. Didn't think about that.

But if I do think of it, why would someone pass `--skip-tls` when using HTTP in the first place? And also, how does one even specify http registries? e.g. How would you tell docker to use http to communicate with `quay.io/foo/bar:latest`? It uses https by default. Searching online didn't lead me to a conclusive answer

omertuc: Also, worth mentioning that seems like `WithPlainHTTP` configures the `plainHTTP` option which is deprecated (although the function WithPlainHTTP itself isn't deprecated which is odd)

omertuc: Here's how it's done:
https://docs.docker.com/registry/insecure/#deploy-a-plain-http-registry

Basically a config file lists all registries where docker should try to communicate with HTTPS, if it gets a cert, ignore any errors, if there's no HTTPS on that server fallback to HTTP.

Doesn't seem like you can specify the protocol in the image address itself in any way I could find

perdasilva: I agree with you here. --skip-tls and http are not related at all. I just don't know if these regopts we are removing here are needed for something specific (e.g. pipeline). We may still need to add a --use-http options or something like that. So, I'll set it to ok-to-test but I'd like @njhale (who I think originally added this) to ok.

omertuc: Original author is here https://github.com/operator-framework/operator-registry/commit/a16399f05c380ba432691d2e1d99a69aa5a269e2

estroz: We can't change/remove current `--skip-tls` functionality, but we can deprecate the flag and add new flags `--skip-tls-verify` and `--use-http` to enable their respective namesake's functionality.

estroz:
This is something I'd definitely like to see fixed. Thanks for your submission @omertuc!

pkg/image/containerdregistry/resolver.go
estroz: Assuming @omertuc does not update this PR soon, @dinhxuanvu @jpower432 can one of you take on the changes in a separate PR?

omertuc: Ah didn't realize your bullets were concrete change suggestions.

Yeah please take it on yourselves as I don't feel confident enough in this code base to make those changes

estroz: Sounds good, thanks for all you've done so far!

jpower432: I can make these changes.

dinhxuanvu:
This PR is replaced by #898

PR #879, opened 2021-12-08
feat(constraint): Add constraint type validation to bundle validation

dinhxuanvu:
A new type of dependency named constraint (olm.constraint) which uses
CEL expression is introduced in OLM. This new type is added to
dependencies.json file in the bundle just like gvk and package
dependencies. As a result, a new validation for constraint type
is added to ensure this type will be validated for bundle
dependencies.

**Reviewer Checklist**


dinhxuanvu:
/close
in favor of #881

PR #881, opened 2021-12-13
feat(constraint): Add constraint type validation to bundle validation

dinhxuanvu:

A new type of dependency named constraint (olm.constraint) which uses
CEL expression is introduced in OLM. This new type is added to
dependencies.json file in the bundle just like gvk and package
dependencies. As a result, a new validation for constraint type
is added to ensure this type will be validated for bundle
dependencies.

**Reviewer Checklist**


anik120 (changes requested):
Looks good, a few test cases should get it ready to go.

pkg/registry/types.go
```diff
@@ -289,6 +301,24 @@ func (pd *PackageDependency) Validate() []error {
...
 }
 
+// Validate will validate constraint type and return error(s)
+func (cc *CelConstraint) Validate() []error {
+	errs := []error{}
+	if cc.Cel == nil {
+		errs = append(errs, fmt.Errorf("The CEL field is missing"))
+	} else {
+		if cc.Cel.Rule == "" {
+			errs = append(errs, fmt.Errorf("The CEL expression is missing"))
```
estroz, line 311:
```suggestion
			errs = append(errs, fmt.Errorf("The CEL expression is missing"))
			continue
```

pkg/registry/types.go
```diff
@@ -329,6 +359,13 @@ func (e *Dependency) GetTypeValue() interface{} {
 			return nil
 		}
 		return dep
+	case ConstraintType:
+		dep := CelConstraint{}
```
estroz, line 363: Shouldn't this be decoded into an api `constraints.Constraint`?

estroz:
Agreed on unit tests.

PR #884, opened 2021-12-13
fix(diff): prune empty channel when old model is empty

### LINKED ISSUE #883: opm: diff is not pruning empty channel when an include config is provided
`opm alpha diff -i config.yaml` is producing empty channels when an include config is providers and no old refs are specified.

Config:

```
packages:
  - name: rhacs-operator
    channels:
    - name: latest
    versions:
    - '3.67.0'
```

Results:
```
{
    "schema": "olm.channel",
    "name": "rhacs-3.66",
    "package": "rhacs-operator",
    "entries": []
}
opm validate of the catalog show this:

$ opm validate catalogs/
FATA[0000] invalid index:
└── invalid package "rhacs-operator":
    ├── invalid channel "rhacs-3.62":
    │   └── channel must contain at least one bundle
    ├── invalid channel "rhacs-3.64":
    │   └── channel must contain at least one bundle
    ├── invalid channel "rhacs-3.65":
    │   └── channel must contain at least one bundle
    └── invalid channel "rhacs-3.66":
        └── channel must contain at least one bundle
```

jpower432:

In this PR, the `opm alpha diff` command will now prune empty channels when an include config is provided in headsOnly mode. 

The motivation for this change to allow the catalog index produced in this instance to pass `opm validate` and be usable with a cluster.

**Reviewer Checklist**

Closes #883


alpha/declcfg/diff.go
```diff
@@ -79,6 +79,19 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
 		return nil
 	}
 
+	// Prunes empty channels
+	pruneEmptyChannels := func() error {
```
estroz, line 83: If this is only called once, in-line the code in that spot. Perhaps this could live in Includer.Run() too.

jpower432: Moved this down to where I was calling it. The only reason I didn't put it in `Includer.Run()` is because this is only an issue when `IncludeAdditively` is false I believe.

estroz (approved):
/approve
/lgtm

PR #890, opened 2021-12-15
(.gitignore): Do not ignore debug/

anik120:

In PR #888, there's vendoring of the cel-go package, one of which
    is named "debug" (github.com/google/cel-go/common/debug). Due to
    the .gitignore file ignoring all [Dd]bug folders, the vendored
    package isn't being checked in by git, causing the build to fail.
    This PR removes the instruction to ignore [Dd]bug/ folders.

**Reviewer Checklist**

Open question to consider: Do we see foresee any drastic consequence of not ignoring all [Dd]ebug folders, apart from unwanted `debug` folders showing up in PRs if `git add .` is invoked? If we do, is an explicit `!github.com/google/cel-go/common/debug` better? Although that is hard-coding the package, and potentially confusing future vendoring efforts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,7 +179,6 @@ tags
 *.userprefs
 
 # Build results
-[Dd]ebug/
 [Dd]ebugPublic/
 x64/
 x86/


anik120:
cc: @dinhxuanvu

dinhxuanvu (approved):
/approve

Issue #885, opened 2021-12-14
opm: Dont pull index images if its present locally?


ashishmax31:
We've been using the `indexer` package at `https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/indexer/indexer.go` in some of our code to extract bundle information from an index image. We've noticed that the package always assumes that the index image is present in a remote registry and tries to pull it[1]. Why can't we have a simple check to determine if the image is present locally before trying to pull it?

I'll be happy to create a fix for this if thats okay, what do you guys think?
 
[1] https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/indexer/indexer.go#L350

/kind feature

dinhxuanvu:
While I understand the reason for your request, I don't think OLM is in the position to support all local image formats that may be different between different toolings/runtimes. For index image, I would recommend you to use the new file-based catalog which essentially is a JSON/YAML blob that you modify as you wish before building an index image later on. In that way, you only need to work with a local file instead of an image. https://olm.operatorframework.io/docs/reference/file-based-catalogs/

joelanford:
I think we can close this based on the comment I left in the PR ([here](https://github.com/operator-framework/operator-registry/pull/892#issuecomment-1006125296))

@ashishmax31 please feel free to open a new issue if you have any more questions or requests!

PR #892, opened 2021-12-16
OPM: Check if image exists locally before pulling

ashishmax31:
Fixes: https://github.com/operator-framework/operator-registry/issues/885.
The the opm package always assumes that the image is present in a remote registry, this PR allows opm to work off of locally present images by checking if an image is present locally before trying to pull it.

 Make opm work with locally build index images.

**Reviewer Checklist**

### DIFF
--- a/pkg/containertools/runner.go
+++ b/pkg/containertools/runner.go
@@ -81,6 +81,11 @@ func (r *ContainerCommandRunner) GetToolName() string {
 // Pull takes a container image path hosted on a container registry and runs the
 // pull command to download it onto the local environment
 func (r *ContainerCommandRunner) Pull(image string) error {
+	if r.imagePresentLocally(image) {
+		// Image present locally, dont pull.
+		return nil
+	}
+
 	args := r.argsForCmd("pull", image)
 
 	command := exec.Command(r.containerTool.String(), args...)
@@ -181,4 +186,18 @@ func (r *ContainerCommandRunner) Inspect(image string) ([]byte, error) {
 	}
 
 	return out, err
+}
+
+// Checks if the passed image is present locally. Returns false
+// if a non-zero exit code (or any other error) is returned when
+// docker image inspect <image> is run.
+func (r *ContainerCommandRunner) imagePresentLocally(image string) bool {
+	args := r.argsForCmd("inspect", image)
+	command := exec.Command(r.containerTool.String(), args...)
+
+	_, err := command.CombinedOutput()
+	if err != nil {
+		return false
+	}
+	return true
 }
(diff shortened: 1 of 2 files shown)


ashishmax31:
I'm not able to run the e2e tests locally on my mac, trying to fix it.

joelanford:
Firstly, the `opm index/registry` commands are deprecated along with all of the underlying code used by those subcommands, so our default stance is that we're not likely to add new features to those commands/packages. They're mainly in bugfix/maintenance mode.

But assuming we make an exception, I do have some questions:

- What problem is this trying to solve? It seems like there are two potential use cases:
    1. I built my index image locally, the image is present in my local image store, but I haven't pushed it yet. I want to continue operating on it locally before I'm ready to push.
   2. I don't want to save time and bandwidth by using locally cached copies of images that I've already pulled.

- What happens if the image exists in a remote repo, but the tag in the remote repo points to a different digest than the the local digest?

- What operations are you doing to extract the bundle metadata from the index?

If you could step back one step and provide a more complete picture of your workflow, we may be able to offer some alternative solutions that would solve this problem in a different way.

ashishmax31:
@joelanford 
>What problem is this trying to solve? It seems like there are two potential use cases:

It's the first(`i`) use-case you've mentioned.

>What happens if the image exists in a remote repo, but the tag in the remote repo points to a different digest than the the local digest?

This seems like a problem. I haven't thought of this scenario.

>What operations are you doing to extract the bundle metadata from the index?

To extract the bundle contents, we create a `indexer.NewIndexExporter()` and then call `ExportFromIndex(<request>)` on the created `IndexExporter` object. After the contents are extracted to the specified directory we read the each of the bundle's contents into a `registry.Bundle`.

What we're building is a cli for running some validations on the bundles present in an indexImage.

joelanford:
> What we're building is a cli for running some validations on the bundles present in an indexImage.

>> I built my index image locally, the image is present in my local image store, but I haven't pushed it yet. I want to continue operating on it locally before I'm ready to push.

> It's the first(i) use-case you've mentioned.

[`alpha/action.Render`](https://github.com/operator-framework/operator-registry/blob/611525be3bd7374d1e55d43c519e5664e0e91c2c/alpha/action/render.go#L51-L57) likely gets you what you need.

It works with both sqlite and file-based config catalog images AND it works with local sqlite files and file-based config directories. What you're using now will only work for the deprecated sqlite catalogs.

Since it sounds like your use case is dealing with locally built images, you likely have access to the source material that was copied into the image. Can you run your validation against the index source material (e.g. the db file or FBC directory) instead of the image content?

Render's `Run()` method returns a `declcfg.DeclarativeConfig` struct, which you can then use to iterate the bundles in the index, extract their bundle images, and then load the contents as `registry.Bundle` structs.

ashishmax31:
@joelanford Sorry for the late response, I kinda missed this. 
Thank you! I will check it out.

PR #893, opened 2022-01-06
Surface the manifest filenames for higher level logic

fgiloux:
Signed-off-by: Frederic Giloux <fgiloux@redhat.com>

This pull request surfaces the manifest filenames. There is currently no metadata, only the content of a manifest file carried over by bundle.Object. By returning the names of the manifest files it is possible to attach properties that can influence OLM behavior in regards to a specific manifest. This can then be leveraged for making [manifests optional](https://github.com/operator-framework/enhancements/blob/master/enhancements/optional-manifest.md) 

**Reviewer Checklist**


go.mod
```diff
@@ -51,7 +51,7 @@ require (
 	golang.org/x/mod v0.4.2
 	golang.org/x/net v0.0.0-20210825183410-e898025ed96a
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
-	google.golang.org/grpc v1.41.0
+	google.golang.org/grpc v1.40.0
```
timflannagan, line 54: Is there a reason we need to downgrade this dependency outside of OLM doesn't work well with the latest v1.41.x version? Can we instead replace pin on the OLM side-of-things while we work through the changes?

timflannagan (commented):
Overall, I think this looks reasonable to me. I think I'd like to see us add some testing around the filenames that get returned from the `Load(...)` method to ensure we don't regress on the registry side-of-things down the line.

go.mod
fgiloux: Some explanation in [these issue comments](https://github.com/golang/go/issues/30993#issuecomment-478755661) and in [this article](https://research.swtch.com/vgo-mvs). TL;DR; you don't pin a version but specify a minimum requirement.
I wish the issue with grpc v1.41.0 would have been further investigated when discovered. It is not a bug in the grpc library but a conscious behavior change. Now if you don't use keepalive it is the normal behavior for a connection to get IDLE if no rpc call is made. The issue is that you are only testing for READY:
~~~
catalog.Status.GRPCConnectionState.LastConnectTime.After(after.LastUpdateTime.Time) && catalog.Status.GRPCConnectionState.LastObservedState == "READY" {
~~~
I could change the line and test for IDLE as well. The one thing is that when connections have been IDLE for a defined time (OS setting) they can be garbage collected. The grpc client would have two options:
1. use keepalive
2. gracefully handle disconnection and reestablish a connection when required
Option 1 is better if it is important not to incur any delay due to the reestablishment of a TCP connection. Option 2 is generally advised as it allows freeing up unused resources (network connections, filedescriptors, memory on the host) and don't unnecessarily create additional network bandwidth consumption (the keepalive packets)

I don't know enough of the way you use the grpc server to say whether the behavior change would break other things than the tests, hence the downgrade of the grpc library. This is however no sustainable approach, just building up some additional technical debt.

fgiloux:
> I think I'd like to see us add some testing around the filenames that get returned from the Load(...) method to ensure we don't regress on the registry side-of-things down the line.

Fair enough. I have added some.

go.mod
exdx: That gRPC explanation makes a lot of sense -- I also started testing for IDLE when trying to address test flakes which did help the situation. It looks like the version bump on the registry side did affect performance. 

I think Option 2 would be better since the client (OLM) needs to talk to the servers (catalog pods) only intermittently, when doing a sync, instead of all the time. It would also help reduce the resource consumption of OLM somewhat versus keeping connections open all the time.

timflannagan: Thanks for the explanation - I think downgrading is reasonable to me, but I wonder whether this points to a need for a simple integration tests with OLM so we can catch this kind of regression quicker? That said, that would increase the coupling between the two repositories, and they both should have their own lifecycle for managing dependencies, so maybe the value add there is watered down.

pkg/configmap/configmap_test.go
```diff
@@ -181,3 +193,23 @@ func loadfromFile(t *testing.T, path string) *corev1.ConfigMap {
 
 	return bundle
 }
+
+// equalSlices compares two string slices and has tolerate different orderings
+func equalSlices(a, b []string) bool {
```
timflannagan, line 198: Do we need a custom testing helper function here? Can we piggy back off of `assert.ElementsMatch(...)` instead?

timflannagan (commented):
I think we're getting close to the finish line here. I see a couple of pkg/lib/indexer/index.Dockerfile* changes that are present in this PR. Any idea what they're needed for, or were they generated locally, and we're not correctly filtering them out in the .gitignore properly?

It looks like codecov is also complaining about a couple of early return statements, and their path isn't being tested. We've had issues with codecov failing in the past due to regressing on overall code coverage on a per-PR-basis over time, which blocked merging entirely. I don't think this is something we'd block the PR from merging given there's a corresponding OLM PR open, but it would be nice to ensure we're not regressing silently when we cut registry releases.

pkg/configmap/configmap_test.go
fgiloux: Good point! I made the change

fgiloux:
> I see a couple of pkg/lib/indexer/index.Dockerfile* changes that are present in this PR. Any idea what they're needed for, or were they generated locally, and we're not correctly filtering them out in the .gitignore properly?

I have done some cleansing. I believe that they are generated by the unit tests for mirror.
When opening files it would be advisable to use: [TemFile](https://pkg.go.dev/io/ioutil#TempFile) without specifying a dir.  It would then use the OS default tmp dir, /tmp for Linux.

perdasilva:

PR #894, opened 2022-01-10
Upgrade github.com/docker/cli to v20.10.12

jeloba:

Upgraded github.com/docker/cli to latest release.

The version previously used contained [CVE-2021-41092](https://nvd.nist.gov/vuln/detail/CVE-2021-41092).

**Reviewer Checklist**

### DIFF
--- a/pkg/image/containerdregistry/resolver.go
+++ b/pkg/image/containerdregistry/resolver.go
@@ -90,6 +90,8 @@ func credential(cfg *configfile.ConfigFile) func(string) (string, string, error)
 }
 
 // protects against a data race inside the docker CLI
+// TODO: upstream issue for 20.10.x is tracked here https://github.com/docker/cli/pull/3410
+// newer versions already contain the fix
 var configMutex sync.Mutex
 
 func loadConfig(dir string) (*configfile.ConfigFile, error) {
--- a/pkg/image/containerdregistry/resolver.go
+++ b/pkg/image/containerdregistry/resolver.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/adrg/xdg"
@@ -88,7 +89,13 @@ func credential(cfg *configfile.ConfigFile) func(string) (string, string, error)
 	}
 }
 
+// protects against a data race inside the docker CLI
+var configMutex sync.Mutex
+
 func loadConfig(dir string) (*configfile.ConfigFile, error) {
+	configMutex.Lock()
+	defer configMutex.Unlock()
+
 	if dir == "" {
 		dir = config.Dir()
 	}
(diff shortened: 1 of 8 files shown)


jeloba:
@timflannagan the data race looks like a questionable use of a global in the docker (CLI) config loader. Would it be acceptable to synchronize use of the config loader here?

https://github.com/operator-framework/operator-registry/blob/da57f13b6d89bb3d905387f53ce14d905da6d9d2/pkg/image/containerdregistry/resolver.go#L93

timflannagan:
@jeloba I haven't been super involved with this codebase, so I'm probably not the best person to answer that. I'd guess burning down any races would be acceptable. Maybe @joelanford or @njhale have a better idea?

jeloba:
I pushed a crude locking solution for loading the docker config. :shrug:

jeloba:
@dinhxuanvu @gallettilance any chance to get this retested or reviewed?

jeloba:
@dinhxuanvu @gallettilance what do I need to do to get this tested, reviewed, or merged?

pkg/image/containerdregistry/resolver.go
```diff
@@ -88,7 +89,13 @@ func credential(cfg *configfile.ConfigFile) func(string) (string, string, error)
 	}
 }
 
+// protects against a data race inside the docker CLI
+var configMutex sync.Mutex
+
 func loadConfig(dir string) (*configfile.ConfigFile, error) {
+	configMutex.Lock()
```
exdx, line 96: I'm not sure how the lock here actually prevents a race condition if this function is called outside of the race context. I see uses of `docker.ConfigFile` and `docker.Credentials` here but I don't see how these benefit from a lock. 

Maybe just bumping the docker dependency to the fixed version is enough to address the issue?

jeloba: The code in the lib uses a global variable during config loading to deal with some stdout messaging. During tests loading is called multiple times and thus triggers the race detector on the global.

I think the global was introduced into the dependency somewhere between the last version we dependent on and the latest one. :shrug: 

So while bumping the docker dependency deals with the issue this MR is about, it introduces a data race in the tests. The lock solves it.

joelanford: @jeloba is there an upstream issue that documents this problem in github.com/docker/cli?

I'd prefer if we had a breadcrumb in this repo that reminds us that the lock is there due to an upstream issue and gives us a pointer to know when it's safe to remove the lock (i.e. when the upstream bug is fixed, we can bump our dep and remove the lock)

jeloba: @joelanford  yes there is https://github.com/docker/cli/pull/3410 should I reference that in the code?

timflannagan: Yeah, I think my preference would be to inline that issue link as a TODO comment if we can't away from having to introduce locking mechanisms in the code.

timflannagan (approved):
/lgtm

PR #897, opened 2022-01-13
feat(registry): add podman auth.json as an alternate location

jpower432:
Adds podman default credential location as an alternate location
when loading the config for registry resolver

This PR changes the `containerd` registry resolver to use `docker` well-known credential locations as a primary credential location for registry authentication and uses the `podman` location as an alternative.

The motivation for this change is to allow podman users to use the default location as a credential location and to help resolve an [issue](https://github.com/openshift/oc-mirror/issues/216) in `oc-mirror` 

**Reviewer Checklist**


jpower432:
cc @dinhxuanvu

pkg/image/containerdregistry/resolver.go
```diff
@@ -85,14 +87,31 @@ func credential(cfg *configfile.ConfigFile) func(string) (string, string, error)
...
+	cfg = configfile.New(dockerConfigJSON)
+
+	switch _, err := os.Stat(dockerConfigJSON); {
+	case err == nil:
+		cfg, err = config.Load(dir)
+		if err != nil {
+			return cfg, err
+		}
+	case os.IsNotExist(err):
+		podmanConfig := filepath.Join(os.Getenv("XDG_RUNTIME_DIR"), "containers/auth.json")
```
joelanford, line 105: The `XDG_RUNTIME_DIR` isn't always set in the environment right? And in nearly all cases, Linux users use the default XDG directories? Should we use something like https://github.com/adrg/xdg?

pkg/image/containerdregistry/resolver.go
```diff
@@ -85,14 +87,31 @@ func credential(cfg *configfile.ConfigFile) func(string) (string, string, error)
...
 	}
 
-	cfg, err := config.Load(dir)
-	if err != nil {
-		return nil, err
+	dockerConfigJSON := filepath.Join(dir, config.ConfigFileName)
+	cfg = configfile.New(dockerConfigJSON)
+
+	switch _, err := os.Stat(dockerConfigJSON); {
+	case err == nil:
```
joelanford, line 99: Does it make sense to always attempt to read both files and then merge them?

pkg/image/containerdregistry/resolver.go
```diff
@@ -85,14 +87,31 @@ func credential(cfg *configfile.ConfigFile) func(string) (string, string, error)
 	}
 }
 
-func loadConfig(dir string) (*configfile.ConfigFile, error) {
+func loadConfig(dir string) (cfg *configfile.ConfigFile, err error) {
```
joelanford, line 90: Nit: any reason to add the named return values? This opens the door for [naked returns](https://go.dev/tour/basics/7), which generally reduce the readability of a function.

joelanford (commented):
Nice!

Just a few comments/questions.

pkg/image/containerdregistry/resolver.go
jpower432: Nice, thanks for sharing this library. I will definitely incorporate this to ensure cross-platform compatibility.

pkg/image/containerdregistry/resolver.go
dinhxuanvu: @jpower432 Any thoughts on this?

jpower432: There is no function in the `config` library that I saw capable of merging, but I could write a helper function to merge the two configs. I definitely think that it would cover more use cases, but we would have to have a preferred option when merging. Assuming we would use the `docker` config as the preferred option?

joelanford: Yeah favoring the docker config option makes sense so we avoid breaking people that already have both and have different credentials for the same registry in the podman auth file.

joelanford: This is a nit, btw. If its hard to do, I don't consider it a must-have.

dinhxuanvu (approved):
Approved.

dinhxuanvu:
@joelanford May you bless this with a lgtm? ;)

pkg/image/containerdregistry/resolver.go
jpower432: Understood. I don't believe it would be overly complicated, but I would like to tackle it in a follow-on PR if that works.

PR #898, opened 2022-01-14
fix(registry): deprecated skip-tls and add use-http and skip-tls-verify

jpower432:
The skip-tls flag currently uses HTTP instead of the expected behavior of skipping
TLS cert validated with HTTPS registries. The new flags seperate untrusted HTTPS from
HTTP registries and behave as expected

This PR deprecated the `skip-tls` flag and create two new flags, `skip-tls-verify` and `use-http`. The new flags disable TLS validation, use plain HTTP, respectively. 

The motivation for the changed is resolve the confusion due to the `skip-tls` flag description and to allow the use of self-signed or untrusted HTTPS when using `opm`

**Reviewer Checklist**

Related to #872

### DIFF
--- a/cmd/opm/alpha/bundle/unpack.go
+++ b/cmd/opm/alpha/bundle/unpack.go
@@ -12,6 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"github.com/operator-framework/operator-registry/cmd/opm/internal/util"
 	"github.com/operator-framework/operator-registry/pkg/image"
 	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
 	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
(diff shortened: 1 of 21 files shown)


jpower432:
cc @dinhxuanvu

dinhxuanvu (approved):
/approve

dinhxuanvu:
@joelanford Would you mind taking a look? Thanks

cmd/opm/alpha/bundle/unpack.go
```diff
@@ -70,14 +75,30 @@ func unpackBundle(cmd *cobra.Command, args []string) error {
...
+	}
+	useHTTP, err = cmd.Flags().GetBool("use-http")
+	if err != nil {
+		return err
+	}
+	if skipTLS {
+		// Set useHTTP when use deprecated skipTlS
+		// for functional parity with existing
+		useHTTP = true
+	}
```
joelanford, line 99: What happens if someone uses `--skip-tls=true --use-http=false`? Should we error out in that case?

Maybe even better, should we just make `skip-tls` mutually exclusive from both `skip-tls-verify` and `use-http`?

joelanford: Also, it seems like `--skip-tls-verify=false --use-http=true` is an invalid combination, right?

jpower432: Agree with doing the flag combination checks. My intent here was to make it simpler to rip out `skip-tls` eventually because it is being deprecated here, but I see how doing that might introduce some unexpected changes since it may not always equate to `use-http` in every case. I will make that change.

joelanford: Not sure if it's possible, but given this code is repeated in so many places, it might be worth taking a look at writing a helper function somewhere in the `cmd` tree (perhaps `cmd/internal/util` or something?) so it can be reused everywhere.

dinhxuanvu:
@joelanford Ready for the final review.

PR #899, opened 2022-01-14
feat(validate): update validate library to warn when multiple properties are defined in properties.yaml as well as CSV

tylerslaton:
To sufficiently address this [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2018095), this PR adds a warning when a user has defined the same property in the CSV's `olm.properties` annotation as well as in a `properties.yaml`. This will act as a soft barrier to users doing this in the future.

As opposed to adding custom logic to the way that OLM handles multiple of the same property being defined, as well as to maintain backwards compatibility, this soft warning was determined to be the best path for the issue.

**Reviewer Checklist**
- [X] Implementation matches the proposed design, or proposal is updated to match implementation
- [X] Sufficient end-to-end test coverage

***Note*** This is currently a work in progress, I am currently working to add a test case which validates this logic is firing correctly. However, I wanted to get opinions on the approach while I do so.


pkg/lib/bundle/validate.go
```diff
@@ -170,6 +192,64 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
+			if len(moreProperties) < len(lessProperties) {
+				moreProperties = fileBasedProperties
+				lessProperties = annotationBasedProperties
+			}
+
+			// For each property in both slices, determine if a property is duplicated, then warn if not done already
+			alreadyWarned := map[string]bool{}
+			for _, mp := range moreProperties {
+				for _, lp := range lessProperties {
+					if mp.Type == lp.Type && !alreadyWarned[mp.Type] {
```
ankitathomas, line 238: nit: can break out of this loop.

pkg/lib/bundle/validate.go
```diff
@@ -170,6 +192,64 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
+			// Determine which props have more values
+			moreProperties := annotationBasedProperties
+			lessProperties := fileBasedProperties
+			if len(moreProperties) < len(lessProperties) {
+				moreProperties = fileBasedProperties
+				lessProperties = annotationBasedProperties
+			}
+
+			// For each property in both slices, determine if a property is duplicated, then warn if not done already
+			alreadyWarned := map[string]bool{}
```
ankitathomas, line 235: If you're only looking at the property types and not checking for multiple values, you can also keep a separate map of types from one property set and validate it against the other.

pkg/lib/bundle/validate.go
```diff
@@ -170,6 +192,64 @@ func (i imageValidator) ValidateBundleFormat(directory string) error {
...
+			for _, mp := range moreProperties {
+				for _, lp := range lessProperties {
+					if mp.Type == lp.Type && !alreadyWarned[mp.Type] {
+						alreadyWarned[mp.Type] = true
+						validationWarnings = append(
+							validationWarnings,
+							fmt.Sprintf("Found property %s defined in both CSV and properties.yaml", mp.Type),
```
ankitathomas, line 242: are we also checking for defining a property multiple times within the same source/defining the property multiple times with the same value?

tylerslaton: This is a good point. Thinking on it, I would say that the main case to worry about (that we don't already cover in this PR) is if the same source has a key/value pair that is identical. It would be good to have a separate warning in that case which shouldn't be too hard to act on. We don't care so much if two different sources have the exact same key/value pair as they're already warned on due to the key. Would you agree?

tylerslaton: Alright, this logic should now be implemented. It also has unit test cases to accompany it.

pkg/lib/bundle/validate.go
tylerslaton: Check the most recent push, I believe it will cover this issue.

tylerslaton:
It was determined that this logic is better served inside of the operator-sdk. As a result, this PR is being closed in favor of this [one](https://github.com/operator-framework/api/pull/217).

PR #900, opened 2022-01-18
adding properties to package and channel

### LINKED ISSUE #895: FBC: Add properties to olm.package and olm.channel
Problem:  There is no simple way to add third-party metadata to File Based Catalogs.  The options include:
1.  Adding labels to the built catalog image.
2. Adding files to the catalog image
3. Adding custom fields to the catalog yaml

The problem with all three of these approaches is that none of the tooling preserves this additional metadata since it is not part of the Go blob schema for these objects.

**Solution:**
To solve this, I'd like to see adding Properties to the `olm.package` and `olm.channel` blobs, just like the properties format for the `olm.bundle` blob.

Example:
```yaml
schema: olm.package
name: foo
defaultChannel: v1
properties:
- type: ibm.epoch
   value: 20220111.223045.00000
- type: ibm.version
   value: 4
```

Slack conversation history:
https://kubernetes.slack.com/archives/C0181L6JYQ2/p1640821737179600

akihikokuroda:
Signed-off-by: akihikokuroda <akihikokuroda2020@gmail.com>

Add Properties to Channel and Package blob for FBC.  They are preserved for the alpha render command.
The unit tests in render_test.go and declcfg_to_model_test.go are updated are updated.

Closes #895 
**Reviewer Checklist**

### DIFF
--- a/alpha/action/render_test.go
+++ b/alpha/action/render_test.go
@@ -3,6 +3,7 @@ package action_test
 import (
 	"context"
 	"embed"
+	"encoding/json"
 	"errors"
 	"io/fs"
 	"os"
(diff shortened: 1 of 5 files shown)


cdjohnson:
Built and tested against my own test catalog and it worked for me.  Thanks!

dinhxuanvu (approved):
/approve

PR #901, opened 2022-01-18
fix(render/diff): Support olm.constraint in opm render/diff cmd

dinhxuanvu:
Currently, the olm.constraint type is ignored and not recognized in
render and diff cmd. This commit is to add supoprt for constraint
type.

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/registry_to_model.go
+++ b/pkg/registry/registry_to_model.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"sort"
 
-	"github.com/operator-framework/api/pkg/constraints"
 	"github.com/operator-framework/operator-registry/alpha/property"
 )
 
@@ -56,11 +55,7 @@ func ObjectsAndPropertiesFromBundle(b *Bundle) ([]string, []property.Property, e
 				return nil, nil, property.ParseError{Idx: i, Typ: p.Type, Err: err}
 			}
 			packageRequiredProps = append(packageRequiredProps, property.MustBuildPackageRequired(v.PackageName, v.Version))
-		case constraints.OLMConstraintType:
-			var v constraints.Constraint
-			if err := json.Unmarshal(p.Value, &v); err != nil {
-				return nil, nil, property.ParseError{Idx: i, Typ: p.Type, Err: err}
-			}
+		default:
 			otherProps = append(otherProps, property.Property{
 				Type:  p.Type,
 				Value: p.Value,
--- a/pkg/registry/registry_to_model.go
+++ b/pkg/registry/registry_to_model.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/operator-framework/api/pkg/constraints"
 	"github.com/operator-framework/operator-registry/alpha/property"
 )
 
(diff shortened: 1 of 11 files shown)


dinhxuanvu:
@joelanford Apparently, `olm.constraint` is ignored by render/diff command for some reason. I'm still testing it out with these changes.

Xia-Zhao-rh:
Hi, @dinhxuanvu
For "opm render", maybe It is needed to add default p.Type here, https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/registry_to_model.go#L57

`	var packageRequiredProps []property.Property
	for i, p := range b.Dependencies {
		switch p.Type {
		case property.TypeGVK:
			...
		case property.TypePackage:
			...
		default:
			packageRequiredProps = append(packageRequiredProps, property.Property{
				Type:  p.Type,
				Value: p.Value,
			})
		}`

dinhxuanvu:
@joelanford Ready for the final review.

pkg/registry/registry_to_model_test.go
```diff
@@ -45,6 +47,10 @@ func testExpectedProperties() []property.Property {
 		property.MustBuildGVK("etcd.database.coreos.com", "v1beta2", "EtcdBackup"),
 		property.MustBuildGVK("etcd.database.coreos.com", "v1beta2", "EtcdRestore"),
 	}
+	props = append(props, property.Property{
+		Type:  "olm.constraint",
+		Value: json.RawMessage(constraint),
+	})
```
joelanford, line 53: Nit: rather using append in a separate statement, in-line this extra property into the original `props` initialization?

pkg/registry/registry_to_model.go
```diff
@@ -55,6 +56,15 @@ func ObjectsAndPropertiesFromBundle(b *Bundle) ([]string, []property.Property, e
...
 			packageRequiredProps = append(packageRequiredProps, property.MustBuildPackageRequired(v.PackageName, v.Version))
+		case constraints.OLMConstraintType:
+			var v constraints.Constraint
+			if err := json.Unmarshal(p.Value, &v); err != nil {
+				return nil, nil, property.ParseError{Idx: i, Typ: p.Type, Err: err}
+			}
+			otherProps = append(otherProps, property.Property{
+				Type:  p.Type,
+				Value: p.Value,
+			})
```
joelanford, line 67: Rather than having special case handling for `olm.constraint` would we lose anything by treating all other `dependency` blobs opaquely?

IIRC, the only reason the existing special case handling exists is because there's translation or FBC-specific validation that we have to do.

```suggestion
		default:
			otherProps = append(otherProps, property.Property{
				Type:  p.Type,
				Value: p.Value,
			})
```

joelanford (commented):
Just to make sure I'm understanding this, the reason we need this logic in the conversion code is because `olm.constraint`s are encoded in bundles as dependencies instead of properties?

I was under the impression that `olm.constraint` was always meant to be introduced as a property, but I also wasn't paying that much attention to that aspect of it.

pkg/registry/registry_to_model.go
dinhxuanvu: I'm fine with this generic approach.

pkg/registry/registry_to_model_test.go
dinhxuanvu: Sounds good.

dinhxuanvu:
> Just to make sure I'm understanding this, the reason we need this logic in the conversion code is because `olm.constraint`s are encoded in bundles as dependencies instead of properties?
> 
> I was under the impression that `olm.constraint` was always meant to be introduced as a property, but I also wasn't paying that much attention to that aspect of it.

The EP said that `olm.constraint` is specified in `dependencies.yaml` as it is intended to be a generic constraint/dependency. If they can be included in `properties.yaml`, it will still work. The fact we do some consolidation behind the scene between dependencies and properties was merely a grpc implementation detail which frankly we just did as a bug fix, not as a proper enhancement change. Up to this point, we still use `olm.gvk` and `olm.package` (and now `olm.constraint) dependencies in `dependencies.yaml` and another other arbitrary properties are in `properties.yaml`. I prefer to keep the convention as it is until we have a plan to consolidate it officially and perhaps deprecate `dependencies.yaml` altogether.

PR #905, opened 2022-01-18
fix(opm): adds removal of empty packages when using opm diff in headsOnly mode

### LINKED ISSUE #904: [BUG] opm: opm diff does not prune empty packages in headsonly mode.
When running `opm alpha diff` without a previous reference, if an empty package is produced due to an input error (e.g., version does not exist in any channels), it should be pruned to ensure the file-based catalog is valid.

Example of a produced catalog (icon removed):
```
{
  "schema": "olm.package",
  "name": "cluster-logging",
  "defaultChannel": "stable",
}

jpower432:
opm diff in headsOnly mode with an include config will prune all empty channels
from a package which can leave the output config with an empty package. This change
will remove any empty packages after the channel pruning.

Fixes #904

This PR removes empty packages when using `opm alpha diff` with no previous reference.

To ensure file-based catalogs are not being created with invalid packages by `opm alpha diff`

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/diff.go
+++ b/alpha/declcfg/diff.go
@@ -101,7 +101,10 @@ func (g *DiffGenerator) Run(oldModel, newModel model.Model) (model.Model, error)
 					if len(ch.Bundles) == 0 {
 						delete(outputPkg.Channels, ch.Name)
 					}
-
+				}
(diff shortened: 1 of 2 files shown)


dinhxuanvu (approved):
/approve

Issue #908, opened 2022-01-26
Add skip-tls-verify and use-http flag support to opm render


ryanemerson:
Currently neither the `skip-tls-verify` or `use-http` are supported by the `opm render` command, meaning it's not possible to pull images from a registry whose CA is unknown or a insecure registry.

PR #909, opened 2022-01-26
Add skip-tls-verify and use-http flag support to opm render #908

ryanemerson:
https://github.com/operator-framework/operator-registry/issues/908

### DIFF
--- a/cmd/opm/render/cmd.go
+++ b/cmd/opm/render/cmd.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/operator-framework/operator-registry/alpha/action"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/cmd/opm/internal/util"
+	containerd "github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
 	"github.com/operator-framework/operator-registry/pkg/sqlite"
 )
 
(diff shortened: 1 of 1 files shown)


alpha/action/render.go
```diff
@@ -56,12 +56,16 @@ type Render struct {
...
+func NullLogger() *logrus.Entry {
 	logger := logrus.New()
 	logger.SetOutput(ioutil.Discard)
 	return logrus.NewEntry(logger)
 }
 
+func NewRenderCacheDir() (string, error) {
+	return os.MkdirTemp("", "render-registry-")
+}
+
```
joelanford, line 68: IMO, we should keep the `NullLogger` and `NewRenderCacheDir` functions out of the exported API of this library.

For such a small amount of code, I would lean toward duplicating these functions in the render CLI code. That would basically mean that the default CLI registry would be decoupled from the default registry created by the action.Render library, and to be honest, I don't see that as a bad thing.

If we think its valuable enough to share this code AND avoid duplication, we could make an `internal/` tree of packages that can only be used by other packages within this repo. e.g. `internal/logutil.NullLogger()` and `internal/renderutil.NewCacheDir()`.

cmd/opm/render/cmd.go
```diff
@@ -44,6 +46,29 @@ func NewCmd() *cobra.Command {
...
+			}
+
+			reg, err := containerd.NewRegistry(
+				containerd.WithCacheDir(cacheDir),
+				containerd.SkipTLSVerify(skipTLSVerify),
+				containerd.WithPlainHTTP(useHTTP),
+				containerd.WithLog(action.NullLogger()),
+			)
+			if err != nil {
+				log.Fatalf("error creating containerd registry: %v", err)
```
joelanford, line 66: really minor nit just to shorten the length of the error message. the context of the log message should make it obvious that an error occurred.
```suggestion
				log.Fatalf("create containerd registry: %v", err)
```

joelanford (commented):
Thanks for the PR! Just a few minor suggestions...

alpha/action/render.go
ryanemerson: I agree that duplicating the code is probably the best way to go. The main reason that I created the `NewRenderCacheDir` was so that the same tmp directory prefix would be used, but I think unless there's another place this is required it's fine just to duplicate the string. I've pushed the changes.

phantomjinx:
Is this likely to be merged soon please?

phantomjinx:
> Is this likely to be merged soon please?

Hey @joelanford, do you have a timeframe for when this is going into a release please?

kevinrizza:
@phantomjinx usually we try to release minor versions of the registry on the same cadence as OLM releases, although our release process hasn't been the most consistent. our last release was only a few weeks ago, so my guess is it may be a little while until the next patch release. but if there's an urgent need for this patch, you could create an issue on this repository, tag this, and we can talk about cutting a release sooner in the upstream olm working group.

ryanemerson:
@kevinrizza I have created https://github.com/operator-framework/operator-registry/issues/925. I have included details of Operatorhub changes that explain why a release is needed by April, thanks.

PR #913, opened 2022-02-01
Using GoReleaser to manage GitHub releases 

aemperador:
Change to use GoReleaser to manage GitHub releases. Originally just wanted to publish ppc64le Github binaries.
@joelanford Please let me know the best way to test this. I ran it locally to ensure minimal errors, but was not able to cover all architectures.

First attempt to add ppc64le binaries to GitHub releases - https://github.com/operator-framework/operator-registry/pull/631 

**Reviewer Checklist**


.goreleaser.yaml
```diff
@@ -82,6 +82,34 @@ builds:
     asmflags: *build-asmflags
     gcflags: *build-gcflags
     ldflags: *build-ldflags
+  - id: darwin-amd64
```
joelanford, line 85: We can't build macos binaries via goreleaser because the macos compiler is not available (and not licensed for use) on Linux.

So the macos binary builds need to continue to be run on `macos-latest` machines via the github action. And then rather than using `goreleaser`'s github release integration, we'll have to tie all this together via github actions to create the release and upload the assets, similar to what we're doing in ` .github/workflows/release.yaml`

joelanford:
Closing since we merged #914 instead. Thanks for helping out!

PR #916, opened 2022-02-09
fix(validate): Fix constraint validation for non-CEL type

dinhxuanvu:
Currently, the constraint type can be either CEL or compound constraint.
The existing constraint validation assumes every constraint is CEL-type
which is not correct. The new validation will parse JSON into constraint
type and will validate CEL expression if it is CEL-type.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/bundle/validate.go
+++ b/pkg/lib/bundle/validate.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 
+	"github.com/operator-framework/api/pkg/constraints"
 	"github.com/operator-framework/api/pkg/manifests"
 	v1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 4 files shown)


pkg/lib/bundle/validate.go
```diff
@@ -237,8 +242,8 @@ func validateDependencies(dependenciesFile *registry.DependenciesFile) []error {
 				errs = dp.Validate()
 			case registry.LabelDependency:
 				errs = dp.Validate()
-			case registry.CelConstraint:
-				errs = dp.Validate()
+			case constraints.Constraint:
+				errs = registry.ValidateCEL(dp)
```
joelanford, line 246: Ideally, we'd be validating the entire constraint here, right?

Can we do one of the following?
1. Go ahead and validate the entire constraint (probably with a recursive validation to handle the compound constraint aspect of it)
2. OR add a TODO comment here and create an issue in the repo so that we remember to follow back up.

Also, do we need to add this same validation for when we encounter a constraint in the bundle properties?

perdasilva:

PR #922, opened 2022-02-22
Bump go to v1.17

awgreene:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.8 AS builder
```
timflannagan, line 1: TIL we still have this Dockerfile and it was missed when we were separating out the downstream builds. Maybe this is something we can rework in the future to avoid using non-external images? cc @joelanford

go.mod
```diff
@@ -65,5 +50,118 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
+require (
```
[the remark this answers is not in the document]
timflannagan: This kind of pattern is more-or-less standard now with how Go 1.17 handles go.mod files now: https://go.dev/doc/go1.17#go-command.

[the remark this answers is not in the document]
awgreene: Odd, going to look into this.

/hold

test/e2e/ctx/provisioner_kubeconfig.go
```diff
@@ -1,3 +1,4 @@
+//go:build !kind
```
timflannagan, line 1: unrelated: This points to the need for a shared operator-framework/test-infra repository so we're not de-duplicating work across the registry and OLM repositories.

go.mod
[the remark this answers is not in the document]
awgreene: /unhold
https://stackoverflow.com/questions/69432877/why-there-are-two-require-blocks-in-go-mod-since-go-1-17?noredirect=1&lq=1

Dockerfile
joelanford: Yeah, in general I think there's a bunch of cleanup of Dockerfiles and quay build triggers we can do in this repo now.

Do we need anything other than the `quay.io/operator-framework/opm` image at this point?

go.mod
```diff
@@ -65,5 +50,118 @@ require (
...
+	k8s.io/component-base v0.22.1 // indirect
+	k8s.io/klog/v2 v2.9.0 // indirect
+	k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e // indirect
+	k8s.io/utils v0.0.0-20210802155522-efc7438f0176 // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.22 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
+)
+
 // latest tag resolves to a very old version. this is only used for spinning up local test registries
 replace github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
```
dinhxuanvu, line 167: Should this `replace` go to the `replace` blob above as well? I know this should work regardless but looks a bit weird. Are we doing this for visibility reason?

dinhxuanvu (approved):
/approve

go.mod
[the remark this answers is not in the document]
exdx: Ok thanks for the info 👍

go.mod
exdx: I had the same comment above, see https://github.com/operator-framework/operator-registry/pull/922#discussion_r813007935 -- basically this is normal starting with Go 1.17


> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies. 
https://go.dev/doc/go1.17#go-command

dinhxuanvu: OK. I can live with that.

PR #923, opened 2022-02-23
fix: add more descriptive error message to error thrown when attempt to prune FBC

jbuchananr:
Provides a more descriptive error when `opm index prune` is called on a file based catalog letting users know that it only supports sqlite-based catalogs. 

Currently, an unclear error message is thrown while calling `opm index prune` on a file based catalog. 
E.g `"Error: index image xxxx missing label operators.operatorframework.io.index.database.v1"`

Fixes [Bug2039135](https://bugzilla.redhat.com/show_bug.cgi?id=2039135#)

**Reviewer Checklist**


pkg/lib/indexer/indexer.go
```diff
@@ -363,6 +363,9 @@ func (i ImageIndexer) getDatabaseFile(workingDir, fromIndex, caFile string, skip
 
 	dbLocation, ok := labels[containertools.DbLocationLabel]
 	if !ok {
+		if _, ok := labels[containertools.ConfigsLocationLabel]; ok {
```
perdasilva, line 366: I think it might be worth adding a regression test

jbuchananr: Working on it! +1

PR #937, opened 2022-03-21
fix(opm diff): handle cyclic dependency graph

### LINKED ISSUE #936: [BUG] opm alpha diff hangs when processing the Red Hat 4.10 catalog
# Version
Version: version.Version{OpmVersion:"v1.21.0", GitCommit:"9999f796", BuildDate:"2022-03-03T21:23:12Z", GoOs:"linux", GoArch:"amd64"}

# What Happened

When running `opm alpha diff` with this Red Hat 4.10 catalog it hangs
```
opm alpha diff registry.redhat.io/redhat/redhat-operator-index:v4.10
...
WARN[0034] DEPRECATION NOTICE:
Sqlite-based catalogs and their related subcommands are deprecated. Support for
them will be removed in a future release. Please migrate your catalog workflows
to the new file-based catalog format. 
...

```

This does not occur with other catalogs (certified catalog) or with an `include.yaml`.

# Expected behavior

The catalog differential to successfully complete

/kind bug

anik120:
While generating a diff from a catalog that has operators that specify dependencies
that are cyclic in nature, eg a->b, b->a, the `opm alpha diff` command hangs. This
was happening because while generating the diff, the command does a breadth-first
search of the dependency graph generated by the operator bundles, but did not keep
a track of the already visited bundles. As a result, when there was a cycle in the
dependency graph, the command was stuck in an infinite loop.

This PR fixes the issue by keeping track of the already visited bundles during the
search, and moving the search forward with only the bundles that haven't been visited
before.

Fixes #936

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/diff.go
+++ b/alpha/declcfg/diff.go
@@ -242,8 +242,10 @@ func bundlesEqual(b1, b2 *model.Bundle) (bool, error) {
 func addAllDependencies(newModel, oldModel, outputModel model.Model) error {
 	// Get every oldModel's bundle's dependencies, and their dependencies, etc. by BFS.
 	providingBundlesByPackage := map[string][]*model.Bundle{}
-	for curr := getBundles(outputModel); len(curr) != 0; {
-		reqGVKs, reqPkgs, err := findDependencies(curr)
+	var visitedBundles []*model.Bundle
+	for currentList := getBundles(outputModel); len(currentList) != 0; {
+		visitedBundles = append(visitedBundles, currentList...)
+		reqGVKs, reqPkgs, err := findDependencies(currentList)
 		if err != nil {
 			return err
 		}
@@ -251,15 +253,16 @@ func addAllDependencies(newModel, oldModel, outputModel model.Model) error {
 		if len(reqGVKs) == 0 && len(reqPkgs) == 0 {
 			break
 		}
-		curr = nil
+		currentList = nil
 		// Get bundles that provide dependencies from newModel, which should have
 		// the latest bundles of each dependency package.
 		for _, pkg := range newModel {
 			providingBundles := getBundlesThatProvide(pkg, reqGVKs, reqPkgs)
-			curr = append(curr, providingBundles...)
+			unvisitedProvidingBundles := difference(visitedBundles, providingBundles)
+			currentList = append(currentList, unvisitedProvidingBundles...)
 
 			oldPkg, oldHasPkg := oldModel[pkg.Name]
-			for _, b := range providingBundles {
+			for _, b := range unvisitedProvidingBundles {
 				// If the bundle is not in oldModel, add it to the set.
 				// outputModel is checked below.
 				add := true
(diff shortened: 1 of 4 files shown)


exdx (commented):
I've reviewed the implementation and I'm not 100% sure I follow everything, but it seems intuitive that ensuring that the bundles in the dependency graph are unique should not impact the diff ultimately. Also the CI is green so more confidence there. 

/lgtm

alpha/declcfg/diff.go
```diff
@@ -242,24 +242,27 @@ func bundlesEqual(b1, b2 *model.Bundle) (bool, error) {
...
 			break
 		}
-		curr = nil
+		currentList = nil
 		// Get bundles that provide dependencies from newModel, which should have
 		// the latest bundles of each dependency package.
 		for _, pkg := range newModel {
 			providingBundles := getBundlesThatProvide(pkg, reqGVKs, reqPkgs)
-			curr = append(curr, providingBundles...)
+			uniqueProvidingBundle := difference(visitedBundles, providingBundles)
```
joelanford, line 261: Minor naming nit:
```suggestion
			unvisitedProvidingBundles := difference(visitedBundles, providingBundles)
```

alpha/declcfg/diff.go
```diff
@@ -242,24 +242,27 @@ func bundlesEqual(b1, b2 *model.Bundle) (bool, error) {
 func addAllDependencies(newModel, oldModel, outputModel model.Model) error {
 	// Get every oldModel's bundle's dependencies, and their dependencies, etc. by BFS.
 	providingBundlesByPackage := map[string][]*model.Bundle{}
-	for curr := getBundles(outputModel); len(curr) != 0; {
-		reqGVKs, reqPkgs, err := findDependencies(curr)
+	var visitedBundles []*model.Bundle
```
joelanford, line 245: Should we store this as a map from the get-go, rather than regenerating the map during every call to `difference()`
```suggestion
	visitedBundles := map[*model.Bundle]struct{}{}
```

anik120: Defining it as a slice here helps with working with the output of `getBundles` [here](https://github.com/operator-framework/operator-registry/pull/937/files#diff-d90d4db089c34d739cc336d4e85f9a924f2ce9c5ddda212f18f00515234513d6R247). The idea was to restrict the time complexity of difference to o(n), so doesn't defining `visitedBundle` as a map mean we won't essentially get any computational reduction, but instead move the complexity from one place to another?

joelanford: I suppose the difference would manifest if there was a way for lots of duplicates to be added to the `visitedBundles` list. But I didn't trace the code enough to know if that happens.

The scenario I'm thinking of:
- the catalog has 1000 bundles in it.
- `foo.v0.1.0` is a bundle in the catalog
- all other bundles depend on `foo.v0.1.0`.

In that scenario, does `visitedBundles` end up with 999 copies of `foo.v0.1.0`?

anik120: In your example, with the 999 bundles existing in level 0 of the graph, `findDependencies` returning a map of the required gvks/packages ensures that `visitedBundles` has only one copy of `foo.v0.1.0`. 

It gets a little more interesting with a graph like this: 

(a, b, c...)-------------
|                             | ---- z
x------------------------

In that case, `visitedBundles` does end up with two copies of z. 

But, is preventing that worth converting `visitedBundles` map (with the 999 entires) to a list every time we enter the loop, i.e when we visit each level of the graph?

joelanford: Perhaps not. You've clearly thought about this more than I have so I'll definitely defer to you. And since this is all just discussion about optimization, I don't think its necessary to keep blocking on that.

If we later run into a case where this becomes a problem, we can take a deeper look then.

PR #941, opened 2022-04-07
`opm alpha veneer-render basic` subcommand

grokspawn:
Implements a "Basic" `Veneer` rendering capability, to illustrate the potential benefits of the use of `veneers` as entry points into the FBC ecosystem.  A `veneer` is identified as an input schema which may be processed to generate an FBC.  The `basic veneer` is a schema which is effectively a subset of an FBC, but allows the user to omit `olm.bundle` metadata which the tool can generate, thus reducing the amount of content a user has to actively maintain. 

In a very simple (non-functioning) example, we define a simple `olm.package` and an `olm.channel` composed of a couple of `olm.bundle` which have an image name attribute but no properties
```yaml
---
schema: olm.package
name: example-operator
defaultChannel: stable
---
schema: olm.channel
package: example-operator
name: stable
entries:
- name: example-operator.v0.1.0
- name: example-operator.v0.2.0
  replaces: example-operator.v0.1.0
---
schema: olm.bundle
image: docker.io/example/example-operator-bundle:0.1.0
---
schema: olm.bundle
image: docker.io/example-operator-bundle:0.2.0
```

Using the `opm alpha veneer render` command on this input generates the full FBC:
```yaml
---
defaultChannel: stable
name: example-operator
schema: olm.package
---
entries:
- name: example-operator.v0.1.0
- name: example-operator.v0.2.0
  replaces: example-operator.v0.1.0
name: stable
package: example-operator
schema: olm.channel
---
image: docker.io/example/example-operator-bundle:0.1.0
name: example-operator.v0.1.0
package: example-operator
properties:
- type: olm.gvk
  value:
    group: example.com
    kind: App
    version: v1
- type: olm.package
  value:
    packageName: example-operator
    version: 0.1.0
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoidjEiLCJraW5kIjoiU2VydmljZSIsIm1ldGFkYXRhIjp7ImNyZWF0aW9uVGltZXN0YW1wIjpudWxsLCJsYWJlbHMiOnsiY29udHJvbC1wbGFuZSI6ImNvbnRyb2xsZXItbWFuYWdlciJ9LCJuYW1lIjoiZXhhbXBsZS1vcGVyYXRvci1jb250cm9sbGVyLW1hbmFnZXItbWV0cmljcy1zZXJ2aWNlIn0sInNwZWMiOnsicG9ydHMiOlt7Im5hbWUiOiJodHRwcyIsInBvcnQiOjg0NDMsInByb3RvY29sIjoiVENQIiwidGFyZ2V0UG9ydCI6Imh0dHBzIn1dLCJzZWxlY3RvciI6eyJjb250cm9sLXBsYW5lIjoiY29udHJvbGxlci1tYW5hZ2VyIn19LCJzdGF0dXMiOnsibG9hZEJhbGFuY2VyIjp7fX19
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoidjEiLCJkYXRhIjp7ImNvbnRyb2xsZXJfbWFuYWdlcl9jb25maWcueWFtbCI6ImFwaVZlcnNpb246IGNvbnRyb2xsZXItcnVudGltZS5zaWdzLms4cy5pby92MWFscGhhMVxua2luZDogQ29udHJvbGxlck1hbmFnZXJDb25maWdcbmhlYWx0aDpcbiAgaGVhbHRoUHJvYmVCaW5kQWRkcmVzczogOjgwODFcbm1ldHJpY3M6XG4gIGJpbmRBZGRyZXNzOiAxMjcuMC4wLjE6ODA4MFxud2ViaG9vazpcbiAgcG9ydDogOTQ0M1xubGVhZGVyRWxlY3Rpb246XG4gIGxlYWRlckVsZWN0OiB0cnVlXG4gIHJlc291cmNlTmFtZTogY2RmNjA0MTIuY29tXG4ifSwia2luZCI6IkNvbmZpZ01hcCIsIm1ldGFkYXRhIjp7Im5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLW1hbmFnZXItY29uZmlnIn19
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoicmJhYy5hdXRob3JpemF0aW9uLms4cy5pby92MSIsImtpbmQiOiJDbHVzdGVyUm9sZSIsIm1ldGFkYXRhIjp7ImNyZWF0aW9uVGltZXN0YW1wIjpudWxsLCJuYW1lIjoiZXhhbXBsZS1vcGVyYXRvci1tZXRyaWNzLXJlYWRlciJ9LCJydWxlcyI6W3sibm9uUmVzb3VyY2VVUkxzIjpbIi9tZXRyaWNzIl0sInZlcmJzIjpbImdldCJdfV19
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoib3BlcmF0b3JzLmNvcmVvcy5jb20vdjFhbHBoYTEiLCJraW5kIjoiQ2x1c3RlclNlcnZpY2VWZXJzaW9uIiwibWV0YWRhdGEiOnsiYW5ub3RhdGlvbnMiOnsiYWxtLWV4YW1wbGVzIjoiW1xuICB7XG4gICAgXCJhcGlWZXJzaW9uXCI6IFwiZXhhbXBsZS5jb20vdjFcIixcbiAgICBcImtpbmRcIjogXCJBcHBcIixcbiAgICBcIm1ldGFkYXRhXCI6IHtcbiAgICAgIFwibmFtZVwiOiBcImFwcC1zYW1wbGVcIlxuICAgIH0sXG4gICAgXCJzcGVjXCI6IHtcbiAgICAgIFwiZm9vXCI6IFwiYmFyXCJcbiAgICB9XG4gIH1cbl0iLCJjYXBhYmlsaXRpZXMiOiJCYXNpYyBJbnN0YWxsIiwib3BlcmF0b3JzLm9wZXJhdG9yZnJhbWV3b3JrLmlvL2J1aWxkZXIiOiJvcGVyYXRvci1zZGstdjEuMTEuMCtnaXQiLCJvcGVyYXRvcnMub3BlcmF0b3JmcmFtZXdvcmsuaW8vcHJvamVjdF9sYXlvdXQiOiJnby5rdWJlYnVpbGRlci5pby92MyJ9LCJuYW1lIjoiZXhhbXBsZS1vcGVyYXRvci52MC4xLjAiLCJuYW1lc3BhY2UiOiJwbGFjZWhvbGRlciJ9LCJzcGVjIjp7ImFwaXNlcnZpY2VkZWZpbml0aW9ucyI6e30sImN1c3RvbXJlc291cmNlZGVmaW5pdGlvbnMiOnsib3duZWQiOlt7ImRlc2NyaXB0aW9uIjoiQXBwIGlzIHRoZSBTY2hlbWEgZm9yIHRoZSBhcHBzIEFQSSIsImRpc3BsYXlOYW1lIjoiQXBwIiwia2luZCI6IkFwcCIsIm5hbWUiOiJhcHBzLmV4YW1wbGUuY29tIiwidmVyc2lvbiI6InYxIn1dfSwiZGlzcGxheU5hbWUiOiJFeGFtcGxlIE9wZXJhdG9yIiwiaWNvbiI6W3siYmFzZTY0ZGF0YSI6IiIsIm1lZGlhdHlwZSI6IiJ9XSwiaW5zdGFsbCI6eyJzcGVjIjp7ImNsdXN0ZXJQZXJtaXNzaW9ucyI6W3sicnVsZXMiOlt7ImFwaUdyb3VwcyI6WyJleGFtcGxlLmNvbSJdLCJyZXNvdXJjZXMiOlsiYXBwcyJdLCJ2ZXJicyI6WyJjcmVhdGUiLCJkZWxldGUiLCJnZXQiLCJsaXN0IiwicGF0Y2giLCJ1cGRhdGUiLCJ3YXRjaCJdfSx7ImFwaUdyb3VwcyI6WyJleGFtcGxlLmNvbSJdLCJyZXNvdXJjZXMiOlsiYXBwcy9maW5hbGl6ZXJzIl0sInZlcmJzIjpbInVwZGF0ZSJdfSx7ImFwaUdyb3VwcyI6WyJleGFtcGxlLmNvbSJdLCJyZXNvdXJjZXMiOlsiYXBwcy9zdGF0dXMiXSwidmVyYnMiOlsiZ2V0IiwicGF0Y2giLCJ1cGRhdGUiXX0seyJhcGlHcm91cHMiOlsiYXV0aGVudGljYXRpb24uazhzLmlvIl0sInJlc291cmNlcyI6WyJ0b2tlbnJldmlld3MiXSwidmVyYnMiOlsiY3JlYXRlIl19LHsiYXBpR3JvdXBzIjpbImF1dGhvcml6YXRpb24uazhzLmlvIl0sInJlc291cmNlcyI6WyJzdWJqZWN0YWNjZXNzcmV2aWV3cyJdLCJ2ZXJicyI6WyJjcmVhdGUiXX1dLCJzZXJ2aWNlQWNjb3VudE5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLWNvbnRyb2xsZXItbWFuYWdlciJ9XSwiZGVwbG95bWVudHMiOlt7Im5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLWNvbnRyb2xsZXItbWFuYWdlciIsInNwZWMiOnsicmVwbGljYXMiOjEsInNlbGVjdG9yIjp7Im1hdGNoTGFiZWxzIjp7ImNvbnRyb2wtcGxhbmUiOiJjb250cm9sbGVyLW1hbmFnZXIifX0sInN0cmF0ZWd5Ijp7fSwidGVtcGxhdGUiOnsibWV0YWRhdGEiOnsibGFiZWxzIjp7ImNvbnRyb2wtcGxhbmUiOiJjb250cm9sbGVyLW1hbmFnZXIifX0sInNwZWMiOnsiY29udGFpbmVycyI6W3siYXJncyI6WyItLXNlY3VyZS1saXN0ZW4tYWRkcmVzcz0wLjAuMC4wOjg0NDMiLCItLXVwc3RyZWFtPWh0dHA6Ly8xMjcuMC4wLjE6ODA4MC8iLCItLWxvZ3Rvc3RkZXJyPXRydWUiLCItLXY9MTAiXSwiaW1hZ2UiOiJnY3IuaW8va3ViZWJ1aWxkZXIva3ViZS1yYmFjLXByb3h5OnYwLjguMCIsIm5hbWUiOiJrdWJlLXJiYWMtcHJveHkiLCJwb3J0cyI6W3siY29udGFpbmVyUG9ydCI6ODQ0MywibmFtZSI6Imh0dHBzIiwicHJvdG9jb2wiOiJUQ1AifV0sInJlc291cmNlcyI6e319LHsiYXJncyI6WyItLWhlYWx0aC1wcm9iZS1iaW5kLWFkZHJlc3M9OjgwODEiLCItLW1ldHJpY3MtYmluZC1hZGRyZXNzPTEyNy4wLjAuMTo4MDgwIiwiLS1sZWFkZXItZWxlY3QiXSwiY29tbWFuZCI6WyIvbWFuYWdlciJdLCJpbWFnZSI6InF1YXkuaW8vam9lbGFuZm9yZC9leGFtcGxlLW9wZXJhdG9yOjAuMS4wIiwibGl2ZW5lc3NQcm9iZSI6eyJodHRwR2V0Ijp7InBhdGgiOiIvaGVhbHRoeiIsInBvcnQiOjgwODF9LCJpbml0aWFsRGVsYXlTZWNvbmRzIjoxNSwicGVyaW9kU2Vjb25kcyI6MjB9LCJuYW1lIjoibWFuYWdlciIsInJlYWRpbmVzc1Byb2JlIjp7Imh0dHBHZXQiOnsicGF0aCI6Ii9yZWFkeXoiLCJwb3J0Ijo4MDgxfSwiaW5pdGlhbERlbGF5U2Vjb25kcyI6NSwicGVyaW9kU2Vjb25kcyI6MTB9LCJyZXNvdXJjZXMiOnsibGltaXRzIjp7ImNwdSI6IjIwMG0iLCJtZW1vcnkiOiIxMDBNaSJ9LCJyZXF1ZXN0cyI6eyJjcHUiOiIxMDBtIiwibWVtb3J5IjoiMjBNaSJ9fSwic2VjdXJpdHlDb250ZXh0Ijp7ImFsbG93UHJpdmlsZWdlRXNjYWxhdGlvbiI6ZmFsc2V9fV0sInNlY3VyaXR5Q29udGV4dCI6eyJydW5Bc05vblJvb3QiOnRydWV9LCJzZXJ2aWNlQWNjb3VudE5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLWNvbnRyb2xsZXItbWFuYWdlciIsInRlcm1pbmF0aW9uR3JhY2VQZXJpb2RTZWNvbmRzIjoxMH19fX1dLCJwZXJtaXNzaW9ucyI6W3sicnVsZXMiOlt7ImFwaUdyb3VwcyI6WyIiXSwicmVzb3VyY2VzIjpbImNvbmZpZ21hcHMiXSwidmVyYnMiOlsiZ2V0IiwibGlzdCIsIndhdGNoIiwiY3JlYXRlIiwidXBkYXRlIiwicGF0Y2giLCJkZWxldGUiXX0seyJhcGlHcm91cHMiOlsiY29vcmRpbmF0aW9uLms4cy5pbyJdLCJyZXNvdXJjZXMiOlsibGVhc2VzIl0sInZlcmJzIjpbImdldCIsImxpc3QiLCJ3YXRjaCIsImNyZWF0ZSIsInVwZGF0ZSIsInBhdGNoIiwiZGVsZXRlIl19LHsiYXBpR3JvdXBzIjpbIiJdLCJyZXNvdXJjZXMiOlsiZXZlbnRzIl0sInZlcmJzIjpbImNyZWF0ZSIsInBhdGNoIl19XSwic2VydmljZUFjY291bnROYW1lIjoiZXhhbXBsZS1vcGVyYXRvci1jb250cm9sbGVyLW1hbmFnZXIifV19LCJzdHJhdGVneSI6ImRlcGxveW1lbnQifSwiaW5zdGFsbE1vZGVzIjpbeyJzdXBwb3J0ZWQiOmZhbHNlLCJ0eXBlIjoiT3duTmFtZXNwYWNlIn0seyJzdXBwb3J0ZWQiOmZhbHNlLCJ0eXBlIjoiU2luZ2xlTmFtZXNwYWNlIn0seyJzdXBwb3J0ZWQiOmZhbHNlLCJ0eXBlIjoiTXVsdGlOYW1lc3BhY2UifSx7InN1cHBvcnRlZCI6dHJ1ZSwidHlwZSI6IkFsbE5hbWVzcGFjZXMifV0sIm1hdHVyaXR5IjoiYWxwaGEiLCJwcm92aWRlciI6e30sInZlcnNpb24iOiIwLjEuMCJ9fQ==
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoiYXBpZXh0ZW5zaW9ucy5rOHMuaW8vdjEiLCJraW5kIjoiQ3VzdG9tUmVzb3VyY2VEZWZpbml0aW9uIiwibWV0YWRhdGEiOnsiYW5ub3RhdGlvbnMiOnsiY29udHJvbGxlci1nZW4ua3ViZWJ1aWxkZXIuaW8vdmVyc2lvbiI6InYwLjYuMSJ9LCJjcmVhdGlvblRpbWVzdGFtcCI6bnVsbCwibmFtZSI6ImFwcHMuZXhhbXBsZS5jb20ifSwic3BlYyI6eyJncm91cCI6ImV4YW1wbGUuY29tIiwibmFtZXMiOnsia2luZCI6IkFwcCIsImxpc3RLaW5kIjoiQXBwTGlzdCIsInBsdXJhbCI6ImFwcHMiLCJzaW5ndWxhciI6ImFwcCJ9LCJzY29wZSI6Ik5hbWVzcGFjZWQiLCJ2ZXJzaW9ucyI6W3sibmFtZSI6InYxIiwic2NoZW1hIjp7Im9wZW5BUElWM1NjaGVtYSI6eyJkZXNjcmlwdGlvbiI6IkFwcCBpcyB0aGUgU2NoZW1hIGZvciB0aGUgYXBwcyBBUEkiLCJwcm9wZXJ0aWVzIjp7ImFwaVZlcnNpb24iOnsiZGVzY3JpcHRpb24iOiJBUElWZXJzaW9uIGRlZmluZXMgdGhlIHZlcnNpb25lZCBzY2hlbWEgb2YgdGhpcyByZXByZXNlbnRhdGlvbiBvZiBhbiBvYmplY3QuIFNlcnZlcnMgc2hvdWxkIGNvbnZlcnQgcmVjb2duaXplZCBzY2hlbWFzIHRvIHRoZSBsYXRlc3QgaW50ZXJuYWwgdmFsdWUsIGFuZCBtYXkgcmVqZWN0IHVucmVjb2duaXplZCB2YWx1ZXMuIE1vcmUgaW5mbzogaHR0cHM6Ly9naXQuazhzLmlvL2NvbW11bml0eS9jb250cmlidXRvcnMvZGV2ZWwvc2lnLWFyY2hpdGVjdHVyZS9hcGktY29udmVudGlvbnMubWQjcmVzb3VyY2VzIiwidHlwZSI6InN0cmluZyJ9LCJraW5kIjp7ImRlc2NyaXB0aW9uIjoiS2luZCBpcyBhIHN0cmluZyB2YWx1ZSByZXByZXNlbnRpbmcgdGhlIFJFU1QgcmVzb3VyY2UgdGhpcyBvYmplY3QgcmVwcmVzZW50cy4gU2VydmVycyBtYXkgaW5mZXIgdGhpcyBmcm9tIHRoZSBlbmRwb2ludCB0aGUgY2xpZW50IHN1Ym1pdHMgcmVxdWVzdHMgdG8uIENhbm5vdCBiZSB1cGRhdGVkLiBJbiBDYW1lbENhc2UuIE1vcmUgaW5mbzogaHR0cHM6Ly9naXQuazhzLmlvL2NvbW11bml0eS9jb250cmlidXRvcnMvZGV2ZWwvc2lnLWFyY2hpdGVjdHVyZS9hcGktY29udmVudGlvbnMubWQjdHlwZXMta2luZHMiLCJ0eXBlIjoic3RyaW5nIn0sIm1ldGFkYXRhIjp7InR5cGUiOiJvYmplY3QifSwic3BlYyI6eyJkZXNjcmlwdGlvbiI6IkFwcFNwZWMgZGVmaW5lcyB0aGUgZGVzaXJlZCBzdGF0ZSBvZiBBcHAiLCJwcm9wZXJ0aWVzIjp7ImZvbyI6eyJkZXNjcmlwdGlvbiI6IkZvbyBpcyBhbiBleGFtcGxlIGZpZWxkIG9mIEFwcC4gRWRpdCBhcHBfdHlwZXMuZ28gdG8gcmVtb3ZlL3VwZGF0ZSIsInR5cGUiOiJzdHJpbmcifX0sInR5cGUiOiJvYmplY3QifSwic3RhdHVzIjp7ImRlc2NyaXB0aW9uIjoiQXBwU3RhdHVzIGRlZmluZXMgdGhlIG9ic2VydmVkIHN0YXRlIG9mIEFwcCIsInR5cGUiOiJvYmplY3QifX0sInR5cGUiOiJvYmplY3QifX0sInNlcnZlZCI6dHJ1ZSwic3RvcmFnZSI6dHJ1ZSwic3VicmVzb3VyY2VzIjp7InN0YXR1cyI6e319fV19LCJzdGF0dXMiOnsiYWNjZXB0ZWROYW1lcyI6eyJraW5kIjoiIiwicGx1cmFsIjoiIn0sImNvbmRpdGlvbnMiOltdLCJzdG9yZWRWZXJzaW9ucyI6W119fQ==
relatedImages:
- image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
  name: ""
- image: docker.io/example/example-operator-bundle:0.1.0
  name: ""
- image: docker.io/example/example-operator:0.1.0
  name: ""
schema: olm.bundle
---
image: docker.io/example/example-operator-bundle:0.2.0
name: example-operator.v0.2.0
package: example-operator
properties:
- type: olm.gvk
  value:
    group: example.com
    kind: App
    version: v1
- type: olm.package
  value:
    packageName: example-operator
    version: 0.2.0
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoidjEiLCJraW5kIjoiU2VydmljZSIsIm1ldGFkYXRhIjp7ImNyZWF0aW9uVGltZXN0YW1wIjpudWxsLCJsYWJlbHMiOnsiY29udHJvbC1wbGFuZSI6ImNvbnRyb2xsZXItbWFuYWdlciJ9LCJuYW1lIjoiZXhhbXBsZS1vcGVyYXRvci1jb250cm9sbGVyLW1hbmFnZXItbWV0cmljcy1zZXJ2aWNlIn0sInNwZWMiOnsicG9ydHMiOlt7Im5hbWUiOiJodHRwcyIsInBvcnQiOjg0NDMsInByb3RvY29sIjoiVENQIiwidGFyZ2V0UG9ydCI6Imh0dHBzIn1dLCJzZWxlY3RvciI6eyJjb250cm9sLXBsYW5lIjoiY29udHJvbGxlci1tYW5hZ2VyIn19LCJzdGF0dXMiOnsibG9hZEJhbGFuY2VyIjp7fX19
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoidjEiLCJkYXRhIjp7ImNvbnRyb2xsZXJfbWFuYWdlcl9jb25maWcueWFtbCI6ImFwaVZlcnNpb246IGNvbnRyb2xsZXItcnVudGltZS5zaWdzLms4cy5pby92MWFscGhhMVxua2luZDogQ29udHJvbGxlck1hbmFnZXJDb25maWdcbmhlYWx0aDpcbiAgaGVhbHRoUHJvYmVCaW5kQWRkcmVzczogOjgwODFcbm1ldHJpY3M6XG4gIGJpbmRBZGRyZXNzOiAxMjcuMC4wLjE6ODA4MFxud2ViaG9vazpcbiAgcG9ydDogOTQ0M1xubGVhZGVyRWxlY3Rpb246XG4gIGxlYWRlckVsZWN0OiB0cnVlXG4gIHJlc291cmNlTmFtZTogY2RmNjA0MTIuY29tXG4ifSwia2luZCI6IkNvbmZpZ01hcCIsIm1ldGFkYXRhIjp7Im5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLW1hbmFnZXItY29uZmlnIn19
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoicmJhYy5hdXRob3JpemF0aW9uLms4cy5pby92MSIsImtpbmQiOiJDbHVzdGVyUm9sZSIsIm1ldGFkYXRhIjp7ImNyZWF0aW9uVGltZXN0YW1wIjpudWxsLCJuYW1lIjoiZXhhbXBsZS1vcGVyYXRvci1tZXRyaWNzLXJlYWRlciJ9LCJydWxlcyI6W3sibm9uUmVzb3VyY2VVUkxzIjpbIi9tZXRyaWNzIl0sInZlcmJzIjpbImdldCJdfV19
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoib3BlcmF0b3JzLmNvcmVvcy5jb20vdjFhbHBoYTEiLCJraW5kIjoiQ2x1c3RlclNlcnZpY2VWZXJzaW9uIiwibWV0YWRhdGEiOnsiYW5ub3RhdGlvbnMiOnsiYWxtLWV4YW1wbGVzIjoiW1xuICB7XG4gICAgXCJhcGlWZXJzaW9uXCI6IFwiZXhhbXBsZS5jb20vdjFcIixcbiAgICBcImtpbmRcIjogXCJBcHBcIixcbiAgICBcIm1ldGFkYXRhXCI6IHtcbiAgICAgIFwibmFtZVwiOiBcImFwcC1zYW1wbGVcIlxuICAgIH0sXG4gICAgXCJzcGVjXCI6IHtcbiAgICAgIFwiZm9vXCI6IFwiYmFyXCJcbiAgICB9XG4gIH1cbl0iLCJjYXBhYmlsaXRpZXMiOiJCYXNpYyBJbnN0YWxsIiwib3BlcmF0b3JzLm9wZXJhdG9yZnJhbWV3b3JrLmlvL2J1aWxkZXIiOiJvcGVyYXRvci1zZGstdjEuMTEuMCtnaXQiLCJvcGVyYXRvcnMub3BlcmF0b3JmcmFtZXdvcmsuaW8vcHJvamVjdF9sYXlvdXQiOiJnby5rdWJlYnVpbGRlci5pby92MyJ9LCJuYW1lIjoiZXhhbXBsZS1vcGVyYXRvci52MC4yLjAiLCJuYW1lc3BhY2UiOiJwbGFjZWhvbGRlciJ9LCJzcGVjIjp7ImFwaXNlcnZpY2VkZWZpbml0aW9ucyI6e30sImN1c3RvbXJlc291cmNlZGVmaW5pdGlvbnMiOnsib3duZWQiOlt7ImRlc2NyaXB0aW9uIjoiQXBwIGlzIHRoZSBTY2hlbWEgZm9yIHRoZSBhcHBzIEFQSSIsImRpc3BsYXlOYW1lIjoiQXBwIiwia2luZCI6IkFwcCIsIm5hbWUiOiJhcHBzLmV4YW1wbGUuY29tIiwidmVyc2lvbiI6InYxIn1dfSwiZGlzcGxheU5hbWUiOiJFeGFtcGxlIE9wZXJhdG9yIiwiaWNvbiI6W3siYmFzZTY0ZGF0YSI6IiIsIm1lZGlhdHlwZSI6IiJ9XSwiaW5zdGFsbCI6eyJzcGVjIjp7ImNsdXN0ZXJQZXJtaXNzaW9ucyI6W3sicnVsZXMiOlt7ImFwaUdyb3VwcyI6WyJleGFtcGxlLmNvbSJdLCJyZXNvdXJjZXMiOlsiYXBwcyJdLCJ2ZXJicyI6WyJjcmVhdGUiLCJkZWxldGUiLCJnZXQiLCJsaXN0IiwicGF0Y2giLCJ1cGRhdGUiLCJ3YXRjaCJdfSx7ImFwaUdyb3VwcyI6WyJleGFtcGxlLmNvbSJdLCJyZXNvdXJjZXMiOlsiYXBwcy9maW5hbGl6ZXJzIl0sInZlcmJzIjpbInVwZGF0ZSJdfSx7ImFwaUdyb3VwcyI6WyJleGFtcGxlLmNvbSJdLCJyZXNvdXJjZXMiOlsiYXBwcy9zdGF0dXMiXSwidmVyYnMiOlsiZ2V0IiwicGF0Y2giLCJ1cGRhdGUiXX0seyJhcGlHcm91cHMiOlsiYXV0aGVudGljYXRpb24uazhzLmlvIl0sInJlc291cmNlcyI6WyJ0b2tlbnJldmlld3MiXSwidmVyYnMiOlsiY3JlYXRlIl19LHsiYXBpR3JvdXBzIjpbImF1dGhvcml6YXRpb24uazhzLmlvIl0sInJlc291cmNlcyI6WyJzdWJqZWN0YWNjZXNzcmV2aWV3cyJdLCJ2ZXJicyI6WyJjcmVhdGUiXX1dLCJzZXJ2aWNlQWNjb3VudE5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLWNvbnRyb2xsZXItbWFuYWdlciJ9XSwiZGVwbG95bWVudHMiOlt7Im5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLWNvbnRyb2xsZXItbWFuYWdlciIsInNwZWMiOnsicmVwbGljYXMiOjEsInNlbGVjdG9yIjp7Im1hdGNoTGFiZWxzIjp7ImNvbnRyb2wtcGxhbmUiOiJjb250cm9sbGVyLW1hbmFnZXIifX0sInN0cmF0ZWd5Ijp7fSwidGVtcGxhdGUiOnsibWV0YWRhdGEiOnsibGFiZWxzIjp7ImNvbnRyb2wtcGxhbmUiOiJjb250cm9sbGVyLW1hbmFnZXIifX0sInNwZWMiOnsiY29udGFpbmVycyI6W3siYXJncyI6WyItLXNlY3VyZS1saXN0ZW4tYWRkcmVzcz0wLjAuMC4wOjg0NDMiLCItLXVwc3RyZWFtPWh0dHA6Ly8xMjcuMC4wLjE6ODA4MC8iLCItLWxvZ3Rvc3RkZXJyPXRydWUiLCItLXY9MTAiXSwiaW1hZ2UiOiJnY3IuaW8va3ViZWJ1aWxkZXIva3ViZS1yYmFjLXByb3h5OnYwLjguMCIsIm5hbWUiOiJrdWJlLXJiYWMtcHJveHkiLCJwb3J0cyI6W3siY29udGFpbmVyUG9ydCI6ODQ0MywibmFtZSI6Imh0dHBzIiwicHJvdG9jb2wiOiJUQ1AifV0sInJlc291cmNlcyI6e319LHsiYXJncyI6WyItLWhlYWx0aC1wcm9iZS1iaW5kLWFkZHJlc3M9OjgwODEiLCItLW1ldHJpY3MtYmluZC1hZGRyZXNzPTEyNy4wLjAuMTo4MDgwIiwiLS1sZWFkZXItZWxlY3QiXSwiY29tbWFuZCI6WyIvbWFuYWdlciJdLCJpbWFnZSI6InF1YXkuaW8vam9lbGFuZm9yZC9leGFtcGxlLW9wZXJhdG9yOjAuMi4wIiwibGl2ZW5lc3NQcm9iZSI6eyJodHRwR2V0Ijp7InBhdGgiOiIvaGVhbHRoeiIsInBvcnQiOjgwODF9LCJpbml0aWFsRGVsYXlTZWNvbmRzIjoxNSwicGVyaW9kU2Vjb25kcyI6MjB9LCJuYW1lIjoibWFuYWdlciIsInJlYWRpbmVzc1Byb2JlIjp7Imh0dHBHZXQiOnsicGF0aCI6Ii9yZWFkeXoiLCJwb3J0Ijo4MDgxfSwiaW5pdGlhbERlbGF5U2Vjb25kcyI6NSwicGVyaW9kU2Vjb25kcyI6MTB9LCJyZXNvdXJjZXMiOnsibGltaXRzIjp7ImNwdSI6IjIwMG0iLCJtZW1vcnkiOiIxMDBNaSJ9LCJyZXF1ZXN0cyI6eyJjcHUiOiIxMDBtIiwibWVtb3J5IjoiMjBNaSJ9fSwic2VjdXJpdHlDb250ZXh0Ijp7ImFsbG93UHJpdmlsZWdlRXNjYWxhdGlvbiI6ZmFsc2V9fV0sInNlY3VyaXR5Q29udGV4dCI6eyJydW5Bc05vblJvb3QiOnRydWV9LCJzZXJ2aWNlQWNjb3VudE5hbWUiOiJleGFtcGxlLW9wZXJhdG9yLWNvbnRyb2xsZXItbWFuYWdlciIsInRlcm1pbmF0aW9uR3JhY2VQZXJpb2RTZWNvbmRzIjoxMH19fX1dLCJwZXJtaXNzaW9ucyI6W3sicnVsZXMiOlt7ImFwaUdyb3VwcyI6WyIiXSwicmVzb3VyY2VzIjpbImNvbmZpZ21hcHMiXSwidmVyYnMiOlsiZ2V0IiwibGlzdCIsIndhdGNoIiwiY3JlYXRlIiwidXBkYXRlIiwicGF0Y2giLCJkZWxldGUiXX0seyJhcGlHcm91cHMiOlsiY29vcmRpbmF0aW9uLms4cy5pbyJdLCJyZXNvdXJjZXMiOlsibGVhc2VzIl0sInZlcmJzIjpbImdldCIsImxpc3QiLCJ3YXRjaCIsImNyZWF0ZSIsInVwZGF0ZSIsInBhdGNoIiwiZGVsZXRlIl19LHsiYXBpR3JvdXBzIjpbIiJdLCJyZXNvdXJjZXMiOlsiZXZlbnRzIl0sInZlcmJzIjpbImNyZWF0ZSIsInBhdGNoIl19XSwic2VydmljZUFjY291bnROYW1lIjoiZXhhbXBsZS1vcGVyYXRvci1jb250cm9sbGVyLW1hbmFnZXIifV19LCJzdHJhdGVneSI6ImRlcGxveW1lbnQifSwiaW5zdGFsbE1vZGVzIjpbeyJzdXBwb3J0ZWQiOmZhbHNlLCJ0eXBlIjoiT3duTmFtZXNwYWNlIn0seyJzdXBwb3J0ZWQiOmZhbHNlLCJ0eXBlIjoiU2luZ2xlTmFtZXNwYWNlIn0seyJzdXBwb3J0ZWQiOmZhbHNlLCJ0eXBlIjoiTXVsdGlOYW1lc3BhY2UifSx7InN1cHBvcnRlZCI6dHJ1ZSwidHlwZSI6IkFsbE5hbWVzcGFjZXMifV0sIm1hdHVyaXR5IjoiYWxwaGEiLCJwcm92aWRlciI6e30sInZlcnNpb24iOiIwLjIuMCJ9fQ==
- type: olm.bundle.object
  value:
    data: eyJhcGlWZXJzaW9uIjoiYXBpZXh0ZW5zaW9ucy5rOHMuaW8vdjEiLCJraW5kIjoiQ3VzdG9tUmVzb3VyY2VEZWZpbml0aW9uIiwibWV0YWRhdGEiOnsiYW5ub3RhdGlvbnMiOnsiY29udHJvbGxlci1nZW4ua3ViZWJ1aWxkZXIuaW8vdmVyc2lvbiI6InYwLjYuMSJ9LCJjcmVhdGlvblRpbWVzdGFtcCI6bnVsbCwibmFtZSI6ImFwcHMuZXhhbXBsZS5jb20ifSwic3BlYyI6eyJncm91cCI6ImV4YW1wbGUuY29tIiwibmFtZXMiOnsia2luZCI6IkFwcCIsImxpc3RLaW5kIjoiQXBwTGlzdCIsInBsdXJhbCI6ImFwcHMiLCJzaW5ndWxhciI6ImFwcCJ9LCJzY29wZSI6Ik5hbWVzcGFjZWQiLCJ2ZXJzaW9ucyI6W3sibmFtZSI6InYxIiwic2NoZW1hIjp7Im9wZW5BUElWM1NjaGVtYSI6eyJkZXNjcmlwdGlvbiI6IkFwcCBpcyB0aGUgU2NoZW1hIGZvciB0aGUgYXBwcyBBUEkiLCJwcm9wZXJ0aWVzIjp7ImFwaVZlcnNpb24iOnsiZGVzY3JpcHRpb24iOiJBUElWZXJzaW9uIGRlZmluZXMgdGhlIHZlcnNpb25lZCBzY2hlbWEgb2YgdGhpcyByZXByZXNlbnRhdGlvbiBvZiBhbiBvYmplY3QuIFNlcnZlcnMgc2hvdWxkIGNvbnZlcnQgcmVjb2duaXplZCBzY2hlbWFzIHRvIHRoZSBsYXRlc3QgaW50ZXJuYWwgdmFsdWUsIGFuZCBtYXkgcmVqZWN0IHVucmVjb2duaXplZCB2YWx1ZXMuIE1vcmUgaW5mbzogaHR0cHM6Ly9naXQuazhzLmlvL2NvbW11bml0eS9jb250cmlidXRvcnMvZGV2ZWwvc2lnLWFyY2hpdGVjdHVyZS9hcGktY29udmVudGlvbnMubWQjcmVzb3VyY2VzIiwidHlwZSI6InN0cmluZyJ9LCJraW5kIjp7ImRlc2NyaXB0aW9uIjoiS2luZCBpcyBhIHN0cmluZyB2YWx1ZSByZXByZXNlbnRpbmcgdGhlIFJFU1QgcmVzb3VyY2UgdGhpcyBvYmplY3QgcmVwcmVzZW50cy4gU2VydmVycyBtYXkgaW5mZXIgdGhpcyBmcm9tIHRoZSBlbmRwb2ludCB0aGUgY2xpZW50IHN1Ym1pdHMgcmVxdWVzdHMgdG8uIENhbm5vdCBiZSB1cGRhdGVkLiBJbiBDYW1lbENhc2UuIE1vcmUgaW5mbzogaHR0cHM6Ly9naXQuazhzLmlvL2NvbW11bml0eS9jb250cmlidXRvcnMvZGV2ZWwvc2lnLWFyY2hpdGVjdHVyZS9hcGktY29udmVudGlvbnMubWQjdHlwZXMta2luZHMiLCJ0eXBlIjoic3RyaW5nIn0sIm1ldGFkYXRhIjp7InR5cGUiOiJvYmplY3QifSwic3BlYyI6eyJkZXNjcmlwdGlvbiI6IkFwcFNwZWMgZGVmaW5lcyB0aGUgZGVzaXJlZCBzdGF0ZSBvZiBBcHAiLCJwcm9wZXJ0aWVzIjp7ImJhciI6eyJkZXNjcmlwdGlvbiI6IkJhciBpcyBhbiBleGFtcGxlIGZpZWxkIG9mIEFwcC4iLCJ0eXBlIjoic3RyaW5nIn0sImZvbyI6eyJkZXNjcmlwdGlvbiI6IkZvbyBpcyBhbiBleGFtcGxlIGZpZWxkIG9mIEFwcC4iLCJ0eXBlIjoic3RyaW5nIn19LCJ0eXBlIjoib2JqZWN0In0sInN0YXR1cyI6eyJkZXNjcmlwdGlvbiI6IkFwcFN0YXR1cyBkZWZpbmVzIHRoZSBvYnNlcnZlZCBzdGF0ZSBvZiBBcHAiLCJ0eXBlIjoib2JqZWN0In19LCJ0eXBlIjoib2JqZWN0In19LCJzZXJ2ZWQiOnRydWUsInN0b3JhZ2UiOnRydWUsInN1YnJlc291cmNlcyI6eyJzdGF0dXMiOnt9fX1dfSwic3RhdHVzIjp7ImFjY2VwdGVkTmFtZXMiOnsia2luZCI6IiIsInBsdXJhbCI6IiJ9LCJjb25kaXRpb25zIjpbXSwic3RvcmVkVmVyc2lvbnMiOltdfX0=
relatedImages:
- image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
  name: ""
- image: docker.io/example/example-operator-bundle:0.2.0
  name: ""
- image: docker.io/example/example-operator:0.2.0
  name: ""
schema: olm.bundle

```

Illustrate a possible effort-saving work flow for operator authors to use to facilitate migration to `FBC`.

**Reviewer Checklist**
- [X] Implementation matches the proposed design, or proposal is updated to match implementation
- [X] Sufficient unit test coverage 
- [X] Sufficient end-to-end test coverage
- [X] Docs updated or added to `/docs` 

### DIFF
--- a/alpha/declcfg/load.go
+++ b/alpha/declcfg/load.go
@@ -31,6 +31,7 @@ func WalkFS(root fs.FS, walkFn WalkFunc) error {
 	if root == nil {
 		return fmt.Errorf("no declarative config filesystem provided")
 	}
+
 	matcher, err := ignore.NewMatcher(root, indexIgnoreFilename)
 	if err != nil {
 		return err
@@ -45,18 +46,12 @@ func WalkFS(root fs.FS, walkFn WalkFunc) error {
 		if info.IsDir() || info.Name() == indexIgnoreFilename || matcher.Match(path, false) {
 			return nil
 		}
-		file, err := root.Open(path)
-		if err != nil {
-			return walkFn(path, nil, err)
-		}
-		defer file.Close()
-		cfg, err := readYAMLOrJSON(file)
+
+		cfg, err := LoadFile(root, path)
 		if err != nil {
 			return walkFn(path, cfg, err)
 		}
-		if err := readBundleObjects(cfg.Bundles, root, path); err != nil {
-			return fmt.Errorf("read bundle objects: %v", err)
-		}
+
 		return walkFn(path, cfg, err)
 	})
 }
@@ -166,5 +161,26 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 			cfg.Others = append(cfg.Others, in)
 		}
 	}
+	return cfg, nil
+}
+
+// LoadFile will unmarshall declarative config components from a single filename provided in 'path'
+// located at a filesystem hierarchy 'root'
+func LoadFile(root fs.FS, path string) (*DeclarativeConfig, error) {
+	file, err := root.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	cfg, err := readYAMLOrJSON(file)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := readBundleObjects(cfg.Bundles, root, path); err != nil {
+		return nil, fmt.Errorf("read bundle objects: %v", err)
+	}
+
 	return cfg, nil
 }
--- a/alpha/veneer/basic/basic.go
+++ b/alpha/veneer/basic/basic.go
@@ -1,0 +1,73 @@
+package basic
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/pkg/image"
+)
+
+type Veneer struct {
+	Registry image.Registry
+}
+
+func (v Veneer) Render(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
+	// only taking first argument as file
+	stat, serr := os.Stat(ref)
+	if serr != nil {
+		return nil, serr
+	}
+
+	if stat.IsDir() {
+		return nil, errors.New("cannot render veneers by directory reference")
+	}
+	return v.renderFile(ctx, ref)
+}
+
+func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
+	// xform any relative to absolute paths
+	abspath, err := filepath.Abs(ref)
+	if err != nil {
+		return nil, err
+	}
+	// xform to break apart dir/file elements
+	rpath, fname := filepath.Split(abspath)
+	root := os.DirFS(rpath)
+
+	cfg, err := declcfg.LoadFile(root, fname)
+	if err != nil {
+		return cfg, err
+	}
+
+	outb := cfg.Bundles[:0] // allocate based on max size of input, but empty slice
+	// populate registry, incl any flags from CLI, and enforce only rendering bundle images
+	r := action.Render{
+		Registry:       v.Registry,
+		AllowedRefMask: action.RefBundleImage,
+	}
+
+	for _, b := range cfg.Bundles {
+		if !isBundleVeneer(&b) {
+			outb = append(outb, b)
+			continue
+		}
+		r.Refs = []string{b.Image}
+		contributor, err := r.Run(ctx)
+		if err != nil {
+			return nil, err
+		}
+		outb = append(outb, contributor.Bundles...)
+	}
+
+	cfg.Bundles = outb
+	return cfg, nil
+}
+
+// isBundleVeneer identifies loaded partial Bundle data from YAML/JSON veneer source as having no properties,
+func isBundleVeneer(b *declcfg.Bundle) bool {
+	return len(b.Properties) == 0
+}
--- a/cmd/opm/alpha/veneer/basic.go
+++ b/cmd/opm/alpha/veneer/basic.go
@@ -1,0 +1,83 @@
+package veneer
+
+import (
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/veneer/basic"
+	"github.com/operator-framework/operator-registry/cmd/opm/internal/util"
+	containerd "github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
+)
+
+func newBasicVeneerRenderCmd() *cobra.Command {
+	var (
+		veneer basic.Veneer
+		output string
+	)
+	cmd := &cobra.Command{
+		Use:   "basic basic-veneer-file",
+		Short: "Generate a declarative config blob from a single 'basic veneer' file",
+		Long:  `Generate a declarative config blob from a single 'basic veneer' file, typified as a declarative configuration file where olm.bundle objects have no properties`,
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+
+			var write func(declcfg.DeclarativeConfig, io.Writer) error
+			switch output {
+			case "yaml":
+				write = declcfg.WriteYAML
+			case "json":
+				write = declcfg.WriteJSON
+			default:
+				log.Fatalf("invalid --output value %q, expected (json|yaml)", output)
+			}
+
+			// The bundle loading impl is somewhat verbose, even on the happy path,
+			// so discard all logrus default logger logs. Any important failures will be
+			// returned from veneer.Render and logged as fatal errors.
+			logrus.SetOutput(ioutil.Discard)
+
+			skipTLSVerify, useHTTP, err := util.GetTLSOptions(cmd)
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			cacheDir, err := os.MkdirTemp("", "veneer-registry-")
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			reg, err := containerd.NewRegistry(
+				containerd.WithCacheDir(cacheDir),
+				containerd.SkipTLSVerify(skipTLSVerify),
+				containerd.WithPlainHTTP(useHTTP),
+				containerd.WithLog(nullLogger()),
+			)
+			if err != nil {
+				log.Fatalf("creating containerd registry: %v", err)
+			}
+			defer reg.Destroy()
+
+			veneer.Registry = reg
+
+			// only taking first file argument
+			cfg, err := veneer.Render(cmd.Context(), args[0])
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			if err := write(*cfg, os.Stdout); err != nil {
+				log.Fatal(err)
+			}
+		},
+	}
+	cmd.Flags().StringVarP(&output, "output", "o", "json", "Output format (json|yaml)")
+	cmd.Flags().Bool("skip-tls-verify", false, "disable TLS verification")
+	cmd.Flags().Bool("use-http", false, "use plain HTTP")
+	return cmd
+}
--- a/alpha/declcfg/load_test.go
+++ b/alpha/declcfg/load_test.go
@@ -821,3 +821,88 @@ present in the .indexignore file.`),
 		"unrecognized-schema.json": unrecognizedSchema,
 	}
 )
+
+func TestLoadFile(t *testing.T) {
+	type spec struct {
+		name              string
+		fsys              fs.FS
+		path              string
+		assertion         require.ErrorAssertionFunc
+		expectNumPackages int
+		expectNumBundles  int
+		expectNumOthers   int
+	}
+	specs := []spec{
+		{
+			name:      "Error/NonExistentDir",
+			fsys:      os.DirFS("non/existent/dir/"),
+			assertion: require.Error,
+		},
+		{
+			name:      "Error/Invalid",
+			fsys:      invalidFS,
+			assertion: require.Error,
+		},
+		{
+			name:      "Error/NotYAMLOrJSON",
+			fsys:      invalidFS,
+			path:      "invalid-format.txt",
+			assertion: require.Error,
+		},
+		{
+			name:      "Error/NotJSONObject",
+			fsys:      invalidFS,
+			path:      "not-object.json",
+			assertion: require.Error,
+		},
+		{
+			name:      "Error/NoSchema",
+			fsys:      invalidFS,
+			path:      "no-schema.yaml",
+			assertion: require.Error,
+		},
+		{
+			name:      "Error/InvalidPackageJSON",
+			fsys:      invalidFS,
+			path:      "invalid-package.json",
+			assertion: require.Error,
+		},
+		{
+			name:      "Error/InvalidBundleJSON",
+			fsys:      invalidFS,
+			path:      "invalid-bundle.json",
+			assertion: require.Error,
+		},
+		{
+			name:              "Success/UnrecognizedSchema",
+			fsys:              validFS,
+			path:              "unrecognized-schema.json",
+			assertion:         require.NoError,
+			expectNumPackages: 1,
+			expectNumBundles:  1,
+			expectNumOthers:   1,
+		},
+		{
+			name:              "Success/ValidFile",
+			fsys:              validFS,
+			path:              "etcd.yaml",
+			assertion:         require.NoError,
+			expectNumPackages: 1,
+			expectNumBundles:  6,
+			expectNumOthers:   0,
+		},
+	}
+
+	for _, s := range specs {
+		t.Run(s.name, func(t *testing.T) {
+			cfg, err := LoadFile(s.fsys, s.path)
+			s.assertion(t, err)
+			if err == nil {
+				require.NotNil(t, cfg)
+				assert.Equal(t, len(cfg.Packages), s.expectNumPackages, "unexpected package count")
+				assert.Equal(t, len(cfg.Bundles), s.expectNumBundles, "unexpected bundle count")
+				assert.Equal(t, len(cfg.Others), s.expectNumOthers, "unexpected others count")
+			}
+		})
+	}
+}
--- a/cmd/opm/alpha/cmd.go
+++ b/cmd/opm/alpha/cmd.go
@@ -6,6 +6,7 @@ import (
 	"github.com/operator-framework/operator-registry/cmd/opm/alpha/bundle"
 	"github.com/operator-framework/operator-registry/cmd/opm/alpha/diff"
 	"github.com/operator-framework/operator-registry/cmd/opm/alpha/list"
+	"github.com/operator-framework/operator-registry/cmd/opm/alpha/veneer"
 )
 
 func NewCmd() *cobra.Command {
@@ -20,6 +21,7 @@ func NewCmd() *cobra.Command {
 		bundle.NewCmd(),
 		list.NewCmd(),
 		diff.NewCmd(),
+		veneer.NewCmd(),
 	)
 	return runCmd
 }
--- a/cmd/opm/alpha/veneer/cmd.go
+++ b/cmd/opm/alpha/veneer/cmd.go
@@ -1,0 +1,26 @@
+package veneer
+
+import (
+	"io/ioutil"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func nullLogger() *logrus.Entry {
+	logger := logrus.New()
+	logger.SetOutput(ioutil.Discard)
+	return logrus.NewEntry(logger)
+}
+
+func NewCmd() *cobra.Command {
+	runCmd := &cobra.Command{
+		Use:   "render-veneer",
+		Short: "Render a veneer type",
+		Args:  cobra.NoArgs,
+	}
+
+	runCmd.AddCommand(newBasicVeneerRenderCmd())
+
+	return runCmd
+}


joelanford (commented):
Looks reasonable from a functional standpoint at first glance. However I'd suggest some tweaks to the code layout.

At this stage, I don't think we should add veneer execution to `action.Render` since that action is not alpha in the same sense that the veneers are (for `render` the Go API is alpha, the CLI is not; for veneers, it's all alpha).

My advice would be to have a `opm alpha veneer <veneerName>` command structure, where each `<veneerName>` gets its own `cobra.Command` with whatever flags/positional arguments it wants. I suspect that each veneer will have a different set of inputs and configurations, so it makes sense to me to carve out a fully isolated way for each veneer to be executed.

I'd also suggest a go package structure that's specific to the veneers. E.g.
- `cmd/opm/alpha/veneer/<veneerName>/` package for the CLI
- `alpha/veneer/<veneerName>/` package for the Go API. Then each veneer package can define all of its types and APIs in that package, which helps us avoid alpha-level veneer stuff creeping into our more mature (and I think more stable) FBC code. 

In my mind, this structure achieves two things:
1. Makes it easy to rip out veneers we decide to get rid of later.
2. Makes it more obvious when there's duplication and gives us an opportunity to specifically address that duplication when it comes up.

grokspawn:
New changes aren't ready to merge yet, but are indicative of my understanding w.r.t. divorcing the basic veneer from the render api/cmd, and I wanted to ensure that they were perceived as moving in the right direction.  
I'd still like to clean up some naming, probably adjust the directory structure as well, and definitely get some clarification about the API.

cmd/opm/alpha/veneer/render.go.bak
```diff
@@ -0,0 +1,82 @@
+package veneer
```
tylerslaton, line 1: Was this file intentionally committed?

cmd/opm/alpha/veneer/render.go
```diff
@@ -0,0 +1,82 @@
...
+
+			var write func(declcfg.DeclarativeConfig, io.Writer) error
+			switch output {
+			case "yaml":
+				write = declcfg.WriteYAML
+			case "json":
+				write = declcfg.WriteJSON
+			default:
+				log.Fatalf("invalid --output value %q, expected (json|yaml)", output)
+			}
```
tylerslaton, line 38: You could mark this `output` flag as required and add another layer of validation to it.

For an example of making a flag required:
https://github.com/operator-framework/operator-registry/blob/9e59597a435341d857283011c8ede1c16e5e292b/cmd/opm/alpha/bundle/build.go#L49

cmd/opm/alpha/veneer/render.go
```diff
@@ -0,0 +1,82 @@
...
+
+func newBasicVeneerRenderCmd() *cobra.Command {
+	var (
+		veneer basic.BasicVeneer
+		output string
+	)
+	cmd := &cobra.Command{
+		Use:   "render veneer-file...",
+		Short: "Generate a declarative config blob from one or more veneer files",
+		Long:  `Generate a declarative config blob from one or more veneer files`,
```
tylerslaton, line 26: I'd consider adding a bit more documentation around this as, at least in general, it is a better user experience. Example:

https://github.com/operator-framework/operator-registry/blob/9e59597a435341d857283011c8ede1c16e5e292b/cmd/opm/alpha/bundle/build.go#L25

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,313 @@
...
+	cfg.Bundles = outb
+	return cfg, nil
+}
+
+// identify loaded partial Bundle data from YAML/JSON veneer source as missing `olm.bundle.object` properties,
+// so `true` if:
+// - no properties
+// - no olm.bundle.object property
+func isBundleVeneer(b *declcfg.Bundle) bool {
+
```
tylerslaton, line 301: super nit:
```suggestion
```

alpha/declcfg/load.go
```diff
@@ -168,3 +168,20 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
...
+
+func LoadFile(root fs.FS, file fs.File, path string) (*DeclarativeConfig, error) {
+	var (
+		cfg = &DeclarativeConfig{}
+		err error
+	)
+
+	if cfg, err = readYAMLOrJSON(file); err != nil {
+		return nil, err
+	}
```
tylerslaton, line 180: Thoughts?
```suggestion
	if cfg, err := readYAMLOrJSON(file); err != nil {
		return nil, err
	}
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return reg, nil
+}
+
+func (v BasicVeneer) renderReference(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
+	if stat, serr := os.Stat(ref); serr == nil {
+		if stat.IsDir() {
+			return nil, fmt.Errorf("cannot ingest veneer directory: %w", ErrNotAllowed) // BUGBUG enhance to walk and eat all files in turn, heeding ignore file
```
tylerslaton, line 92: I'd advocate for a more descriptive error message and variable name here. Something like

```suggestion
			return nil, ErrCannotIngestDirectory // BUGBUG enhance to walk and eat all files in turn, heeding ignore file
```

Where `ErrCannotIngestDirectory` is the exact same error you defined above (maybe with a "why", but that's up to you) ^

```go
var ErrCannotIngestDirectory = errors.New("cannot ingest veneer directory") // Also potentially add a why here or in the return.
```

tylerslaton (commented):
Sending out an initial round of review your way. Will try to get some more out tomorrow!

cmd/opm/alpha/veneer/render.go
grokspawn: I'm following the pattern of the 'opm alpha render' command, where we default to JSON output unless countermanded with this option.

alpha/declcfg/load.go
grokspawn: I tried that, but then I get errors:
```alpha/declcfg/load.go:173:5: cfg declared but not used
alpha/declcfg/load.go:177:30: undefined: cfg
alpha/declcfg/load.go:178:10: undefined: cfg
alpha/declcfg/load.go:181:9: undefined: cfg
make: *** [bin/opm] Error 2
```

tylerslaton: Ah, no you had it right before. Forgot that you're checking the error inline here.

joelanford: The more idiomatic way to do this is:
```suggestion
	cfg, err := readYAMLOrJSON(file)
	if err != nil {
		return nil, err
	}
```

alpha/veneer/basic/basic.go
joelanford: IMO, the motivating factor for specific error variables like that is "do I need to check for that specific error from a calling scope?" 

If the answer is yes, it's better to create the variable such that callers can use `errors.Is(err, ErrCannotIngestDirectory)` instead of `strings.Contain(err.Error(), "cannot ingest directory")

If the answer is no, it's better to just use `fmt.Errorf`. Also, it should be noted that using `%w` in the error message format instead of `%v` is a defacto public API because it enables callers to use `errors.Is` and `errors.As` to match against wrapped errors. I generally always default to `%v` for this reason. It's simple to change a `%v` to a `%w` when necessary without breaking any callers, but not the other way around. See https://go.dev/blog/go1.13-errors#whether-to-wrap

cmd/opm/alpha/veneer/render.go.bak
grokspawn: Um... SURE ... *that's* the ticket ...

alpha/declcfg/load.go
```diff
@@ -168,3 +168,20 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	}
 	return cfg, nil
 }
+
+func LoadFile(root fs.FS, file fs.File, path string) (*DeclarativeConfig, error) {
```
joelanford, line 172:
```suggestion
func LoadFile(root fs.FS, file io.Reader, path string) (*DeclarativeConfig, error) {
```

Does this work?

alpha/declcfg/load.go
```diff
@@ -168,3 +168,20 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
...
+	var (
+		cfg = &DeclarativeConfig{}
+		err error
+	)
+
+	if cfg, err = readYAMLOrJSON(file); err != nil {
+		return nil, err
+	}
+
+	if err = readBundleObjects(cfg.Bundles, root, path); err != nil {
```
joelanford, line 182:
```suggestion
	if err := readBundleObjects(cfg.Bundles, root, path); err != nil {
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,101 @@
...
+	var cfg = &declcfg.DeclarativeConfig{}
+
+	// xform any relative to absolute paths
+	abspath, err := filepath.Abs(ref)
+	if err != nil {
+		return nil, err
+	}
+	// xform to break apart dir/file elements
+	rpath := filepath.Dir(abspath)
+	fname := filepath.Base(abspath)
```
joelanford, line 45:
```suggestion
	rpath, fname := filepath.Split(abspath)
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,101 @@
...
+	// only taking first argument as file
+	stat, serr := os.Stat(ref)
+	if serr == nil {
+		if stat.IsDir() {
+			return nil, ErrCannotIngestDirectory // BUGBUG enhance to walk and eat all files in turn, heeding ignore file
+		} else {
+			return v.renderFile(ctx, ref)
+		}
+	}
+	return nil, serr
```
joelanford, line 32: In Go, its idiomatic to keep indents to a minimum, so you see bail out kind of patterns such that happy path typically flows to the end of the function. I'd suggest:
```suggestion
	stat, err := os.Stat(ref)
	if err != nil {
		return nil, serr
	}
	if stat.IsDir() {
		return nil, ErrCannotIngestDirectory // BUGBUG enhance to walk and eat all files in turn, heeding ignore file
	}
	return v.renderFile(ctx, ref)
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,101 @@
...
+
+	cfg.Bundles = outb
+	return cfg, nil
+}
+
+// identify loaded partial Bundle data from YAML/JSON veneer source as missing `olm.bundle.object` properties,
+// so `true` if:
+// - no properties
+// - no olm.bundle.object property
+func isBundleVeneer(b *declcfg.Bundle) bool {
```
joelanford, line 89: This isn't a good detection mechanism -- in lots of catalogs only the channel head bundles will have `olm.bundle.object` properties.

I think a better detection mechanism would be simply `len(b.Properties) == 0`

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,101 @@
...
+	"path/filepath"
+
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
+	"github.com/operator-framework/operator-registry/pkg/image"
+)
+
+var ErrNotAllowed = errors.New("not allowed")
+var ErrCannotIngestDirectory = errors.New("cannot ingest veneers by directory reference")
```
joelanford, line 16: Do we need to check for these specific errors anywhere? If not, I'd suggest either unexporting them or calling `errors.New()` directly in the return statement.

alpha/declcfg/load.go
grokspawn: Yes.  Committed after testing both new `veneer render` and "regular" render opm commands.

alpha/veneer/basic/basic.go
grokspawn: I did not know.  I've updated to look for empty properties instead.

alpha/veneer/basic/basic.go
grokspawn: The error would be needed in one place, and on an alpha-level tool I'm disinclined to invest a lot of effort in it.  Based on other comments to unexport it if not needed externally, IMHO this is better off simpler.  
If not, drop a comment and we'll update it.

alpha/declcfg/load.go
grokspawn: Works.  Tested for both regular and veneer rendering.

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,83 @@
...
+	if cfg, err = declcfg.LoadFile(root, file, ref); err != nil {
+		return cfg, err
+	}
+
+	var outb = cfg.Bundles[:0] // allocate based on max size of input, but empty slice
+	var r action.Render
+	// populate registry, incl any flags from CLI
+	if v.Registry != nil {
+		r.Registry = v.Registry
+	}
```
joelanford, line 59: This can be simplified to:
```suggestion
	// populate registry, incl any flags from CLI
	r := action.Render{
		Registry: v.Registry
	}
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+
+	defer file.Close()
+
+	if cfg, err = declcfg.LoadFile(root, file, ref); err != nil {
+		return cfg, err
+	}
+
+	var outb = cfg.Bundles[:0] // allocate based on max size of input, but empty slice
```
joelanford, line 54: Nit: 
```suggestion
	outb := cfg.Bundles[:0] // allocate based on max size of input, but empty slice
```

alpha/veneer/basic/basic.go
joelanford: I'd also suggest enforcing that we only allow rendering bundle images, by setting `AllowedRefMask: action.RefBundleImage` on the `action.Render` struct

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,83 @@
...
+			contributor, err := r.Run(ctx)
+			if err != nil {
+				// abort processing
+				return cfg, err
+			} else {
+				outb = append(outb, contributor.Bundles...)
+			}
+		} else {
+			outb = append(outb, b)
+		}
```
joelanford, line 73: Nit: outdent the happy path

```suggestion
		if !isBundleVeneer(&b) {
			outb = append(outb, b)
			continue
		}
		r.Refs = []string{b.Image}
		contributor, err := r.Run(ctx)
		if err != nil {
			// abort processing
			return cfg, err
		}
		outb = append(outb, contributor.Bundles...)
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,83 @@
...
+		r.Registry = v.Registry
+	}
+
+	for _, b := range cfg.Bundles {
+		if isBundleVeneer(&b) {
+			r.Refs = []string{b.Image}
+			contributor, err := r.Run(ctx)
+			if err != nil {
+				// abort processing
+				return cfg, err
```
joelanford, line 67: Nit: Always return `nil` for the first return value when errors occur unless there's a good reason not to.
```suggestion
				return nil, err
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,83 @@
...
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/pkg/image"
+)
+
+type BasicVeneer struct {
```
joelanford, line 14: Nit: avoid unnecessary stuttering between the package name `basic` and the type `BasicVeneer`
```suggestion
type Veneer struct {
```

When used in other packages, this looks like `basic.Veneer`

alpha/veneer/basic/basic.go
grokspawn: Before the big refactoring yesterday, I had the code structured to accumulate between loop iterations, so we might have partial good results to report.  I figured that giving partial results and the error allowed the caller to determine what they wanted to do with the results. 

Now, of course all this is moot, because we only accumulate on successful iteration over ALL bundles.  

... long way of saying "I missed it in the refactoring / will fix". :S

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,83 @@
...
+	Registry image.Registry
+}
+
+func (v BasicVeneer) Ingest(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
```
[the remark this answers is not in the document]
grokspawn: Original design had the `basic.Veneer` holding the DeclarativeConfig as a member, for further transformation based on re-use by other veneer variants, however now that I understand that variants can wildly vary on composition, it didn't make any more sense to Ingest() ... Transform() ... Render() any more.

Now we jump straight to `Render`. :)

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,81 @@
...
+	file, err := root.Open(fname)
+	if err != nil {
+		return nil, err
+	}
+
+	defer file.Close()
+
+	if cfg, err = declcfg.LoadFile(root, file, ref); err != nil {
+		return cfg, err
+	}
```
joelanford, line 52: Nit:
```suggestion
	cfg, err := declcfg.LoadFile(root, file, ref)
	if err != nil {
		return cfg, err
	}
```

And drop the `var cfg = ...` declaration on line 32.

alpha/declcfg/load.go
```diff
@@ -168,3 +168,16 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	}
 	return cfg, nil
 }
+
+func LoadFile(root fs.FS, file io.Reader, path string) (*DeclarativeConfig, error) {
```
joelanford, line 172: Add a GoDoc comment, which should take the form: `<name of symbol> <description of symbol>`, so:

```go 
// LoadFile <... what LoadFile does, what are its inputs/outputs (if not obvious from the types), any interesting behaviors or side effects, etc. ...> 
```

alpha/declcfg/load.go
joelanford: Seems like this suggestion and fix may have been lost?

alpha/declcfg/load.go
```diff
@@ -168,3 +168,16 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
...
+	cfg, err := readYAMLOrJSON(file)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = readBundleObjects(cfg.Bundles, root, path); err != nil {
+		return cfg, err
```
joelanford, line 179: Return nil here instead of a partial cfg?
```suggestion
		return nil, err
```

cmd/opm/alpha/veneer/render.go
```diff
@@ -0,0 +1,83 @@
+package veneer
```
joelanford, line 1: rename this file to "basic.go"?

cmd/opm/alpha/veneer/render.go
```diff
@@ -0,0 +1,83 @@
...
+	containerd "github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
+)
+
+func newBasicVeneerRenderCmd() *cobra.Command {
+	var (
+		veneer basic.Veneer
+		output string
+	)
+	cmd := &cobra.Command{
+		Use:   "render veneer-file",
```
joelanford, line 24: Given my earlier comment about supporting multiple veneers, should we have the `opm alpha veneer` subcommand names be the names of the veneers?

That way, when we add the `semver` veneer, we'd end up with `opm alpha veneer basic` and `opm alpha veneer semver`.

```suggestion
		Use:   "basic <veneer-file>",
```

cmd/opm/alpha/veneer/render.go
```diff
@@ -0,0 +1,83 @@
...
+				write = declcfg.WriteYAML
+			case "json":
+				write = declcfg.WriteJSON
+			default:
+				log.Fatalf("invalid --output value %q, expected (json|yaml)", output)
+			}
+
+			// The bundle loading impl is somewhat verbose, even on the happy path,
+			// so discard all logrus default logger logs. Any important failures will be
+			// returned from veneer.Ingest and logged as fatal errors.
```
joelanford, line 42:
```suggestion
			// returned from veneer.Render and logged as fatal errors.
```

grokspawn: Dangit.  I knew I had missed one, but couldn't find it. Thanks.

alpha/declcfg/load.go
```diff
@@ -168,3 +168,18 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
...
+// LoadFile will unmarshall declarative config components from a single 'file' located
+// at a filesystem hierarchy 'root' represented in string form as 'path'
+func LoadFile(root fs.FS, file io.Reader, path string) (*DeclarativeConfig, error) {
+	cfg, err := readYAMLOrJSON(file)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := readBundleObjects(cfg.Bundles, root, path); err != nil {
+		return nil, err
```
joelanford, line 181: To keep the same behavior as before, should we keep the same error context?
```suggestion
		return nil, fmt.Errorf("read bundle objects: %v", err)
```

alpha/declcfg/load.go
```diff
@@ -168,3 +168,18 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	}
 	return cfg, nil
 }
+
+// LoadFile will unmarshall declarative config components from a single 'file' located
+// at a filesystem hierarchy 'root' represented in string form as 'path'
+func LoadFile(root fs.FS, file io.Reader, path string) (*DeclarativeConfig, error) {
```
joelanford, line 174: Looking at this a bit closer, I'd suggest this function signature:
```suggestion
func LoadFile(root fs.FS, path string) (*DeclarativeConfig, error) {
```
where path is the path of the DC file we want to parse, relative to the root. We can do the `root.Open(path)` step inside `LoadFile`, I think. That would make it impossible for a caller is accidentally mess up sending in an io.Reader that isn''t loaded from the file at `path`.

joelanford (commented):
Hopefully my last few comments. Great job with this!

alpha/declcfg/load.go
grokspawn: Yes, writing the godoc comment made this redundancy super clear.  I had planned to touch again in a later PR, but why not now? :)

alpha/declcfg/load.go
```diff
@@ -168,3 +163,24 @@ func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
 	}
 	return cfg, nil
 }
+
+// LoadFile will unmarshall declarative config components from a single filename provided in 'path'
+// located at a filesystem hierarchy 'root'
+func LoadFile(root fs.FS, path string) (*DeclarativeConfig, error) {
+	file, err := root.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("read bundle objects: %v", err)
```
joelanford, line 172:
```suggestion
		return nil, err
```

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (v Veneer) Render(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
+	// only taking first argument as file
+	stat, serr := os.Stat(ref)
+	if serr != nil {
+		return nil, serr
+	}
+
+	if stat.IsDir() {
+		return nil, errors.New("cannot render veneers by directory reference") // BUGBUG enhance to walk and eat all files in turn, heeding ignore file
```
tylerslaton, line 26: Should this comment stick around?

cmd/opm/alpha/veneer/basic.go
```diff
@@ -0,0 +1,83 @@
...
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			if err := write(*cfg, os.Stdout); err != nil {
+				log.Fatal(err)
+			}
+		},
+	}
+	cmd.Flags().StringVarP(&output, "output", "o", "json", "Output format (json|yaml)")
```
tylerslaton, line 79: nit: Generally our repos tend to place the flags closer to the top of the file.

Example:
https://github.com/operator-framework/operator-registry/blob/64c2f24da4f01f4a6eb1f3a87aa0bf7676e241e9/cmd/opm/index/add.go#L59

tylerslaton (commented):
This implementation is looking great. Just had a few comments but I'm going to go ahead and LGTM this.

/lgtm

cmd/opm/alpha/veneer/basic.go
joelanford: This is actually the same thing as what `opm/index/add.go` does. We're adding the flags to `cmd` as soon as we have the `cmd` variable. The only real difference between these two is that the `Run` field of the basic veneer is inlined.

I personally prefer inlined Run functions when they're relatively short, so I don't see this as a problem. My two cents 🤷‍♂️

tylerslaton: Ah, yeah I see what you're saying. I got confused looking at the two as one is inlined. I think this is totally fine with that information, you're right.

alpha/declcfg/load.go
grokspawn: Whoops!

alpha/veneer/basic/basic.go
grokspawn: Nope.  Fire-sale time.  
I had thought that Joe would see this comment, go "yes, do that", and away we would go.  
Instead, he said "do you understand the complexity you're enabling by opening up the possibility of collisions .... for an alpha demo project???"  
I took out the rest of the functionality, but left in the comment.

cmd/opm/alpha/veneer/basic.go
grokspawn: I'm going to push back on this one, because it seems like there's ample evidence that the preference is mutable (over time, folks, etc) and there's a bunch of prior art for this idiom.  This one follows examples of `opm/index/add.go` and `opm/alpha/render/cmd.go`.

alpha/veneer/basic/basic.go
```diff
@@ -0,0 +1,80 @@
...
+	// xform to break apart dir/file elements
+	rpath, fname := filepath.Split(abspath)
+
+	root := os.DirFS(rpath)
+	file, err := root.Open(fname)
+	if err != nil {
+		return nil, err
+	}
+
+	defer file.Close()
```
joelanford, line 46: Remove this code now?

grokspawn: Yes.  \<\/sheepish\>

tylerslaton:
Final LGTM, looks great @grokspawn.

/lgtm

PR #943, opened 2022-04-19
feat(opm): Add semver range support for olm alpha diff

dinhxuanvu:
Currently, opm diff allows an array of versions or bundles for
diff operation. This commit will enable a semver range being specified
for package or channel (but not both) in diff configuration. The diff
operator will return bundles that match the range from channel(s).

**Reviewer Checklist**

### DIFF
--- a/alpha/action/diff.go
+++ b/alpha/action/diff.go
@@ -117,6 +117,12 @@ type DiffIncludePackage struct {
 	// are parsed for an upgrade graph.
 	// Set this field only if the named bundle has no semantic version metadata.
 	Bundles []string `json:"bundles,omitempty" yaml:"bundles,omitempty"`
+	// Semver range of versions to include. All channels containing these versions
+	// are parsed for an upgrade graph. If the channels don't contain these versions,
+	// they will be ignored. This range can only be used with package exclusively
+	// and cannot combined with `Range` in `DiffIncludeChannel`.
+	// Range setting is mutually exclusive with channel versions/bundles/range settings.
+	Range string `json:"range,omitempty" yaml:"range,omitempty"`
 }
 
 // DiffIncludeChannel contains a name (required) and versions (optional)
@@ -129,6 +135,11 @@ type DiffIncludeChannel struct {
 	// Bundles are bundle names to include.
 	// Set this field only if the named bundle has no semantic version metadata.
 	Bundles []string `json:"bundles,omitempty" yaml:"bundles,omitempty"`
+	// Semver range of versions to include in the channel. If the channel don't contain
+	// these versions, an error will be raised. This range can only be used with
+	// channel exclusively and cannot combined with `Range` in `DiffIncludePackage`.
+	// Range setting is mutually exclusive with Versions and Bundles settings.
+	Range string `json:"range,omitempty" yaml:"range,omitempty"`
 }
 
 // LoadDiffIncludeConfig loads a (YAML or JSON) DiffIncludeConfig from r.
@@ -148,10 +159,32 @@ func LoadDiffIncludeConfig(r io.Reader) (c DiffIncludeConfig, err error) {
 			errs = append(errs, fmt.Errorf("package at index %v requires a name", pkgI))
 			continue
 		}
+		if pkg.Range != "" && (len(pkg.Versions) != 0 || len(pkg.Bundles) != 0) {
+			errs = append(errs, fmt.Errorf("package at index %v has an invalid settings for range/versions/bundles", pkgI))
+		}
+		if pkg.Range != "" {
+			_, err := semver.ParseRange(pkg.Range)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("package at index %v has an invalid version range %s", pkgI, pkg.Range))
+			}
+		}
 		for chI, ch := range pkg.Channels {
 			if ch.Name == "" {
 				errs = append(errs, fmt.Errorf("package %s: channel at index %v requires a name", pkg.Name, chI))
 				continue
+			}
+			if ch.Range == "" {
+				continue
+			}
+			if ch.Range != "" && (len(ch.Versions) != 0 || len(ch.Bundles) != 0) {
+				errs = append(errs, fmt.Errorf("package %s: channel at index %v has an invalid settings for range/versions/bundles", pkg.Name, chI))
+			}
+			if pkg.Range != "" && ch.Range != "" {
+				errs = append(errs, fmt.Errorf("version range settings in package (%s) and in channel (%s) must be mutually exclusive", pkg.Name, ch.Name))
+			}
+			_, err := semver.ParseRange(ch.Range)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("package %s: channel at index %v has an invalid version range %s", pkg.Name, chI, pkg.Range))
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


dinhxuanvu:
@jpower432 FYI

jpower432:
@dinhxuanvu This looks good to me. Verified it also works as expected vendored into `oc-mirror`.

joelanford (commented):
It'll take me a little longer to do a full review, but one question I came up with is:

If a channels entries are not semver ordered (e.g.  v1.0.0 -> v3.0.0 -> v4.0.0 -> v2.0.0), and I specify my range as `<=v2.0.0`, will it pick up v3.0.0 and v4.0.0 since they sit between versions that _are_ in the range? If so, is that expected behavior?

Should we have a test case for that scenario?

dinhxuanvu:
> It'll take me a little longer to do a full review, but one question I came up with is:
> 
> If a channels entries are not semver ordered (e.g. v1.0.0 -> v3.0.0 -> v4.0.0 -> v2.0.0), and I specify my range as `<=v2.0.0`, will it pick up v3.0.0 and v4.0.0 since they sit between versions that _are_ in the range? If so, is that expected behavior?
> 
> Should we have a test case for that scenario?

This is strictly for semver-based operator which should be the standard order for operators and the expectation here is that they choose (knowingly know these versions will suit their needs) to use semver range to specifically select versions that they want to for a specific package/channel. This is not a requirement. If they do not wish to use semver range due to non-semver order (which frankly oc-mirror wouldn't care), they can use other settings such as `bundles` and `versions` which is more discrete. My original intention was to get rid of any other settings and use semver range only because those other settings are not helpful. However, just as a caution, I ended up keeping everything as it and add semver range additively so users may have other choices.

alpha/action/diff.go
```diff
@@ -148,11 +159,33 @@ func LoadDiffIncludeConfig(r io.Reader) (c DiffIncludeConfig, err error) {
 			errs = append(errs, fmt.Errorf("package at index %v requires a name", pkgI))
 			continue
 		}
+		if pkg.Range != "" && (len(pkg.Versions) != 0 || len(pkg.Bundles) != 0) {
+			errs = append(errs, fmt.Errorf("package at index %v has an invalid settings for range/versions/bundles", pkgI))
```
perdasilva, line 163: Could we suggest a correction here? Or point to docs or something? Something like "Please only use Range OR Versions and Bundles, or some such...

Or maybe: `package at index %v contains range %s, %d versions and %d bundles. range and versions and/or bundles are mutually exclusive. Please only specify one (or please see docs at ...)`. 

It would just be nice to know the remedy.

alpha/action/diff.go
```diff
@@ -148,11 +159,33 @@ func LoadDiffIncludeConfig(r io.Reader) (c DiffIncludeConfig, err error) {
...
 				continue
 			}
+			if ch.Range == "" {
+				continue
+			}
+			if ch.Range != "" && (len(ch.Versions) != 0 || len(ch.Bundles) != 0) {
+				errs = append(errs, fmt.Errorf("package %s: channel at index %v has an invalid settings for range/versions/bundles", pkg.Name, chI))
+			}
+			if pkg.Range != "" && ch.Range != "" {
+				errs = append(errs, fmt.Errorf("version range settings in package (%s) and in channel (%s) must be mutually exclusive", pkg.Name, ch.Name))
```
perdasilva, line 183: Maybe add the ranges as well to the error message?

alpha/action/diff_test.go
```diff
@@ -311,6 +311,77 @@ packages:
...
+			name: "Fail/InvalidChannelRange",
+			input: `
+			{
+			  "packages": [
+			    {
+			      "name": "foo",
+			      "channels": [
+			        {
+			          "name": "stable",
+								"range": "test"
```
perdasilva, line 337: nit: formatting

alpha/action/diff_test.go
```diff
@@ -311,6 +311,77 @@ packages:
...
+			input: `
+			{
+			  "packages": [
+			    {
+			      "name": "foo",
+						"range": "test",
+			      "channels": [
+			        {
+			          "name": "stable",
+								"range": "test"
```
perdasilva, line 356: nit: format

alpha/action/diff_test.go
```diff
@@ -311,6 +311,77 @@ packages:
...
+		{
+			name: "Fail/InvalidRangeSetting/MixedRange&OtherVersions",
+			input: `
+			{
+				"packages": [
+			    {
+			      "name": "foo",
+			      "channels": [
+			        {
+								"name": "stable",
```
perdasilva, line 373: nit: format

alpha/action/diff_test.go
```diff
@@ -311,6 +311,77 @@ packages:
 `,
 			assertion: require.Error,
 		},
+		{
```
perdasilva, line 314: Maybe it's worth adding a happy path test too?

alpha/declcfg/diff_include.go
```diff
@@ -59,7 +134,7 @@ func (i DiffIncluder) Run(newModel, outputModel model.Model) error {
 	return nil
 }
 
-// includeNewInOutputModel adds all packages, channels, and versions (bundles)
+// includeNewInOutputModel adds all packages, channels, and range (or versions/bundles)
```
perdasilva, line 137: Nice! Thanks for remembering to update the description ^^

alpha/action/diff.go
dinhxuanvu: I did add the comment on this exclusivity on the struct (see above) but I will improve the message on the error a bit for clarity.

alpha/action/diff.go
```diff
@@ -148,11 +159,33 @@ func LoadDiffIncludeConfig(r io.Reader) (c DiffIncludeConfig, err error) {
...
 		for chI, ch := range pkg.Channels {
 			if ch.Name == "" {
 				errs = append(errs, fmt.Errorf("package %s: channel at index %v requires a name", pkg.Name, chI))
 				continue
 			}
+			if ch.Range == "" {
+				continue
+			}
+			if ch.Range != "" && (len(ch.Versions) != 0 || len(ch.Bundles) != 0) {
+				errs = append(errs, fmt.Errorf("package %s: channel at index %v has an invalid settings for range/versions/bundles", pkg.Name, chI))
```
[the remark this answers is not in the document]
dinhxuanvu: Will improve the error message

alpha/action/diff.go
dinhxuanvu: Roger.

alpha/action/diff_test.go
dinhxuanvu: Unfortunately, `semver.Range` is a func so asserting equality of two ranges as objects in this test is just not possible. I prefer not to change this test just for that. Instead, on the other `diff_test`, the test cases will validate that correct range will behave as expected.

dinhxuanvu:
@perdasilva Ready for final lgtm.
I would like to emphasis that this whole diff feature will eventually move over to oc-mirror given it doesn't provide any value to registry nor it is relevant to registry functionality.

alpha/declcfg/diff_include.go
```diff
@@ -72,26 +147,44 @@ func (ipkg DiffIncludePackage) includeNewInOutputModel(newModel, outputModel mod
 	pkgLog := logger.WithField("package", newPkg.Name)
 
 	// No channels or versions were specified, meaning "include the full package".
```
perdasilva, line 149: nit: stale comment...?

alpha/declcfg/diff_include.go
```diff
@@ -72,26 +147,44 @@ func (ipkg DiffIncludePackage) includeNewInOutputModel(newModel, outputModel mod
...
-	if len(ipkg.Channels) == 0 && len(ipkg.AllChannels.Versions) == 0 && len(ipkg.AllChannels.Bundles) == 0 {
+	if len(ipkg.Channels) == 0 && len(ipkg.AllChannels.Versions) == 0 && len(ipkg.AllChannels.Bundles) == 0 && ipkg.Range == nil {
```
perdasilva, line 150: nit: stale comment...?

alpha/declcfg/diff_include.go
perdasilva, line 149: nit: stale comment...?

PR #946, opened 2022-04-22
Smooth phrasing in under-the-hood graph update passage

grokspawn:
Attempt to make verbiage less complex by:
- removing commentary
- eliminating passive voice
- adding emphasis to distinguish between `new` and `already-present` CSV cases during graph rebuild
- eliminated paragraph references
- add whitespace to delimit complete thoughts and focus the eye on the fact that they were discrete

As a first-touch user reading the docs, I found it difficult to follow the graph update section and thought it could be simplified while retaining intent

**Reviewer Checklist**

### DIFF
--- a/docs/design/opm-tooling.md
+++ b/docs/design/opm-tooling.md
@@ -116,23 +116,23 @@ Running this command will still generate the updated registry database, but it w
 
 In an effort to make channel head selection understandable and deterministic when bulk-adding bundles to an index using `--mode=replaces` (the default), the following heuristic has been adopted: the bundles with the highest version within a package are considered the heads of the channels they belong to.
 
-<ul>
-
 #### Under the Hood
 
-`opm` effectively decomposes bundle addition into three steps for each package:
+`opm` effectively decomposes bundle addition into the following steps for each package:
 
-1. Add bundles to the underlying data store
-2. Choose the channel heads and default channel
-3. Rebuild the update graph starting at the new heads
-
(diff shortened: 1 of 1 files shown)


docs/design/opm-tooling.md
```diff
@@ -124,15 +124,17 @@ In an effort to make channel head selection understandable and deterministic whe
 
 1. Add bundles to the underlying data store
 2. Choose the channel heads and default channel
-3. Rebuild the update graph starting at the new heads
-
-Channel head -- the "latest" operator in a channel -- selection is now informed by [semver](https://semver.org/). The heurstic is simple, the bundle with the highest version in each channel becomes the new head. The default channel is then taken from the maximum versioned bundle which defines a default channel.
+3. Rebuild the update graph starting at the new heads  
+<br />    
+Channel head -- the "latest" operator in a channel -- selection is now informed by [semver](https://semver.org/). The heurstic is that the bundle with the highest version in each channel becomes the new head. The default channel is then taken from the maximum versioned bundle which defines a default channel.
```
timflannagan, line 129: It looks like this breaks the semver inline link when viewing this in the markdown preview tab.

timflannagan: I _think_ adding inline html may also cause IDEs that have marker linters to complain too: https://github.com/updownpress/markdown-lint/blob/master/rules/033-no-inline-html.md.

PR #948, opened 2022-04-25
implement `opm alpha render-veneer semver`

grokspawn:
**NOTE: Updated comments/docs from yesterday's peer review.**

Implements a "SemVer" `Veneer` rendering capability, to illustrate the potential benefits of the use of `veneer` as entry points into the FBC ecosystem.  

 Since a `veneer` is identified as an input schema which may be processed to generate a valid FBC, we can define a `semver veneer` as a schema which uses channel conventions to facilitate the auto-generation of channels along `semver` delimiters.

Please see the [README.md](https://github.com/operator-framework/operator-registry/pull/948/files?short_path=d76b7e6#diff-d76b7e65c589a18f29fb6ce199a3d3bd4da743423fe41de8476052cb10bfe4e2) for complete description.
  
Illustrate a possible effort-saving work flow for operator authors to use to facilitate migration to `FBC`.

**Reviewer Checklist**


joelanford:
> channels need to include a replace value from the head of the previously-versioned channel

this is true

>  (e.g. stable-v1 should replace stable-v0 -- if defined)

but this isn't. we shouldn't include that replaces value if we're crossing over a major version number transition.

---

So only minor version channels will have these channel cross-over replaces edges.

joelanford:
> should we error out if the necessary channel hasn't been generated because user did not supply any v0 bundles?

No, there should only be edges that connect to actual bundles listed in the veneer. If the veneer contains some v1.1 bundles, but no v1.0 bundles, then the leaf nodes of that v1.1 channels will all be skips and there will be no replace to the a previous v1.0 bundle (because there isn't one).

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,335 @@
...
+
+// isvd.Channels["stable"] --> quay.io/foo/foo-bundle[0.1.0]
+type ChannelsDataMap map[string]BundlesDataMap  // channel-name --> BundlesDataMap
+type BundlesDataMap map[string][]semver.Version // bundle-name --> []bundle-versions
+// cname --> bname --> [v0, v1, v2, ...]
+
+type _tokenized_bundle_entry struct {
+	_path    string
+	_version semver.Version
+}
```
joelanford, line 66: AFAIK, the `_` prefix is not idiomatic, so I'd drop the `_`

identifiers that begin with lowercase letter are unexported (and can therefore only be used in the same package).

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,335 @@
...
+
+const (
+	CandidateChannelName = "candidate"
+	FastChannelName      = "fast"
+	StableChannelName    = "stable"
+)
+
+// isvd.Channels["stable"] --> quay.io/foo/foo-bundle[0.1.0]
+type ChannelsDataMap map[string]BundlesDataMap  // channel-name --> BundlesDataMap
+type BundlesDataMap map[string][]semver.Version // bundle-name --> []bundle-versions
```
joelanford, line 60: Do we use these types outside this package? If not, let's unexport them. Ditto for anything else that isn't directly used in `cmd/op/alpha/veneer`

cmd/opm/alpha/veneer/semver.go
```diff
@@ -0,0 +1,115 @@
...
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	// "github.com/blang/semver/v4"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	veneer_semver "github.com/operator-framework/operator-registry/alpha/veneer/semver"
```
joelanford, line 14: Idiomatic import strings are all lowercase concatenated words (so no underscores) and generally only with import aliases if you're importing two different packages with the same name.

```suggestion
	"github.com/operator-framework/operator-registry/alpha/veneer/semver"
```

cmd/opm/alpha/veneer/semver.go
```diff
@@ -0,0 +1,115 @@
...
+			if err != nil {
+				log.Fatalf("creating containerd registry: %v", err)
+			}
+			defer reg.Destroy()
+
+			veneer := veneer_semver.Veneer{
+				Ref:           ref,
+				SkipPatch:     skipPatch,
+				ChannelsMajor: majorChannels,
+				ChannelsMinor: minorChannels,
```
joelanford, line 88: My opinion: we should make major channels optional, but minor channels always on. Production users will want to have assurances they can be on a channel that receives only back-compat patches.

alpha/veneer/semver/semver.go
grokspawn: The original plan was to include this in declcfg types, so I totally would have been crossing packages, but ... there's really no need to do that for this tool.

cmd/opm/alpha/veneer/semver.go
grokspawn: I agree, but the only thing which prevented me from taking this position is trying to anticipate the UX where only the major channels are desired, and the fact that it would make the flags conflict (e.g. --no-minor-channels, --major-channels). Instead of many variations on the "no on one, yes on the other" I felt it was most consistent if both were explicitly requested by the user.

cmd/opm/alpha/veneer/semver.go
grokspawn: Yeah, looks like I was trying to resolve a situation that I thought was path resolution, when actually it was "dummy used the wrong scope, and didn't notice when he corrected that mistake while aliasing the imports". 
#OpportunitiesForHumility

cmd/opm/alpha/veneer/semver.go
grokspawn: For the moment (in my local), I've converted these flags to 'omitempty' fields in the veneer itself, and provided a default of (major: false | minor: true | skips: true) for the values.  
Any deviation from default would then be specifically overridden by user-provided values in the veneer, and transparent since it's in the veneer.
How does that sound?

cmd/opm/alpha/veneer/semver.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	cmd.Flags().StringVarP(&output, "output", "o", "json", "Output format (json|yaml)")
+	cmd.Flags().BoolVar(&skipPatch, "skip-patch", false, "Add skips for intermediate semver patch versions")
+	// cmd.Flags().StringSliceVarP(&templates, "templates", "t", []string{"default"}, "Template strings evaluated against semver versions to generate channel names")
+	// cmd.Flags().StringVarP(&semverRangeStr, "semver-range", "r", "", "Semver range of bundles to consider when building channels")
+	cmd.Flags().BoolVarP(&majorChannels, "major-channels", "M", false, "generate channels including edges with the same major version")
+	cmd.Flags().BoolVarP(&minorChannels, "minor-channels", "m", false, "generate channels including edges with the same minor version")
+	cmd.Flags().Bool("skip-tls-verify", false, "disable TLS verification")
+	cmd.Flags().Bool("use-http", false, "use plain HTTP")
```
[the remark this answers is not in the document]
grokspawn: Concur.  Making it so.

cmd/opm/alpha/veneer/semver.go
joelanford: Sounds perfect for the "let's get feedback phase" for sure. 👍

cdjohnson:
It seems like we'll want sort of categorization of veneers by channel taxonomy models.

I see at least three models, of which you can combine them:
- Maturity (dev/alpha/beta/stable)
- Stability (fast/candidate/stable)
- Version (v1, v1.0, v1.1, v2...)

This particular one is a Version + Stability model generator.  

That said, I think a more strategic approach (probably for a different PR) would be to store the model parameters within the input veneer file rather than separate veneers or CLI parameters.  Here is some research I did with these conclusions for our (100+) operator standards.

Rough example for illustration (not much thought into this):

```yaml 
schema: olm.semver.model
name: stability
type: stability # This is a model type coded into the veneer
categories:
- name: fast
- name: candidate
- name: stable
---
schema: olm.semver.model
name: maturity
type: maturity
categories:
- name: dev
- name: alpha
- name: beta
- name: stable
---
schema: olm.semver.model
name: versionminor
type: version
pattern: v{MAJOR}.{MINOR}
---
schema: olm.semver.model
name: versionmajor
type: version
pattern v{MAJOR}
---
schema: olm.semver.model
name: versionbystability
type: combo
patterns: 
- {maturity}-{versionmajorminor}
- {maturity}-{versionmajor}
---

File: alpha/veneer/semver/semver.go [hunk 6]
```diff
@@ -0,0 +1,386 @@
...
+		} else {
+			maxIndex := len(c.Entries) - 1
+			curSkips := sets.NewString()
+			for i := 0; i < maxIndex; i++ {
+				curSkips = curSkips.Insert(c.Entries[i].Name)
+			}
+			c.Entries[maxIndex] = declcfg.ChannelEntry{
+				Name:  c.Entries[maxIndex].Name,
+				Skips: curSkips.List(),
+			}
```
joelanford, line 247: This appears to add all non-head entries in the channel as skips of the head. This is undesirable for major version channels because it means we could jump over several minor versions at a time.

For major channels we need:
1. A `replaces` chain connecting the highest patch version of each of the minor versions.
2. Each of the members of the replaces chain from 1 should skip other members of their minor version.

So we need:
```mermaid
graph LR
   1.2.2 -- replaces --> 1.1.3
   1.2.2 -- skips --> 1.2.1
   1.2.2 -- skips --> 1.2.0

   1.1.3 -- replaces --> 1.0.1
   1.1.3 -- skips --> 1.1.2
   1.1.3 -- skips --> 1.1.1
   1.1.3 -- skips --> 1.1.0

   1.0.1 -- skips --> 1.0.0



```

joelanford: I _think_ this code hooked the edges up correctly when I was experimenting with something like this: https://github.com/joelanford/declcfg/blob/d1ebdcf2717649e9b867710a8166b738fa63966e/internal/action/semver.go#L67-L90

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,386 @@
...
+	return &decomposedBundleEntry{
+		img: b.Name,
+		ver: v,
+	}, nil
+}
+
+func addBundlesToChannel(bundles []semverVeneerBundleEntry, cfg *declcfg.DeclarativeConfig) (*bundlesDataMap, error) {
+	bdm := make(bundlesDataMap)
+	for _, b := range bundles {
+		fmt.Printf("  <--> adding bundle %q\n", b.Image)
```
joelanford, line 123: This causes the stdout to be unparsable as FBC. Remove it (or move to stderr)?

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,386 @@
...
+
+func (v Veneer) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
+	var out declcfg.DeclarativeConfig
+	// fmt.Printf("<--> Received config: skip(%t) major(%t) minor(%t) ref(%s)\n", v.SkipPatch, v.ChannelsMajor, v.ChannelsMinor, v.Ref)
+
+	sv, err := ReadFile(v.Ref)
+	if err != nil {
+		log.Fatalf("semver-render: unable to read file: %v", err)
+	}
+	fmt.Printf("Semver-Veneer parsed\n")
```
joelanford, line 184: Ditto (remove to avoid invalid FBC)

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,386 @@
...
+			return nil, err
+		}
+		cfgs = append(cfgs, *c)
+	}
+	out = *combineConfigs(cfgs)
+
+	fmt.Printf("all bundles parsed\n")
+	for _, b := range out.Bundles {
+		fmt.Printf("{ bundle Name:%q Image:%q }\n", b.Name, b.Image)
+	}
```
joelanford, line 204: Ditto (remove to avoid invalid FBC)

alpha/veneer/semver/semver.go
grokspawn: Updated with this, plus `-o mermaid` mermaid output support so we can visually debug (plus some traditional unit tests ;) ) :

```mermaid
graph LR
testoperator.v0.1.0
testoperator.v0.1.1
testoperator.v0.1.2
testoperator.v0.1.3
testoperator.v0.1.3-- skips --> testoperator.v0.1.0
testoperator.v0.1.3-- skips --> testoperator.v0.1.1
testoperator.v0.1.3-- skips --> testoperator.v0.1.2
testoperator.v0.2.0
testoperator.v0.2.1
testoperator.v0.2.2
testoperator.v0.2.2-- replaces --> testoperator.v0.1.3
testoperator.v0.2.2-- skips --> testoperator.v0.2.0
testoperator.v0.2.2-- skips --> testoperator.v0.2.1
testoperator.v0.3.0
testoperator.v0.3.0-- replaces --> testoperator.v0.2.2
```
This is channel `candidate-v0`

ankitathomas (approved):
/lgtm

grokspawn:
> A few questions before I start reviewing the code:
> 
>     1. I know this is likely a artifact of `render` being out of the `alpha` subcommand now, but when this particular feature graduates, I wonder if we can improve the UX a bit.
>        `opm render-veneer semver <filename>` -> `opm render --veneer semver <filename>`
>        This would feel pretty intuitive for me personally. That being said, the impact of such a change is minimal. Let me know if there are some concerns I'm not considering.

Yeah, the separation largely stems from the fact that the new functionality is alpha and we didn't want to cross the streams.  On promotion, I think we want to more tightly integrate things. 

>     2. Probably fine for a follow-up and _definitely_ not blocking but I notice in the demos that there is a long pause between invocation and render generation. Could we render some form of loading bar/icon to let a user know that we're working on completing the request?

On a fast machine with fast network connection, we're seeing ~30s between query submission and output results.  If we agree on the functionality, I think the next steps include seeing how we can optimize that.   If we can't make it any better, then we can certainly be more communicative in a later iteration.  No one wants the "tap, tap ... is this thing on?" experience. :)

anik120 (changes requested):
~Did we decide that we're okay with these functions implementing the veneer are in shell scripts? Why not implement them in Go, instead of bringing in shell script dependency to the tool?~

Just realized that the shell scripts are inside `semver/demo` 😄 jumped the gun early on this one without looking more at the code, my apologies. 

Although, we might want to discuss what we want to decide as acceptable for including in the official repos for demos. That's definitely lot more code to review, and make sure is secured enough to be checked into the repo + a lot more content being checked into git. If my demo needs manifests, do I include that in the repo? That'll blow up the repo soon if a considerable number of demos include files like that.

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,428 @@
...
+}
+
+type semverVeneer struct {
+	Schema                string           `json:"schema"`
+	GenerateMajorChannels bool             `json:"generateMajorChannels,omitempty"`
+	GenerateMinorChannels bool             `json:"generateMinorChannels,omitempty"`
+	AvoidSkipPatch        bool             `json:"avoidSkipPatch,omitempty"`
+	Candidate             candidateBundles `json:"candidate,omitempty"`
+	Fast                  fastBundles      `json:"fast,omitempty"`
+	Stable                stableBundles    `json:"stable"`
```
grokspawn, line 56: remove this req.  Instead, enforce that we function as long as at least one channel has a single bundle in it.

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	versions[StableChannelName] = bdm
+
+	return &versions, nil
+}
+
+func (sv *semverVeneer) getVersionsFromChannel(semverBundles []semverVeneerBundleEntry, cfg *declcfg.DeclarativeConfig) (map[string]semver.Version, error) {
+	entries := make(map[string]semver.Version)
+
+	for _, semverBundle := range semverBundles {
```
joelanford, line 165: Do we even need `semverBundles` here? I wonder if we can just iterate `cfg.Bundles` to build the map of `cfg.Bundles[i].Image` to `cfg.Bundles[i].Properties[?(type == olm.package)].Version`?

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,428 @@
...
+			sv.pkg = props.Packages[0].PackageName
+		}
+
+		entries[b.Name] = v
+	}
+
+	return entries, nil
+}
+
+type highwaterChannel struct {
```
grokspawn, line 215: add comment/description:
highwater-mark captures here the concept of the most-stable channel-head.  Like a real hwm, it will 'float up' to the greatest of the core comparison.

alpha/veneer/semver/README.md
```diff
@@ -0,0 +1,222 @@
...
+  ***Schema Anatomy***  
+    This veneer type currently supports channel names `candidate`, `fast`, and `stable`. The `stable` channel is required.
+    `generatemajorchannels` and `generateminorchannels` dictate whether this veneer will generate X-stream- or Y-stream-based channels. (Both attributes can be set simultaneously.)
+    Under each channel are a list of bundle image references which contribute to that channel.  
+
+With the (functional) example:
+```yaml
+schema: olm.semver
+generatemajorchannels: true
+generateminorchannels: true
```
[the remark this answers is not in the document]
joelanford: TIL that Go's json decoder is case insensitive wrt field names.

I'm +1 on camel casing these in our docs though. That seems to be the standard in YAML (k8s-related YAML at least), and it would align with the JSON tags on the struct.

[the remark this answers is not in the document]
grokspawn: Done.

alpha/veneer/semver/semver_test.go
```diff
@@ -0,0 +1,548 @@
...
+			Name:    "stable-v3.1",
+			Package: "a",
+			Entries: []declcfg.ChannelEntry{
+				{Name: "a-v3.1.0"},
+				{Name: "a-v3.1.1"},
+			},
+		},
+	}
+
+	tests := []struct {
```
awgreene, line 136: It'd be nice to add a testcase for [prerelease versions](https://semver.org/#spec-item-9) and [build metadata tags](https://semver.org/#spec-item-10).

awgreene (changes requested):
This looks good to me, no major functional changes requested. Great work!

alpha/veneer/semver/README.md
```diff
@@ -0,0 +1,238 @@
...
+- simple channel promotion for maturing bundles
+- demonstration of a common type of channel maturity model
+- Y-stream, X-stream versioning controls
+
+The resulting FBC must clearly indicate how generated channels relate to veneer entities
+
+### Schema Anatomy
+For convenience and simplicity, this veneer currently supports hard-coded channel names `Candidate`, `Fast`, and `Stable`, in order of increasing channel stability.  We leverage this relationship to calculate the default channel for the package. 
+
+`GenerateMajorChannels` and `GenerateMinorChannels` dictate whether this veneer will generate X-stream- or Y-stream-based channels (attributes can be set independently).  If omitted, only minor (Y-stream) channels will be generated.
```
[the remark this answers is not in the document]
grokspawn: https://en.wikipedia.org/wiki/Hyphen#Suspended_hyphens

upstream-opm-builder.Dockerfile
```diff
@@ -3,9 +3,13 @@
 ##             GoReleaser to build and push multi-arch images for opm
 ##
 
-FROM quay.io/operator-framework/golang:1.18-alpine AS builder
+FROM golang:1.18-alpine AS builder
 
+RUN apk update && apk add ca-certificates
+COPY ["nsswitch.conf", "/etc/nsswitch.conf"]
 RUN apk update && apk add sqlite build-base git mercurial bash
+RUN set -eux;     apk add --no-cache --virtual .build-deps         bash         gcc
```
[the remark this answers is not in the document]
joelanford: Also, pretty sure these dockerfiles are mostly defunct - at least upstream.

The only one used upstream now is: https://github.com/operator-framework/operator-registry/blob/master/release/goreleaser.opm.Dockerfile

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,423 @@
...
+// map of channels : bundles : bundle-version
+// channels --> bundles --> version
+type semverRenderedChannelVersions map[string]map[string]semver.Version // e.g. d["stable-v1"]["example-operator/v1.0.0"] = 1.0.0
+
+func (v Veneer) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
+	var out declcfg.DeclarativeConfig
+
+	sv, err := ReadFile(v.Ref)
+	if err != nil {
+		log.Fatalf("semver-render: unable to read file: %v", err)
```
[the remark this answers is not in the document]
joelanford: Oh, also, we shouldn't do any `Fatal` or `os.Exit` stuff in the library side of this code. We should always return an error and do the `Fatal` and `os.Exit` stuff only in the cobra code.

alpha/veneer/semver/README.md
```diff
@@ -0,0 +1,238 @@
...
+  - Image: quay.io/foo/olm:testoperator.v0.2.1
+  - Image: quay.io/foo/olm:testoperator.v0.2.2
+  - Image: quay.io/foo/olm:testoperator.v0.3.0
+  - Image: quay.io/foo/olm:testoperator.v1.0.1
+  - Image: quay.io/foo/olm:testoperator.v1.1.0
+Stable:
+  Bundles:
+  - Image: quay.io/foo/olm:testoperator.v1.0.1
+```
+we define a mock bundle which has 11 versions, represented across each of the channel types.  In this example, `Candidate` has the entire version range of bundles,  `Fast` has a mix of older and more-recent versions, and `Stable` channel only has a single published entry.
```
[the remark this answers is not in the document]
grokspawn: This is a continuation of the previous clause "With the (hypothetical) example:" not a new thought.  I was trying to avoid using the word 'following' since it's a pet peeve, but I need to keep the full sentence together to help it be clearer.

alpha/veneer/semver/README.md
```diff
@@ -0,0 +1,238 @@
+## Description of the change:
+
+  Since a `veneer` is identified as an input schema which may be processed to generate a valid FBC, we can define a `semver veneer` as a schema which uses channel conventions to facilitate the auto-generation of channels along `semver` delimiters.
+
+### Schema Goals
+The `semver veneer` must have:
+- compact grammar to minimize creation/maintenance effort
```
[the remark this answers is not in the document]
grokspawn: "terse" 😄

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func getMinorVersion(v semver.Version) semver.Version {
+	return semver.Version{
+		Major: v.Major,
+		Minor: v.Minor,
+	}
+}
+
+func MermaidChannelWriter(cfg declcfg.DeclarativeConfig, out io.Writer) error {
```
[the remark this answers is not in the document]
joelanford: Not for this PR, but I'd prefer to see this writer moved to something usable in `opm render`. That way _any_ veneer's output could make use of it. That can be done as a follow-up IMO though.

alpha/veneer/semver/semver_test.go
joelanford: Pre-release versions are interesting from the perspective of this veneer. I wonder if we should have a check that ensures pre-release versions are only present in `candidate*` channels?

alpha/veneer/semver/README.md
```diff
@@ -0,0 +1,238 @@
...
+- deterministic output
+- simple channel promotion for maturing bundles
+- demonstration of a common type of channel maturity model
+- Y-stream, X-stream versioning controls
```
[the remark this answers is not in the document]
grokspawn: I tried really hard to always list the colloquialism and the technical together as if defining new jargon (e.g. minor stream [Y-stream]).  I missed it here, but adding it to be consistent.

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,423 @@
...
+		if sv.AvoidSkipPatch {
+			for i := 1; i < len(channel.Entries); i++ {
+				channel.Entries[i] = declcfg.ChannelEntry{
+					Name:     channel.Entries[i].Name,
+					Replaces: channel.Entries[i-1].Name,
+				}
+			}
+		} else {
+			curIndex := len(channel.Entries) - 1
+			curMinor := getMinorVersion((*semverChannels)[(*channelMapping)[channelName]][channel.Entries[curIndex].Name])
```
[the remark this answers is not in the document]
grokspawn: In back-and-forth on slack, arrived at a more communicative form of this.

alpha/veneer/semver/semver.go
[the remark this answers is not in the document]
grokspawn: In a diff PR I wanted to add that.  @ankitathomas was working on a related story during hackday to integrate this, so I was not.

alpha/veneer/semver/semver_test.go
grokspawn: Introduced prerelease versioning to a couple of the utests since the sequencing is so very important, but I really have to remind folks that the original AC here did not include docs/e2e/utest.

alpha/veneer/semver/README.md
```diff
@@ -0,0 +1,238 @@
...
+- name: testoperator.v1.0.1
+name: stable-v1
+package: testoperator
+schema: olm.channel
+```
+
+We generated a channel for each veneer channel entity corresponding to each of the 0.\#.\#, 1.\#.\# major version ranges with skips to the head of the highest semver in a channel.  We also generated a replaces edge to traverse across minor version transitions within each major channel.  Finally, we generated an `olm.package` object, setting as default the most-stable channel head we created.  This process will prefer `Stable` channel over `Fast`, over `Candidate` and then a higher bundle version over a lower version.   
+(Please note that the naming of the generated channels indicates the digits of significance for that channel.  For example, `fast-v1` is a decomposed channel of the `fast` type which contains only major versions of contributing bundles matching `v1`.)  
+
+For contrast, with the veneer attribute `GenerateMinorChannels: true` set and running the command again (again skipping rendered bundle image output) we get a bunch more channels:
```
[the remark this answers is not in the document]
grokspawn: The link doesn't work for me for some reason, but if neither attribute is set, then we will act as if the `GenerateMinorChannels` is set to `true`.  
If the user sets both attributes to false ... then we won't generate any output, or override their choice.

cmd/opm/alpha/veneer/semver.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			cacheDir, err := os.MkdirTemp("", "veneer-registry-")
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			reg, err := containerd.NewRegistry(
```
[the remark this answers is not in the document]
grokspawn: Good catch.  The whole purpose is to allow us to pass in the `--use-http` and `--skip-tls-verify` options for the declcfg.Render actions we use inside the semver.Render.  In the multiple iterations of this PR, this fell off and I lost the thread.

cmd/opm/alpha/veneer/semver.go
```diff
@@ -0,0 +1,101 @@
...
+			// if semverRangeStr == "" {
+			// 	semverRange = func(semver.Version) bool { return true }
+			// } else {
+			// 	semverRange, err = semver.ParseRange(semverRangeStr)
+			// 	if err != nil {
+			// 		return fmt.Errorf("invalid semver range %q", semverRangeStr)
+			// 	}
+			// }
+
+			var write func(declcfg.DeclarativeConfig, io.Writer) error
```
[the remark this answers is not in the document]
grokspawn: This is the same pattern expressed in `opm render` and `opm alpha render-veneer basic`.  It's a convenience function to allow us to react to the switch given by the user, requesting json/yaml/mermaid output formats.

alpha/veneer/semver/README.md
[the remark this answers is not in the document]
awgreene: What is the logic behind setting GenerateMinorChannels to true by default if it isn't set?

cmd/opm/alpha/veneer/semver.go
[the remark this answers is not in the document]
awgreene: Gotcha, seems reasonable if its an existing pattern.

alpha/veneer/semver/semver.go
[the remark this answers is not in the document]
awgreene: Created an [issue](https://github.com/operator-framework/operator-registry/issues/973) to track the work.

alpha/veneer/semver/semver_test.go
awgreene: >Build metadata tags are intentionally not sortable according to the semver spec. What are you thinking for a test involving build metadata versions?

@joelanford This isn't entirely true, `v0.0.1` and `v0.0.1+metadata` are still sortable but they are given the same precedence.  One of the goals of the proposed schema was to be deterministic, I was hoping that to test that the order was consistent even when multiple entries with different semver build metadata were included in a channel, provided that the semver veneer schema hadn't changed. My hope that the output is consistent provided that no changes where made to the semver schema (the order of entries in the channel did not change).

awgreene: >Introduced prerelease versioning to a couple of the utests since the sequencing is so very important, but I really have to remind folks that the original AC here did not include docs/e2e/utest.

In general, we push for a lot of unit tests before passing code to other teams. However, this is an alpha feature so I'm not opposed to merging this with the set of tests you've provided, assuming you can answer the consistency question proposed in the previous comment.

alpha/veneer/semver/README.md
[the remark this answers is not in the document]
grokspawn: It's just our anticipation as the most common case, and the one that makes the most sense as a default.

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,436 @@
...
+	stableChannelName    string = "stable"
+)
+
+// mapping channel name --> stability, where higher values indicate greater stability
+var channelPriorities = map[string]int{candidateChannelName: 0, fastChannelName: 1, stableChannelName: 2}
+
+// sorting capability for a slice according to the assigned channelPriorities
+type byChannelPriority []string
+
+func (b byChannelPriority) Len() int { return len(b) }
```
tylerslaton, line 69: Do we need a custom `.Len()` reciever on this struct? Why not just use `len()` like normal when we want the length?

awgreene (approved):
/lgtm

dinhxuanvu (approved):
/approve

tylerslaton (commented):
Had some left-over comments after this got merged. Not blocking at all so we can address later if we feel the need.

/lgtm

alpha/veneer/semver/semver.go
```diff
@@ -0,0 +1,436 @@
...
+
+	pkg            string `json:"-"` // the derived package name
+	defaultChannel string `json:"-"` // detected "most stable" channel head
+}
+
+// IO structs -- END
+
+// channel "kinds", restricted in this iteration to just these
+const (
+	candidateChannelName string = "candidate"
```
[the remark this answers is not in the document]
grokspawn: The goal of this particular veneer was to focus on and encourage the candidate < fast < stable channel set as a MVP which endorsed/encourages our internal practice.  "Guidelines and guardrails" is how I like to refer to it. 

It wouldn't be hard, but it was counter to our goals at this stage.  

We can iterate into a variant or leave it as an exercise for the operator author who needs it.  Early feedback will help guide us.

alpha/veneer/semver/semver.go
grokspawn: A certain amount of "monkey see, monkey do" here, as [online](https://thedeveloperblog.com/go/sort-go) [references](https://www.callicoder.com/golang-sorting-custom-functions/) [all](https://stackoverflow.com/questions/30336616/how-to-avoid-re-implementing-sort-interface-for-similar-golang-structs) [say that](https://stackoverflow.com/questions/4947161/is-there-a-way-to-avoid-the-implementation-of-the-full-sort-interface-for-slices). 

Evidently the whole of the `data.Interface` must be implemented for any custom slice sort to be used by `sort.Sort()`.

tylerslaton: Interesting! Thanks for that, learned something new 😄

PR #950, opened 2022-04-29
feat(diff): change HeadsOnly to be an explicit setting

dinhxuanvu:
Currently the headsOnly mode is implicitly set if newRef is not provided.
This behavior is unclear and leads to confusion. This commit will change the current
headsOnly to be explicit setting that will be converted to package setting that will
be executed accordingly.

**Reviewer Checklist**


dinhxuanvu:
@jpower432 PTAL

jpower432:
Review and tested. Working as expected from an `oc-mirror` perspective.
/lgtm

Issue #952, opened 2022-05-11
error loading bundle into db: FOREIGN KEY constraint failed


camilamacedo86:
**Description**

```
Error: error loading bundle into db: FOREIGN KEY constraint failed
Usage:
  opm registry add [flags]
```

**Steps to reproduce**

```
opm registry add -d /database/index.db -b quay.io/operatorhubio/hive-operator:v2.5.3508-6cb94c6 --mode=semver --skip-tls-verify=false --use-http=false
```

**What/Where it is causing an impact? (Context)**

We have been facing issues with the `operator-sdk run bundle` and `operator-idk run bundle-upgrade`. These commands use OLM as dep/lib (version v1.21.0 latest release). The issue does not occur for all/any scenario, but it has been affecting SDK users and the CVP pipeline in some edge cases. 

More info and details can be found in https://github.com/operator-framework/operator-sdk/issues/5410

**Where the error occurs**

The error `error loading bundle into db: FOREIGN KEY constraint failed` came from: [here](https://github.com/operator-framework/operator-registry/blob/master/pkg/sqlite/directory.go#L174). So, it is around [here](https://github.com/operator-framework/operator-registry/blob/aa8db09278a0ea30dd7cf07e3b197a3d7b5ce034/pkg/sqlite/load.go#L550-L745) in the opm code. But, unfortunately, we are returning all errors without a context to know where it fails exactly. 

**Assumptions made so far**

```
operators.operatorframework.io.bundle.channel.default.v1: alpha
operators.operatorframework.io.bundle.channels.v1: mce-2.0
```

From @joelanford : I'm guessing this is the issue. OPM puts the bundle in the mce-2.0 channel, but the bundle says "the default channel is alpha".  except there is no alpha channel.

PR #953, opened 2022-05-16
✨ adds some extra text context to sql statement errors

camilamacedo86:
**Description**
- Add some extra context to the errors, mainly to retrieve the data (SQL)

**Motivation**
- Try to reduce the toll until SQL not be removed. I.e., https://github.com/operator-framework/operator-registry/issues/952. Further info can be helpful for troubleshooting and trying to find workarounds. 

**Tested locally**

Screenshot 2022-05-18 at 14 10 44

Closes https://github.com/operator-framework/operator-registry/issues/952 because, via OPM, we can only provide extra info to clarify the scenario/why the error occurs.

### DIFF
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -255,7 +255,7 @@ func (i *DirectoryPopulator) loadManifestsSemver(bundle *Bundle, skippatch bool)
 	}
 
 	if err := i.loader.AddBundleSemver(updatedGraph, bundle); err != nil {
-		return fmt.Errorf("error loading bundle into db: %s", err)
+		return fmt.Errorf("error loading bundle %s into db: %s", bundle.Name, err)
 	}
 
 	return nil
@@ -268,7 +268,7 @@ func (i *DirectoryPopulator) loadOperatorBundle(manifest PackageManifest, bundle
 	}
 
 	if err := i.loader.AddBundlePackageChannels(manifest, bundle); err != nil {
-		return fmt.Errorf("error loading bundle into db: %s", err)
+		return fmt.Errorf("error loading bundle %s into db: %s", bundle.Name, err)
 	}
 
 	return nil
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -88,13 +88,13 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 
 	addImage, err := tx.Prepare("insert into related_image(image, operatorbundle_name) values(?,?)")
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to insert related image: %s", err)
 	}
 	defer addImage.Close()
 
 	csvName, bundleImage, csvBytes, bundleBytes, _, err := bundle.Serialize()
 	if err != nil {
-		return err
+		return fmt.Errorf("unable to serialize the bundle : %s", err)
 	}
 
 	if csvName == "" {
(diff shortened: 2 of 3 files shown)


pkg/sqlite/load.go
```diff
@@ -404,7 +404,7 @@ func (s *sqlLoader) appendSkips(tx *sql.Tx, skips []string, csvName string) erro
 func (s *sqlLoader) AddPackageChannelsFromGraph(graph *registry.Package) error {
 	tx, err := s.db.Begin()
 	if err != nil {
-		return err
+		return fmt.Errorf("unable to starts a transaction: %s", err)
```
grokspawn, line 407: please change to "unable to start a transaction: %s"

pkg/sqlite/load.go
```diff
@@ -435,10 +435,26 @@ func (s *sqlLoader) AddPackageChannelsFromGraph(graph *registry.Package) error {
...
+		}
+		// just remove the channel_entry when it exists
+		defer rows.Close()
+		if rows.Next() {
+			err := truncChannelGraph(tx, channelName, graph.Name)
+			if err != nil {
+				errs = append(errs, err)
+				break
+			}
+			return err
```
grokspawn, line 457: nit: err will be nil, but elsewhere we always explicitly return nil for the sunny-day path.  
Is it better to do 'return nil' here?

pkg/sqlite/load.go
```diff
@@ -530,7 +546,7 @@ func (s *sqlLoader) AddPackageChannelsFromGraph(graph *registry.Package) error {
 func (s *sqlLoader) AddPackageChannels(manifest registry.PackageManifest) error {
 	tx, err := s.db.Begin()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to starts a db transaction: %s", err)
```
grokspawn, line 549: Please update to "failed to start a db transaction: %s"

pkg/sqlite/load.go
```diff
@@ -424,7 +424,7 @@ func (s *sqlLoader) AddPackageChannelsFromGraph(graph *registry.Package) error {
 	}
 
 	if err := updateDefaultChannel(tx, graph.DefaultChannel, graph.Name); err != nil {
-		errs = append(errs, err)
+		//errs = append(errs, err)
```
grokspawn, line 427: Do we need to remove this commented-out code?

grokspawn (changes requested):

pkg/sqlite/load.go
camilamacedo86: It is wrong. Good catcher. 
I am retesting and updating the first comment

grokspawn (approved):

ankitathomas (approved):
/lgtm

pkg/registry/populator.go
```diff
@@ -255,7 +255,7 @@ func (i *DirectoryPopulator) loadManifestsSemver(bundle *Bundle, skippatch bool)
 	}
 
 	if err := i.loader.AddBundleSemver(updatedGraph, bundle); err != nil {
-		return fmt.Errorf("error loading bundle into db: %s", err)
+		return fmt.Errorf("error loading bundle %s into db: %s", bundle.Name, err)
```
dinhxuanvu, line 258: It seems your intention here to print the error message then it should be `err.Error()`?
Same applied for all below ones as well.

dinhxuanvu (commented):
Looking good. Just one question.

pkg/registry/populator.go
camilamacedo86: The fmt will call the string interface for the errors, see: https://pkg.go.dev/fmt#hdr-Printing:

> If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

Because of this, before this PR we get the errors that came from the SQL. The problem is that without context, we have no additional info to try troubleshooting.

PR #964, opened 2022-05-25
(chore): bump k8s to 1.24

everettraven:
Made changes to bump k8s dependencies to support k8s 1.24. 

Changes:
- bump dependencies in go.mod to support k8s 1.24
- changes in `vendor` directory from running `go mod vendor`
- update testdata CRDs  `spec` to have the `preserveUnknownFields: false` value.
    - There were some additional errors that were occurring during the unit tests related to the `preserveUnknownFields` field being automatically set to `true`. I believe this has to do with the v1beta1 CRDs being converted to v1 since v1beta1 is no longer supported for CRDs (more info: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning). If this should not be done I am happy to try and find a different solution, but this seemed to resolve the issues.
- modifications to test cases in validate_test.go and bundle_test.go
    - There was some test failures happening that expected a warning like `CRD etcdclusters.etcd.database.coreos.com/v1beta2 is present in bundle "etcdoperator.v0.9.4" but not defined in CSV` to show up as an error. I found that the following shows that it should be a warning and not an error: https://github.com/operator-framework/operator-registry/blob/1436816926abd87b9c9151b9ce3ff0c07f9f9be1/pkg/lib/validation/bundle.go#L63-L66 If this is not the case and it should be showing up as an error, I am happy to try and investigate this a bit further but would likely need some direction as to where that warning should be converted to an error instead of a warning.

Begin supporting k8s 1.24

**Reviewer Checklist**


joelanford:
How hard would it be to bump the CRD versions from v1beta1 to v1? v1beta1 CRDs are no longer supported (since k8s 1.22), so it seems like those should be bumped at some point.

It seems like this could be done in a follow-up since the tests seem to have all passed.

everettraven:
> How hard would it be to bump the CRD versions from v1beta1 to v1? v1beta1 CRDs are no longer supported (since k8s 1.22), so it seems like those should be bumped at some point.
> 
> It seems like this could be done in a follow-up since the tests seem to have all passed.

I don't *think* it should be to difficult to do that, but can't say for sure. I agree that this would be a good follow up item. May even be able to get away with an update from `v1beta1` --> `v1alpha1` if it seems like that would be better than a `v1beta1` --> `v1` upgrade

dinhxuanvu (approved):
/approve

dinhxuanvu:
@everettraven Would you mind rebasing this PR?

dinhxuanvu:
The podman installation issue seems to be a known issue that is pending: https://github.com/containers/podman/issues/14278#issuecomment-1131743956
We may need to wait for a bit to see if things are sorted out. Stay tuned.

everettraven:
@dinhxuanvu is there anything else that needs to be done to get this merged?

PR #968, opened 2022-05-27
add offline pprof data for `opm serve` command

grokspawn:
Add pprof cpu/mem profiling to `opm serve` command.  Profile files will be generated if user sets the `--debug` flag on execution.  The usage:
```
% ./bin/opm serve -h
This command serves declarative configs via a GRPC server.

NOTE: The declarative config directory is loaded by the serve command at
startup. Changes made to the declarative config after the this command starts
will not be reflected in the served content.

Usage:
  opm serve <source_path> [flags]

Flags:
      --cpuprofile string        output file for CPU profiling (default "/tmp/serve.cpu.prof")
      --debug                    enable debug logging
  -h, --help                     help for serve
      --memprofile string        output file for memory profiling (default "/tmp/serve.mem.prof")
  -p, --port string              port number to serve on (default "50051")
  -t, --termination-log string   path to a container termination log file (default "/dev/termination-log")

Global Flags:
      --skip-tls-verify   skip TLS certificate verification for container image registries while pulling bundles
      --use-http          use plain HTTP for container image registries while pulling bundles
```

The `--cpuprofile` and `--memprofile` options may be used to set per-execution file names/paths to output the profiling data.  (Default is displayed, in /tmp/ directory.)

Update the Dockerfile to generate the profiling data (and set optional paths/filenames):
```
...
COPY amd64 /configs
COPY --from=builder /bin/opm/ /bin/opm
COPY --from=builder /bin/grpc_health_probe /bin/grpc_health_probe
WORKDIR /tmp
ARG user=1001
RUN microdnf clean all && chmod -vR g=u /etc/passwd
USER ${user}
ENTRYPOINT ["/bin/opm"]
CMD ["serve", "--debug", "--cpuprofile",  "/tmp/cpu-profile-info.prof", /configs"]
```

Once the files are retrieved, use the `go tool pprof [filename]` profiler CLI for examination of the CPU time / memory allocation time.

Downstream issues with FBC serving on alternate architectures, needing more information which could be generally useful.

**Reviewer Checklist**


joelanford:
Should we run a `/pprof` http endpoint instead? That would make it easier to pull down the pprof data in the context of a pod running in the cluster.

grokspawn:
> Should we run a `/pprof` http endpoint instead? That would make it easier to pull down the pprof data in the context of a pod running in the cluster.

I gave this some thought, and based on our current diagnosis needs, I thought offline info would be easier to stick on a bz.   

However, I have no broad data to support this, and could go either way if you feel strongly.

cmd/opm/serve/serve.go
```diff
@@ -59,12 +68,32 @@ will not be reflected in the served content.
...
+	cmd.Flags().StringVar(&s.memProfile, "memprofile", "", "output file for memory profiling data dump")
 	cmd.Flags().StringVarP(&s.port, "port", "p", "50051", "port number to serve on")
 	cmd.Flags().StringVarP(&s.terminationLog, "termination-log", "t", "/dev/termination-log", "path to a container termination log file")
 	return cmd
 }
 
 func (s *serve) run(ctx context.Context) error {
+	if s.profilePort != "" {
+		go func() {
+			s.logger.Info(http.ListenAndServe("localhost:"+s.profilePort, nil))
```
exdx, line 82: I saw this in the pprof docs but generally it's not recommended to start a goroutine without any way of stopping it. Maybe we could add a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) with a value of one and then call `Done()` on it in the teardown logic below where the grpc server is torn down? 

Also, this logger would capture the server shutdown as an info-level log, which isn't a big deal but isn't great IMO. Could we swallow the shutdown error somehow?

cmd/opm/serve/serve.go
```diff
@@ -59,12 +68,32 @@ will not be reflected in the served content.
...
+		}()
+	}
+	if s.cpuProfile != "" {
+		f, err := os.Create(s.cpuProfile)
```
exdx, line 86: Should we sanitize the user-input somehow before processing it?

cmd/opm/serve/serve.go
```diff
@@ -59,12 +68,32 @@ will not be reflected in the served content.
 	}
 
 	cmd.Flags().BoolVar(&s.debug, "debug", false, "enable debug logging")
+	cmd.Flags().StringVar(&s.profilePort, "profileport", "", "pprof HTTP endpoint port")
```
exdx, line 71: My personal opinion is that we could have a high-level `enable-pprof` flag and then some sane defaults for the  flags, WDYT? Having to specify specific ports and files might be annoying if a user just wants to see some data.

cmd/opm/serve/serve.go
```diff
@@ -59,12 +68,32 @@ will not be reflected in the served content.
...
+			s.logger.Info(http.ListenAndServe("localhost:"+s.profilePort, nil))
+		}()
+	}
+	if s.cpuProfile != "" {
+		f, err := os.Create(s.cpuProfile)
+		if err != nil {
+			s.logger.WithError(err).Fatalf("unable to open cpu-profile output file: %q", s.cpuProfile)
+		}
+		defer f.Close()
+		if err := pprof.StartCPUProfile(f); err != nil {
```
exdx, line 91: Might be helpful to emit a log here stating the pprof CPU profiling has started writing to file f

cmd/opm/serve/serve.go
```diff
@@ -107,5 +136,18 @@ func (s *serve) run(ctx context.Context) error {
...
 		grpcServer.GracefulStop()
+
+		if s.memProfile != "" {
+			f, err := os.Create(s.memProfile)
+			if err != nil {
+				s.logger.Fatalf("could not create memory profile: %v", err)
+			}
+			defer f.Close()
+			runtime.GC()
+			if err := pprof.WriteHeapProfile(f); err != nil {
```
exdx, line 147: Might be helpful to write a log here that the heap profile has been written to f

exdx (approved):
This looks really good. We can definitely use a pprof implementation like this in the newer set of APIs.

cmd/opm/serve/serve.go
grokspawn: Done

cmd/opm/serve/serve.go
grokspawn: I went both ways on this, because the write paths could be vastly different between the pod and local executions of this tool (plus some unanticipated runtime environment).  
At the end, the choices came down to versatility (and vetting the data for security/access/etc) or using some well-communicated defaults which should hit the majority of cases.  

I landed with high-level file/endpoint enabling flags, and sane defaults. :)

cmd/opm/serve/serve.go
grokspawn: Done

cmd/opm/serve/serve.go
grokspawn: We should, so at least for this PR we're just going to use well-known/communicated output locations, so that we don't have to.

cmd/opm/serve/serve.go
grokspawn: On discussion, we decided that this iteration didn't need to implement goroutine lifecycle management for pprof.  
There is no shutdown error, since it seems that the logger has been gc'd before the function returns. 

Restructured to output good diagnostics but no server shutdown error.

exdx: SGTM

joelanford:
Sorry for the driveby, but a potential option for the "where to put the startup profile" is to capture it to a `bytes.Buffer{}`, and then run an http endpoint that writes that buffer to the http response.

cmd/opm/serve/serve.go
```diff
@@ -59,12 +73,37 @@ will not be reflected in the served content.
...
 	return cmd
 }
 
 func (s *serve) run(ctx context.Context) error {
+
+	if s.pprofEnabledHttp {
+		// pprof tool HTTP endpoint
+		go func() {
+			s.logger.Infof("starting pprof endpoint at %s:%s", "localhost", defaultProfPort)
+			http.ListenAndServe("localhost:"+defaultProfPort, nil)
```
joelanford, line 89: This won't be queryable in a cluster environment unless we run a sidecar proxy that binds on something other than `localhost`. Is that intentional?

cmd/opm/serve/serve.go
```diff
@@ -59,12 +73,37 @@ will not be reflected in the served content.
...
+		if err != nil {
+			s.logger.WithError(err).Fatalf("unable to open cpu-profile output file: %q", defaultProfCpuFilename)
+		}
+		defer f.Close()
+		if err := pprof.StartCPUProfile(f); err != nil {
+			s.logger.Fatalf("could not start CPU profile: %v", err)
+		}
+		s.logger.Infof("writing CPU pprof data to %q", defaultProfCpuFilename)
+
+		defer pprof.StopCPUProfile()
```
joelanford, line 104: Would it make sense to make this a "startup" profile that we explicitly close just before we start the grpc server (and thus after all of the startup parsing has happened)?

That way we could write the profile data (or host it via an http handler) while still keeping the main server running.

cmd/opm/serve/serve.go
grokspawn: Ignorance on my part as what the desired configuration should be.  When I'm operating in an unknown area (to me), I tend to try to force the errors to be early & loud, so they're not missed.  
Should this be just `:port` instead?

tylerslaton (approved):
/lgtm

cmd/opm/serve/serve.go
```diff
@@ -103,9 +122,70 @@ func (s *serve) run(ctx context.Context) error {
 	health.RegisterHealthServer(grpcServer, server.NewHealthServer())
 	reflection.Register(grpcServer)
 	s.logger.Info("serving registry")
+	p.stopStartupCpuProfiling()
```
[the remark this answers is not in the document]
grokspawn: It's a question of scope.  We want to stop profiling CPU info for our custom endpoint once the GRPC server is ready.  If we defer in this scope, then it will stop profiling on process exit, which won't help. 
The intent with the custom endpoint is to profile CPU usage between the process startup and GRPC server readiness.

cmd/opm/serve/serve.go
```diff
@@ -103,9 +122,70 @@ func (s *serve) run(ctx context.Context) error {
...
+func (p *profilerInterface) stopStartupCpuProfiling() {
+	if !p.isEnabled() {
+		return
+	}
+	pprof.StopCPUProfile()
+	p.logger.Info("stop caching startup cpu profile data")
+}
+
+func (p *profilerInterface) httpHandler(w http.ResponseWriter, r *http.Request) {
+	w.Write(p.cpu.Bytes())
```
joelanford, line 190: Is it possible to query this endpoint before the profile data has been written?

If so, should we return a different http error code (maybe 503 Service Unavailable)?

grokspawn: It is *technically* possible to query before any data has been written, but I felt that:

- timing is fast enough that one would have to specifically target that interval (at least for this freshman effort)
- the existing error message was sufficient: `Fetching profile over HTTP from http://localhost:60066/debug/pprof/startup/cpu http://localhost:60066/debug/pprof/startup/cpu: parsing profile: empty input file`

I understand this as the use-case.  If you think that we need to support a general access attempt instead of the specific tool (e.g. curl), then there would be a benefit to providing a different status before the data is available/complete.

cmd/opm/serve/serve.go
```diff
@@ -59,12 +71,19 @@ will not be reflected in the served content.
...
+	cmd.Flags().StringVar(&s.pprofAddr, "pprof-addr", "", "address of startup profiling endpoint")
 	return cmd
 }
 
 func (s *serve) run(ctx context.Context) error {
+	p := newProfilerInterface(s.pprofAddr, s.logger)
+	p.startDefaultEndpoint()
+	if err := p.startCustomEndpoint(); err != nil {
+		return fmt.Errorf("could not start CPU profile: %v", err)
+	}
```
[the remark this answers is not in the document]
joelanford: These are mostly style/idiomatic nits.

> we are installing a custom handler in the default mux in addition to the pprof handlers we get for just doing the import _ "net/http/pprof".

Yeah, my minor style nit here is that the handlers for the server are configured in several different places, so it's a little hard to build the mental model of "what handlers are configured right now on my server?" I'm also not a huge fan of the "side-effect import" style of stuff because it obfuscates what's happening. Being explicit about the built-in pprof handlers we're running is nicer IMO.

The other nice thing about using `http.Server` is that you have more control over other aspects of it (e.g. server timeouts). I generally use `http.Server` rather than `http.ListenAndServe` because:
1. If I ever need to do more fine-grained configuration, I'm already setup for that.
2. It enables the graceful shutdown that `http.ListenAndServe` precludes.

`http.ListenAndServe` is actually just a [small wrapper](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/net/http/server.go;l=3220-3223) around the snippet I posted above.

> I don't think it matters here. I've run shutdowns where I had a pprof connection to the startup profile endpoint and it continued working just fine

`go tool pprof` doesn't actually hold an open connection to the pprof endpoints as far as I know. The concern here is that if a client is in the middle of downloading a profile when we shutdown, we won't wait for that download to complete. This problem would be more apparent if the profiles we served were large enough that it would be easy to initiate the connection and then kill the server before the download completes. For a debug endpoint like this, I don't think that matters really at all, but I just wanted to bring this up as an FYI for "here's the hardened way to configure/run a webserver using Go's stdlib" just for future reference.

cmd/opm/serve/serve.go
joelanford: Ideally, we'd have a `p.isProfileReady` (protected by a mutex) such that we'd do:

```go
p.mutex.RLock()
isProfileReady := p.isProfileReady
p.mutex.RUnlock()
if !isProfileReady  {
	http.Error(w, errors.New("profile is not yet ready"), http.StatusServiceUnavailable)
	return
}
w.Write(p.cpu.Bytes())
```

Again, this is more of a "this is how a hardened server should work", but in my mind this is low-effort enough that we don't have to cut this corner.

cmd/opm/serve/serve.go
```diff
@@ -103,9 +122,70 @@ func (s *serve) run(ctx context.Context) error {
...
+	// pprof tool HTTP endpoint
+	if !p.isEnabled() {
+		return
+	}
+
+	// only logs fatal error on listen failure, otherwise swallowed on nominal closure
+	// goroutine exits with main
+	go func() {
+		p.logger.Info("starting default pprof endpoint")
+		p.logger.Fatal(http.ListenAndServe(p.addr+":"+defaultProfPort, nil))
```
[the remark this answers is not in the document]
joelanford: The expectation would be that a user would provide host:port. That's pretty standard I think.

cmd/opm/serve/serve.go
grokspawn: Done

cmd/opm/serve/serve.go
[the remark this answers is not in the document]
grokspawn: Done

cmd/opm/serve/serve.go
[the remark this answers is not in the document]
grokspawn: Done.  It's inconsistent with the fact that only the port is provided as part of normal operation, but I think I sufficiently mitigated it with help output text.

cmd/opm/serve/serve.go
```diff
@@ -88,10 +108,10 @@ func (s *serve) run(ctx context.Context) error {
 		return fmt.Errorf("could not build index model from declarative config: %v", err)
 	}
 	store, err := registry.NewQuerier(m)
-	defer store.Close()
 	if err != nil {
 		return err
 	}
+	defer store.Close()
```
joelanford, line 114: In this case, I think we should put this back to _beforE_ the error check.

Under the hood, `registry.NewQuerier` creates a temp dir, and begins marshaling all of the FBC over to the `api.Bundle` format to disk. If an error occurs halfway through, we still return a valid `Querier` on which we can call `Close` to ensure that the temp stuff can be cleaned up, even when where was an error.

If we wait until _after_ the error check, we will no longer clean up the temp dir if an error occurs.

grokspawn: Good discussion.  I felt it was probably the case, but generally vscode makes good recommendations so I took this one.  
Reverted in latest.

timflannagan:
I mentioned in slack that I think we can refactor this implementation, but it seems sufficient for now.

/lgtm

cmd/opm/serve/serve.go
```diff
@@ -1,10 +1,17 @@
...
+	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"os"
+	"sync"
+
+	"net/http"
+	http_pprof "net/http/pprof"
```
timflannagan, line 13: > Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:

> time (provides functionality for measuring and displaying time)
> list (implements a doubly linked list)
> http (provides HTTP client and server implementations)
> The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:

> computeServiceClient
> priority_queue

nit: https://go.dev/blog/package-names.

grokspawn: updating, so now we can discuss the merits of my new replacement

cmd/opm/serve/serve.go
```diff
@@ -26,11 +33,17 @@ type serve struct {
 
 	port           string
 	terminationLog string
-	debug          bool
+
```
awgreene, line 36: Do we need this newline?

cmd/opm/serve/serve.go
```diff
@@ -103,9 +123,121 @@ func (s *serve) run(ctx context.Context) error {
...
+	mux.HandleFunc(defaultCpuStartupPath, p.httpHandler)
+
+	p.server = http.Server{
+		Addr:    p.addr,
+		Handler: mux,
+	}
+
+	// goroutine exits with main
+	go func() {
+
```
awgreene, line 186: Nit: remove newline.

cmd/opm/serve/serve.go
```diff
@@ -103,9 +123,121 @@ func (s *serve) run(ctx context.Context) error {
 	health.RegisterHealthServer(grpcServer, server.NewHealthServer())
 	reflection.Register(grpcServer)
 	s.logger.Info("serving registry")
+	p.stopCpuProfileCache()
```
awgreene, line 126: So the goal is to only log the cpu profile for the startup?

awgreene (commented):
Overall this looks great, nice work!

cmd/opm/serve/serve.go
grokspawn: Correct.  The pprof endpoint does not cache any information until/unless connected-to and requested to do so.  This generates a cache of the CPU profile data from init until the grpc server is ready, accessible for the lifetime of the session.

dinhxuanvu (approved):
/approve

PR #969, opened 2022-06-01
Enhance diff logic to set a new defaultChannel by adding a priority property to a channel

joehuizenga:
This PoC was driven from a discussion in the oc-mirror project
https://github.com/openshift/oc-mirror/issues/453

There are scenarios where the default channel for a package is not updated when diff filters out all the packages within original default channel.

This PR proposes a solution to this issue by adding a channel priority property and logic in the diff code that will use this new priority to set the defaultChannel 

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/declcfg_to_model.go
+++ b/alpha/declcfg/declcfg_to_model.go
@@ -53,9 +53,10 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
 		}
 
 		mch := &model.Channel{
-			Package: mpkg,
-			Name:    c.Name,
-			Bundles: map[string]*model.Bundle{},
+			Package:    mpkg,
+			Name:       c.Name,
+			Bundles:    map[string]*model.Bundle{},
+			Properties: c.Properties,
 		}
 
 		cde := sets.NewString()
--- a/alpha/declcfg/declcfg_to_model.go
+++ b/alpha/declcfg/declcfg_to_model.go
@@ -53,9 +53,12 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
 		}
 
 		mch := &model.Channel{
-			Package:    mpkg,
-			Name:       c.Name,
-			Bundles:    map[string]*model.Bundle{},
+			Package: mpkg,
+			Name:    c.Name,
+			Bundles: map[string]*model.Bundle{},
+			// NOTICE: The field Properties of the type Channel is for internal use only.
+			//   DO NOT use it for any public-facing functionalities.
+			//   This API is in alpha stage and it is subject to change.
 			Properties: c.Properties,
 		}
 
--- a/alpha/declcfg/model_to_declcfg.go
+++ b/alpha/declcfg/model_to_declcfg.go
@@ -60,10 +60,11 @@ func traverseModelChannels(mpkg model.Package) ([]Channel, []Bundle) {
 	for _, ch := range mpkg.Channels {
 		// initialize channel
 		c := Channel{
-			Schema:  SchemaChannel,
-			Name:    ch.Name,
-			Package: ch.Package.Name,
-			Entries: []ChannelEntry{},
+			Schema:     SchemaChannel,
+			Name:       ch.Name,
+			Package:    ch.Package.Name,
+			Entries:    []ChannelEntry{},
+			Properties: ch.Properties,
 		}
 
 		for _, chb := range ch.Bundles {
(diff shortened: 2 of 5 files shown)


joelanford:
I don't think this makes sense in the context of the core design of FBC. FBC is declarative and on-cluster FBC clients don't have any concept of channel priority.

Any tool that operates on FBC should always generate a valid FBC as output. If that tool filters out the default channel, that tool is responsible for changing the default channel name specified in the package. That means that tool needs to do one of two things:
1. Allow the new default channel to be defined as input
2. Use a heuristic that chooses a new default channel.

I suppose its conceivable that one such heuristic could be "choose the channel with the highest priority as defined by a new priority property", but I'm not convinced this should be something we support in this repo.

I think there's a larger discussion to be had around the design of `opm diff` and `oc mirror`. I don't know _all_ of the history, but my impression of the goal of `opm diff` is to provide `oc mirror` with a "slimmed down" catalog that's the minimum thing necessary for mirroring.

IMO, I think we've built some APIs incorrectly that lead us to co-mingling concepts. I think mirroring should be three explicit stages:
1. Execute $someProcess that produces the full FBC that you want to be available in the disconnected environment. Our extensible veneer concept is basically this (a veneer is _anything_ that produces valid FBC)
2. Run a diff of the $alreadyMirroredFBCImage and the $toBeMirroredFBCImage; produce a list of images/relatedImages that are present only in the $toBeMirroredFBCImage set. The $toBeMirroredFBCImage and this set of images are fed into the actual mirroring process
3. Mirror the full set of images (or at least their layer differences) produced in step 2.

joehuizenga:
In my mind I do view `oc-mirror` and its ISC as a veneer. From some prior discussions, I was under the impression that maybe the alpha diff code was going to move into the oc-mirror code base and then this "opinionated" use of a channel priority would be ok (ie it would live and be supported in oc-mirror).  For this PoC I implemented this in operator-registry and then pulled it into oc-mirror

joelanford:
> I was under the impression that maybe the alpha diff code was going to move into the oc-mirror code base and then this "opinionated" use of a channel priority would be ok (ie it would live and be supported in oc-mirror)

Yep, that all sounds about right, and is a fair take IMO, assuming the oc-mirror folks are on board of course.

joehuizenga:
@jpower432  @dinhxuanvu @joelanford  in the OLM workgroup meeting last week I briefly covered this PR, it sounded like it might be feasible if its something that the oc-mirror team is "ok" with.  I know this work is transitioning to a new team and that the alpha diff code currently in this repo may move but I would like to understand if this approach is worth taking forward or if we need to look to some other way.  I am new to this process so welcome any and all suggestions as to the best way to proceed.

joehuizenga:
@dmesser @joelanford We would like to proceed with this enhancement so I have removed the WIP reference in the title for this issues.  I am happy to have @pawicao-ibm onboarding to help work thru the process.  This is the first time either of us has gone thru this process so any guidance or direction will be greatly appreciated!

joehuizenga:
@dinhxuanvu what are your thoughts about the changes that aren't in the diff code?  The new channel priority property for example? Are you suggesting that all traces of this PR need to move to oc-mirror?

image

dinhxuanvu:
> @dinhxuanvu what are your thoughts about the changes that aren't in the diff code? The new channel priority property for example? Are you suggesting that all traces of this PR need to move to oc-mirror?
> 

I haven't reviewed the code line-by-line but from the initial review I think everything looks fine. Probably a few polishing here and there and it should be ready for the final review.
Yes. I suggest we move the entire `diff` code from registry to oc-mirror including the changes in your PR.

alpha/declcfg/declcfg_to_model.go
```diff
@@ -165,7 +166,8 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
 	}
 
 	if err := mpkgs.Validate(); err != nil {
-		return nil, err
+		fmt.Println("Joe - ignore errors in validation for testing")
+		//return nil, err
```
jmrodri, line 170: Is this still needed?

joehuizenga: oops,  no, we can clean that up once we get concurrence that we want to move all this into oc-mirror directly, I will add this to the olm call tomorrow if room on the agenda

pawicao-ibm:
https://github.com/operator-framework/operator-registry/pull/1006 - follow-up PR with a suggestion of migration. I'd suggest to close this one and move the discussions to the new one

dinhxuanvu:
@pawicao-ibm Just an update for you. We are moving forward with this PR. Please add explicit comments on all channel `Properties` fields to indicate that this field is for internal use and do not use for public-facing functionalities and also remind that this API is in alpha so it is subject to change. Thanks.

pawicao-ibm:
@dinhxuanvu is it okay now?

dinhxuanvu:
@pawicao-ibm Looking good to me.

dinhxuanvu (approved):
/approve

dinhxuanvu:
@joelanford PTAL.

PR #972, opened 2022-06-22
Add `add_mode` column to package table

joelanford:
This PR introduces:
- an sqlite migration that adds an `add_mode` column to the `package` table.
- a new global sanity check that's part of the `add` command that causes failures when attempting to add with a mode that is incompatible with a package's previously used add mode.
- logic to ensure the `add_mode` column is correctly set when any of the existing supported add modes is used (replaces, semver, semver-skippatch)

This change enshrines our existing guidance that add modes cannot be mixed and matched, by comparing the previously-used add mode with the currently requested add mode and emitting an error if they do not match. This functionality is crucial for ensuring that `opm` can maintain the integrity of the upgrade graph of each package in the catalog.

This is also a necessary pre-requisite for #949 , which will add a new add mode called "fbc" that will be able to idempotently convert FBC catalogs to sqlite. We need to be able to prevent declarative FBC-derived packages from being tampered with via imperative add commands.

**Downstream note**: This change will likely need to be backported to all versions we expect to support catalogs containing reverse-migrated packages so that we can ask users to upgrade to the `opm` binary that contains these extra add_mode checks. Users that use versions of `opm` without these checks will still be at risk for mixing/matching add modes despite our guidance that they should not.

**Reviewer Checklist**

### DIFF
--- a/pkg/lib/registry/registry.go
+++ b/pkg/lib/registry/registry.go
@@ -91,6 +91,9 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		simpleRefs = append(simpleRefs, image.SimpleReference(ref))
 	}
 
+	if request.Mode == "" {
+		request.Mode = registry.ReplacesMode
+	}
 	if err := populate(context.TODO(), dbLoader, graphLoader, dbQuerier, reg, simpleRefs, request.Mode, request.Overwrite); err != nil {
 		r.Logger.Debugf("unable to populate database: %s", err)
 
--- a/pkg/registry/populator.go
+++ b/pkg/registry/populator.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strings"
 
 	"github.com/blang/semver/v4"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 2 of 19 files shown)


tylerslaton (approved):
Nice, this looks good to me. The comments really helped me parse what was going on here 🙏🏻  Thanks, Joe! 

Note: Putting on hold so that any current reviewers can keep reviewing.

/hold
/lgtm

pkg/lib/registry/registry.go
```diff
@@ -91,6 +91,9 @@ func (r RegistryUpdater) AddToRegistry(request AddToRegistryRequest) error {
 		simpleRefs = append(simpleRefs, image.SimpleReference(ref))
 	}
 
+	if request.Mode == "" {
+		request.Mode = registry.ReplacesMode
+	}
```
joelanford, line 96: This is required due to the change from an integer-based AddMode type to a string based type. Previously `ReplacesMode` value was 0, so the default value of request.Mode was `0`/`ReplacesMode`

Now, the default value is `""`, so we need to check for that case and translate to `ReplacesMode`

pkg/registry/populator.go
```diff
@@ -63,14 +64,52 @@ func (i *DirectoryPopulator) Populate(mode Mode) error {
...
 
+	// check package add modes
+	for _, image := range imagesToAdd {
+		_, err := i.querier.GetPackage(context.TODO(), image.Bundle.Package)
+		if err != nil {
+			// Package not found errors are okay. That means we're adding this
+			// package for the first time. If the error is anything else, we
+			// should report it.
+			if !strings.Contains(err.Error(), "not found") {
+				errs = append(errs, AddModeError{fmt.Sprintf("failed checking for package %q: %v", image.Bundle.Package, err)})
```
grokspawn, line 83:
```suggestion
				errs = append(errs, AddModeError{ErrorString: fmt.Sprintf("failed checking for package %q: %v", image.Bundle.Package, err)})
```

grokspawn:
We definitely need some docs on this, or customers are just not going to understand the interactions here, particularly mapping the error messages into remediation actions.

perdasilva:

PR #974, opened 2022-06-23
Create database copy in tmp directory

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>

opm registry serve creates a writable copy of the db to a tmp file. This file was being created at the root of the filesystem. This breaks on cluster due to permissions (can't write on /). This change creates the file in the tmp directory to not get caught by fs permissions.

The recent security updates require us to run as non-root rendering '/' read-only. Also it's not good practice to write to the root.

**Reviewer Checklist**


dinhxuanvu (commented):
/lgtm

PR #977, opened 2022-06-23
make 'opm serve' initialization asynchronous

joelanford:

This PR modifies `opm serve` so that loading FBC is handled asynchronously in the background so that the grpc server can start immediately, thus enabling the health probes to return successfully faster.

All of the grpc endpoints that require the FBC to be loaded now block, waiting for the initialization to complete. Clients can now initiate a connection and the server will block the response until initialization completes. Clients that don't with to wait can set timeouts on their requests.

While we shipped a change in catalog-operator to make use of the startupProbe functionality, that fix will not solve problems caused by `opm serve`'s long startup times on clusters without the startupProbe.

This fix solves that problem by ensuring the grpc server starts immediately (much like the sqlite-based grpc server).

**Reviewer Checklist**


Makefile
```diff
@@ -60,7 +60,7 @@ static: build
 
 .PHONY: unit
 unit:
-	$(GO) test -coverprofile=coverage.out $(SPECIFIC_UNIT_TEST) $(TAGS) $(TEST_RACE) -count=1 ./pkg/... ./alpha/...
+	$(GO) test -coverprofile=coverage.out -coverpkg=./... $(SPECIFIC_UNIT_TEST) $(TAGS) $(TEST_RACE) -count=1 ./pkg/... ./alpha/...
```
joelanford, line 63: This change means we now consider cross-package coverage. (i.e. the tests in `pkg/server/server_test.go` can contribute to coverage in `pkg/registry/query.go`)

This very likely means there's an artificial increase in the repo-wide coverage, but I did my best to add specific unit tests related to this PR in `pkg/registry/query_test.go` to make sure the new code paths are relatively well-covered.

pkg/registry/query.go
```diff
@@ -95,15 +150,21 @@ func (q Querier) loadAPIBundle(k apiBundleKey) (*api.Bundle, error) {
...
 	var packages []string
 	for pkgName := range q.pkgs {
 		packages = append(packages, pkgName)
 	}
 	return packages, nil
 }
 
-func (q Querier) ListBundles(ctx context.Context) ([]*api.Bundle, error) {
+func (q *Querier) ListBundles(ctx context.Context) ([]*api.Bundle, error) {
+	if err := q.Wait(ctx); err != nil {
```
dinhxuanvu, line 165: At the moment, gRPC client on OLM side doesn't have a deadline setting so the rRPC calls can wait here for a long time. I wonder if we should set a deadline now so that if it goes for too long, it will fail out with DEADLINE_EXCEED error. Or perhaps "waiting until ready regardless how long it is" is a desire behavior? I would love this to be tested in a real cluster to see if things work well.

dinhxuanvu (commented):
Looking good. Just one question.

jdockter:
@joelanford is this issue on permanent hold or do you see this getting resolved as is?

joelanford:
@jdockter I don't think it's permanently blocked. We just need to prioritize checking the implications of this change on client of the GRPC server (catalog-operator and packageserver).

joelanford:
Closing in favor of #1005

PR #985, opened 2022-07-06
Add /opm directory dedicated to the 1001 user

perdasilva:
Signed-off-by: perdasilva <perdasilva@redhat.com>

Adds a /opm directory to the opm container with ownership to the 1001 user. 
Right now this will be used to fix a bug in the operator-sdk project for legacy (sqlite) catalogs. Though it pays to have a user owner directory in the container for the future in case it is needed.

Recent changes that added a user to the opm container. This broke the operator-sdk `run bundle` command because it wants to create a directory in the container as part of the pod command.

**Reviewer Checklist**

### DIFF
--- a/release/goreleaser.opm.Dockerfile
+++ b/release/goreleaser.opm.Dockerfile
@@ -1,11 +1,16 @@
 # NOTE: This Dockerfile is used in conjuction with GoReleaser to
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
+FROM alpine as fsutil
+RUN mkdir -p /opm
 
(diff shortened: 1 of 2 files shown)


release/goreleaser.opm.Dockerfile
```diff
@@ -1,11 +1,16 @@
 # NOTE: This Dockerfile is used in conjuction with GoReleaser to
 #   build opm images. See the configurations in .goreleaser.yaml
 #   and .github/workflows/release.yaml.
+FROM alpine as fsutil
+RUN mkdir -p /opm
```
joelanford, line 5: Is it necessary to do this with a separate image? I thought the `gcr.io/distroless/static:debug` had busybox utils that could help us here.

PR #988, opened 2022-07-08
`semver veneer`: validate bundle versions differ by more than build metadata

grokspawn:

A quick PR to change behavior processing a list of bundles in a channel from
(before) rejection if any bundle versions contain build metadata
(after) rejection IFF any two bundles differ by only build metadata

Example captured `error` type with three package that collide based only on build-metadata versioning:
```"encountered bundle versions which differ only by build metadata, which cannot be ordered: [bundle version \"1.3.1-alpha\" cannot be compared to \"1.3.1-alpha+2001Jan21\", bundle version \"1.3.1-alpha\" cannot be compared to \"1.3.1-alpha+2003May06\"]"```

**Reviewer Checklist**

### DIFF
--- a/alpha/veneer/semver/semver.go
+++ b/alpha/veneer/semver/semver.go
@@ -13,6 +13,7 @@ import (
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/alpha/property"
 	"github.com/operator-framework/operator-registry/pkg/image"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
@@ -140,16 +141,25 @@ func (sv *semverVeneer) getVersionsFromStandardChannels(cfg *declcfg.Declarative
 	if err != nil {
 		return nil, err
 	}
+	if err = validateVersions(&bdm); err != nil {
+		return nil, err
+	}
 	versions[candidateChannelName] = bdm
 
 	bdm, err = sv.getVersionsFromChannel(sv.Fast.Bundles, cfg)
 	if err != nil {
 		return nil, err
 	}
+	if err = validateVersions(&bdm); err != nil {
+		return nil, err
+	}
 	versions[fastChannelName] = bdm
 
 	bdm, err = sv.getVersionsFromChannel(sv.Stable.Bundles, cfg)
 	if err != nil {
+		return nil, err
+	}
+	if err = validateVersions(&bdm); err != nil {
 		return nil, err
 	}
 	versions[stableChannelName] = bdm
@@ -189,9 +199,6 @@ func (sv *semverVeneer) getVersionsFromChannel(semverBundles []semverVeneerBundl
 		if err != nil {
 			return nil, fmt.Errorf("bundle %q has invalid version %q: %v", b.Name, props.Packages[0].Version, err)
 		}
-		if len(v.Build) > 0 {
-			return nil, fmt.Errorf("bundle %q uses build metadata in its versioning, which is not sortable: %v", b.Name, v.String())
-		}
 
 		// package name detection
 		if sv.pkg != "" {
@@ -204,6 +211,10 @@ func (sv *semverVeneer) getVersionsFromChannel(semverBundles []semverVeneerBundl
 			p := newPackage(props.Packages[0].PackageName)
 			cfg.Packages = append(cfg.Packages, *p)
 			sv.pkg = props.Packages[0].PackageName
+		}
+
+		if _, ok := entries[b.Name]; ok {
+			return nil, fmt.Errorf("duplicate bundle name %q", b.Name)
 		}
 
 		entries[b.Name] = v
@@ -255,7 +266,7 @@ func (sv *semverVeneer) generateChannels(semverChannels *semverRenderedChannelVe
 
 		// sort the bundle names according to their semver, so we can walk in ascending order
 		bundleNamesByVersion := []string{}
-		for b, _ := range bundles {
+		for b := range bundles {
 			bundleNamesByVersion = append(bundleNamesByVersion, b)
 		}
 		sort.Slice(bundleNamesByVersion, func(i, j int) bool {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: alpha/veneer/semver/semver.go [hunk 1]
```diff
@@ -434,3 +440,32 @@ func MermaidChannelWriter(cfg declcfg.DeclarativeConfig, out io.Writer) error {
 	}
 	return nil
 }
+
+func withoutBuildMetadataConflict(versions *map[string]semver.Version) error {
```
joelanford, line 444: Might be able to simplify this function a bit by:
1. making a `seen := map[string]semver.Version{}` map
2. iterating the versions and for each of them, get the `base` version by setting `Build` to `nil`, then check if that `base` version's `String()` value already exists in the seen map. if so, fail. if not, add it.

joelanford: If you wanted to get really fancy, you could have `seen := map[string][]semver.Version{}` (note the slice of versions). With that, the keys would be the semver version sans `Build`, and the values would be all of the versions that share that base.

Then you could collect and report all of the dups (i.e. all of the slices that are length > 1).

grokspawn: I did the former.  Ironically, I did the latter in May as part of the getBundleVersions implementation, but there were assertions that it was more complex than necessary and we iterated away from it.

joelanford: 🤷‍♂️ 

I don't have strong opinions. From a UX perspective, if there are multiple duplicates, a user will have to run and re-run the veneer multiple times to discover all of them. But I think adding multiple versions such that multiple duplicates would be possible will be rare (except for maybe the initial setup of the veneer), so probably not a big deal.

alpha/veneer/semver/semver.go
```diff
@@ -434,3 +445,43 @@ func MermaidChannelWriter(cfg declcfg.DeclarativeConfig, out io.Writer) error {
...
+	errs := []error{}
+
+	fmt.Printf("<--> input versions is %v\n", versions)
```
dinhxuanvu, line 452: Is this for debugging reason? If it is, then remove it.

alpha/veneer/semver/semver.go
```diff
@@ -434,3 +445,43 @@ func MermaidChannelWriter(cfg declcfg.DeclarativeConfig, out io.Writer) error {
...
+
+	// using the stringified semver because the semver package generates deterministic representations,
```
dinhxuanvu, line 454: Typo: stringfield

dinhxuanvu (commented):
Looking good. A few nits.

alpha/veneer/semver/semver.go
grokspawn: https://en.wiktionary.org/wiki/stringify#English
Not typo.  Just means to convert to a string format.

alpha/veneer/semver/semver.go
grokspawn: Dangit!  Fair caught.  I was sure I'd cleared all of those out. :(

dinhxuanvu (commented):
/lgtm

alpha/veneer/semver/semver.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/image"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"sigs.k8s.io/kind/pkg/errors"
```
timflannagan, line 18: Can we use the "k8s.io/apimachinery/pkg/util/errors" package instead, and use that package's NewAggregate functionality? I think we'll want to move off having a sigs.k8s.io/kind direct dependency given it's only used for testing reasons, and minimizing the number of packages that use that dependency can help ease that process over time.

alpha/veneer/semver/semver.go
```diff
@@ -255,7 +266,7 @@ func (sv *semverVeneer) generateChannels(semverChannels *semverRenderedChannelVe
 
 		// sort the bundle names according to their semver, so we can walk in ascending order
 		bundleNamesByVersion := []string{}
-		for b, _ := range bundles {
```
timflannagan, line 258: :+1:

alpha/veneer/semver/semver.go
```diff
@@ -434,3 +445,41 @@ func MermaidChannelWriter(cfg declcfg.DeclarativeConfig, out io.Writer) error {
 	}
 	return nil
 }
+
+func withoutBuildMetadataConflict(versions *map[string]semver.Version) error {
```
timflannagan, line 449: Do we need to pass a pointer here if we aren't manipulating that variable anywhere in this function? It's a shame Go doesn't have const refing primitives in the language.

alpha/veneer/semver/semver.go
```diff
@@ -434,3 +445,41 @@ func MermaidChannelWriter(cfg declcfg.DeclarativeConfig, out io.Writer) error {
...
+	errs := []error{}
+
+	// using the stringified semver because the semver package generates deterministic representations,
+	// and because the semver.Version contains slice fields which make it unsuitable as a map key
+	//      stringified-semver.Version ==> incidence count
+	seen := make(map[string]int)
```
timflannagan, line 455: It seems like we could have a map of string -> struct{} or string -> bool k/v pairs here instead. Are we consuming the number of seen semver strings anywhere, or are we just checking for whether we've seen and processing this semver string before?

timflannagan:
Do we have any guidance around codecov for this repository? It looks like the UI is complaining about untested paths despite it being loosely tested with those extra semver versions being added to existing tests.

dinhxuanvu:
> Do we have any guidance around codecov for this repository? It looks like the UI is complaining about untested paths despite it being loosely tested with those extra semver versions being added to existing tests.

The test code should try to cover as much as possible but given this is on alpha, I'm not too concerned about it. Potential subsequent PRs to improve test coverage further.

timflannagan:
Yeah, that's a good call out. I'm cool with that approach.

dinhxuanvu (approved):
/approve

timflannagan:
I mentioned this in slack, but probably the only blocking thing from my perspective is removing the sigs.k8s.io/kind package import in favor of the k8s.io/apimachinery's NewAggregate implementation given it's widely used throughout the rest of the OLM ecosystem.

dinhxuanvu:
> Yeah, that's a good call out. I'm cool with that approach.

I would usually stress a bit more on test coverage but this veneer code is so new and more changes coming, I

alpha/veneer/semver/semver.go
grokspawn: We're not consuming the incidence anywhere, but this is me #futureproofing.  I figure if we get to a place where we want to report the collision count as well, we're fortified.

alpha/veneer/semver/semver.go
dinhxuanvu: I thought we always use apimachinery error. I don't think we use kind error anywhere. Probably just a wrong reference here.

alpha/veneer/semver/semver.go
grokspawn: IKR?  I was being really conservative and passing the minimum necessary overhead.  You let me know if that's a concern.

alpha/veneer/semver/semver.go
grokspawn: Can do.  I cribbed this approach from other code and wasn't sensitive to the import used.

PR #999, opened 2022-07-28
Remove unused Dockerfiles

timflannagan:
Signed-off-by: timflannagan <timflannagan@gmail.com>

Remove the opm-example.Dockerfile and index.Dockerfile that are unused.

Tech debt.

**Reviewer Checklist**


dinhxuanvu:
These two Dockerfiles are more or less examples that can be used for development. I did use then from time to time to build index images. It's not a bad thing to keep some example manifests around given this is an upstream repo.

timflannagan:
The problem is there's some many Dockerfiles in the repository, and having to potentially accommodate them while doing routine maintenance just increases that effort from my perspective. In terms of these specific Dockerfiles: if their role is simply to be consumed as examples, then I'm less inclined to keep them around.

dinhxuanvu:
> The problem is there's some many Dockerfiles in the repository, and having to potentially accommodate them while doing routine maintenance just increases that effort from my perspective. In terms of these specific Dockerfiles: if their role is simply to be consumed as examples, then I'm less inclined to keep them around.

If the rationale is to remove examples, then there are others to purge as well. If the concern is they are not organized, then I recommend to put them in a directory dedicated for development/examples or alike.

timflannagan:
I think my underlying motivation here is to start cutting around the edges of the registry repository, and remove some of the cruft to help work towards removing the vendor/ directory over time. That should help unlock upstream-downstream automation given the frequency of merge conflicts in that domain.

If we'd prefer to house the example Dockerfiles in a dedicated directory, then that's fine with me, but I'd still argue those dockerfile still require maintenance over time.

grokspawn:
It seems that we need to commit that either these dockerfiles are:
1. useful reference and should be maintained
2. with no special merit and an update burden

I'm of the opinion that the downstream one (index.Dockerfile) falls into the second case and doesn't need to be maintained in an upstream repo.  
Given that the other one is "opm-example.Dockerfile" and doesn't have dependencies outside this repo, it sounds like it falls into the first case.

PR #1004, opened 2022-07-31
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

**Reviewer Checklist**

### DIFF
--- a/alpha/action/migrate_test.go
+++ b/alpha/action/migrate_test.go
@@ -37,7 +37,7 @@ func TestMigrate(t *testing.T) {
 	err := generateSqliteFile(dbFile, sqliteBundles)
 	require.NoError(t, err)
 
-	reg, err := newMigrateRegistry(sqliteBundles)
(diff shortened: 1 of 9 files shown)


dinhxuanvu (approved):
/approve

PR #1006, opened 2022-08-02
Migration of the diff logic to oc-mirror

pawicao-ibm:
This PR is a follow-up of https://github.com/operator-framework/operator-registry/pull/969 and is being worked on simultaneously with https://github.com/openshift/oc-mirror/pull/497

The discussion on the diff logic started with https://github.com/openshift/oc-mirror/issues/453 issue where the defaultChannels for operators where invalid while using minVersion/maxVersion.

Further work and discussions on adding the Priority property (see https://github.com/operator-framework/operator-registry/pull/969) resulted in a conclusion to migrate the entirety of diff logic to oc-mirror.

This work has been prepared in the following PR: https://github.com/openshift/oc-mirror/pull/497

Some changes in the operator-registry have to be done though together with the migration (e.g. changing schema variables to public), in order for oc-mirror's diff to be able to take advantage of needed operator-registry structures.

Lastly this migration results in changes to `opm diff`. See [here](https://github.com/openshift/oc-mirror/pull/497#issuecomment-1201173644) for the latest thoughts on deprecating this command

**Reviewer Checklist**


dinhxuanvu:
Please do not merge this PR until the oc-mirror PR is merged. I will put a hold here just in case.
/hold

dinhxuanvu:
I would like to ask you to squash the commits to reduce the noises here. Grouping the commits together based on their purposes to reduce maybe 2 main commits (one for adding channel properties and one for removing the diff logic).
I would prefer having the two proposed commits to be 2 separate PRs given they are not necessarily linked together but I'll let others chime in.
@joelanford FYI

dinhxuanvu:
One a second look, in fact there is already a PR for priority channel which is linked here (#969) so I recommend to move all commits associated to that PR out of this PR and squash the rest into one.

pawicao-ibm:
It made sense and I followed your suggestion. This PR now includes the migration of the diff logic to oc-mirror, https://github.com/operator-framework/operator-registry/pull/969 includes only adding the channel priority.

There are two things to note here:
1. Unfortunately it is not that easy to just wait for oc-mirror PR to be merged first as it also depends on changes we are making here - refactoring some variables to public in order for them to be accessible from other projects (schemas mostly)
2. The question here is what would be done with `opm alpha diff`. This PR deletes it completely. Should it stay here, have a dependency to diff from oc-mirror and be left with a deprecation notice for some time? Or is the complete removal okay?

dinhxuanvu:
> It made sense and I followed your suggestion. This PR now includes the migration of the diff logic to oc-mirror, #969 includes only adding the channel priority.
> 
> There are two things to note here:
> 
> 1. Unfortunately it is not that easy to just wait for oc-mirror PR to be merged first as it also depends on changes we are making here - refactoring some variables to public in order for them to be accessible from other projects (schemas mostly)
> 2. The question here is what would be done with `opm alpha diff`. This PR deletes it completely. Should it stay here, have a dependency to diff from oc-mirror and be left with a deprecation notice for some time? Or is the complete removal okay?

If you remove the diff code, then `alpha diff` command on `opm` will need to get removed as well.

pawicao-ibm:
> If you remove the diff code, then `alpha diff` command on `opm` will need to get removed as well.

yes and that's exactly how it is prepared now. I was asking if for sure you are okay with that

dinhxuanvu:
@pawicao-ibm Hey Oskar, do me a favor and do a few things for this PR:

- Rebase against master
- Run `make vendor` and ensure the `sanity` check pass
- Sign your commit

Thanks.

pawicao-ibm:
> @pawicao-ibm Hey Oskar, do me a favor and do a few things for this PR:
> 
>     * Rebase against master
> 
>     * Run `make vendor` and ensure the `sanity` check pass
> 
>     * Sign your commit
> 
> 
> Thanks.

Is it okay now?

dinhxuanvu (approved):
/approve

dinhxuanvu:
@joelanford Can you please take a look and give it a lgtm if everything is good? Thanks.

alpha/declcfg/declcfg.go
```diff
@@ -7,9 +7,9 @@ import (
 )
 
 const (
-	schemaPackage = "olm.package"
-	schemaChannel = "olm.channel"
-	schemaBundle  = "olm.bundle"
+	SchemaPackage = "olm.package"
```
grokspawn, line 10: Do these consts need to be exported?  I checked existing references, and it's only used w/in the declcfg package, which would not require exporting.

alpha/declcfg/model_to_declcfg.go
```diff
@@ -115,7 +115,7 @@ func traverseModelChannels(mpkg model.Package) ([]Channel, []Bundle) {
 	return channels, bundles
 }
 
-func modelRelatedImagesToRelatedImages(relatedImages []model.RelatedImage) []RelatedImage {
+func ModelRelatedImagesToRelatedImages(relatedImages []model.RelatedImage) []RelatedImage {
```
grokspawn, line 118: Does this need to be exported?  All references I could find were local, and we'd like to minimize exports due to Hyrum's Law.

grokspawn (changes requested):
Appreciate it if we can run to ground the new exports before concluding this.  Unless required by an API consumer not represented here, we'd like to minimize exports when possible.

dinhxuanvu:
> Appreciate it if we can run to ground the new exports before concluding this. Unless required by an API consumer not represented here, we'd like to minimize exports when possible.

I'm pretty sure the diff code uses them. Now the diff code is moving out of the package, these variables and functions need to be exportable so they can be used on mirror repo.

dinhxuanvu:
@pawicao-ibm Can you confirm that you need those variables and the func to be exportable for mirror usage? Or there is another reason for it?

pawicao-ibm:
Now that I look into that, they are used but within tests of the diff logic, while creating DeclarativeConfigs. They are also used here: https://github.com/openshift/oc-mirror/blob/a1440e623b1f6c9652c3053ec78661bfe223ecae/pkg/operator/diff/internal/diff.go#L542 - this function is called for comparing bundles. Maybe there would be a way around that tho

dinhxuanvu:
> Now that I look into that, they are used but within tests of the diff logic, while creating DeclarativeConfigs. They are also used here: https://github.com/openshift/oc-mirror/blob/a1440e623b1f6c9652c3053ec78661bfe223ecae/pkg/operator/diff/internal/diff.go#L542 - this function is called for comparing bundles. Maybe there would be a way around that tho

I see. The main question is if these changes are needed. If they ain't, we should reverse them. I personally don't mind the changes but let keep the changes to the minimum and avoid unnecessary ones.

pawicao-ibm:
I am looking into that

pawicao-ibm:
> > Now that I look into that, they are used but within tests of the diff logic, while creating DeclarativeConfigs. They are also used here: https://github.com/openshift/oc-mirror/blob/a1440e623b1f6c9652c3053ec78661bfe223ecae/pkg/operator/diff/internal/diff.go#L542 - this function is called for comparing bundles. Maybe there would be a way around that tho
> 
> I see. The main question is if these changes are needed. If they ain't, we should reverse them. I personally don't mind the changes but let keep the changes to the minimum and avoid unnecessary ones.

These changes are needed if we want to have the diff functions in oc-mirror and not have redundant duplications of these schemas. I see some possibility of manouvering functions here and there. For example some of the internal diff functions (https://github.com/openshift/oc-mirror/blob/a1440e623b1f6c9652c3053ec78661bfe223ecae/pkg/operator/diff/internal/diff.go#L542:L629) could go to operator-registry - `model.go` or `model_to_declcfg.go` yet in the end they would have to be exported anyway

grokspawn (approved):

PR #1015, opened 2022-08-22
Add pipe support to `opm alpha render-veneer semver`

### LINKED ISSUE #1011: feature: opm alpha render-veneer support for pipe stdin
enhance the veneer rendering to be able to read the input veneer data from stdin, via a pipe, in a manner similar to https://dev.to/napicella/linux-pipes-in-golang-2e8j

then the command could be used in a manner similar to many k8s examples like

```shell
opm alpha render-veneer semver -o yaml < infile > outfile
```

oceanc80:
Signed-off-by: Catherine Chan-Tse <cchantse@redhat.com>

Adds pipe support to the `olm alpha render-veneer semver` command. An input veneer file can be passed in as a positional argument to the command and alternatively, if no file is passed or `-` is passed as the file, the input veneer is read from standard input allowing for commands like the following to work:
```
// Example with no file argument passed in
opm alpha render-veneer semver -o yaml < infile.semver.veneer.yaml > outfile.yaml

// Example with `-` as the file argument passed in
cat infile.semver.veneer.yaml | opm alpha render-veneer semver -o yaml -
```

Closes #1011 

Enhances the veneer rendering command to be able to read the input veneer data from stdin.-o

**Reviewer Checklist**

### DIFF
--- a/alpha/veneer/semver/semver.go
+++ b/alpha/veneer/semver/semver.go
@@ -3,7 +3,7 @@ package semver
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"sort"
 
 	"github.com/blang/semver/v4"
@@ -18,8 +18,8 @@ import (
 
 // data passed into this module externally
 type Veneer struct {
-	Ref string
-	Reg image.Registry
+	Data     io.Reader
+	Registry image.Registry
 }
 
 // IO structs -- BEGIN
@@ -78,7 +78,7 @@ type semverRenderedChannelVersions map[string]map[string]semver.Version // e.g. 
 func (v Veneer) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
 	var out declcfg.DeclarativeConfig
 
-	sv, err := readFile(v.Ref)
+	sv, err := readFile(v.Data)
 	if err != nil {
 		return nil, fmt.Errorf("semver-render: unable to read file: %v", err)
 	}
@@ -88,7 +88,7 @@ func (v Veneer) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error) 
 		r := action.Render{
 			AllowedRefMask: action.RefBundleImage,
 			Refs:           []string{b.Image},
-			Registry:       v.Reg,
+			Registry:       v.Registry,
 		}
 		c, err := r.Run(ctx)
 		if err != nil {
@@ -114,8 +114,8 @@ func (v Veneer) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error) 
 	return &out, nil
 }
 
-func readFile(ref string) (*semverVeneer, error) {
-	data, err := ioutil.ReadFile(ref)
+func readFile(data io.Reader) (*semverVeneer, error) {
+	fileData, err := io.ReadAll(data)
 	if err != nil {
 		return nil, err
 	}
@@ -126,7 +126,7 @@ func readFile(ref string) (*semverVeneer, error) {
 		GenerateMinorChannels: true,
 		AvoidSkipPatch:        false,
 	}
-	if err := yaml.Unmarshal(data, &sv); err != nil {
+	if err := yaml.Unmarshal(fileData, &sv); err != nil {
 		return nil, err
 	}
 	return &sv, nil
--- a/cmd/opm/alpha/veneer/semver.go
+++ b/cmd/opm/alpha/veneer/semver.go
@@ -18,16 +18,28 @@ import (
 func newSemverCmd() *cobra.Command {
 	output := ""
 	cmd := &cobra.Command{
-		Use:   "semver <filename>",
-		Short: "Generate a file-based catalog from a single 'semver veneer' file",
-		Long:  `Generate a file-based catalog from a single 'semver veneer' file`,
-		Args:  cobra.MinimumNArgs(1),
+		Use:   "semver [FILE]",
+		Short: "Generate a file-based catalog from a single 'semver veneer' file \nWhen FILE is '-' or not provided, the veneer is read from standard input",
+		Long:  "Generate a file-based catalog from a single 'semver veneer' file \nWhen FILE is '-' or not provided, the veneer is read from standard input",
+		Args:  cobra.MaximumNArgs(1),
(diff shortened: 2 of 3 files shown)


cmd/opm/alpha/veneer/semver.go
```diff
@@ -18,17 +18,24 @@ import (
 func newSemverCmd() *cobra.Command {
 	output := ""
 	cmd := &cobra.Command{
-		Use:   "semver <filename>",
-		Short: "Generate a file-based catalog from a single 'semver veneer' file",
-		Long:  `Generate a file-based catalog from a single 'semver veneer' file`,
-		Args:  cobra.MinimumNArgs(1),
+		Use:   "semver [FILE]",
+		Short: "Generate a file-based catalog from a single 'semver veneer' file \nWith no File, or when FILE is -, the file is read from standard input",
+		Long:  "Generate a file-based catalog from a single 'semver veneer' file \nWith no File, or when FILE is -, the file is read from standard input",
```
grokspawn, line 23: Nit:  consistency of parameter
```suggestion
		Short: "Generate a file-based catalog from a single 'semver veneer' file \nWhen FILE is '-' or not provided, the veneer is read from standard input",
		Long:  "Generate a file-based catalog from a single 'semver veneer' file \nWhen FILE is '-' or not provided, the veneer is read from standard input",
```

alpha/veneer/semver/semver.go
```diff
@@ -18,7 +18,7 @@ import (
 
 // data passed into this module externally
 type Veneer struct {
-	Ref string
+	Ref io.Reader
 	Reg image.Registry
```
grokspawn, line 22: Totally non-blocking, but this was crap when I wrote it, and I meant to get back to it, but ... I lost track.  
If I had a wish-list, I would change this to something like 
```
type Veneer struct {
   Data     io.Reader
   Registry image.Registry
}
```
... and usage accordingly.  
I mention it because in here there's already a breaking API change, so it wouldn't introduce one. ;)

grokspawn:
Looks great now.  Just waiting to approve until sanity passes.

cmd/opm/alpha/veneer/semver.go
```diff
@@ -18,17 +18,25 @@ import (
...
 			)
 
+			if len(args) == 0 || args[0] == "-" {
+				data = cmd.InOrStdin()
+			} else {
+				data, err = os.Open(args[0])
+				if err != nil {
+					return err
+				}
+			}
```
timflannagan, line 38: Move this logic to it's own function that returns an io.Reader? It would help remove the if/else conditional structure that can help improve code readability.

bentito:
The example usages in the PR are better than the current README additions. I'd add them to the README as-is for a better README...

alpha/veneer/semver/semver.go
```diff
@@ -114,8 +114,8 @@ func (v Veneer) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error)
 	return &out, nil
 }
 
-func readFile(ref string) (*semverVeneer, error) {
-	data, err := ioutil.ReadFile(ref)
+func readFile(data io.Reader) (*semverVeneer, error) {
```
bentito, line 117: I think `file` is a better name here than `data` at this point in the code, right? Then below, `data` can go back to being `data`, or maybe change to `fileContents` even?  Trying to avoid `data` becoming `d` for sure.

oceanc80: I was using it to stay consistent with the name of the field in the Veneer struct, but as long as @grokspawn is okay with it (I don't have enough context to know if data vs. file makes more sense in the api as a whole), I'm happy to change it to `file` everywhere.

bentito: I don't have that context either, how anything but `data` then, such that `data` can stay `data` below and not become `d`?

oceanc80: I can change the `d` to `fileData`. I feel like keeping `data` in the function signature makes it easier to understand if someone is doing a go-def inspection on the function as it matches the Veneer struct that way.

cmd/opm/alpha/veneer/semver.go
oceanc80: I'm worried that pulling that logic out into its own function will obfuscate things more because it seems like it'd be more confusing to have to find/look in a separate function to understand what the command's input argument logic is. I can add a comment explaining the logic to help improve readability of the if/else structure.

alpha/veneer/semver/semver.go
bentito: that sounds good, really small variable names with no meaning just make me have to go look (and have to remember)

cmd/opm/alpha/veneer/semver.go
```diff
@@ -18,17 +18,29 @@ import (
...
-			ref := args[0]
-
 			var (
-				err error
+				data io.Reader
+				err  error
 			)
 
+			// Handle different input argument types
+			if len(args) == 0 || args[0] == "-" {
```
bentito, line 32: IntelliJ/GoLand offers this method extraction (naming is mine). I think with good function naming, people can just read right over it, as it's a utility function:
```suggestion
			data, err, err2 := openFileOrPipe(cmd, args, data, err)
			if err2 != nil {
				return err2
			}
		
func openFileOrPipe(cmd *cobra.Command, args []string, data io.Reader, err error) (io.Reader, error, error) {
	if len(args) == 0 || args[0] == "-" {
		// When no arguments or "-" is passed to the command,
		// assume input is coming from stdin
		data = cmd.InOrStdin()
	} else {
		// Otherwise open the file passed to the command
		data, err = os.Open(args[0])
		if err != nil {
			return nil, nil, err
		}
	}
	return data, err, nil
}	
			
```

exdx (commented):
/lgtm

alpha/veneer/semver/semver.go
grokspawn: So ... I know the last thing that anyone wants is to include someone else who introduces a new opinion, but ... :) 

IMHO, my preference is for 
```
func readFile(reader io.Reader) (*semverVeneer, error) {
```

and 

```
data, err := io.ReadAll(reader)
```

PR #1016, opened 2022-08-24
do not fatally exit from serve commands

joelanford:

This commit ensures that we always cleanly return an error from the
serve function, thus ensuring that defered cleanups always happen.

When we fatally exit directly in the serve functions, we do not give a
chance for the registered cleanup functions to run, resulting in
temporary files being left behind.

**Reviewer Checklist**


dinhxuanvu (approved):
/lgtm

PR #1018, opened 2022-08-30
(bug) fix permissions problems with pregen cache access

grokspawn:
Signed-off-by: Jordan <jordan@nimblewidget.com>

create `opm serve` cache artifacts to serve for any uid by generating as 0777 (dirs) / 0666 (files). 

Several options were evaluated, and we landed on a combo of relaxing the umask during cache generation coupled with updated permission of created artifacts to provide world read-/write-ability. 

| characterization | URI | verdict |
| ------------------- | ---------------------------- | --------- |
| original perms conflict | docker.io/grokspawn/catalog-permissions:orig | 0700/0600 *FAIL* |
| 0770/0660 perms | docker.io/grokspawn/catalog-permissions:permbump | 0700/0600 *FAIL* |
| 0777/0666 perms | docker.io/grokspawn/catalog-permissions:permbumpall | 0700/0600 *FAIL* |
| 0777/0666 perms + umask | docker.io/grokspawn/catalog-permissions:final-umask | 0777/0666 *SUCCESS* |
| 0750/0640 perms + umask | docker.io/grokspawn/opm:umask-no-other | 0750/0640 *SUCCESS* | 

To build a candidate opm replacement image, !!IN LINUX!! do 

```shell
cd $your-operator-registry-clone
GOENV=CGO_ENABLED=0 make static 
cp bin/opm .
docker build -f release/goreleaser.opm/Dockerfile -t [tag] .
docker push [tag]
```

Then create the catalog image:

```shell
mkdir -p /tmp/ctest
cd /tmp/ctest
mkdir catalog
[path-to-opm] generate dockerfile catalog
```

and finally, edit the `FROM` line in the `catalog.Dockerfile` to point to the pushed tag for the opm replacement image.

c.f. OCPBUGS-650. 

Cache generation code was creating for uid 1001 and permissions of 0700 (dirs) / 0600 (files), so later deployment in a pod would fail unless uid matched.  

**Reviewer Checklist**

### DIFF
--- a/pkg/registry/query.go
+++ b/pkg/registry/query.go
@@ -17,6 +17,11 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/api"
 )
 
+const (
+	cachePermissionDir  = 0750
+	cachePermissionFile = 0640
+)
+
 type Querier struct {
 	*cache
 }
@@ -423,7 +428,7 @@ func newEphemeralCache() (*cache, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err := os.MkdirAll(filepath.Join(baseDir, "cache"), 0750); err != nil {
+	if err := os.MkdirAll(filepath.Join(baseDir, "cache"), cachePermissionDir); err != nil {
 		return nil, err
 	}
 	return &cache{
@@ -434,7 +439,7 @@ func newEphemeralCache() (*cache, error) {
 }
 
 func newPersistentCache(baseDir string) (*cache, error) {
-	if err := os.MkdirAll(baseDir, 0750); err != nil {
+	if err := os.MkdirAll(baseDir, cachePermissionDir); err != nil {
 		return nil, err
 	}
 	qc := &cache{baseDir: baseDir, persist: true}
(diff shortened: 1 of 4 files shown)


bentito:
Yeah I was wondering same, are any lesser RWX for any of the Group and Other perms possible.

joelanford:
This _does_ mean that if a non-owner encounters a stale cache, they'll fail instead of repopulating the cache because they won't be able to delete the existing stale files. IMO, that's reasonable though. Outcome is "file/dir owner is responsible for making sure cache is not stale".

grokspawn:
As there was previously a 'user is root' implicit aspect in the past (which has been replaced with never-be-root-if-you-can-help-it), there appears to be a 'group is root' implicit aspect presently which allows `other` to have no privileges. 

If/when that changes, this will also need to change.

joelanford:
https://github.com/operator-framework/operator-registry/runs/8100687382?check_suite_focus=true#step:4:155

Oh joy! Looks like we get to experience the fun of using a windows-specific source code file to do something different than `syscall.Umask` (which does not exist in Windows).  My guess is that we'll just make that a no-op on Windows.

pkg/registry/query.go
```diff
@@ -524,7 +528,7 @@ func (qc *cache) repopulateCache(model digestableModel) error {
 					return err
 				}
 				filename := filepath.Join(qc.baseDir, "cache", fmt.Sprintf("%s_%s_%s.json", p.Name, ch.Name, b.Name))
-				if err := os.WriteFile(filename, jsonBundle, 0666); err != nil {
+				if err := os.WriteFile(filename, jsonBundle, 0640); err != nil {
```
perdasilva, line 531: would it be worth exporting the permission octals to a constant?

PR #1022, opened 2022-09-09
add skipRange support to render-graph

grokspawn:

mermaid output of channels will express skipRange relationships between nodes

tired of the header/footer always expressing the feature's shortcomings ;) 

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/write.go
+++ b/alpha/declcfg/write.go
@@ -19,12 +19,7 @@ import (
 // output is sorted lexicographically by package name, and then by channel name
 // if provided, minEdgeName will be used as the lower bound for edges in the output graph
 //
-// NB:  Output has wrapper comments stating the skipRange edge caveat in HTML comment format, which cannot be parsed by mermaid renderers.
-//
-//	This is deliberate, and intended as an explicit acknowledgement of the limitations, instead of requiring the user to notice the missing edges upon inspection.
-//
 // Example output:
-// <!-- PLEASE NOTE:  skipRange edges are not currently displayed -->
 // graph LR
 //
 //	  %% package "neuvector-certified-operator-rhmp"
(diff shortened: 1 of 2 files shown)


cdjohnson:
Thanks for doing this.  I tried it out against one our packages:  `ibm-mq`. 

 Question:  This clearly shows the graph between bundles.  Is there any thought of showing the upgrade graphs between channels, so one can see the intra-channel upgrade paths a bit more obviously?  

Maybe it would be a different render command, or it would get too busy?

E.g. Make it obvious that you can move from any channel to the v2.0 channel.  We have other examples, where one needs to traverse from v1.0 to v1.1 to v2.0 and directly from v1.0 to v2.0.  

Here's what the `ibm-mq` one looks like:
mq

grokspawn:
I'm abandoning this PR in favor of #1023.  There started to be complex interactions between the different filtering options, and an increasing burden of sequencing the PRs was just more than I wanted to countenance.  
I'll read through the feedback so far and see what I can incorporate...  I'm not ignoring it.  It's just that I had to consolidate to reduce complexity.

grokspawn:
> Thanks for doing this. I tried it out against one our packages: `ibm-mq`.
> 
> Question: This clearly shows the graph between bundles. Is there any thought of showing the upgrade graphs between channels, so one can see the intra-channel upgrade paths a bit more obviously?
> 
> Maybe it would be a different render command, or it would get too busy?
> 
> E.g. Make it obvious that you can move from any channel to the v2.0 channel. We have other examples, where one needs to traverse from v1.0 to v1.1 to v2.0 and directly from v1.0 to v2.0.
> 
> Here's what the `ibm-mq` one looks like: mq

That is really cool!  

I hadn't really ever thought about inter-channel linkages before, honestly.  It would definitely make the graph more noisy.  I'll give it a look.

bentito:
The CNV folks have expressed a desire for inter-index graphs as well. That is, it would show the way to proceed when you move to another index.

PR #1023, opened 2022-09-09
(feature) add skipRange support to `opm alpha render-graph` and provide the capability to filter mermaid output to a single package name

grokspawn:

add rendering of skipRange relationships to `opm alpha render-graph` and support an additional filter to allow users to limit results to a specific package in the catalog. 

From usage:
```
Flags:
  -h, --help                  help for render-graph
      --minimum-edge string   the channel edge to be used as the lower bound of the set of edges composing the upgrade graph; default is to include all edges
  -p, --package-name string   a specific package name to filter output; default is to include all packages in reference

Global Flags:
      --skip-tls-verify   skip TLS certificate verification for container image registries while pulling bundles
      --use-http          use plain HTTP for container image registries while pulling bundles
 ```     

skipRange was supported in an earlier PR, but it started having complex interactions with this one, so I merged them here. 
skipRange was a glaring inadequacy of the tool, and there have been multiple requestors for a package-filter option. 

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/write.go
+++ b/alpha/declcfg/write.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by timflannagan** [comment]
Holding in case another review is needed.

/lgtm
/hold

**Comment by grokspawn** [comment]
Blech.  Multiple filters got too complex, so collapsed other PR into this one.

**Comment by grokspawn** [comment]
@cdjohnson  try this one out.  There was a bug where we would drop some nodes that I fixed.  It makes for a slightly longer graph-text, but doesn't really impact the graph-pic at all.  
I just didn't like the fact that it was incomplete.  :)

**Comment by grokspawn** [comment]
@cdjohnson You know ... it helps if you're considering the correct pool from which to get skipRange destination endpoints from ...  I was pulling from the list of all bundles in the catalog, instead of the list of all other entities in the channel.  
It appears to work for me now. :)

**Comment by cdjohnson** [comment]
Verified fixed!

File: alpha/declcfg/write.go [hunk 1]
```diff
@@ -109,31 +161,50 @@ func WriteMermaidChannels(cfg DeclarativeConfig, out io.Writer, minEdgeName stri
...
 	}
-	// convert the edge name to the version so we don't have to duplicate the lookup
-	minVersion := versionMap[minEdgeName]
+
+	// short-circuit if channel's package doesn't match filter
+	if writer.SpecifiedPackageName != "" && c.Package != writer.SpecifiedPackageName {
+		return nil
+	}
+
+	// short-circuit where the minEdgeName is not in this channel
```
bentito, line 181: Not sure if this is really a "short" short-circuit. It's a lot of work for a check isn't it? We're looping on `c.Entries` right below as well, can't any checking for the MinEdgeName just be combined into that loop?

grokspawn: It evaluates a single criterion across the range of entries in the channel in order to recognize a case where there is nothing pertinent in the channel (when the minimumEdge is specified).  

I had attempted to include this in the logic of the larger loop, but it complicated the logic something fierce, as well as introducing inefficiencies like having to parse the skipRange twice for validation.  

I'll take a look at it though to see if I can smooth it out.  If you have any ideas, I'd love to hear them as well.

PR #1024, opened 2022-09-19
bump k8s to 1.25

asmacdo:
Signed-off-by: Austin Macdonald <austin@redhat.com>

~I also switched the github actions to pull the go version from go.mod so we don't have to update all of them every golang bump.~ (Pulling go version from go.mod merged in https://github.com/operator-framework/operator-registry/pull/1025)


laxmikantbpandhare:
`go-apidiff` errors are expected ones.

laxmikantbpandhare:
This has to do with gofmt updates related to bumping the Go version to Go 1.19.

If you take a look at the Headings section of this documentation https://tip.golang.org/doc/comment you can see that as of Go 1.19 the new heading format is similar to Markdown and uses the # prefix:

grokspawn (approved):
Not a blocker, but since you retreated from the position of adjusting workflows to pull go version from go.mod, it would be nice if you edited the PR desc ... it got my hopes up when I saw you using go-version-file... but alas, it was not to be.

I understand if you'd like that to happen separately from this PR, but I was able to get it working in https://github.com/operator-framework/operator-registry/pull/1025

timflannagan (approved):
LGTM once this is rebased.

go.mod
```diff
@@ -12,55 +12,56 @@ require (
...
 	github.com/mattn/go-sqlite3 v1.14.10
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.2.2
-	github.com/onsi/ginkgo/v2 v2.1.3
-	github.com/onsi/gomega v1.18.1
+	github.com/onsi/ginkgo/v2 v2.1.4
+	github.com/onsi/gomega v1.19.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.2
-	github.com/operator-framework/api v0.15.0
+	github.com/operator-framework/api v0.17.2-0.20220915200120-ff2dbc53d381
```
timflannagan, line 26: This looks like an unrelated o-f/api psuedoversion?

timflannagan: Would it be possible to bump this version outside of these changes? It looks like there's been some API changes in the past couple of minor version releases, and we may need to be aware of any order of operation issues here.

timflannagan:
Ended up convincing myself that bumping o-f/api here should be relatively harmless, and I'm more concerned with doing due diligence when bumping that dependency in the OLM repository.

/lgtm

PR #1026, opened 2022-09-27
OCPBUGS-1272 Add pipe support to "opm alpha render-veneer basic"

oceanc80:
Add pipe support to the render-veneer basic command and expand basic veneer validation checks.

Signed-off-by: Catherine Chan-Tse <cchantse@redhat.com>

- Add pipe support to "opm alpha render-veneer basic"
- Expand basic veneer validation checks and fail when non-veneer bundles are passed in

- Addresses OCPBUGS-1272

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/load.go
+++ b/alpha/declcfg/load.go
@@ -118,7 +118,9 @@ func extractCSV(objs []string) string {
 	return ""
 }
 
-func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
+// LoadReader reads yaml or json from the passed in io.Reader and unmarshals it into a DeclarativeConfig struct.
+// Path references will not be de-referenced so callers are responsible for de-referencing if necessary.
+func LoadReader(r io.Reader) (*DeclarativeConfig, error) {
 	cfg := &DeclarativeConfig{}
 	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
 	for {
@@ -173,7 +175,7 @@ func LoadFile(root fs.FS, path string) (*DeclarativeConfig, error) {
 	}
 	defer file.Close()
 
-	cfg, err := readYAMLOrJSON(file)
+	cfg, err := LoadReader(file)
 	if err != nil {
 		return nil, err
 	}
--- a/alpha/veneer/basic/basic.go
+++ b/alpha/veneer/basic/basic.go
@@ -2,9 +2,8 @@ package basic
 
 import (
 	"context"
-	"errors"
-	"os"
-	"path/filepath"
+	"fmt"
+	"io"
 
 	"github.com/operator-framework/operator-registry/alpha/action"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
@@ -15,30 +14,8 @@ type Veneer struct {
 	Registry image.Registry
 }
 
-func (v Veneer) Render(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
-	// only taking first argument as file
-	stat, serr := os.Stat(ref)
-	if serr != nil {
-		return nil, serr
-	}
-
-	if stat.IsDir() {
-		return nil, errors.New("cannot render veneers by directory reference")
-	}
-	return v.renderFile(ctx, ref)
-}
-
-func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.DeclarativeConfig, error) {
-	// xform any relative to absolute paths
-	abspath, err := filepath.Abs(ref)
-	if err != nil {
-		return nil, err
-	}
-	// xform to break apart dir/file elements
-	rpath, fname := filepath.Split(abspath)
-	root := os.DirFS(rpath)
-
-	cfg, err := declcfg.LoadFile(root, fname)
+func (v Veneer) Render(ctx context.Context, reader io.Reader) (*declcfg.DeclarativeConfig, error) {
+	cfg, err := declcfg.LoadReader(reader)
 	if err != nil {
 		return cfg, err
 	}
@@ -52,8 +29,7 @@ func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.Declarativ
 
 	for _, b := range cfg.Bundles {
 		if !isBundleVeneer(&b) {
-			outb = append(outb, b)
-			continue
+			return nil, fmt.Errorf("unexpected fields present in basic veneer bundle")
 		}
 		r.Refs = []string{b.Image}
 		contributor, err := r.Run(ctx)
@@ -67,7 +43,8 @@ func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.Declarativ
 	return cfg, nil
 }
 
-// isBundleVeneer identifies loaded partial Bundle data from YAML/JSON veneer source as having no properties,
+// isBundleVeneer identifies a Bundle veneer source as having a Schema and Image defined
+// but no Properties, RelatedImages or Package defined
 func isBundleVeneer(b *declcfg.Bundle) bool {
-	return len(b.Properties) == 0
+	return b.Schema != "" && b.Image != "" && b.Package == "" && len(b.Properties) == 0 && len(b.RelatedImages) == 0
 }
--- a/cmd/opm/alpha/veneer/basic.go
+++ b/cmd/opm/alpha/veneer/basic.go
@@ -20,11 +20,22 @@ func newBasicVeneerRenderCmd() *cobra.Command {
 		output string
 	)
 	cmd := &cobra.Command{
-		Use:   "basic basic-veneer-file",
-		Short: "Generate a declarative config blob from a single 'basic veneer' file",
-		Long:  `Generate a declarative config blob from a single 'basic veneer' file, typified as a declarative configuration file where olm.bundle objects have no properties`,
-		Args:  cobra.ExactArgs(1),
+		Use: "basic basic-veneer-file",
+		Short: `Generate a file-based catalog from a single 'basic veneer' file
+When FILE is '-' or not provided, the veneer is read from standard input`,
+		Long: `Generate a file-based catalog from a single 'basic veneer' file
+When FILE is '-' or not provided, the veneer is read from standard input`,
+		Args: cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
+			// Handle different input argument types
+			// When no arguments or "-" is passed to the command,
+			// assume input is coming from stdin
+			// Otherwise open the file passed to the command
+			data, source, err := openFileOrStdin(cmd, args)
+			if err != nil {
+				log.Fatalf("unable to open %q: %v", source, err)
+			}
+			defer data.Close()
 
 			var write func(declcfg.DeclarativeConfig, io.Writer) error
 			switch output {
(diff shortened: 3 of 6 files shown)


alpha/declcfg/load.go
```diff
@@ -118,7 +118,7 @@ func extractCSV(objs []string) string {
 	return ""
 }
 
-func readYAMLOrJSON(r io.Reader) (*DeclarativeConfig, error) {
+func LoadReader(r io.Reader) (*DeclarativeConfig, error) {
```
joelanford, line 121: This should have some GoDoc comments that describe what it does. And part of that description should mention the fact that path references will not be de-referenced and that it's up to callers to do that de-referencing if necessary.

alpha/veneer/basic/basic.go
```diff
@@ -52,8 +29,7 @@ func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.Declarativ
 
 	for _, b := range cfg.Bundles {
 		if !isBundleVeneer(&b) {
-			outb = append(outb, b)
-			continue
+			return nil, fmt.Errorf("bundle veneers must define only a schema and an image; all other fields must not be defined")
 		}
```
grokspawn, line 33: The original intent of this code was to allow the author to provide a combination of completely-rendered olm.bundle references as well as 'shorthand-ed' ones which contained only the image references.  
If a bundle contained properties, it was presumed to be of the completely-rendered variety and just accumulated (i.e. the author was responsible for the completeness of the rendering). 
If a bundle was without properties, it was presumed to be a shorthand bundle, and we would render it ourselves and accumulate it to the output. 

I don't love the idea of refusing to accept any completely-rendered bundles in the veneer, but there is some question of why any author would choose to mix-and-match in that fashion anyway.

alpha/declcfg/load.go
grokspawn, line 121: What was the benefit of the rename?  I may be missing some background but `readYAMLOrJSON`, while verbose, does tell us what it does while `LoadReader` is more vague. 

Reading between the lines ... is it because this interface was not exported?

alpha/veneer/basic/basic.go
```diff
@@ -15,30 +14,8 @@ type Veneer struct {
...
-	if err != nil {
-		return nil, err
-	}
-	// xform to break apart dir/file elements
-	rpath, fname := filepath.Split(abspath)
-	root := os.DirFS(rpath)
-
-	cfg, err := declcfg.LoadFile(root, fname)
+func (v Veneer) Render(ctx context.Context, reader io.Reader) (*declcfg.DeclarativeConfig, error) {
+	cfg, err := declcfg.LoadReader(reader)
```
grokspawn, line 18: We were originally using declcfg.LoadFile here because we wanted the follow-on call to readBundleObjects.  Do we no longer want that?

alpha/veneer/basic/basic.go
joelanford: My perspective here is to make things as strict and locked down as possible at first so that we can attempt to keep the scope of what we need to support small. We can always open things up to support that use case in the future, but it we support it now, we're signing up for it now and we might not understand the ways that people might use/abuse it.

alpha/veneer/basic/basic.go
joelanford: `LoadFile` doesn't work with an arbitrary `io.Reader` because `readBundleObjects` needs to have some awareness of the directory in which the file containing the bundle object was found. That way it can de-reference those bundle object paths, using that directory as the root.

alpha/declcfg/load.go
oceanc80: I guess the reason I'd lean towards reader vs yaml or json is just because I feel like I'd be more likely to assume that readYAMLOrJSON would take yaml or json as an input vs a reader, but I'm also not super wedded to it.

joelanford: IMO, its because we're going to export it, and since we're exporting it, we're following the pattern of the other exported functions that return `(declcfg.DeclarativeConfig, error)`

alpha/veneer/basic/basic.go
```diff
@@ -52,8 +29,7 @@ func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.Declarativ
 
 	for _, b := range cfg.Bundles {
 		if !isBundleVeneer(&b) {
-			outb = append(outb, b)
-			continue
+			return nil, fmt.Errorf("bundle veneers must define only a schema and an image; all other fields must not be defined")
```
joelanford, line 32: I think I would just say something more generic, something like: "unexpected fields present in basic veneer bundle"

"all other fields must not be defined" gives the impression that there _are_ other fields that should be left empty. In my mind, that's probably the wrong message. Basic veneer is its own format that just happens to look a bit like FBC.

cmd/opm/alpha/veneer/basic.go
```diff
@@ -21,10 +21,19 @@ func newBasicVeneerRenderCmd() *cobra.Command {
 	)
 	cmd := &cobra.Command{
 		Use:   "basic basic-veneer-file",
-		Short: "Generate a declarative config blob from a single 'basic veneer' file",
-		Long:  `Generate a declarative config blob from a single 'basic veneer' file, typified as a declarative configuration file where olm.bundle objects have no properties`,
-		Args:  cobra.ExactArgs(1),
+		Short: "Generate a declarative config blob from a single 'basic veneer' file \nWhen FILE is '-' or not provided, the veneer is read from standard input",
+		Long:  "Generate a declarative config blob from a single 'basic veneer' file, typified as a declarative configuration file where olm.bundle objects have no properties \nWhen FILE is '-' or not provided, the veneer is read from standard input",
```
joelanford, line 25: Two in one suggestion:
1. You can use backticks and actual newlines in the code (rather than escape newline characters), which can help with readability.
2. Drop the mention of FBC, minus bundle properties. re, my previous comment.

```suggestion
		Long:  `Generate a declarative config blob from a single 'basic veneer' file.
		
When FILE is '-' or not provided, the veneer is read from standard input.`,
```

alpha/veneer/basic/basic.go
joelanford: I say this because it gives us more flexibility in the future if we define the basic veneer input file separately from FBC. The two formats don't necessarily have to evolve in lock-step.

cmd/opm/alpha/veneer/basic.go
```diff
@@ -21,10 +21,19 @@ func newBasicVeneerRenderCmd() *cobra.Command {
...
+		Args:  cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
+			// Handle different input argument types
+			// When no arguments or "-" is passed to the command,
+			// assume input is coming from stdin
+			// Otherwise open the file passed to the command
+			data, source, err := OpenFileOrReadStdin(cmd, args)
+			if err != nil {
+				log.Fatalf("unable to read %q: %v", source, err)
```
joelanford, line 34: couple of nits here:
1. we're not reading stdin yet, we're just opening it for reading.
2. do we need `OpenFileOrReadStdin` to be consumed outside this package? If not, make it unexported.

```suggestion
			data, source, err := OpenFileOrStdin(cmd, args)
			if err != nil {
				log.Fatalf("unable to open %q: %v", source, err)
```

alpha/veneer/basic/basic.go
grokspawn: I lean heavily toward getting the basic veneer on its own schema, so we could do `opm alpha render-veneer` instead of `opm alpha render-veneer [basic|semver]`.

cmd/opm/alpha/veneer/cmd.go
```diff
@@ -16,3 +19,18 @@ func NewCmd() *cobra.Command {
...
+	var reader io.ReadCloser
+
+	if len(args) == 0 || args[0] == "-" {
+		reader = io.NopCloser(cmd.InOrStdin())
+		source = "stdin"
+	} else {
+		reader, err = os.Open(args[0])
+		source = args[0]
+	}
+	return reader, source, err
```
joelanford, line 35: Nit: might be able to shorten this a bit, but take it or leave it depending on your feel for readability.

```suggestion
	if len(args) == 0 || args[0] == "-" {
		return io.NopCloser(cmd.InOrStdin()), "stdin", nil
	}
	file, err := os.Open(args[0])
	return file, args[0], err
```

cmd/opm/alpha/veneer/basic.go
```diff
@@ -21,10 +21,21 @@ func newBasicVeneerRenderCmd() *cobra.Command {
 	)
 	cmd := &cobra.Command{
 		Use:   "basic basic-veneer-file",
-		Short: "Generate a declarative config blob from a single 'basic veneer' file",
-		Long:  `Generate a declarative config blob from a single 'basic veneer' file, typified as a declarative configuration file where olm.bundle objects have no properties`,
-		Args:  cobra.ExactArgs(1),
+		Short: `Generate a declarative config blob from a single 'basic veneer' file
+When FILE is '-' or not provided, the veneer is read from standard input`,
+		Long:  `Generate a declarative config blob from a single 'basic veneer' file
+When FILE is '-' or not provided, the veneer is read from standard input`,
```
grokspawn, line 27: It would be nice if -- since we're touching this file -- we could be consistent across both supported veneers.  There were many objections to "blob" for semver, and what we settled on was something like this:
```suggestion
		Short: `Generate a file-based catalog from a single 'basic veneer' file
When FILE is '-' or not provided, the veneer is read from standard input`,
		Long:  `Generate a file-based catalog from a single 'basic veneer' file
When FILE is '-' or not provided, the veneer is read from standard input`,
```

cmd/opm/alpha/veneer/cmd.go
```diff
@@ -16,3 +19,11 @@ func NewCmd() *cobra.Command {
...
 	return runCmd
 }
+
+func openFileOrStdin(cmd *cobra.Command, args []string) (io.ReadCloser, string, error) {
+	if len(args) == 0 || args[0] == "-" {
+		return io.NopCloser(cmd.InOrStdin()), "stdin", nil
+	}
+	reader, err := os.Open(args[0])
+	return reader, args[0], err
+}
```
grokspawn, line 29: yay, re-use!

grokspawn (approved):

alpha/veneer/basic/basic.go
```diff
@@ -67,7 +43,8 @@ func (v Veneer) renderFile(ctx context.Context, ref string) (*declcfg.Declarativ
 	return cfg, nil
 }
 
-// isBundleVeneer identifies loaded partial Bundle data from YAML/JSON veneer source as having no properties,
+// isBundleVeneer identifies a Bundle veneer source as having a Schema and Image defined
+// but no Properties, RelatedImages or Package defined
 func isBundleVeneer(b *declcfg.Bundle) bool {
-	return len(b.Properties) == 0
+	return b.Schema != "" && b.Image != "" && b.Package == "" && len(b.Properties) == 0 && len(b.RelatedImages) == 0
```
ankitathomas, line 49: if package isn't defined, wouldn't that limit render-veneer to one package at a time? Are we counting on the bundle to declare its package, and if so, does that limit a bundle to only one package?

ankitathomas, line 49: We should have the field restrictions for the simple veneer documented, specifically the hard requirement for prohibited fields. The error message isn't very helpful here.

ankitathomas (approved):
/lgtm

PR #1032, opened 2022-10-17
change to log image ref value when resolving `name`

acornett21:
Updating error log when calling `Resolve` method to log `ref.String()` since if there is an error `name` will be `nil` and the logging an empty value is not helpful for troubleshooting.

When running `opm add index` when `Resolve` fails, since `name` is nil the log message does not indicate which registry there is an issue with. Which leads to trying to trouble shoot is it the registry where the index resides, or the registry where the bundle exists (since in alot of cases these are different).

**Reviewer Checklist**

### DIFF
--- a/alpha/action/list_test.go
+++ b/alpha/action/list_test.go
@@ -28,7 +28,7 @@ foo   Foo Operator  beta
 		{
 			name:        "Error/UnknownIndex",
 			list:        ListPackages{IndexReference: "unknown-index"},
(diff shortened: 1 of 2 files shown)


grokspawn:
Could you also add a unit test which captures the scenario where resolution fails?  It feels like an important edge case.

acornett21:
> Could you also add a unit test which captures the scenario where resolution fails? It feels like an important edge case.

@grokspawn It looks like there were already test cases for this, I was just unsure on how to run them. They are now updated and passing locally.

PR #1040, opened 2022-11-14
Unmarshall error details

grokspawn:
Capture JSON unmarshal type violations in greater detail to help authors spot the issue. 

for e.g., when we set a properties type entry as type 4 instead of a string, we get a more descriptive blob to help us spot the violation:

```shell
./bin/opm render -o yaml /tmp/community-operator-index-4.12
2022/11/14 14:49:08 render reference "/tmp/community-operator-index-4.12": unmarshal error: json: cannot unmarshal number into Go struct field Property.properties.type of type string at offset 230 (indicated by <==)
 {"image":"quay.io/openshift-community-operators/zookeeper-operator@sha256:c55e04596b97a1571675d59a7594e6c28f819c5ba1b2907561219ef334f50002","name":"zookeeper-operator.v0.10.3","package":"zookeeper-operator","properties":[{"type":4 <==
 ```

Solves #1039 

**Reviewer Checklist**

### CODE REVIEW

File: alpha/declcfg/load.go [hunk 1]
```diff
@@ -186,3 +186,25 @@ func LoadFile(root fs.FS, path string) (*DeclarativeConfig, error) {
...
+	if e, ok := err.(*json.UnmarshalFieldError); ok {
+		return e.Error()
+	}
+	if e, ok := err.(*json.InvalidUnmarshalError); ok {
+		return e.Error()
+	}
+	if e, ok := err.(*json.SyntaxError); ok {
+		return e.Error()
+	}
+	return err.Error()
```
joelanford, line 209: All the error type assertions that end up returning `.Error()` could be condensed down to `return err.Error()`, right?

(Also for future reference, it's typically preferential to use `errors.As` in cases like this, because it means you can still get to the error you're after even if some code in between you and the error site has wrapped it).

grokspawn: This was reviewer-bait.  I was hoping that someone would be opinionated about some of these other error types.  Otherwise, you're right ... they're just a series of "why didn't you just do the single fallthrough for all them" instead?  😄

grokspawn: Thanks for the  reviews, guys!  
I updated the code to support both of the errors we're initially interested in, and also reformatted the output to make it more human-readable.  
The error scenario reflected in the PR description now displays as:

```shell
./bin/opm render -o yaml /tmp/community-operator-index-4.12
2022/11/15 16:00:38 render reference "/tmp/community-operator-index-4.12": unmarshal error: json: cannot unmarshal number into Go struct field Property.properties.type of type string at offset 230 (indicated by <==)
 {
    "image": "quay.io/openshift-community-operators/zookeeper-operator@sha256:c55e04596b97a1571675d59a7594e6c28f819c5ba1b2907561219ef334f50002",
    "name": "zookeeper-operator.v0.10.3",
    "package": "zookeeper-operator",
    "properties": [
        {
            "type": 4 <== ,
            "value": {
                "group": "zookeeper.streamnative.io",
                "kind": "ZooKeeperCluster",
                "version": "v1alpha1"
            }
        },
        {
            "type": "olm.package",
            "value": {
                "packageName": "zookeeper-operator",
                "version": "0.10.3"
            }
        }
    ],
    "relatedImages": [
        {
            "image": "docker.cloudsmith.io/streamnative/operators/zookeeper-operator:v0.10.3",
            "name": ""
        },
        {
            "image": "gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0",
            "name": ""
        },
        {
            "image": "quay.io/openshift-community-operators/zookeeper-operator@sha256:c55e04596b97a1571675d59a7594e6c28f819c5ba1b2907561219ef334f50002",
            "name": ""
        }
    ],
    "schema": "olm.bundle"
}
```

everettraven (approved):
Looks good to me!

PR #1049, opened 2023-01-03
fix: don't panic when closing registry querier

joelanford:

This commit refactors the `registry.Querier` load behavior such that:
1. If any error occurs creating a `Querier`, any necessary cleanup happens and a nil `Querier` is returned.
2. `Querier.Close` checks for a nil cache before attempting a close of the `Querier`'s underlying cache.

When [certain errors](https://github.com/operator-framework/operator-registry/blob/249ae621bb8fa6fc8a8e4a5ae26355577393f127/pkg/registry/query.go#L414-L416) occur loading a `registry.Querier`, it is possible for a `Querier` to be returned with a nil cache. This causes a panic when `Close` is later called on the `Querier`.

**Reviewer Checklist**


joelanford:
Closing this one out. I'm taking at stab at #1050, and that work renders this change obsolete.

PR #1051, opened 2023-01-16
Refactor FBC caching

### LINKED ISSUE #1050: RFE:  enhance error handling leading to catalog cache rebuilding
When a catalog cache's digest is not readable (file does not exist, or permissions prevent reading) or when the digest does not match the locally-calculated digest, `opm` will try to regenerate the cache.  In typical published catalog images with a pre-generated cache, the cache directory is deliberately lacking delete permissions. 

This produces messaging like:

```shell
2023-01-03T08:42:27.514784820Z Error: compute digest: compute hash: write tar: open /tmp/cache/cache: permission denied
2023-01-03T08:42:27.515439730Z Usage:
2023-01-03T08:42:27.515439730Z   opm serve <source_path> [flags]
```

The message could be improved by distinguishing between:
1. [failure to create cache directory](https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/query.go#L443)
2. [failure to read digest](https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/query.go#L447), and [here](https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/query.go#L454)
3. other failure modes in the `load` method related to digest mismatch, recompute errors

In addition, it would be extremely useful to provide a new flag to `opm serve` which would enforce process exit on cache load error, instead of attempts to regenerate the cache.  This would allow us to align flags explicitly with the specific "no updates allowed" use-case.

joelanford:
This commit improves maintainability and extensibility of the FBC caching used to serve the GRPC API.

It:
- introduces a new cache subpackage and interface, which defines primitives for checking cache integrity, building the cache, and loading the cache.
- moves the existing registry.Querier to an implementation of the cache interface, called cache.JSON

This refactor also resolves two outstanding issues:
1. The current code contains a bug that causes panics when certain failures occur loading the cache.
2. The current code is hardcoded to always rebuild the cache if it is unreadable or its digest doesn't contain the expected value.

This commit:
1. Refactors and removes the cases the led to the panic situation
2. Adds a new flag, --cache-enforce-integrity, that causes `opm serve` to exit with a failure if the cache is unreadable or its digest doesn't contain the expected value. This is useful in production contexts when one always expects the cache to be pre-populated correctly.

Closes #1049
Closes #1050 

**Reviewer Checklist**


joelanford:
Updated coverage report is here: https://app.codecov.io/github/operator-framework/operator-registry/commit/712627b1d32deeeb29773629724dbb785321d68d

Trying to track down what I did to make coverage go down... 🤔

everettraven (approved):
These changes look good and seem reasonable to me. I like the approach of making the cache an interface which will make it nice if we ever feel we need to extend to support different types of caches.

PR #1054, opened 2023-01-17
Add composite veneer to `opm alpha render-veneer ...`

everettraven:
- Adds new subcommand `opm alpha render-veneer composite [FILE] [OPTIONS]`
- Adds a new `composite` package that adds:
    - A new composite veneer type
    - Builders for building different veneer types in the composite configuration
    - Schema for catalog configuration
    - Schema for catalog component configuration
    - Unit tests

**Reviewer Checklist**

**Note for Reviewers**:
- There are quite a few TODO comments sprinkled around with some questions - please feel free to include your inputs on these questions as part of your review!

### DIFF
--- a/alpha/veneer/composite/builder.go
+++ b/alpha/veneer/composite/builder.go
@@ -1,0 +1,181 @@
+package composite
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/veneer/basic"
+	"github.com/operator-framework/operator-registry/alpha/veneer/semver"
+)
+
+const (
+	BasicVeneerBuilderSchema  = "olm.veneer.basic"
+	SemverVeneerBuilderSchema = "olm.veneer.semver"
+	RawVeneerBuilderSchema    = "olm.veneer.raw"
+)
+
+type ContainerConfig struct {
+	ContainerTool string
+	BaseImage     string
+	WorkingDir    string
+}
+
+// TODO: update this to use docker ...
+type Builder interface {
+	Build(ctx context.Context, vd VeneerDef, containerCfg ContainerConfig) (*declcfg.DeclarativeConfig, string, error)
+}
+
+type BasicBuilder struct {
+	veneer basic.Veneer
+}
+
+var _ Builder = &BasicBuilder{}
+
+func NewBasicBuilder() *BasicBuilder {
+	return &BasicBuilder{
+		veneer: basic.Veneer{},
+	}
+}
+
+func (bb *BasicBuilder) Build(ctx context.Context, vd VeneerDef, containerCfg ContainerConfig) (*declcfg.DeclarativeConfig, string, error) {
+	// Parse out the basic veneer configuration
+	basicConfig := &BasicVeneerConfig{}
+	err := json.Unmarshal(vd.Config, basicConfig)
+	if err != nil {
+		return nil, "", fmt.Errorf("unmarshalling basic veneer config: %w", err)
+	}
+
+	// get the current working directory
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, "", fmt.Errorf("getting current working directory: %w", err)
+	}
+
+	// build the container command
+	containerCmd := exec.Command(containerCfg.ContainerTool,
+		"run",
+		"--rm",
+		"-v",
+		fmt.Sprintf("%s:%s", wd, containerCfg.WorkingDir),
+		containerCfg.BaseImage,
+		"alpha",
+		"render-veneer",
+		"basic",
+		path.Join(containerCfg.WorkingDir, basicConfig.Input))
+
+	out, err := containerCmd.Output()
+	if err != nil {
+		return nil, "", fmt.Errorf("running command %q | STDERR: %s", containerCmd.String(), err.(*exec.ExitError).Stderr)
+	}
+
+	// parse out to dcfg
+	dcfg, err := declcfg.LoadReader(bytes.NewReader(out))
+	if err != nil {
+		return nil, "", fmt.Errorf("parsing basic veneer render output: %w", err)
+	}
+
+	return dcfg, basicConfig.Output, nil
+}
+
+type SemverBuilder struct {
+	veneer semver.Veneer
+}
+
+var _ Builder = &SemverBuilder{}
+
+func NewSemverBuilder() *SemverBuilder {
+	return &SemverBuilder{
+		veneer: semver.Veneer{},
+	}
+}
+
+func (sb *SemverBuilder) Build(ctx context.Context, vd VeneerDef, containerCfg ContainerConfig) (*declcfg.DeclarativeConfig, string, error) {
+	// Parse out the semver veneer configuration
+	semverConfig := &SemverVeneerConfig{}
+	err := json.Unmarshal(vd.Config, semverConfig)
+	if err != nil {
+		return nil, "", fmt.Errorf("unmarshalling semver veneer config: %w", err)
+	}
+
+	// get the current working directory
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, "", fmt.Errorf("getting current working directory: %w", err)
+	}
+
+	// build the container command
+	containerCmd := exec.Command(containerCfg.ContainerTool,
+		"run",
+		"--rm",
+		"-v",
+		fmt.Sprintf("%s:%s", wd, containerCfg.WorkingDir),
+		containerCfg.BaseImage,
+		"alpha",
+		"render-veneer",
+		"semver",
+		path.Join(containerCfg.WorkingDir, semverConfig.Input))
+
+	out, err := containerCmd.Output()
+	if err != nil {
+		return nil, "", fmt.Errorf("running command %q | STDERR: %s", containerCmd.String(), err.(*exec.ExitError).Stderr)
+	}
+
+	// parse out to dcfg
+	dcfg, err := declcfg.LoadReader(bytes.NewReader(out))
+	if err != nil {
+		return nil, "", fmt.Errorf("parsing semver veneer render output: %w", err)
+	}
+
+	return dcfg, semverConfig.Output, nil
+}
+
+type RawBuilder struct{}
+
+var _ Builder = &RawBuilder{}
+
+func NewRawBuilder() *RawBuilder {
+	return &RawBuilder{}
+}
+
+func (rb *RawBuilder) Build(ctx context.Context, vd VeneerDef, containerCfg ContainerConfig) (*declcfg.DeclarativeConfig, string, error) {
+	// Parse out the basic veneer configuration
+	rawConfig := &RawVeneerConfig{}
+	err := json.Unmarshal(vd.Config, rawConfig)
+	if err != nil {
+		return nil, "", fmt.Errorf("unmarshalling raw veneer config: %w", err)
+	}
+	// get the current working directory
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, "", fmt.Errorf("getting current working directory: %w", err)
+	}
+
+	// build the container command
+	containerCmd := exec.Command(containerCfg.ContainerTool,
+		"run",
+		"--rm",
+		"-v",
+		fmt.Sprintf("%s:%s", wd, containerCfg.WorkingDir),
+		"--entrypoint=cat", // This assumes that the `cat` command is available in the container -- Should we also build a `... render-veneer raw` command to ensure consistent operation? Does OPM already have a way to render a raw FBC?
+		containerCfg.BaseImage,
+		path.Join(containerCfg.WorkingDir, rawConfig.Input))
+
+	out, err := containerCmd.Output()
+	if err != nil {
+		return nil, "", fmt.Errorf("running command %q | STDERR: %s", containerCmd.String(), err.(*exec.ExitError).Stderr)
+	}
+
+	// parse out to dcfg
+	dcfg, err := declcfg.LoadReader(bytes.NewReader(out))
+	if err != nil {
+		return nil, "", fmt.Errorf("parsing raw veneer render output: %w", err)
+	}
+
+	return dcfg, rawConfig.Output, nil
+}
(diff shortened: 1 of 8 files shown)


alpha/veneer/composite/builder.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+func (bb *BasicBuilder) Build(dir string, vd VeneerDefinition) error {
+	// Parse out the basic veneer configuration
+	basicConfig := &BasicVeneerConfig{}
+	err := json.Unmarshal(vd.Config, basicConfig)
+	if err != nil {
+		return fmt.Errorf("unmarshalling basic veneer config: %w", err)
+	}
+	// TODO: should we ensure that the schema matches and error if it does not?
```
grokspawn, line 57: Yes, please.  We should in general always verify that the input schema matches our expectations w.r.t. type and component fields, as well as erroring on any unexpected fields, like in semver.go:
```
	if err := yaml.Unmarshal(data, &sv, func(decoder *json.Decoder) *json.Decoder {
		decoder.DisallowUnknownFields()
		return decoder
	}); err != nil {
		return nil, err
	}
	if sv.Schema != schema {
		return nil, fmt.Errorf("readFile: input file has unknown schema, should be %q", schema)
	}
```

alpha/veneer/composite/builder.go
```diff
@@ -0,0 +1,287 @@
...
+		"--rm",
+		"-v",
+		fmt.Sprintf("%s:%s", wd, bb.builderCfg.ContainerCfg.WorkingDir),
+		bb.builderCfg.ContainerCfg.BaseImage,
+		"alpha",
+		"render-veneer",
+		"basic",
+		path.Join(bb.builderCfg.ContainerCfg.WorkingDir, basicConfig.Input))
+
+	return build(containerCmd, path.Join(dir, basicConfig.Output), bb.builderCfg.OutputType)
```
grokspawn, line 77: Do we panic if the bb.* fields are not present in the file?  Do we need to verify that we have valid values (at a minimum, not empty-string) before we use them?

cmd/opm/alpha/veneer/composite.go
```diff
@@ -0,0 +1,128 @@
...
+		validate      bool
+		configFile    string
+	)
+	cmd := &cobra.Command{
+		Use: "composite composite-veneer-file",
+		Short: `Generate a file-based catalog from a single 'composite veneer' file
+When FILE is '-' or not provided, the veneer is read from standard input`,
+		Long: `Generate a file-based catalog from a single 'composite veneer' file
+When FILE is '-' or not provided, the veneer is read from standard input`,
+		Args: cobra.MaximumNArgs(1),
```
grokspawn, line 29: From the doc:
>opm alpha render-veneer composite -f catalog/config.yaml -c .../catalogs.yaml

We need to have two inputs to this process:
1. the composite veneer itself
2. the catalogs config to define valid builders, source opm environment, and working dir for generated artifacts

So wouldn't we have two inputs and `cobra.MinimumNArgs(2)`?

everettraven: I don't think so. The way I have it setup is to run like so:
```
opm alpha render-veneer composite catalogs.yaml -c catalog/config.yaml
```
The catalogs.yaml file is fed in via a positional argument and config.yaml file can be specified by the `--composite-config, -c` flag with it's default value being to look for `catalog/config.yaml`

alpha/veneer/composite/builder.go
everettraven: It shouldn't panic because they aren't pointers but running the command would likely fail if it was an empty string. I could probably update the structs to have some kind of json/yaml parsing tags that requires fields to be present or else it errors when unmarshalling.

If the command fails for any reason it will return an error here: https://github.com/operator-framework/operator-registry/pull/1054/files#diff-932fc031fb99963dd1cedfe8ae41f2c1531dcfedba6e36f2c396e28e7403ba21R263-R266

cmd/opm/alpha/veneer/composite.go
grokspawn: Right, and the alternative format of this can use the input redirect as 
`opm alpha render-veneer composite - -c catalog/config.yaml <catalogs.yaml`
which gets a little confusing.  
Isn't it clearer to drop the pipe capability since we have two mandatory inputs and no way of knowing which one the user will want to pipe in?

I got thrown initially by the use of the positional vs. enumerated arg passing.

everettraven: I guess that makes sense but I would suspect that someone wanting to do a form of input redirection may end up wanting to organize their command a bit more like:
```
opm alpha render-veneer composite -c catalog/config.yaml - <catalogs.yaml
```
(i.e organize it such that all the flags are declared first and then do the file or input redirect at the end)

I'm also not against just forcing the use of flags and files and setting default values for the flags if that is what we think is the best UX flow. I was just trying to keep this as similar to the other `opm alpha render-veneer ...` subcommands as possible

> Isn't it clearer to drop the pipe capability since we have two mandatory inputs and no way of knowing which one the user will want to pipe in?
This is also a very good point with the perspective that the current implementation would only let you pipe the catalogs.yaml file and not the catalog/config.yaml equivalent.

everettraven: Went ahead and updated the implementation for this in commit https://github.com/operator-framework/operator-registry/pull/1054/commits/606acd5eef65c62ae77a29e0eddaaf55b8332a86

everettraven: However I forgot to change the args :facepalm:

alpha/veneer/composite/builder.go
everettraven: Done in commit https://github.com/operator-framework/operator-registry/pull/1054/commits/606acd5eef65c62ae77a29e0eddaaf55b8332a86

alpha/veneer/composite/builder.go
everettraven: So I took a brief look into this and I was correct that running the command ends up failing. I also did some research on Go struct tags and it doesn't look like there is an easy way to specify that a field is required without adding additional dependencies and extra struct tags.

What I am thinking to do is just add some checks that will fail fast if the WorkingDir and BaseImage fields are empty strings prior to the creation of a builder in this loop:
https://github.com/operator-framework/operator-registry/pull/1054/files#diff-f0ad1e448a4ece45a9aa1e9e28bafc35d5c88ffd8517dadbe2a679baa02ba89eR58-R78

joelanford: fail fast is good as a first pass for sure. but something to keep in mind from a UX perspective. if someone has multiple errors in their input file, its nice to tell them about all of them rather than just the first one you encounter.

alpha/veneer/composite/builder.go
everettraven: Added even better validations for everything in commit https://github.com/operator-framework/operator-registry/pull/1054/commits/f33598f9912c7e2cfa9d55ec79d54c0fd174e1b3

alpha/veneer/composite/builder.go
everettraven: Added validations and instead of failing fast I created the validations in such a way that will check all of the fields before failing and append them to an error message.

All this and the respective test cases were added in https://github.com/operator-framework/operator-registry/pull/1054/commits/f33598f9912c7e2cfa9d55ec79d54c0fd174e1b3

cmd/opm/alpha/veneer/composite.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+
+			err = veneer.Render(cmd.Context(), compositeConfig, validate)
+			if err != nil {
+				log.Fatalf("rendering the composite veneer: %s", err)
+			}
+		},
+	}
+	cmd.Flags().StringVarP(&output, "output", "o", "json", "Output format (json|yaml)")
+	// TODO: Investigate ways to do this without using a cli tool like docker/podman
```
grokspawn, line 148: I was going to opine that we should allow use of (and with the same precedence ordering, default, etc.) as other `opm` sub-commands, but I think for this first pass I'd rather if we removed the option to specify (just comment out, don't delete) and always defaulted to docker.

alpha/veneer/composite/builder.go
```diff
@@ -0,0 +1,317 @@
...
+		return fmt.Errorf("unmarshalling custom veneer config: %w", err)
+	}
+
+	// validate the custom config fields
+	if customConfig.Command == "" {
+		return errors.New("custom veneer configuration is invalid: custom veneer config must have a non-empty command (veneerDefinition.config.command)")
+	}
+
+	// build the command to execute
+	// TODO: should the command be run within the container?
```
grokspawn, line 245: The intent here was just to provide an 'escape hatch' for custom processing.  I think it's totally safe for now to assume that it runs locally.

alpha/veneer/composite/builder.go
```diff
@@ -0,0 +1,317 @@
...
+	cmd := exec.Command(customConfig.Command, customConfig.Args...)
+	cmd.Dir = cb.builderCfg.CurrentDirectory
+
+	// TODO: Should we capture the output here for any reason?
```
grokspawn, line 249: I think just capturing the exit status is sufficient for now.

alpha/veneer/composite/builder.go
```diff
@@ -0,0 +1,317 @@
...
+	// Should the custom veneer output an FBC to STDOUT like the other veneer outputs?
```
grokspawn, line 250: IMHO, we should assume that the custom approach to produce valid FBC in the target destination.  I wouldn't even attempt to capture any files or move anything around.  If folks want to use the custom approach, they really need to be responsible for everything we'd otherwise do.

alpha/veneer/composite/composite.go
```diff
@@ -0,0 +1,48 @@
...
+				}
+
+				if validate {
+					// run the validation for the builder
+					err = builder.Validate(component.Destination.Path)
+					if err != nil {
+						return fmt.Errorf("validating component %q: %w", component.Name, err)
+					}
+				}
+				// TODO(everettraven): Should we remove the built FBC if validation fails?
```
grokspawn, line 35: If we presume that the failure reporting is useful, then IMHO don't delete the erroneous FBC.  We don't error if they overwrite it.

alpha/veneer/composite/builder.go
joelanford: Does the custom script even know what the configured destination is?  Seems like we'd have to feed the destination is an an envvar or something.

Another option is that we say "if exit status is 0, stdout is streamed into a single catalog file in the configured destination". I sort of like this idea because it means that it follows the pattern of other veneers (e.g. output stuff to stdout and let the caller decide what to do with it).

everettraven: > I sort of like this idea because it means that it follows the pattern of other veneers (e.g. output stuff to stdout and let the caller decide what to do with it).

@joelanford This is the reason why I wrote that comment, so that all the builders process everything the exact same way :).

> Does the custom script even know what the configured destination is? Seems like we'd have to feed the destination is an an envvar or something.

Nope, in my demo composite repo I show an example that pipes output to a particular file: https://github.com/everettraven/fbc-composite-example/blob/107df84f1618d95905f32d6708062409c5e2cb3d/catalog/config.yaml#L33-L44

That being said, we could update the custom config spec to require defining an output location and then we process it just like all the other builders.

alpha/veneer/composite/builder.go
everettraven: Sgtm, removed TODO comment in commit https://github.com/operator-framework/operator-registry/pull/1054/commits/bfde6b3074cc7e7a083048f085c2aac7e22c4c70

cmd/opm/alpha/veneer/composite.go
everettraven: Fair enough, done in commit https://github.com/operator-framework/operator-registry/pull/1054/commits/bfde6b3074cc7e7a083048f085c2aac7e22c4c70

alpha/veneer/composite/builder.go
everettraven: Updated the custom veneer builder to work the same as all the other builders in commit https://github.com/operator-framework/operator-registry/pull/1054/commits/a9e6f68adc310362ea221b29fe56ae7047c98107

alpha/veneer/composite/builder.go
everettraven: https://github.com/operator-framework/operator-registry/pull/1054#discussion_r1082628767

grokspawn (approved):

cmd/opm/alpha/veneer/composite.go
```diff
@@ -0,0 +1,173 @@
...
+	// cmd.Flags().StringVar(&containerTool, "container-tool", "docker", "container tool to be used when rendering veneers (should be an equivalent replacement to docker - similar to podman)")
+	cmd.Flags().BoolVar(&validate, "validate", true, "whether or not the created FBC should be validated (i.e 'opm validate')")
+	cmd.Flags().StringVarP(&compositeFile, "composite-config", "c", "catalog/config.yaml", "File to use as the composite configuration file")
+	cmd.Flags().StringVarP(&catalogFile, "catalog-config", "f", "catalogs.yaml", "File to use as the catalog configuration file")
+	return cmd
```
gallettilance, line 154: could we add a flag to account for a previous catalog generation so we dont have to pull all images again? or an overwrite option for the output in case it's non empty so we can warn users if they make a veneer change that would strand users on a particular version for example?

cmd/opm/alpha/veneer/composite.go
gallettilance: i remember iib having issues with docker - can we check with them that omitting podman is ok?

gallettilance (approved):
/lgtm - just one comment maybe for a follow up.

cmd/opm/alpha/veneer/composite.go
grokspawn: Are you thinking about the use-case where we want to make incremental updates only to an existing catalog?  According to the document (https://docs.google.com/document/d/1uEZdVA06GGahPl3ygnGrFLRkrohaQjs25mVcXToYakE/edit#heading=h.kjc5gjiy2pkd) SOP will be to overwrite the specific contribution in all cases. 

I'd be interested in talking more about the use-case, but def not barring this PR merge.

cmd/opm/alpha/veneer/composite.go
anik120: The ideal thing to do here will be to have the containerd implementation as the underlying component, that way we don't have dependency on either docker/podman/etc. 

But as long as it's called out here as the "must do before graduating from alpha" it doesn't need to block this particular PR.

PR #1060, opened 2023-01-26
use mount flag to force consistent selinux labelling between host/guest on generated files

grokspawn:
Signed-off-by: Jordan <jordan@nimblewidget.com>

without this flag, there is inconsistent labeling between intermediate composite template products on the host and the container which will (read-only) access them. 
When selinux is in enforcing mode, these steps will fail to execute. 
For more info, https://projectatomic.io/blog/2015/06/using-volumes-with-docker-can-cause-problems-with-selinux/

**Reviewer Checklist**


everettraven (approved):
Changes look good to me. Thanks for digging into this!

PR #1063, opened 2023-02-17
OPECO-2646: exclude bundles with `olm.deprecated` property when rendering

rashmigottipati:

- Previous bundle deprecation was handled by assigning a property to the olm.bundle object of `olm.deprecated`.  SQLite DBs had to have all valid upgrade edges supported by olm.bundle information in order to prevent foreign key violations.  This property meant that the bundle was to be ignored & never installed. FBC has a simpler method for achieving the same goal which is to not include the bundle.  Upgrade edges from it may still be specified, and the bundle will not be installable.

- Excluding bundles with the `olm.deprecated` property when rendering as this property isn't relevant for FBC, so bundles containing this property should be omitted from the catalog rendered by opm.

### DIFF
--- a/pkg/sqlite/conversion.go
+++ b/pkg/sqlite/conversion.go
@@ -76,7 +76,19 @@ func populateModelChannels(ctx context.Context, pkgs model.Model, q *SQLQuerier)
 	if err != nil {
 		return err
 	}
+	deprecatedValue, err := json.Marshal(registry.DeprecatedProperty{})
+	if err != nil {
+		return err
+	}
+ConvertBundles:
 	for _, bundle := range bundles {
+		for _, prop := range bundle.Properties {
+			if prop.Type == registry.DeprecatedType && prop.Value == string(deprecatedValue) {
+				// bundle contains `olm.Deprecated` property
+				// exclude this bundle from being rendered
+				continue ConvertBundles
+			}
+		}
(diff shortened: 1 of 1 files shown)


pkg/sqlite/conversion.go
```diff
@@ -76,7 +76,19 @@ func populateModelChannels(ctx context.Context, pkgs model.Model, q *SQLQuerier)
 	if err != nil {
 		return err
 	}
+	deprecatedValue, err := json.Marshal(registry.DeprecatedProperty{})
+	if err != nil {
+		return err
+	}
```
everettraven, line 82: Is this necessary? Could we just get away with checking the property type and ignoring the value?

pkg/sqlite/conversion.go
```diff
@@ -76,7 +76,19 @@ func populateModelChannels(ctx context.Context, pkgs model.Model, q *SQLQuerier)
...
+ConvertBundles:
 	for _, bundle := range bundles {
+		for _, prop := range bundle.Properties {
+			if prop.Type == registry.DeprecatedType && prop.Value == string(deprecatedValue) {
```
everettraven, line 86: Is it necessary to check `prop.Value`? I *think* we could probably just get away with checking for the `olm.deprecated` property type.

everettraven (approved):
LGTM!

grokspawn (approved):

PR #1066, opened 2023-02-20
OPECO-2664: rename veneer to template

grokspawn:
Renaming all references from veneer to template.  Catalog Template is now the accepted term, and this just enforces all code references to be consistent. 

Along the way, we discovered that 
- cobra command hierarchies will only exit with non-zero status if there is an unbroken chain of provided Run[E] functions from parent to child, so this includes some empty Run methods to preserve this chain
- the base image for opm to render fragments from composite template does not support the name-change, but we have to merge this PR to be able to release an image which will, so this includes temporary skips of utests which were failing due to the image dependency.

**Reviewer Checklist**


oceanc80 (approved):

alpha/template/composite/builder_test.go
```diff
@@ -17,14 +17,16 @@ func TestBasicBuilder(t *testing.T) {
...
+		templateDefinition TemplateDefinition
 		files              map[string]string
 		buildAssertions    func(t *testing.T, dir string, buildErr error)
 		validateAssertions func(t *testing.T, validateErr error)
+		skipme             bool
 	}
 
 	testCases := []testCase{
 		{
+			skipme:   true,
```
grokspawn, line 29: Adding a temporary flag to basic/semver tests which currently because they depend on an image which has the updated naming, but we cannot publish a new image until we merge this PR. 

Immediately after the PR merges we'll cut a release and create a new PR to remove this mechanism and any failures it hides.

PR #1068, opened 2023-02-21
OPECO-2664: updating image pullspec to rename-aware baseimage

grokspawn:

**Reviewer Checklist**


everettraven (approved):
It might be nice to actually have the unit tests create their own base image to use based on the current changes. It may add some additional complexity to the tests but at the advantage of not having to rely on prebuilt images.

I'm not entirely sure what changes this would entail but just some food for thought. If we don't want to worry about going this direction these changes look good to me.

PR #1075, opened 2023-03-01
OPECO-2735: initial update to avoid nested docker execution

grokspawn:

**Reviewer Checklist**

### DIFF
--- a/alpha/template/composite/builder.go
+++ b/alpha/template/composite/builder.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"sigs.k8s.io/yaml"
@@ -16,6 +17,7 @@ import (
 	basictemplate "github.com/operator-framework/operator-registry/alpha/template/basic"
 	semvertemplate "github.com/operator-framework/operator-registry/alpha/template/semver"
 	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/pkg/lib/config"
 )
 
 const (
@@ -32,9 +34,9 @@ type ContainerConfig struct {
 }
 
 type BuilderConfig struct {
-	ContainerCfg     ContainerConfig
-	OutputType       string
-	CurrentDirectory string
+	ContainerCfg   ContainerConfig
+	OutputType     string
+	InputDirectory string
 }
 
 type Builder interface {
@@ -84,23 +86,23 @@ func (bb *BasicBuilder) Build(ctx context.Context, reg image.Registry, dir strin
 
 	b := basictemplate.Template{Registry: reg}
 	reader, err := os.Open(basicConfig.Input)
-	defer reader.Close()
 	if err != nil {
 		return fmt.Errorf("error reading basic template: %v", err)
 	}
+	defer reader.Close()
 
 	dcfg, err := b.Render(ctx, reader)
 	if err != nil {
 		return fmt.Errorf("error rendering basic template: %v", err)
 	}
 
-	destPath := path.Join(bb.builderCfg.ContainerCfg.WorkingDir, basicConfig.Input)
+	destPath := path.Join(bb.builderCfg.ContainerCfg.WorkingDir, dir, basicConfig.Output)
 
 	return build(dcfg, destPath, bb.builderCfg.OutputType)
 }
 
 func (bb *BasicBuilder) Validate(dir string) error {
-	return validate(bb.builderCfg.ContainerCfg, path.Join(bb.builderCfg.CurrentDirectory, dir))
+	return validate(bb.builderCfg.ContainerCfg, dir)
 }
 
 type SemverBuilder struct {
@@ -156,13 +158,13 @@ func (sb *SemverBuilder) Build(ctx context.Context, reg image.Registry, dir stri
 		return fmt.Errorf("error rendering semver template: %v", err)
 	}
 
-	destPath := path.Join(sb.builderCfg.ContainerCfg.WorkingDir, semverConfig.Input)
+	destPath := path.Join(sb.builderCfg.ContainerCfg.WorkingDir, dir, semverConfig.Output)
 
 	return build(dcfg, destPath, sb.builderCfg.OutputType)
 }
 
 func (sb *SemverBuilder) Validate(dir string) error {
-	return validate(sb.builderCfg.ContainerCfg, path.Join(sb.builderCfg.CurrentDirectory, dir))
+	return validate(sb.builderCfg.ContainerCfg, dir)
 }
 
 type RawBuilder struct {
@@ -205,14 +207,6 @@ func (rb *RawBuilder) Build(ctx context.Context, _ image.Registry, dir string, t
 		return fmt.Errorf("raw template configuration is invalid: %s", strings.Join(validationErrs, ","))
 	}
 
-	istat, err := os.Stat(rawConfig.Input)
-	if err != nil {
-		return fmt.Errorf("raw template input file does not exist: %s, %v", rawConfig.Input, err)
-	}
-	if !istat.Mode().IsRegular() {
-		return fmt.Errorf("raw template input file is not a normal file: %s, %v", rawConfig.Input, err)
-	}
-
 	reader, err := os.Open(rawConfig.Input)
 	if err != nil {
 		return fmt.Errorf("error reading raw input file: %s, %v", rawConfig.Input, err)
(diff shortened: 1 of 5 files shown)


alpha/template/composite/builder.go
```diff
@@ -258,7 +278,23 @@ func (cb *CustomBuilder) Build(dir string, td TemplateDefinition) error {
 
 	// custom template should output a valid FBC to STDOUT so we can
 	// build the FBC just like all the other templates.
-	return build(cmd, path.Join(dir, customConfig.Output), cb.builderCfg.OutputType)
+	v, err := cmd.Output()
+	reader := bytes.NewReader(v)
+
+	dcfg, err := declcfg.LoadReader(reader)
+	cmdString := []string{customConfig.Command}
+	cmdString = append(cmdString, customConfig.Args...)
```
everettraven, line 286: Nit: I think you could this in one line by doing something like:
```suggestion
	cmdString := append([]string{customConfig.Command}, customConfig.Args...)
```

alpha/template/composite/builder.go
```diff
@@ -258,7 +278,23 @@ func (cb *CustomBuilder) Build(dir string, td TemplateDefinition) error {
...
+	}
+
+	destPath := path.Join(dir, customConfig.Output)
+
+	return build(dcfg, destPath, cb.builderCfg.OutputType)
+
+	// custom template should output a valid FBC to STDOUT so we can
+	// build the FBC just like all the other templates.
+	return build(dcfg, destPath, cb.builderCfg.OutputType)
 }
```
everettraven, line 298: Double return here

alpha/template/composite/builder.go
```diff
@@ -78,19 +82,21 @@ func (bb *BasicBuilder) Build(dir string, td TemplateDefinition) error {
...
-		fmt.Sprintf("%s:%s:Z", bb.builderCfg.CurrentDirectory, bb.builderCfg.ContainerCfg.WorkingDir),
-		bb.builderCfg.ContainerCfg.BaseImage,
-		"alpha",
-		"render-template",
-		"basic",
-		path.Join(bb.builderCfg.ContainerCfg.WorkingDir, basicConfig.Input))
-
-	return build(containerCmd, path.Join(dir, basicConfig.Output), bb.builderCfg.OutputType)
+	b := basictemplate.Template{Registry: reg}
+	reader, err := os.Open(basicConfig.Input)
```
everettraven, line 86: I think we may need to also include the current directory like we did when mounting volumes previously:
```suggestion
	reader, err := os.Open(path.Join(bb.builderCfg.CurrentDirectory, basicConfig.Input))
```

alpha/template/composite/builder.go
```diff
@@ -137,19 +143,22 @@ func (sb *SemverBuilder) Build(dir string, td TemplateDefinition) error {
...
-		"-v",
-		fmt.Sprintf("%s:%s:Z", sb.builderCfg.CurrentDirectory, sb.builderCfg.ContainerCfg.WorkingDir),
-		sb.builderCfg.ContainerCfg.BaseImage,
-		"alpha",
-		"render-template",
-		"semver",
-		path.Join(sb.builderCfg.ContainerCfg.WorkingDir, semverConfig.Input))
-
-	return build(containerCmd, path.Join(dir, semverConfig.Output), sb.builderCfg.OutputType)
+	reader, err := os.Open(semverConfig.Input)
```
everettraven, line 146: Ditto:
```suggestion
	reader, err := os.Open(path.Join(sb.builderCfg.CurrentDirectory, semverConfig.Input))
```

alpha/template/composite/builder.go
```diff
@@ -196,17 +205,28 @@ func (rb *RawBuilder) Build(dir string, td TemplateDefinition) error {
...
-		path.Join(rb.builderCfg.ContainerCfg.WorkingDir, rawConfig.Input))
+	istat, err := os.Stat(rawConfig.Input)
+	if err != nil {
+		return fmt.Errorf("raw template input file does not exist: %s, %v", rawConfig.Input, err)
+	}
+	if !istat.Mode().IsRegular() {
+		return fmt.Errorf("raw template input file is not a normal file: %s, %v", rawConfig.Input, err)
+	}
+
+	reader, err := os.Open(rawConfig.Input)
```
everettraven, line 216: Ditto:
```suggestion
	reader, err := os.Open(path.Join(rb.builderCfg.CurrentDirectory, rawConfig.Input))
```

everettraven (changes requested):

alpha/template/composite/builder.go
```diff
@@ -196,17 +205,28 @@ func (rb *RawBuilder) Build(dir string, td TemplateDefinition) error {
...
-		"--rm",
-		"-v",
-		fmt.Sprintf("%s:%s:Z", rb.builderCfg.CurrentDirectory, rb.builderCfg.ContainerCfg.WorkingDir),
-		"--entrypoint=cat", // This assumes that the `cat` command is available in the container -- Should we also build a `... render-template raw` command to ensure consistent operation? Does OPM already have a way to render a raw FBC?
-		rb.builderCfg.ContainerCfg.BaseImage,
-		path.Join(rb.builderCfg.ContainerCfg.WorkingDir, rawConfig.Input))
+	istat, err := os.Stat(rawConfig.Input)
+	if err != nil {
+		return fmt.Errorf("raw template input file does not exist: %s, %v", rawConfig.Input, err)
+	}
```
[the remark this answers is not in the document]
grokspawn: The reasoning behind this check was to check against someone putting a pipe in there and we tried to open it.  But you cannot check for a regular file if you can't stat the file, and stat returns an error, so it's just polite to handle it.

alpha/template/composite/builder.go
```diff
@@ -196,17 +205,28 @@ func (rb *RawBuilder) Build(dir string, td TemplateDefinition) error {
...
+	if !istat.Mode().IsRegular() {
+		return fmt.Errorf("raw template input file is not a normal file: %s, %v", rawConfig.Input, err)
+	}
```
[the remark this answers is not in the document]
grokspawn: It would completely fail, unless someone was super clever and provided a named pipe that they put in the dir that generated passable input schema... maybe this check isn't necessary.

alpha/template/composite/builder.go
grokspawn: I think this loses in readability what it gains in brevity, but maybe it's just my relative youth w.r.t. golang?

alpha/template/composite/builder.go
```diff
@@ -196,21 +207,26 @@ func (rb *RawBuilder) Build(dir string, td TemplateDefinition) error {
...
+
+	destPath := path.Join(dir, rawConfig.Output)
 
-	return build(containerCmd, path.Join(dir, rawConfig.Output), rb.builderCfg.OutputType)
+	return build(dcfg, destPath, rb.builderCfg.OutputType)
 }
 
 func (rb *RawBuilder) Validate(dir string) error {
-	return validate(rb.builderCfg.ContainerCfg, path.Join(rb.builderCfg.CurrentDirectory, dir))
+	return validate(rb.builderCfg.ContainerCfg, path.Join(rb.builderCfg.InputDirectory, dir))
```
grokspawn, line 229: This namechange really illustrates how moving from container-in-container to container-less really moves us in the direction of re-examining the entire schema.  

Logically, this should be `builderCfg.ContainerCfg.WorkingDir` unless `CompositeConfig.Components[].Destination.Path` (dir here) is canonical, in which case `builderCfg.InputDirectory` is superfluous.   

I think we need to eliminate the `WorkingDir` element from the schema since it only really makes sense w/in a container.  

Similar for `InputDirectory` where we really need to be able to rely on the input path from the author. 

Still nibbling on this, but might just choose to leave as-is for this PR.

grokspawn: On later thought, `WorkingDir` makes a lot of sense from the use-case from the maintainer of "I will look for contributions generated **here**".  

I think `InputDirectory` is still obsolete, though.

alpha/template/composite/builder.go
```diff
@@ -28,13 +34,13 @@ type ContainerConfig struct {
 }
 
 type BuilderConfig struct {
-	ContainerCfg     ContainerConfig
-	OutputType       string
-	CurrentDirectory string
+	ContainerCfg   ContainerConfig
```
everettraven, line 37: Nit: Ideally we would simplify this by removing the `ContainerConfig` struct and instead just defining the one field we use from it directly in the `BuilderConfig` struct.

everettraven (approved):
Looks good to me overall. Only nits I have is leaving in the `BuilderConfig.InputDirectory` field if it isn't going to be used and maybe simplifying the `BuilderConfig` struct by removing the `ContainerConfig` struct since we only use a string field from it. I'm OK with it going in like this as long as the plan is to clean it up in follow up PRs.

cmd/opm/alpha/template/composite.go
```diff
@@ -89,8 +90,9 @@ and a 'composite template' file`,
 								BaseImage:     catalog.Destination.BaseImage,
 								WorkingDir:    catalog.Destination.WorkingDir,
 							},
-							OutputType:       output,
-							CurrentDirectory: wd,
+							OutputType: output,
+							// BUGBUG: JEK: This is a strong assumption that input is always local to execution which we need to eliminate in a later PR
+							InputDirectory: wd,
```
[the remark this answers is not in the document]
grokspawn: `BUGBUG` is an old habit of mine... it's essentially the same as your `TODO(everettraven)` :) 
And like the comment says, I think we should handle streamlining the schema in a later PR.

PR #1079, opened 2023-03-22
Add Support for Fetching Remote Catalog

theishshah:
Contains implementation changes, needs utest and documentation updates.

### DIFF
--- a/cmd/opm/alpha/template/composite.go
+++ b/cmd/opm/alpha/template/composite.go
@@ -39,13 +39,13 @@ and a 'composite template' file`,
 			if err != nil {
 				tempCatalog, err = os.Open(catalogFile)
 				if err != nil {
-					log.Fatalf("opening catalog config file %q: %s", catalogFile, err)
+					log.Fatalf("opening catalog config file %q: %v", catalogFile, err)
 				}
 				defer tempCatalog.Close()
 			} else {
 				tempResp, err := http.Get(catalogURI.String())
 				if err != nil {
-					log.Fatalf("fetching remote catalog config file %q: %s", catalogFile, err)
+					log.Fatalf("fetching remote catalog config file %q: %v", catalogFile, err)
 				}
 				tempCatalog = tempResp.Body
 				defer tempCatalog.Close()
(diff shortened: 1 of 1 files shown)


cmd/opm/alpha/template/composite.go
```diff
@@ -31,11 +34,23 @@ and a 'composite template' file`,
 		Args: cobra.MaximumNArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			containerTool = "docker"
-			catalogData, err := os.Open(catalogFile)
+			var tempCatalog io.ReadCloser
```
grokspawn, line 37: Probably a stupid question w.r.t. golang and deferred calls, but is there any reason that we couldn't do 
```golang
var catalogData io.ReadCloser
```
here, and in the conditionals, assign to it/defer closure to it as needed? 

If it's imposed by the language, fine, but L53 feels like we could optimize it away with this approach.

cmd/opm/alpha/template/composite.go
```diff
@@ -31,11 +34,23 @@ and a 'composite template' file`,
...
-				log.Fatalf("opening catalog config file %q: %s", catalogFile, err)
+				tempCatalog, err = os.Open(catalogFile)
+				if err != nil {
+					log.Fatalf("opening catalog config file %q: %s", catalogFile, err)
+				}
+				defer tempCatalog.Close()
+			} else {
+				tempResp, err := http.Get(catalogURI.String())
+				if err != nil {
+					log.Fatalf("fetching remote catalog config file %q: %s", catalogFile, err)
```
grokspawn, line 48: Unrelated to this PR, per se, but one of the confusing bits of nomenclature we have in this space is the names that we use to designate the spec/files from the operator author and catalog maintainer which fulfill the roles of this conversation.  `catalog/config` and `catalogs.yaml` are too similar.  
We're effectively canonicalizing `catalog config file` here, which I think is fine, so maybe we start referring to the operator author's file as `contributions spec/file` WDYT?

everettraven: That sounds reasonable to me

theishshah: /resolve

cmd/opm/alpha/template/composite.go
theishshah: /resolve

everettraven (approved):
/lgtm

PR #1085, opened 2023-04-07
Remove vendor dir

dtfranz:
Removes the vendor directory from source control and removes the need for it during builds so we don't have to keep re-populating it. Also removes a couple of unused dockerfiles which relied on the vendor directory's presence.

Upgrades to `actions/setup-go@v4` which caches the `go mod` packages by default.

Makes the process of downstreaming easier and speeds up builds since we don't have to re-download the vendor directory and run setup-go fresh on each run.


dtfranz:
Curiously, the `goreleaser / build-windows` job gets a different hash from the `go.sum` file. Anybody know why that might be?

.github/workflows/unit.yaml
```diff
@@ -14,9 +14,20 @@ jobs:
...
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-vendor-dir
+        with:
+          path: ./vendor
+          key: go-vendor-${{ hashFiles('go.sum') }}
+          enableCrossOsArchive: true
+      - if: ${{ steps.cache-vendor.outputs.cache-hit != 'true' }}
+        name: Run make vendor
+        run: make vendor
```
grokspawn, line 30: Is this just a weird display or are L30, L31 a nesting problem?

Makefile
```diff
@@ -35,18 +35,20 @@ endif
 .PHONY: all
 all: clean test build
 
-$(CMDS):
+$(CMDS): vendor
```
grokspawn, line 38: Understanding that we're just adding a single new target, it almost seems like we should recognize the "all targets" use-case here and add a 
```make
.PHONY: environment
environment: vendor
```
and add the dependency `environment` to every target instead.  This would allow us to scale out if we discover future "for all targets" kind of things.

grokspawn:
> Curiously, the `goreleaser / build-windows` job gets a different hash from the `go.sum` file. Anybody know why that might be?

Is the hash calc OS/ARCH-dependent?

grokspawn:
Completed a review pass.  While the comments I left are more of the question/nit variety, I'm left feeling like I'd like to see some proof of concept of this before I'm comfortable merging it.  
For e.g. all the build-${OS} actions currently fail to use the cache, missing uniformly.

.github/workflows/unit.yaml
dtfranz: It just looks weird since the `make vendor` step has the `-` up above on line 28. It would probably look better if it was:
```yaml
      - name: Run make vendor
        run: make vendor
        if: ${{ steps.cache-vendor.outputs.cache-hit != 'true' }}
      - run: make unit
```

dtfranz:
> > Curiously, the `goreleaser / build-windows` job gets a different hash from the `go.sum` file. Anybody know why that might be?
> 
> Is the hash calc OS/ARCH-dependent?

Looks as though it is, based on the docs. So for our case, we would have a version of the cache for each OS.

> Completed a review pass. While the comments I left are more of the question/nit variety, I'm left feeling like I'd like to see some proof of concept of this before I'm comfortable merging it. For e.g. all the build-${OS} actions currently fail to use the cache, missing uniformly.

That's understandable, and I'm happy to prove this out through another project. It should be expected however that the first run of each build action would fail to find a cache, since all of them ran simultaneously with no previous existing cache (setup-go did not use a cache until v4). On subsequent re-runs however, all three build-${OS} actions succeeded in using the previous cache [here](https://github.com/operator-framework/operator-registry/actions/runs/4640349959).

Makefile
dtfranz: This could be coming from my lack of familiarity with the project but for the `vendor` target at least this may not be necessary, since not all targets actually require that the vendor directory be present. That may add some unnecessary churn to some of the make targets here and there.

everettraven (approved):
Changes make sense and look good to me

grokspawn:
> On subsequent re-runs however, all three build-${OS} actions succeeded in using the previous cache [here](https://github.com/operator-framework/operator-registry/actions/runs/4640349959).

That's what I was looking for.  Thanks!

dtfranz:
> Re: os-specific cache.
> 
> Here's the suggested setup for go modules, which seems to account for that difference: https://github.com/actions/cache/blob/main/examples.md#go---modules
> 
> (I'm afraid to click the diff link, so not sure if this is what you already have)

I understand your fears :smile:  I'll just paste the addition here. It's essentially the same except that it only targets the vendor dir in this case:
```yaml
      ...
      - id: cache-vendor
        uses: actions/cache@v3
        env:
          cache-name: cache-vendor-dir
        with:
          path: ./vendor
          key: ${{ runner.os }}-go-vendor-${{ hashFiles('**/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-go-vendor
      - if: ${{ steps.cache-vendor.outputs.cache-hit != 'true' }}
        name: Run make vendor
        run: make vendor
      ...
```
We can add more paths to the caching in the future if you think it's worthwhile.

m1kola (changes requested):
It seems like our job restores & saves cache 2 times: as part of `actions/setup-go` and as part of `actions/cache`:

Screenshot 2023-04-12 at 10 32 01

Do we need both?

[Documentation for `actions/setup-go`](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs) says that it has a built-in functionality for caching and restoring go modules based on `go.sum`. And it seems to be enabled by default in v4.

I haven't looked into this properly, but I assume that once cache is restored `go mod vendor` just copies everything from restored cache into vendor dir. Should give us aproximately the same results as caching & restoring vendor directory.

Is there any reason why we can not rely on cache from `actions/setup-go`?

Makefile
tmshort, line 38: Not sure why we are adding the vendor dependency here? Which seems to eventually run `go mod vendor` to fill/create the vendor directory (which is .gitignored)? That seems like quite a slowdown; especailly for those who run `git clean -fdxq` a lot.

tmshort:
My concern here is that all this does is get rid of the vendor directory; building recreates the directory. It should be possible to do this without recreating the vendor directory.

dtfranz:
> Is there any reason why we can not rely on cache from `actions/setup-go`?

I'm glad you brought this up; my understanding was that `go mod vendor` was not covered by this action, which is partially true. Running the `setup-go` action will not give you the vendor directory, which is why I had the additional cache action setup. However, you were still correct that the additional cache action is unnecessary because `setup-go` will still download or restore the required packages from cache, then when `go mod vendor` is run it will pull those packages locally from `~/go/pkg/mod` instead of downloading them again. I've updated the branch to simply use `setup-go` now.

Makefile
dtfranz: The builds for this repo are all dependent on the presence of the vendor directory and will fail without it. Running `make vendor` again when the `vendor` dir is already there is essentially a no-op since it won't need to re-download them. From my testing `go mod vendor` will not re-download the entire folder every time; it will first grab them from `~/go/pkg/mod` which is a very fast operation. The presence of this folder is also a requirement for the downstream repos, since everything the project is built from needs to exist in the repo, or at least that's my understanding of why.

tmshort: I would argue this is still vendoring; this PR just gets rid of the directory and the build step has to recreate it each time...

tmshort: I guess my point is, if everything is already in the GOPATH, why are we copying it over?

dtfranz:
> My concern here is that all this does is get rid of the vendor directory; building recreates the directory. It should be possible to do this without recreating the vendor directory.

The goal of this task was to remove the vendor directory and recreate it from cache when the github actions are run. I think that reworking the build process to remove the need for the vendor directory would be out of scope of the tasks intended purpose; primarily we just wanted to reduce the number of merge conflicts we get from downstreaming with the vendor folder inside source control. I personally don't see any issue with frequent re-runs of `go mod vendor`, particularly because it won't actually need to download anything at all in the vast majority of cases.

That being said though if there was something else you were getting at here that I misunderstood please let me know!

tmshort:
> That being said though if there was something else you were getting at here that I misunderstood please let me know!

No, that's basically it. If we want to _get rid_ of vendoring, there's more to do.

Makefile
m1kola: I agree with Todd. I think there is no benefit of in copying into vendor since we have local cache for modules.

> The builds for this repo are all dependent on the presence of the vendor directory and will fail without it.

I think this is because our `Makefile` has `GOFLAGS="-mod=vendor"` which is forcing `go build` to use vendor dir.

Try running this in master:

```
sed -i 's/GO := GOFLAGS="-mod=vendor" go/GO := go/g' ./Makefile
rm -r ./vendor
go mod download
make build
```

Even `go mod download` might not be necessary. I think `go build` without `-mod=vendor` will be to download modules (if not present) and build.

dtfranz: I agree with you both here and would prefer to not have the `vendor/` dir at all. But I do want to ask the following, not as a challenge but a genuine question as somebody still not terribly familiar with up/down-streaming processes: If we change the build process for the upstream repo to stop using `vendor/`, will that not make the downstream cherry-picking process more complicated? Since the downstream repo _must_ use the `vendor/` dir for the build (somebody correct me if this is wrong), then aren't we beginning to create a significant divergence between the up and downstream repos? Or is this not what you're recommending?

m1kola: I'm also not super familiar with downstreaming, but I'm happy to look into it together on next week.

tmshort: The vendor directory doesn't exist in the downstream staging directory. But there is a root-level vendor.

tmshort (approved):
No more `vendor`!

PR #1089, opened 2023-04-22
Add WalkMetas* functions and meta.Name field

### LINKED ISSUE #1080: [RFE] Add optional "name" field to FBC schema
## Feature Request

**Is your feature request related to a problem? Please describe.**

The File Based Catalogs (FBC) [schema](https://olm.operatorframework.io/docs/reference/file-based-catalogs/#schema) doesn't include a `name` field as an optional object identifier, although all of the [olm-defined schemas](https://olm.operatorframework.io/docs/reference/file-based-catalogs/#olm-defined-schemas) currently require it. This makes it difficult for us to create a consistent merge without having some consistent rule about how "name" is used.

**Describe the solution you'd like**

Add optional `name` field to FBC _Meta schema

joelanford:
These function are like WalkFS except that callers provide a function that handles a Meta object at a time, rather than an entire file's worth of FBC objects.

This PR also adds `Name string` to the Meta struct, which is necessary information to distinguish meta objects that have the same schema and package (e.g. `olm.channel` and `olm.bundle` objects)

For callers that can operate on a single meta at a time, these functions will provide significant reductions in memory usage. The catalogd project will likely make use of this function.

Closes #1080 

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/declcfg.go
+++ b/alpha/declcfg/declcfg.go
@@ -1,7 +1,13 @@
 package declcfg
 
 import (
+	"bytes"
 	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"go4.org/bytereplacer"
 
 	"github.com/operator-framework/operator-registry/alpha/property"
 )
@@ -94,6 +100,8 @@ func (m Meta) MarshalJSON() ([]byte, error) {
 }
 
 func (m *Meta) UnmarshalJSON(blob []byte) error {
+	blob = bytereplacer.New(`\u003c`, "<", `\u003e`, ">", `\u0026`, "&").Replace(blob)
+
 	type tmp struct {
 		Schema     string              `json:"schema"`
 		Package    string              `json:"package,omitempty"`
@@ -102,11 +110,58 @@ func (m *Meta) UnmarshalJSON(blob []byte) error {
 	}
 	var t tmp
 	if err := json.Unmarshal(blob, &t); err != nil {
-		return err
+		// TODO: return an error that includes the the full JSON message,
+		//    the offset of the error, and the error message. Let callers
+		//    decide how to format it.
+		return errors.New(resolveUnmarshalErr(blob, err))
 	}
 	m.Schema = t.Schema
 	m.Package = t.Package
 	m.Name = t.Name
 	m.Blob = blob
 	return nil
+}
+
+func resolveUnmarshalErr(data []byte, err error) string {
+	var te *json.UnmarshalTypeError
+	if errors.As(err, &te) {
+		return formatUnmarshallErrorString(data, te.Error(), te.Offset)
+	}
+	var se *json.SyntaxError
+	if errors.As(err, &se) {
+		return formatUnmarshallErrorString(data, se.Error(), se.Offset)
+	}
+	return err.Error()
+}
+
+func formatUnmarshallErrorString(data []byte, errmsg string, offset int64) string {
+	sb := new(strings.Builder)
+	_, _ = sb.WriteString(fmt.Sprintf("%s at offset %d (indicated by <==)\n ", errmsg, offset))
+	// attempt to present the erroneous JSON in indented, human-readable format
+	// errors result in presenting the original, unformatted output
+	var pretty bytes.Buffer
+	err := json.Indent(&pretty, data, "", "    ")
+	if err == nil {
+		pString := pretty.String()
+		// calc the prettified string offset which correlates to the original string offset
+		var pOffset, origOffset int64
+		origOffset = 0
+		for origOffset = 0; origOffset < offset; {
+			if pString[pOffset] != '\n' && pString[pOffset] != ' ' {
+				origOffset++
+			}
+			pOffset++
+		}
+		_, _ = sb.WriteString(pString[:pOffset])
+		_, _ = sb.WriteString(" <== ")
+		_, _ = sb.WriteString(pString[pOffset:])
+	} else {
+		for i := int64(0); i < offset; i++ {
+			_ = sb.WriteByte(data[i])
+		}
+		_, _ = sb.WriteString(" <== ")
+		_, _ = sb.Write(data[offset:])
+	}
+
+	return sb.String()
 }
--- a/alpha/declcfg/declcfg.go
+++ b/alpha/declcfg/declcfg.go
@@ -84,6 +84,7 @@ type RelatedImage struct {
 type Meta struct {
 	Schema  string
 	Package string
+	Name    string
 
 	Blob json.RawMessage
 }
@@ -96,6 +97,7 @@ func (m *Meta) UnmarshalJSON(blob []byte) error {
 	type tmp struct {
 		Schema     string              `json:"schema"`
 		Package    string              `json:"package,omitempty"`
+		Name       string              `json:"name,omitempty"`
 		Properties []property.Property `json:"properties,omitempty"`
 	}
 	var t tmp
@@ -104,6 +106,7 @@ func (m *Meta) UnmarshalJSON(blob []byte) error {
 	}
 	m.Schema = t.Schema
 	m.Package = t.Package
+	m.Name = t.Name
 	m.Blob = blob
 	return nil
 }
--- a/alpha/declcfg/load.go
+++ b/alpha/declcfg/load.go
@@ -1,14 +1,12 @@
 package declcfg
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"path/filepath"
-	"strings"
 
 	"github.com/joelanford/ignore"
 	"github.com/operator-framework/api/pkg/operators"
@@ -47,18 +45,12 @@ type WalkMetasReaderFunc func(meta *Meta, err error) error
 func WalkMetasReader(r io.Reader, walkFn WalkMetasReaderFunc) error {
 	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
 	for {
-		doc := json.RawMessage{}
-		if err := dec.Decode(&doc); err != nil {
+		var in Meta
+		if err := dec.Decode(&in); err != nil {
 			if errors.Is(err, io.EOF) {
 				break
 			}
 			return walkFn(nil, err)
-		}
-		doc = []byte(strings.NewReplacer(`\u003c`, "<", `\u003e`, ">", `\u0026`, "&").Replace(string(doc)))
-
-		var in Meta
-		if err := json.Unmarshal(doc, &in); err != nil {
-			return walkFn(nil, fmt.Errorf("unmarshal error: %s", resolveUnmarshalErr(doc, err)))
 		}
 
 		if err := walkFn(&in, nil); err != nil {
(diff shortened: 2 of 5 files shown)


joelanford:
/hold
I want a few reviews on this one, so I don't want a single lgtm allowing this to merge.

alpha/declcfg/load.go
```diff
@@ -22,13 +22,74 @@ const (
...
+	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
+	for {
+		doc := json.RawMessage{}
+		if err := dec.Decode(&doc); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return walkFn(nil, err)
+		}
+		doc = []byte(strings.NewReplacer(`\u003c`, "<", `\u003e`, ">", `\u0026`, "&").Replace(string(doc)))
```
anik120, line 57: Could we have some documentation for what the values for the `NewReplacer` method are? I.e what exactly are we achieving with this line?

alpha/declcfg/load.go
```diff
@@ -22,13 +22,74 @@ const (
...
+		cfg, err := LoadFile(root, path)
+		if err != nil {
+			return walkFn(path, cfg, err)
+		}
+
+		return walkFn(path, cfg, nil)
+	})
+}
+
+func walkFiles(root fs.FS, fn func(root fs.FS, path string, err error) error) error {
```
everettraven, line 92: Nit: Can we make the `fn` parameter a separate type? I find it easier to read when parameters that are functions use a type to represent that function:
```suggestion
type walkFileFunc func(root fs.FS, path string, err error)
func walkFiles(root fs.FS, fn walkFilesFunc error) error {
```

alpha/declcfg/load.go
```diff
@@ -213,10 +261,10 @@ func formatUnmarshallErrorString(data []byte, errmsg string, offset int64) strin
 		var pOffset, origOffset int64
 		origOffset = 0
 		for origOffset = 0; origOffset < offset; {
-			pOffset++
 			if pString[pOffset] != '\n' && pString[pOffset] != ' ' {
 				origOffset++
 			}
+			pOffset++
```
everettraven, line 267: Is there a reason this was moved? Could it have been intentional that the `pOffset` increment occurs before it's use?

anik120 (commented):
Overall looks good to me. 

> The catalogd project will likely make use of this function.

I was a little mislead by this. If I'm not wrong, nothing changes in terms of wirings for clients, I.e clients still [call LoadReader](https://github.com/operator-framework/catalogd/blob/main/pkg/controllers/core/catalogsource_controller.go#L371) like they have been. So it's more `The catalogd project will likely benefit from this function`. Is that correct?

alpha/declcfg/load.go
```diff
@@ -22,13 +22,74 @@ const (
...
+
+func WalkMetasReader(r io.Reader, walkFn WalkMetasReaderFunc) error {
+	dec := yaml.NewYAMLOrJSONDecoder(r, 4096)
+	for {
+		doc := json.RawMessage{}
+		if err := dec.Decode(&doc); err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return walkFn(nil, err)
```
everettraven, line 55: Nit: Could we signal that the error occurred while attempting to decode like we do with the unmarshal error below? Could be useful for debugging.
```suggestion
			return walkFn(nil, fmt.Errorf("decoding error: %s", err))
```

Another thought, I could see it being useful to return a custom error type that the walkFn can use to easily determine what step the error occurred in (i.e a `DecodeError` and/or `UnmarshalError` types). Would be useful if the walkFn want's to perform some specific logic based on a specific error. That being said, maybe being that specific isn't useful in this case? I don't have any use cases for that off the top of my head.

everettraven (approved):
Looks good overall, just have a few nits/comments - I don't think any of them should block this PR from merging though.

alpha/declcfg/load.go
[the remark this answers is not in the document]
joelanford: I don't see a `bytes.NewReplacer`. There _is_ `bytes.Replace(All)`, but as far as I can tell, they only handle one substitution at a time. So I think options are:
1. Roll with this and tolerate the double-allocation
2. Call `bytes.Replace` in a loop for each replacement. (seems worse from an allocation standpoint since we have 3 replacements)
3. Roll our own bytes replacer.
4. Figure out a way to not require this replacement in the first place. It's been awhile, but iirc, its because an encoder is used somewhere along the line that we don't control and therefore can't set [`enc.SetEscapeHTML(false)`](https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML)
5. Not bother with the replacement and let the unicode show up in the output.

alpha/declcfg/load.go
joelanford: No need, IMO, because this particular walk function is private.

alpha/declcfg/load.go
joelanford: Yeah I moved it because I was getting off-by-one panics in the tests I implemented, which included an error at the last byte of the JSON string, and as this tried to map to the pretty-printed offset, it incremented `pOffset` beyond the length of `pString`.

Maybe @grokspawn can confirm for sure if there's something we should be aware of. But I tested before an after of this change to make sure the resulting error looked the same. All the other tests also still passed, so I'm thinking I was just the unlucky person to run into this untested edge case?

alpha/declcfg/load.go
joelanford: I could maybe see this:
```suggestion
			return walkFn(nil, fmt.Errorf("decoding error: %w", err))
```

I think it's important to give the walk function access to the underlying error if we wrap it with some context. In fact I was going to comment that in the case of the formatting we do for json unmarshaling errors, we might want to move that to a different spot so we don't obfuscate the original error.

alpha/declcfg/load.go
```diff
@@ -213,10 +261,10 @@ func formatUnmarshallErrorString(data []byte, errmsg string, offset int64) strin
 		var pOffset, origOffset int64
 		origOffset = 0
 		for origOffset = 0; origOffset < offset; {
-			pOffset++
```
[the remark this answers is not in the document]
joelanford: Don't disagree. Seems out of scope for this PR though? Can we make an issue for this?

alpha/declcfg/load_test.go
```diff
@@ -92,6 +92,73 @@ func TestLoadReader(t *testing.T) {
...
+		{
+			name:      "Error/Invalid",
+			fsys:      invalidFS,
+			assertion: require.Error,
+		},
+		{
+			name:              "Success/ValidDir",
+			fsys:              validFS,
+			assertion:         require.NoError,
+			expectNumPackages: 3,
```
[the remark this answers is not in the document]
joelanford: Would that leave a testing gap where both implementations are victims of the same underlying bug, so they end up seeing the same data, but its the wrong data?

alpha/declcfg/declcfg.go
```diff
@@ -96,6 +97,7 @@ func (m *Meta) UnmarshalJSON(blob []byte) error {
 	type tmp struct {
```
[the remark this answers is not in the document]
joelanford: Because `m` implements `json.Unmarshaler`, which would just circle back to this function being called again from the top. And we'd have a stack overflow.

The reason we need to implement a custom unmarshal function is to get the full content of the object set in `m.Blob`.

alpha/declcfg/load.go
joelanford: See some context here: https://github.com/operator-framework/operator-registry/pull/1089#discussion_r1175737819

joelanford:
> clients still [call LoadReader](https://github.com/operator-framework/catalogd/blob/main/pkg/controllers/core/catalogsource_controller.go?rgh-link-date=2023-04-24T16%3A44%3A07Z#L371) like they have been

That section of code reads the entire contents of the catalog into memory (twice actually: once into a bytes.Buffer, and a second time into declcfg.DeclarativeConfig), returns it from the function, and then the caller calls `createPackage` and `createBundleMetadata`, each of which basically iterate the appropriate schema-based lists.

It would be _way_ more performant if we did `declcfg.LoadMetasReader(podLogs, walkFunc)`, where walk func could:
1. directly create `BundleMetadata` when it sees an `olm.bundle`
2. collect packages and channels up to eventually be used to create `Package` objects.

Later, if/when we treat metas a bit more opaquely, we can just sync every object directly to the underlying storage without having to collect anything up at all.

alpha/declcfg/load.go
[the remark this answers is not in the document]
stevekuznetsov: I mis-remembered this, sorry - I used [this library](https://pkg.go.dev/go4.org/bytereplacer) in the past and found it nice.

alpha/declcfg/load_test.go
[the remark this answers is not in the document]
stevekuznetsov: Yeah, six of one, half dozen of the other, I guess - this approach has the bug where the expected packages are incorrect.

alpha/declcfg/declcfg.go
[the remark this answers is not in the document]
stevekuznetsov: Ah, I see. Tricky!

alpha/declcfg/load_test.go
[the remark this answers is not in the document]
joelanford: Fair. To be honest, I cribbed from an existing test. I think [this test](https://github.com/operator-framework/operator-registry/blob/45f491b932997c20ac9c6e54b4b4fa822994cefb/alpha/declcfg/load_test.go#L119) gives us what both of us are asking for. We have lots of overlapping tests because all these functions now basically build on each other.

It wouldn't be an unreasonable argument to mimic that test instead though. Because in this case, we want to see if our `Meta` objects look like we expect. I'll take a look.

alpha/declcfg/declcfg.go
```diff
@@ -93,17 +100,68 @@ func (m Meta) MarshalJSON() ([]byte, error) {
 }
 
 func (m *Meta) UnmarshalJSON(blob []byte) error {
+	blob = bytereplacer.New(`\u003c`, "<", `\u003e`, ">", `\u0026`, "&").Replace(blob)
```
joelanford, line 103: I moved the bytes replacement here, which meant I could drop the extra unmarshal into `json.RawMessage`. Between the allocation improvement of a direct bytes replacer and the removal of the other unmarshal, we shave off ~7% of the runtime of `opm render` when I run it against the operatorhub catalog.

Not too shabby!

In a follow-up, I might take a look at an alternate mode of `action.Render` (maybe a new function alongside `Run()`?), which lets the caller use a walk function. That would save a bunch on memory and might shave off more time as well.

alpha/declcfg/declcfg.go
```diff
@@ -93,17 +100,68 @@ func (m Meta) MarshalJSON() ([]byte, error) {
...
+		Name       string              `json:"name,omitempty"`
 		Properties []property.Property `json:"properties,omitempty"`
 	}
 	var t tmp
 	if err := json.Unmarshal(blob, &t); err != nil {
-		return err
+		// TODO: return an error that includes the the full JSON message,
+		//    the offset of the error, and the error message. Let callers
+		//    decide how to format it.
+		return errors.New(resolveUnmarshalErr(blob, err))
```
joelanford, line 116: I moved this here because it was the only place access to the full json message existed after I removed the extra unmarshal to json.RawMessage.

alpha/declcfg/load.go
joelanford: For now I think I'm just going to leave this as directly passing an unaltered error, which is what we were doing before. We can revisit if it turns out this extra context would be important in this case.

joelanford:
/hold cancel
I think this is ready to merge (just needs `/lgtm`)

For follow-ups, we have:
- [ ] Refactor the unmarshal error formatting to return an error struct to give callers more formatting flexibility.
- [ ] Look at improving WalkMetas* tests to assert on the contents of the objects.

alpha/declcfg/load.go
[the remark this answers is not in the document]
grokspawn: One producer of these artifacts is `opm render` itself. :)  If we output YAML, it escapes html-hostile runes like these, so we have to replace them on ingest.

PR #1090, opened 2023-04-25
comment out CatalogDestination.BaseImage while we mature the concept

grokspawn:
comment out reliance on BaseImage

The original concept os the BaseImage was intended to use the same version of opm for build/verify/serve, but the initial implementation had to be reverted.  While we mature that concept separately, we thought it seemed like a good idea to remove reliance on it (but neglected to in a previous PR). 

**Reviewer Checklist**

### DIFF
--- a/alpha/template/composite/composite_test.go
+++ b/alpha/template/composite/composite_test.go
@@ -42,7 +42,6 @@ schema: olm.composite.catalogs
 catalogs:
   - name: first-catalog
     destination:
-      baseImage: quay.io/operator-framework/opm:latest
       workingDir: contributions/first-catalog
     builders:
       - olm.builder.test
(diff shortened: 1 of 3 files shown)


alpha/template/composite/composite_test.go
everettraven, line 1: I think there are some YAML strings we use for testing here that look like they still use the `baseImage` field - would be good to just remove that field from those strings for now.

A few places I saw:
- https://github.com/operator-framework/operator-registry/pull/1090/files#diff-0299f965ef1d5dc96c54539eccbbcbf7a35e5f38255886bcbb6f8c39edf3364fR45
- https://github.com/operator-framework/operator-registry/pull/1090/files#diff-0299f965ef1d5dc96c54539eccbbcbf7a35e5f38255886bcbb6f8c39edf3364fR301
- https://github.com/operator-framework/operator-registry/pull/1090/files#diff-0299f965ef1d5dc96c54539eccbbcbf7a35e5f38255886bcbb6f8c39edf3364fR328

everettraven (changes requested):
One comment that I think may lead to unit test failures if not taken care of

everettraven (commented):
Based on the current unit test failure looks like we need to update https://github.com/operator-framework/operator-registry/pull/1090/files#diff-0299f965ef1d5dc96c54539eccbbcbf7a35e5f38255886bcbb6f8c39edf3364fR328

everettraven (approved):
/lgtm

PR #1094, opened 2023-05-16
render: improve olm.bundle.object rendering for bundles

joelanford:
This pull request is a series of commits that progressively improves FBC parsing. Commits are:
1. Render `olm.bundle.object` _just_ for CSVs with `opm render <bundleImage>`
2. Render `olm.bundle.object` _just_ for CSVs with `opm render <sqliteImageOrFile>` (but only if the bundle has an image reference in the database).
3. Use $numCPUs goroutines to parse files in an FBC directory concurrently.
4. Introduce an `olm.csv.metadata` property, and use it in place of the `olm.bundle.object` properties in the scenarios covered in commits 1 and 2. When serving the GRPC API, this property is used to generate a synthetic CSV to be served.
5. Fix a bug in `Meta` object unmarshaling that was made apparent when testing commit 4 on real-world catalogs

A few notes:
- Commits 1 and 2 are somewhat overtaken by commit 4 (no code path generates full `olm.bundle.object` properties for _just_ CSVs after commit 4).
- Only commit 3 will realize gains for existing FBCs. Commits 1, 2, and 4 require re-rendering or re-migrating from bundles/sqlite catalogs (or manually pruning/converting `olm.bundle.object` properties)
- This PR does _not_ attempt to proactively remove objects from non-channel-heads. Despite the inability for existing OLMv0 components to make use of this metadata, it is nonetheless conceivably valuable for it to remain in catalogs to be readable by future versions of OLM.
- Commit 4's functionality is only activated if a bundle both:
    - has no CSV in any defined `olm.bundle.object` properties
    - has exactly 1 `olm.csv.metadata` property.
- Commit 4 injects the icon defined in `olm.package` into synthetic CSVs, if defined in the package.
- Commit 4 injects the description defined in `olm.package` into synthetic CSVs, if not defined in the `olm.csv.metadata` object.

Open Questions:
- How much of the `olm.csv.metadata` object belongs in the bundle vs. the package (e.g. display name, keywords, maintainers, provider)? In this PR, I'm hesitant to add new root level fields or package properties, but maybe that's an unfounded fear. It would be a further optimization because catalogs wouldn't have to repeat those same things over and over again in each bundle.
- Should we update the `render` and `migrate` commands to do any conversion of `olm.bundle.object` to `olm.csv.metadata` for existing FBCs? It seems like this would be pretty useful, so I've got another commit in a separate branch that would do this. I can include it in this PR or make a separate PR for it after we finalize this PR. Thoughts?
- This does PR does break an invariant we had originally tried to keep. That is "an sqlite db and its migrated FBC equivalent will serve the same GRPC API". Now that `opm migrate` intentionally loses fidelity, that invariant no longer holds. Are we okay with this? IMO, I think this is okay. SQLite serving has been deprecated for several years, so I think it is okay to start diverging. Perhaps we even consider dropping the SQLite server prior to the SQLite-based catalog building commands?

/hold
There's a bunch here to discuss, evaluate, and manually test. I updated our tests as best I could, and I think I've covered old code paths + new code paths. But this is a pretty significant change, so it would be good to get some detail-oriented eyes on this.

**Reviewer Checklist**


joelanford:
Looks like a deadlock I introduced in the LoadFS concurrency commit: https://github.com/operator-framework/operator-registry/actions/runs/4992141512/jobs/8940057783?pr=1094#step:4:172

Investigating now.

joelanford:
Alright, I think the concurrency deadlock issue is fixed up!

alpha/declcfg/declcfg.go
```diff
@@ -100,25 +98,27 @@ func (m Meta) MarshalJSON() ([]byte, error) {
...
-	type tmp struct {
-		Schema     string              `json:"schema"`
-		Package    string              `json:"package,omitempty"`
-		Name       string              `json:"name,omitempty"`
-		Properties []property.Property `json:"properties,omitempty"`
+	blobMap := map[string]interface{}{}
+	if err := json.Unmarshal(blob, &blobMap); err != nil {
+		return err
+	}
+	for key, ptr := range map[string]*string{"schema": &m.Schema, "package": &m.Package, "name": &m.Name} {
```
everettraven, line 105: Do we care about about processing the `properties` field as well? What happens with this implementation if the meta object's YAML has a populated `properties` field?

alpha/template/semver/README.md
everettraven, line 1: The changes in this file are fine and make sense to me but seem completely irrelevant to this PR. Is there a specific reason these changes are in this PR?

everettraven (commented):
Overall looks reasonable to me, just have a couple questions

alpha/template/semver/README.md
joelanford: It turns out they are not quite irrelevant because the changes in handling JSON unmarshaling in declcfg.go mean that only lowercase key fields are handled now. That change broke a test where the semver template using upper camel case keys.

alpha/declcfg/declcfg.go
joelanford: In the original code, we are parsing to check for expected syntax of the properties field of _all_ schema types, regardless of whether its a known type or not. Then we throw those parsed properties away.

I'm not sure we should require _all_ schemas that want to have a properties field to follow our properties schema. Ideally, we could have this discussion separately from this PR. We could do that, but it would require a bit more complexity and resource utilization (potentially requiring multiple parsing passes) in this function.

everettraven: Fair enough, I'm definitely open to having that discussion separately from this PR

joelanford:
I've uncovered and fixed a few more issues:
- The concurrency logic needed one more fix to prevent deadlocks. Now channels reads and writes can be preempted by the context being canceled
- In testing an FBC that uses `olm.csv.metadata`, I found that the `csv.metadata.labels` needed to be preserved.

With these changes, I:
- Pulled down the latest operatorhub catalog and migrated it to use `olm.csv.metadata` instead of `olm.bundle.object` properties, rebuilt it with an opm binary from this branch, and pushed to a personal quay repo.
- Created a kind cluster and installed OLM using `operator-sdk olm install` (this includes a catalog source for the real operatorhubio catalog)
- Created a catalog source pointing to my personal rebuilt operatorhub catalog
- Queried for all packagemanifests (once for the real catalog, once for my rebuilt catalog)
- Filtered out  `spec.metadata.creationTimestamp` and `spec.metadata.labels[catalog*]` which are not expected to be the same
- Performed a diff and noticed that the packagemanifests output for the rebuilt catalog contained an extra relatedImage for the bundle image
- Filtered out `status.channels[].currentCSVDesc.relatedImages`, performed another diff, which turned nothing up.

The last test I want to do is to put the `opm` binary from this branch into the existing operatorhubio catalog without changing any other contents, and verify that that image produces the same packagemanifests as the original. Once I verify that, I'll unhold.

joelanford:
Ok I was able to confirm that the packagemanifest output of the existing operatorhub catalog using the opm binary from this branch is identical. So I think this is ready to go from a back-compat standpoint.

/hold cancel

alpha/declcfg/load.go
```diff
@@ -111,20 +115,113 @@ func walkFiles(root fs.FS, fn func(root fs.FS, path string, err error) error) er
...
+	// before the cfgChan is closed.
+	var wg sync.WaitGroup
+	for i := 0; i < concurrency; i++ {
+		wg.Add(1)
+		eg.Go(func() error {
+			defer wg.Done()
+			return parsePaths(ctx, root, pathChan, cfgChan)
+		})
+	}
+	eg.Go(func() error {
```
stevekuznetsov, line 151: Why does this need to be async?

alpha/declcfg/load.go
```diff
@@ -111,20 +115,113 @@ func walkFiles(root fs.FS, fn func(root fs.FS, path string, err error) error) er
...
+			if !ok {
+				return nil
+			}
+			cfg, err := LoadFile(root, path)
+			if err != nil {
+				return err
+			}
+			select {
+			case cfgChan <- cfg:
+			case <-ctx.Done(): // don't block on sending to cfgChan
```
stevekuznetsov, line 202: Technically this still blocks forever if, for instance, you run out of room without any errors. If you really want to not block, you could add a timeout here.

alpha/declcfg/load.go
```diff
@@ -111,20 +115,113 @@ func walkFiles(root fs.FS, fn func(root fs.FS, path string, err error) error) er
...
+func mergeCfgs(ctx context.Context, cfgChan <-chan *DeclarativeConfig, fcfg *DeclarativeConfig) error {
+	for {
+		select {
+		case <-ctx.Done(): // don't block on receiving from cfgChan
+			return ctx.Err()
+		case cfg, ok := <-cfgChan:
+			if !ok {
+				return nil
+			}
+			fcfg.Packages = append(fcfg.Packages, cfg.Packages...)
```
stevekuznetsov, line 218: `go test -race` might false positive on this write if it's accessed elsewhere

stevekuznetsov (commented):
JSON bits continue to break my brain but the parallel stuff looks mostly good!

stevekuznetsov:
A Go benchmark would be sweet to capture this improvement.

alpha/declcfg/load.go
joelanford: This is the third iteration of this code, so I'm by no means 100% confident there isn't another deadlock scenario here, but I don't think I'm seeing it.

- If there's an error that happens in any of the `eg`'s goroutines, the context gets cancelled.
- If no error happens, then eventually `mergeCfgs` will read from `cfgChan` and unblock this select call.

Is there another scenario I'm missing?

alpha/declcfg/load.go
joelanford: Standard `make unit` uses `-race`, and I haven't seen any issues, so hopefully all good.

I know you said false positive, but and I'm sure how the race detector works, but I don't see any race possibilities with `fcfg`.

alpha/declcfg/load.go
```diff
@@ -111,20 +115,113 @@ func walkFiles(root fs.FS, fn func(root fs.FS, path string, err error) error) er
 // If LoadFS encounters an error loading or parsing any file, the error will be
 // immediately returned.
 func LoadFS(root fs.FS) (*DeclarativeConfig, error) {
-	cfg := &DeclarativeConfig{}
-	if err := WalkFS(root, func(path string, fcfg *DeclarativeConfig, err error) error {
+	if root == nil {
+		return nil, fmt.Errorf("no declarative config filesystem provided")
+	}
+
+	concurrency := runtime.NumCPU()
```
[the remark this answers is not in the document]
joelanford: Yeah, I didn't explore this at all. I'd vote for follow-up to keep scope tight.

alpha/declcfg/load.go
```diff
@@ -111,20 +115,113 @@ func walkFiles(root fs.FS, fn func(root fs.FS, path string, err error) error) er
...
+		fcfg     = &DeclarativeConfig{}
+		pathChan = make(chan string, concurrency)
+		cfgChan  = make(chan *DeclarativeConfig, concurrency)
+	)
+
+	// Create an errgroup to manage goroutines. The context is closed when any
+	// goroutine returns an error. Goroutines should check the context
+	// to see if they should return early (in the case of another goroutine
+	// returning an error).
+	eg, ctx := errgroup.WithContext(context.Background())
```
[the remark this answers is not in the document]
joelanford: Yeah, I considered that. This function never had anything in its call stack that had a context parameter, so it was never plumbed through. Perhaps it's worth the breaking API change to go ahead and do this (this package is in an `alpha` subtree). If we were concerned about that, we could deprecate this function and add a new `LoadFSContext` function or something.

I think I'd vote for just changing `LoadFS` to accept a context.

alpha/declcfg/load.go
joelanford: Yeah I guess we could move this to be called just before `eg.Wait()`?

joelanford:
> JSON bits continue to break my brain

I _think_ I understand it much better now that I know that `sigs.k8s.io/yaml.Unmarshal` into `json.RawMessage` does this: https://github.com/kubernetes-sigs/yaml/issues/94

joelanford:
I added a go benchmark in 4cfabfe, just before the changes that added concurrency. Then I:
- ran the benchmark on 4cfabfe
- ran the same exact benchmark on aef5a6b – the commit that added concurrency
- updated the benchmark to use `olm.csv.metadata` properties instead of `olm.bundle.object` properties in fc25695, and then ran it.

`benchstat` results below:
```
goos: darwin
goarch: arm64
pkg: github.com/operator-framework/operator-registry/alpha/declcfg
          │ 00-4cfabfe.txt │           01-aef5a6b.txt            │           02-fc25695.txt           │
          │     sec/op     │    sec/op     vs base               │   sec/op     vs base               │
LoadFS-10    10667.6m ± 0%   1355.1m ± 1%  -87.30% (p=0.002 n=6)   109.4m ± 1%  -98.97% (p=0.002 n=6)
```

TL;DR
- _Just_ concurrency gets us ~8x faster on my 8-core machine
- Using `olm.csv.metadata` and dropping `olm.bundle.object` gets another ~12x faster using the generated data
- So overall improvement with generated data on my machine is ~97x faster

alpha/declcfg/load.go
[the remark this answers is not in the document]
joelanford: I attempted this in another commit and realized two things:
1. The change cascades all over the place
2. It may make sense to add context to other walk/load functions in this package.

Given that, I'd like to handle passing context in a separate PR.

alpha/declcfg/load.go
joelanford: Done.

everettraven (approved):

alpha/declcfg/load_benchmark_test.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
+)
+
+func BenchmarkLoadFS(b *testing.B) {
+	fbc := generateFBC(b, 300, 450, 3000)
+	fsys := &memFS{MapFS: fstest.MapFS{}}
```
stevekuznetsov, line 25: Filesystem interactions are likely important in loading, so this benchmark is IMHO less useful than it could be if it just used a temporary directory and real files.

alpha/declcfg/load_benchmark_test.go
```diff
@@ -0,0 +1,313 @@
...
+
+func generateFBC(b *testing.B, numPackages, numChannels, numBundles int) *declcfg.DeclarativeConfig {
```
stevekuznetsov, line 41: Could be nice to use fuzzing here

alpha/declcfg/load_benchmark_test.go
stevekuznetsov: And ideally it's not the same filesystem for each benchmark since they'll be memory-mapped if it is :)

joelanford:
Re: the open questions from the PR description:
1. I chatted some with @tlwu2013. Both of us had come to the same conclusion: let's not over-complicate this PR with other optimizations related to moving more fields out of the bundle and into parent FBC blobs. We can handle that sort of thing separately if/when there is more clarity around it
2. I opened #1104
3. No discussion on this point yet. My opinion remains that we can break this invariant. I'm still interested in having a conversation about removing `opm registry serve` (but still keeping the other deprecated `opm` commands that write the database for now). Opened #1105

stevekuznetsov:
@joelanford any updates here?

joelanford:
I think this is ready to merge. Just waiting on `lgtm` label.

grokspawn (approved):

grokspawn:
I think we have more conversation to explore the case-insensitivity of schema keys (FWIW, folks filed a bug with go/json about it, and they basically said "too late... Hyrum's law"), but I don't want to hold this heavy PR up because of that discussion.  We just need to not forgot about it.  

Waiting for the CI to finish to push in.

AdrianVasiliu:
@joelanford For some reason, for the catalog of our operator, upgrading operator-registry from 1.27.1 to 1.28.0 and testing in OpenShift 4.12.7, I observe the following:
* The `CatalogSource` passes in `Ready` status as expected.
* But the operator doesn't show up in OperatorHub anymore.

Can this be a side-effect of the changes brought by this PR? We do observe (with `opm render`) large changes consistent with those involved here - just that we weren't expecting it would break our catalog. 
The notes in https://github.com/operator-framework/operator-registry/releases/tag/v1.28.0 don't seem to mention breaking changes.
All in one:
* Is it a bug, or is it an expected breaking change?
* If so, what adjustments should we bring to our catalog/bundle build to make it work with operator-registry 1.28.0?

joelanford:
The changes require the opm server to be using 1.28 as well. Can you confirm the version of open used I. The catalog image to serve the new FBC?

AdrianVasiliu:
@joelanford Thanks a lot for the feedback.
Our catalog image gets the following opm version from `registry.redhat.io/openshift4/ose-operator-registry:v4.13`:
```
$ opm version 
Version: version.Version{OpmVersion:"ce46f5b97", GitCommit:"ce46f5b97e4838c4508f6dc471cabc1199a8f0b9", BuildDate:"2023-06-07T20:53:52Z", GoOs:"linux", GoArch:"amd64"}
```
So build date June 7, but on https://github.com/openshift/operator-framework-olm/tree/master/staging/operator-registry. , while https://github.com/operator-framework/operator-registry/releases/tag/v1.28.0 is from June 9 and the code base is different.

I don't know what is the policy of keeping both in sync. Anyway, we'll use your 1.28 in our catalog too, and report back the testing results.

AdrianVasiliu:
@joelanford I confirm using operator-registry 1.28.0 in the catalog image and at our build time does the trick! All good. Thanks a lot!

joelanford:
Awesome! Thanks for confirming!

PR #1097, opened 2023-05-17
Add `generate devfile` command

theishshah:

**Reviewer Checklist**

### DIFF
--- a/alpha/action/generate_devfile.go
+++ b/alpha/action/generate_devfile.go
@@ -8,6 +8,9 @@ import (
 
 type GenerateDevfile struct {
 	IndexDir string
+	Name     string
+	BuildCTX string
+	Provider string
 	Writer   io.Writer
 }
 
@@ -34,18 +37,18 @@ func (i GenerateDevfile) validate() error {
 
 const devfileTmpl = `schemaVersion: 2.2.0
 metadata:
-  name: fbc-4.11
-  displayName: FBC 4.11
+  name: {{.Name}}
+  displayName: {{.Name}}
   description: 'File based catalog'
   language: fbc
-  provider: Red Hat
+  provider: {{.Provider}}
 components:
   - name: image-build
     image:
       imageName: fbc:latest
       dockerfile:
         uri: {{.IndexDir}}.Dockerfile
-        buildContext: "4.11"
+        buildContext: {{.BuildCTX}}
 commands:
   - id: build-image
     apply:
--- a/alpha/action/generate_devfile.go
+++ b/alpha/action/generate_devfile.go
@@ -1,0 +1,53 @@
+package action
+
+import (
+	"fmt"
+	"io"
+	"text/template"
+)
+
+type GenerateDevfile struct {
+	IndexDir string
+	Writer   io.Writer
+}
+
+func (i GenerateDevfile) Run() error {
+	if err := i.validate(); err != nil {
+		return err
+	}
+
+	t, err := template.New("devfile").Parse(devfileTmpl)
+	if err != nil {
+		// The template is hardcoded in the binary, so if
+		// there is a parse error, it was a programmer error.
+		panic(err)
+	}
+	return t.Execute(i.Writer, i)
+}
+
+func (i GenerateDevfile) validate() error {
+	if i.IndexDir == "" {
+		return fmt.Errorf("index directory is unset")
+	}
+	return nil
+}
+
+const devfileTmpl = `schemaVersion: 2.2.0
+metadata:
+  name: fbc-4.11
+  displayName: FBC 4.11
+  description: 'File based catalog'
+  language: fbc
+  provider: Red Hat
+components:
+  - name: image-build
+    image:
+      imageName: fbc:latest
+      dockerfile:
+        uri: {{.IndexDir}}.Dockerfile
+        buildContext: "4.11"
+commands:
+  - id: build-image
+    apply:
+      component: image-build
+`
--- a/cmd/opm/alpha/template/composite.go
+++ b/cmd/opm/alpha/template/composite.go
@@ -17,6 +17,8 @@ func newCompositeTemplateCmd() *cobra.Command {
 		validate      bool
 		compositeFile string
 		catalogFile   string
+		dockerfileDir string
+		devfileDir    string
 	)
 	cmd := &cobra.Command{
 		Use: "composite",
@@ -73,5 +75,7 @@ and a 'composite template' file`,
 	cmd.Flags().BoolVar(&validate, "validate", true, "whether or not the created FBC should be validated (i.e 'opm validate')")
 	cmd.Flags().StringVarP(&compositeFile, "composite-config", "c", "composite.yaml", "File to use as the composite configuration file")
 	cmd.Flags().StringVarP(&catalogFile, "catalog-config", "f", "catalogs.yaml", "File to use as the catalog configuration file")
+	cmd.Flags().StringVar(&dockerfileDir, "dockerfile", "", "Generates a Dockerfile in the provided directory")
+	cmd.Flags().StringVar(&devfileDir, "devfile", "", "Generates a Devfile in the provided directory")
 	return cmd
 }
--- a/cmd/opm/generate/cmd.go
+++ b/cmd/opm/generate/cmd.go
@@ -87,6 +87,56 @@ value of each duplicate key will be added to the generated Dockerfile.
 	return cmd
 }
 
+func newDevfileCmd() *cobra.Command {
+	var (
+		baseImage string
+	)
+	cmd := &cobra.Command{
+		Use:   "devfile <dcRootDir>",
+		Args:  cobra.ExactArgs(1),
+		Short: "Generate a Devfile for a declarative config index",
+		Long: `Generate a Devfile for a declarative config index.
+
+This command creates a Devfile in the same directory as the <dcRootDir>
+(named <dcDirName>.Dockerfile) that can be used to build the index. If a
+Devfile with the same name already exists, this command will fail.
+`,
+		RunE: func(_ *cobra.Command, args []string) error {
+			fromDir := filepath.Clean(args[0])
+
+			dir, indexName := filepath.Split(fromDir)
+			dockerfilePath := filepath.Join(dir, fmt.Sprintf("%s.Devfile", indexName))
+
+			if err := ensureNotExist(dockerfilePath); err != nil {
+				logrus.Fatal(err)
+			}
+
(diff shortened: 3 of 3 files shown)


alpha/action/generate_devfile.go
```diff
@@ -0,0 +1,53 @@
...
+    image:
+      imageName: fbc:latest
+      dockerfile:
+        uri: {{.IndexDir}}.Dockerfile
+        buildContext: "4.11"
+commands:
+  - id: build-image
+    apply:
+      component: image-build
+`
```
everettraven, line 53: We are hardcoding a lot of values in the template that would probably be better to parameterize. Things like the name, display name, provider, and build context are likely things that we are going to want to be configurable and be based on the actual catalog contributions being generated.

cmd/opm/alpha/template/composite.go
```diff
@@ -26,6 +32,37 @@ and a 'composite template' file`,
 and a 'composite template' file`,
 		Args: cobra.MaximumNArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
+			containerTool = "docker"
```
everettraven, line 35:
```suggestion
			containerTool := "docker"
```

cmd/opm/alpha/template/composite.go
```diff
@@ -26,6 +32,37 @@ and a 'composite template' file`,
...
+				}
+				tempCatalog = tempResp.Body
+				defer tempCatalog.Close()
+			}
+			catalogData := tempCatalog
+
+			// get catalog configurations
+			catalogConfig := &composite.CatalogConfig{}
+			catalogDoc := json.RawMessage{}
+			catalogDecoder := yaml.NewYAMLOrJSONDecoder(catalogData, 4096)
```
everettraven, line 57: Looks like the `yaml` package isn't being imported

cmd/opm/alpha/template/composite.go
```diff
@@ -36,6 +73,70 @@ and a 'composite template' file`,
...
+			if err != nil {
+				log.Fatalf("getting current working directory: %v", err)
+			}
+
+			// setup the builders for each catalog
+			setupFailed := false
+			setupErrors := map[string][]string{}
+			for _, catalog := range catalogConfig.Catalogs {
+				errs := []string{}
+				if catalog.Destination.BaseImage == "" {
```
everettraven, line 88: IIRC, this field was commented out as part of https://github.com/operator-framework/operator-registry/pull/1090

cmd/opm/alpha/template/composite.go
```diff
@@ -73,5 +180,7 @@ and a 'composite template' file`,
 	cmd.Flags().BoolVar(&validate, "validate", true, "whether or not the created FBC should be validated (i.e 'opm validate')")
 	cmd.Flags().StringVarP(&compositeFile, "composite-config", "c", "composite.yaml", "File to use as the composite configuration file")
 	cmd.Flags().StringVarP(&catalogFile, "catalog-config", "f", "catalogs.yaml", "File to use as the catalog configuration file")
+	cmd.Flags().StringVar(&dockerfileDir, "dockerfile", "", "Generates a Dockerfile in the provided directory")
+	cmd.Flags().StringVar(&devfileDir, "devfile", "", "Generates a Devfile in the provided directory")
```
everettraven, line 184: I could be wrong here, but I thought we already know what directories we would need to output these files to? If so, I think these should be boolean flags to signal whether or not to include the creation of them in the correct output directories. IIRC the structure of the output with both devfiles and a dockerfile should look something like:
```sh
catalogs
├── foo
│   ├── catalog
│   │   └── README.md
│   └── devfile.yaml
├── bar
│   ├── catalog
│   │   └── README.md
│   └── devfile.yaml
└── catalog.Dockerfile
```
With this expected directory structure, I *think* we should be able to parse all the correct directory information from the various config files

cmd/opm/generate/cmd.go
```diff
@@ -87,6 +87,56 @@ value of each duplicate key will be added to the generated Dockerfile.
...
+		baseImage string
+	)
+	cmd := &cobra.Command{
+		Use:   "devfile <dcRootDir>",
+		Args:  cobra.ExactArgs(1),
+		Short: "Generate a Devfile for a declarative config index",
+		Long: `Generate a Devfile for a declarative config index.
+
+This command creates a Devfile in the same directory as the <dcRootDir>
+(named <dcDirName>.Dockerfile) that can be used to build the index. If a
```
everettraven, line 101: devfile.yaml?
```suggestion
(named devfile.yaml) that can be used to build the index. If a
```

cmd/opm/generate/cmd.go
```diff
@@ -87,6 +87,56 @@ value of each duplicate key will be added to the generated Dockerfile.
...
+Devfile with the same name already exists, this command will fail.
+`,
+		RunE: func(_ *cobra.Command, args []string) error {
+			fromDir := filepath.Clean(args[0])
+
+			dir, indexName := filepath.Split(fromDir)
+			dockerfilePath := filepath.Join(dir, fmt.Sprintf("%s.Devfile", indexName))
```
everettraven, line 108:
```suggestion
			devfilePath := filepath.Join(dir, "devfile.yaml")
```

cmd/opm/generate/cmd.go
```diff
@@ -87,6 +87,56 @@ value of each duplicate key will be added to the generated Dockerfile.
...
+				IndexDir: indexName,
+				Writer:   f,
+			}
+			if err := gen.Run(); err != nil {
+				log.Fatal(err)
+			}
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&baseImage, "binary-image", "i", containertools.DefaultBinarySourceImage, "Image in which to build catalog.")
```
everettraven, line 136: I don't see anywhere this flag value is being used by the devfile command

everettraven (commented):
Looks like maybe this needs a rebase? There are a lot of errors that look like they are from this PR being based on an out of date branch

alpha/action/generate_devfile.go
```diff
@@ -0,0 +1,53 @@
...
+func (i GenerateDevfile) validate() error {
+	if i.IndexDir == "" {
+		return fmt.Errorf("index directory is unset")
+	}
+	return nil
+}
+
+const devfileTmpl = `schemaVersion: 2.2.0
+metadata:
+  name: fbc-4.11
```
grokspawn, line 37: IMOs...

`name` should be a parameter, which in the composite case should be provided from components[].name
`displayName` should be a parameter, which in the composite case should be the same as `name`
`provider` should be a parameter since it's unbounded text and (I don't think) related to the processing logic connected to devfiles in RHTAP. 
`buildContext` should be a parameter, which in the composite case should be from catalogs.yaml `catalogs[].destination.workingDir`. 
`components[].image.ImageName` _may_ be desired to be parameterized, but since this is alpha and we're dry-fitting it I think it's probably fine not to expose it for now, and to iterate as necessary in future.

grokspawn (changes requested):

cmd/opm/alpha/template/composite.go
```diff
@@ -26,6 +32,37 @@ and a 'composite template' file`,
...
+				defer tempCatalog.Close()
+			} else {
+				tempResp, err := http.Get(catalogURI.String())
+				if err != nil {
+					log.Fatalf("fetching remote catalog config file %q: %v", catalogFile, err)
+				}
+				tempCatalog = tempResp.Body
+				defer tempCatalog.Close()
+			}
+			catalogData := tempCatalog
```
grokspawn, line 52: This all moves into the API: https://github.com/operator-framework/operator-registry/blob/7629c6f991fc8dcce91bcb5ee4e998678bbd7779/alpha/template/composite/composite.go#L88

cdjohnson:
Can I ask what this new feature is for?  I don't see an EP, Issue or any other breadcrumbs around this and the use case.

grokspawn:
> Can I ask what this new feature is for? I don't see an EP, Issue or any other breadcrumbs around this and the use case.

We've had several requests about how operator authors can prep their FBC repos for catalog publication pipelines, including Eclipse Che / RH Trusted Application Pipeline / Tekton.   This is an early evaluation to see if there's benefit in providing the capability to produce OCI-generating devfiles to facilitate these connections.  

The idea is that authors already have to manage their FBC upgrade graphs separately from their bundles, so if we can find a way to do it which doesn't require them to hand-manage additional artifacts or additional tools, there'd be reduced friction to adoption.  Since the composite template helps authors coordinate multiple catalog contributions, this subcommand leverages that knowledge to generate contribution-specific devfiles.

cmd/opm/generate/cmd.go
```diff
@@ -87,6 +87,56 @@ value of each duplicate key will be added to the generated Dockerfile.
...
+
+This command creates a Devfile in the same directory as the <dcRootDir>
+(named devfile.yaml) that can be used to build the index. If a
+Devfile with the same name already exists, this command will fail.
+`,
+		RunE: func(_ *cobra.Command, args []string) error {
+			fromDir := filepath.Clean(args[0])
+
+			dir, indexName := filepath.Split(fromDir)
+			dockerfilePath := filepath.Join(dir, "devfile.yaml")
```
everettraven, line 108: Nit:
```suggestion
			devfilePath := filepath.Join(dir, "devfile.yaml")
```

cmd/opm/generate/cmd.go
everettraven: If we are going to continue with this route of adding this new subcommand, I think we ought to add a few more flags that correspond to different options for the generated devfile and provide some sensible defaults. Mostly for the `Name`, `Provider`, `BuildCTX` fields in the `GenerateDevfile` type.

grokspawn:
Last we tagged up, I believe the consensus was that we would not merge this PR.  Instead we would ticket new effort to capture the goal in a separate structure (for e.g. instead of `opm alpha generate devfile` we would have something like `pipeinator generate-devfile`. 
The remaining effort on this PR should just be related to ensuring that we have either responded to (and marked as resolved) all comments, and pushed here all the latest-and-greatest. 
Then we'll close-not-merged this PR, and spin a new one that is exclusive to pipeline-fitting so that we keep the catalog versus pipeline concerns separated.

grokspawn:
with so many caveats about immaturity, etc., [this early draft PR](https://github.com/operator-framework/operator-registry/pull/1121) is the direction I see this going.  

That PR generates a new top-level command 'pipe-fitter' out of the operator-registry repo, generated via `go build -o bin/pipe-fitter cmd/alpha/pipe-fitter/main.go`

Issue #1114, opened 2023-06-16
namespace-conflict when importing both client and bundle 


KevFan:
On Go1.19, importing both` "github.com/operator-framework/operator-registry/pkg/client"` and `"github.com/operator-framework/operator-registry/pkg/lib/bundle"` will result in a `#namespace-conflict` error:

```shell
panic: proto: file "grpc/health/v1/health.proto" has a name conflict over grpc.health.v1.HealthCheckResponse
        previously from: "github.com/operator-framework/operator-registry/pkg/api/grpc_health_v1"
        currently from:  "google.golang.org/grpc/health/grpc_health_v1"
See https://protobuf.dev/reference/go/faq#namespace-conflict


goroutine 1 [running]:
google.golang.org/protobuf/reflect/protoregistry.glob..func1({0x1065e3b60?, 0x14000621050?}, {0x1065e3b60?, 0x14000621090})
        /Users/chfan/.gvm/pkgsets/go1.20.1/global/pkg/mod/google.golang.org/protobuf@v1.29.1/reflect/protoregistry/registry.go:56 +0x200
google.golang.org/protobuf/reflect/protoregistry.(*Files).RegisterFile.func1({0x1066037b8, 0x14000682410})
        /Users/chfan/.gvm/pkgsets/go1.20.1/global/pkg/mod/google.golang.org/protobuf@v1.29.1/reflect/protoregistry/registry.go:154 +0x23c
google.golang.org/protobuf/reflect/protoregistry.rangeTopLevelDescriptors({0x1066098e0, 0x14000550c40}, 0x140008beb40)
        /Users/chfan/.gvm/pkgsets/go1.20.1/global/pkg/mod/google.golang.org/protobuf@v1.29.1/reflect/protoregistry/registry.go:417 +0x13c
google.golang.org/protobuf/reflect/protoregistry.(*Files).RegisterFile(0x1400000e150, {0x1066098e0?, 0x14000550c40?})
        /Users/chfan/.gvm/pkgsets/go1.20.1/global/pkg/mod/google.golang.org/protobuf@v1.29.1/reflect/protoregistry/registry.go:149 +0x5f4
google.golang.org/protobuf/internal/filedesc.Builder.Build({{0x1062c6b09, 0x2c}, {0x1072d0d00, 0x22d, 0x22d}, 0x1, 0x2, 0x0, 0x1, {0x1065e86a8, ...}, ...})
[...]

PR #1115, opened 2023-06-16
fix: resolve namespace conflict with google.golang.org/grpc/health/grpc_health_v1

KevFan:

There is a `namespace-conflict` issue with `google.golang.org/grpc/health/grpc_health_v1` when importing both "github.com/operator-framework/operator-registry/pkg/client" and "github.com/operator-framework/operator-registry/pkg/lib/bundle"

Renaming the conflicts will prevent this and allows importing both these packages. 

This can be testing by checking out this branch and adding a `replaces` field to the `go.mod` in https://github.com/KevFan/operator-registry-health-conflict
```
replace github.com/operator-framework/operator-registry => ../operator-registry
```
Running it should no longer give the error:
```
go mod tidy; go mod vendor; go run ./main/main.go
```

**Reviewer Checklist**

Closes https://github.com/operator-framework/operator-registry/issues/1114


joelanford:
Hi @KevFan! Thanks for reporting this issue and attempting a fix for it! I also encountered this problem while implementing #1165.

I have fixed it there by just removing the package we copied into `pkg/api/grpc_health_v1` and changing all importers to use the official package.

Do you think that fix will resolve the problem you were having?

PR #1128, opened 2023-07-25
[BUGFIX] Allow absolute filepaths for resolving catalog config

theishshah:
Adds an additional check to the Catalog Config fetching function to allow absolute file paths to be resolved. 

When adding support for remote catalog configurations, the resolver lost the ability to use absolute filepaths and could only handle relative paths.

### DIFF
--- a/alpha/template/composite/composite.go
+++ b/alpha/template/composite/composite.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 
 	"github.com/operator-framework/operator-registry/pkg/image"
 	"k8s.io/apimachinery/pkg/util/yaml"
@@ -88,7 +89,7 @@ type HttpGetter interface {
 func FetchCatalogConfig(path string, httpGetter HttpGetter) (io.ReadCloser, error) {
 	var tempCatalog io.ReadCloser
 	catalogURI, err := url.ParseRequestURI(path)
-	if err != nil {
+	if err != nil || filepath.IsAbs(path) {
 		tempCatalog, err = os.Open(path)
 		if err != nil {
 			return nil, fmt.Errorf("opening catalog config file %q: %v", path, err)
--- a/alpha/template/composite/composite.go
+++ b/alpha/template/composite/composite.go
@@ -98,9 +98,9 @@ func FetchCatalogConfig(path string, httpGetter HttpGetter) (io.ReadCloser, erro
 		if err != nil {
 			return nil, fmt.Errorf("opening catalog config file %q: %v", path, err)
 		}
+	} else {
 		// Evalute remote catalog config
 		// If URi is valid, execute fetch
-	} else {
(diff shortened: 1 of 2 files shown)


everettraven (commented):
The changes look good to me. Do we need to make any test updates to verify that this change does indeed function as expected?

grokspawn:
I'd like to see a utest for this case since this (fetch or use local reference) is central to this template's usefulness.

everettraven:
>I'd like to see a utest for this case since this (fetch or use local reference) is central to this template's usefulness.

FWIW I think this is a test case that should be updated to force the `fakeGetter` to error and ensure we never see any errors (i.e we are using `os.Open` instead of the HTTP `.Get()`):

https://github.com/operator-framework/operator-registry/blob/68e13df96590977370ffcd1a8e9ff76e0f2a03f2/alpha/template/composite/composite_test.go#L567-L579

alpha/template/composite/composite_test.go
```diff
@@ -589,10 +591,21 @@ func TestFetchCatalogConfig(t *testing.T) {
...
+			fakeGetter: &fakeGetter{
+				catalog:     validCatalog,
+				shouldError: true,
+			},
+			createFile: true,
+			assertions: func(t *testing.T, rc io.ReadCloser, err error) {
+				require.NoError(t, err)
+				require.NotNil(t, rc)
+			},
+		},
```
everettraven, line 606: I'm actually not sure this whole test case is necessary. Looking at lines 612 - 621 it looks like if we end up creating the file we actually set the file path provided to the `FetchCatalogConfig` function to be the absolute path. This makes me think that we should add `shouldError: true` under https://github.com/operator-framework/operator-registry/pull/1128/files#diff-0299f965ef1d5dc96c54539eccbbcbf7a35e5f38255886bcbb6f8c39edf3364fR574 and that would meet this criteria.

I think the "Failed file fetch" actually tests the relative path case by ensuring it attempts to get a relative path that doesn't exist and errors out.

everettraven (approved):

alpha/template/composite/composite.go
```diff
@@ -85,14 +86,20 @@ type HttpGetter interface {
...
-	if err != nil {
+	// Evalute local catalog config
+	// URI parse will fail on relative filepaths
+	// Check if path is an absolute filepath
+	if err != nil || filepath.IsAbs(path) {
 		tempCatalog, err = os.Open(path)
 		if err != nil {
 			return nil, fmt.Errorf("opening catalog config file %q: %v", path, err)
 		}
+		// Evalute remote catalog config
```
dinhxuanvu, line 101: Are those comments for the `else` case? If so, please correct the indentation or move the comments inside the else bracket is also fine.

dinhxuanvu (changes requested):

dinhxuanvu (approved):
/approve
The bot will take care of the commit squashing after merge.

PR #1137, opened 2023-08-22
Enforce Argument Count in Alpha Template Command

theishshah:
This is a simple change in order to force users to provide the appropriate context as an argument when running the command preventing the tool from hanging without it.

This solution was arrived upon in place of checking for the context's existence post CLI.


theishshah:
Updated to include this fix for basic templates as well

theishshah:
Closing as this is working as intended, mis-reported mis-interpereted bug

PR #1138, opened 2023-08-22
fix CSV unmarshaling error with custom unmarshaler with pretty errors

grokspawn:

**Reviewer Checklist**


grokspawn:
I sorted out the sequence, and maybe the multiple 'helper' functions aren't necessary, but it aided my investigation.  It turns out that because there is only one named field in the struct, and it is a `json.RawMessage` we have to be careful to extract it by name.  

Now only a single utest is failing: `TestExportForBundleWithBadImage`, because it attempts a registry pull of image `foo` and it never comes back from `syscall`. 

I'm investigating, but here's the trace in case something jumps out to someone. 

```goroutine 69 [syscall, 10 minutes]:
syscall.syscall6(0xc000723728?, 0x104bd5e94?, 0x1081818e0?, 0x10986cf40?, 0xc0007237a8?, 0x104bd60fc?, 0x104b8bfd0?)
	/Users/jordan/.gvm/gos/go1.20.7/src/runtime/sys_darwin.go:45 +0x68
syscall.wait4(0x108149640?, 0xc0007237a8, 0x0?, 0xc000726000)
	/Users/jordan/.gvm/gos/go1.20.7/src/syscall/zsyscall_darwin_arm64.go:43 +0x64
syscall.Wait4(0xc0002ba318?, 0xc000723808, 0xc000121b30?, 0xc000294660?)
	/Users/jordan/.gvm/gos/go1.20.7/src/syscall/syscall_bsd.go:144 +0x48
os.(*Process).wait(0xc000294660)
	/Users/jordan/.gvm/gos/go1.20.7/src/os/exec_unix.go:43 +0xbc
os.(*Process).Wait(...)
	/Users/jordan/.gvm/gos/go1.20.7/src/os/exec.go:132
os/exec.(*Cmd).Wait(0xc000121a20)
	/Users/jordan/.gvm/gos/go1.20.7/src/os/exec/exec.go:890 +0x74
os/exec.(*Cmd).Run(0xc00011e930?)
	/Users/jordan/.gvm/gos/go1.20.7/src/os/exec/exec.go:590 +0x58
os/exec.(*Cmd).CombinedOutput(0xc000121a20)
	/Users/jordan/.gvm/gos/go1.20.7/src/os/exec/exec.go:1005 +0x218
github.com/operator-framework/operator-registry/pkg/containertools.(*ContainerCommandRunner).Pull(0xc00000d560, {0x1069093d1, 0x3})
	/Users/jordan/devel/operator-registry/pkg/containertools/runner.go:90 +0x318
github.com/operator-framework/operator-registry/pkg/image/execregistry.(*Registry).Pull(0xc00000d590, {0x3?, 0xc00070dd10?}, {0x1071d3b68, 0xc0003782b0})
	/Users/jordan/devel/operator-registry/pkg/image/execregistry/registry.go:38 +0x78
github.com/operator-framework/operator-registry/pkg/lib/bundle.(*BundleExporter).Export(0xc000723e60, 0x1, 0x0)
	/Users/jordan/devel/operator-registry/pkg/lib/bundle/exporter.go:67 +0x4c0
github.com/operator-framework/operator-registry/pkg/lib/bundle.TestExportForBundleWithBadImage(0x0?)
	/Users/jordan/devel/operator-registry/pkg/lib/bundle/exporter_test.go:12 +0xb0
testing.tRunner(0xc0001a0680, 0x1071c7878)
	/Users/jordan/.gvm/gos/go1.20.7/src/testing/testing.go:1576 +0x18c
created by testing.(*T).Run
	/Users/jordan/.gvm/gos/go1.20.7/src/testing/testing.go:1629 +0x5e8
```

grokspawn:
/unhold

> Now only a single utest is failing: `TestExportForBundleWithBadImage`, because it attempts a registry pull of image `foo` and it never comes back from `syscall`.

Evidently this issue only exists on mac(??) and it runs cleanly on linux.  Going to get this into review while I continue to investigate.

pkg/registry/csv.go
```diff
@@ -412,3 +414,61 @@ func (csv *ClusterServiceVersion) GetSubstitutesFor() string {
...
+
+	if err := csv.UnmarshalSpec(data); err != nil {
+		return err
+	}
+	if err := csv.UnmarshalTypeMeta(data); err != nil {
+		return err
+	}
+	if err := csv.UnmarshalObjectMeta(data); err != nil {
+		return err
+	}
```
everettraven, line 428: It looks like this will always return on the first error. Something maybe worth considering for a better dx is having some form of error aggregation that can return all errors that occurred in unmarshalling. Non blocking though and could be done in a follow-up IMO

pkg/registry/csv.go
```diff
@@ -412,3 +414,61 @@ func (csv *ClusterServiceVersion) GetSubstitutesFor() string {
...
+	for k, v := range m {
+		switch k {
+		case "spec":
+			var spec json.RawMessage
+			if err := json.Unmarshal(v, &spec); err != nil {
+				return errors.New(prettyunmarshaler.NewJSONUnmarshalError(v, err).Pretty())
+			}
+			csv.Spec = spec
+		}
+	}
```
everettraven, line 448: If we only care about the `spec` key is this loop necessary? Maybe something like
```go
specVal, ok := m["spec"]
if !ok {
  // seems like the current intention is to continue on if the spec field isn't found
  // but could also return an error here if we wanted
  return nil
}

var spec json.RawMessage
if err := json.Unmarshal(v, &spec); err != nil {
	return errors.New(prettyunmarshaler.NewJSONUnmarshalError(v, err).Pretty())
}
csv.Spec = spec
```
is better suited for this case?

grokspawn:
> >    Now only a single utest is failing: TestExportForBundleWithBadImage, because it attempts a registry pull of image foo and it never comes back from syscall.

> Evidently this issue only exists on mac(??) and it runs cleanly on linux. Going to get this into review while I continue to investigate.

Turns out to be a bad docker install on my mac. :( 
Resolved, and all tests pass now.

pkg/registry/csv.go
grokspawn: In this case we specifically wanted to return on first error, since it means less output. :) 
But I can definitely agree that we could extend this to an aggregate and return all types in future.

grokspawn: Note though that the original code using the default unmarshaler would have the same behavior where it would unmarshal only that which was present in the blob; a missing spec would still not be an error until we tried to use it later.  So this PR preserves the earlier behavior.

everettraven: Ah, good to know. Will mark this as resolved since it wasn't really meant to be an action item now but just something to consider for the future

pkg/registry/csv.go
```diff
@@ -412,3 +414,60 @@ func (csv *ClusterServiceVersion) GetSubstitutesFor() string {
...
+
+	specRaw, ok := m["spec"]
+	if !ok {
+		return nil
+	}
+	var spec json.RawMessage
+	if err := json.Unmarshal(specRaw, &spec); err != nil {
+		return errors.New(prettyunmarshaler.NewJSONUnmarshalError(specRaw, err).Pretty())
+	}
```
everettraven, line 446: Do we need this unmarshal if the first unmarshal does a `map[string]json.RawMessage`? `specRaw` would already be a `json.RawMessage` here so could we just set `csv.Spec = specRaw` and it work the same?

awgreene (approved):
/approve

awgreene:
Ah @everettraven already approved
/lgtm

PR #1139, opened 2023-09-05
export parsed composite specs

grokspawn:
export parsed yaml/json specs so that other tooling can leverage it

tap-fitter tooling will be able to generate some pipeline-specific pieces from the composite template, but the current implementation parses and executes the specs in an uninterrupted flow

example:
```golang
c := composite.NewTemplate(
         composite.WithCatalogFile(tempCatalog),
         composite.WithContributionFile(compositeReader)
)
specs, err := c.Parse()
```

**Reviewer Checklist**

### DIFF
--- a/alpha/template/composite/composite.go
+++ b/alpha/template/composite/composite.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


everettraven (approved):
Looks good to me.

I do have a teeny nit (totally non-blocking) - I _personally_ prefer to have methods co-located with the type they are associated with because it makes it easier for me to dig into the code due to less jumping between different files.

grokspawn:
> Looks good to me.
> 
> I do have a teeny nit (totally non-blocking) - I _personally_ prefer to have methods co-located with the type they are associated with because it makes it easier for me to dig into the code due to less jumping between different files.

I'm of two minds on this.  But we already had a `types.go` file for composite templates, and I thought it was confusing to have _some_ types there, but not others.  So when considering whether to obsolete the module and move all the types here, or add a couple more types to that one .... I landed on the 'consolidation' move.  

If we find that it causes issue later, I'm totally OK with a future PR to move them out of that file so we can have everything in one place.

PR #1143, opened 2023-09-22
Remove defunct ref-style `olm.bundle.object`

joelanford:
This commit removes a feature of FBC that has never actually been used in practice: the ability to reference a file in an `olm.bundle.object` property, where the path is relative to the directory in which the file containing the `olm.bundle.object` property exists. This was originally intended to be a way to avoid bloating the FBC, but it's presence has caused two problems:

  1. It hinted that it would be okay for third-party properties and schemas to reference files in a similar way.
  2. Because of (1), we have never really been able to make assumptions that would enable us to migrate and re-write FBC in a different hierarchy, which has been limiting.

In short, it imposes a burden on catalog maintainers to keep a catalog in a filesystem structure that is imposed by the author of the catalog contribution.

In practice, ref-style `olm.bundle.object` properties have never been used (as far as I'm aware), because no tooling has ever produced that style, and no one I have heard of is using other methods to render bundles into an FBC.

Lastly, with the recent addition of the `olm.csv.metadata` property, the useful life of the `olm.bundle.object` property (which has always been alpha) is nearing an end.

NOTE: If this PR merges, we need to update [docs about `olm.bundle.object`](https://github.com/operator-framework/olm-docs/blob/master/content/en/docs/Reference/file-based-catalogs.md#olmbundleobject-alpha)

**Reviewer Checklist**

### DIFF
--- a/alpha/action/migrate.go
+++ b/alpha/action/migrate.go
@@ -31,9 +31,7 @@ func (m Migrate) Run(ctx context.Context) error {
 	r := Render{
 		Refs: []string{m.CatalogRef},
 
-		// Only allow sqlite images and files to be migrated. Other types cannot
-		// always be migrated cleanly because they may contain file references.
-		// Rendered sqlite databases never contain file references.
+		// Only allow sqlite images and files to be migrated.
 		AllowedRefMask: RefSqliteImage | RefSqliteFile,
 
 		skipSqliteDeprecationLog: true,
--- a/alpha/action/render_test.go
+++ b/alpha/action/render_test.go
@@ -294,8 +294,8 @@ func TestRender(t *testing.T) {
 							property.MustBuildGVKRequired("test.bar", "v1alpha1", "Bar"),
 							property.MustBuildPackage("foo", "0.1.0"),
 							property.MustBuildPackageRequired("bar", "<0.1.0"),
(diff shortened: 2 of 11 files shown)


fgiloux:
It would probably be a good idea to open a PR to amend [the doc](https://github.com/operator-framework/olm-docs/blob/master/content/en/docs/Reference/file-based-catalogs.md#olmbundleobject-alpha) accordingly.

everettraven (approved):

joelanford:
Proposed documentation updates here: https://github.com/operator-framework/olm-docs/pull/312

cdjohnson:
I'll reach out to our teams to verify we aren't building any FBC catalogs with references.  We definitely toyed with this at the beginning, but also ran into various issues.

cdjohnson:
@joelanford:  Looks like we're OK from the IBM side.  The products that have adopted FBC are using `opm render` for the most part and using `data` vs. `ref`

Note:  I realize that the `olm.bundle.object` properties are "alpha", but because they are required to be used to use FBC with OLM, which is GA, it seems like it would be prudent to rev the major version when a release is cut.  In hindsight, this probably should have also been done when `opm render <bundle-image>` was changed to stop using `olm.bundle.object` in favor of `olm.csv.metadata`

PR #1144, opened 2023-09-22
migrate: support migration of FBC to latest preferred FBC

joelanford:
This commit adds support for migrating FBC to the latest preferred FBC
contents. Note that sqlite and bundle inputs are always rendered using
the latest preferred FBC contents.

The migrate command is updated to now support FBC images and directories
as input (only sqlite was supported prior), such that the written output
will always be migrated.

The render command is updated with a `--migrate` flag that allows a
caller to opt into migration during rendering.

Under the hood, both of these subcommands use the action.Render struct,
which has a new `Migrate` boolean field that callers can use to
enable/disable the migration behavior.

---

This is specifically useful right now to convert existing FBCs using `olm.bundle.object` to use of `olm.csv.metadata`, which can reduce catalog size significantly. At the time of the PR being opened, it reduces the operatorhub catalog from 98M to 23M (~76% reduction in size)

NOTE: This builds upon #1143, which is necessary to align on the tenet that FBC directory structure can change without changing the validity or meaning of the catalog.

**Reviewer Checklist**

### DIFF
--- a/alpha/action/migrate.go
+++ b/alpha/action/migrate.go
@@ -31,9 +31,7 @@ func (m Migrate) Run(ctx context.Context) error {
 	r := Render{
 		Refs: []string{m.CatalogRef},
 
-		// Only allow sqlite images and files to be migrated. Other types cannot
-		// always be migrated cleanly because they may contain file references.
-		// Rendered sqlite databases never contain file references.
+		// Only allow sqlite images and files to be migrated.
 		AllowedRefMask: RefSqliteImage | RefSqliteFile,
 
 		skipSqliteDeprecationLog: true,
--- a/alpha/action/migrate.go
+++ b/alpha/action/migrate.go
@@ -29,12 +29,11 @@ func (m Migrate) Run(ctx context.Context) error {
 	}
 
 	r := Render{
-		Refs: []string{m.CatalogRef},
+		Refs:    []string{m.CatalogRef},
+		Migrate: true,
 
-		// Only allow sqlite images and files to be migrated.
-		AllowedRefMask: RefSqliteImage | RefSqliteFile,
-
-		skipSqliteDeprecationLog: true,
+		// Only allow catalogs to be migrated.
+		AllowedRefMask: RefSqliteImage | RefSqliteFile | RefDCImage | RefDCDir,
 	}
 	if m.Registry != nil {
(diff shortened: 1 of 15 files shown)


everettraven (approved):

cdjohnson:
@joelanford Thanks for this PR.  I think this will be very useful.

I did some testing, and I noticed that you may have inadvertently changed the default for rendering a bundle image.  Is this on purpose?  I would think that the `--migrate` option also applied to a bundle image reference.

```
img="icr.io/cpopen/ibm-mq-operator-bundle@sha256:11639e73211324978bde65fbe35d7f71ee2b7efd96e4d5d1c8158855256c2cae"

opmnew=~/tools/opm/dev/opm
opmold=~/tools/opm/1.26.1/opm
fbcdir=./configs/ibm-mq

# Renders olm.bundle.properties - OK
$opmold render $img > bundle-render-old.yaml

# Renders olm.csv.metadata - ERROR
$opmnew render $img > bundle-render-new.yaml

# Renders olm.csv.metadata - OK
$opmnew render --migrate $img  > bundle-render-migrate-new.yaml

# Renders olm.bundle.properties - OK
$opmold render $fbcdir > catalog-render-old.yaml

# Renders olm.bundle.properties - OK
$opmnew render $fbcdir > catalog-render-new.yaml

# Renders olm.csv.metadata - OK
$opmnew render --migrate $fbcdir > catalog-render-migrate-new.yaml
```

joelanford:
> \# Renders olm.csv.metadata - ERROR
> $opmnew render $img > bundle-render-new.yaml

You're saying that this command produces `olm.csv.metadata` even without `--migrate=true`? That's the behavior I intended. That has been the case since 1.28+.

The `--migrate` flag is a no-op when the inputs to `render` are sqlite DBs and bundles. The reasoning is essentially that `opm` should be converting sqlite DBs and bundles directly to the latest/best formats, rather than doing the extra work of converting them to the oldest format and then running all of the migrations.

cdjohnson:
> You're saying that this command produces olm.csv.metadata even without --migrate=true? That's the behavior I intended. That has been the case since 1.28+.

Yes, I didn't expect that.  I happened to be testing with 1.26.1.  I retested with 1.29.0, and you're right, and it's clear in that PR.  Just missed it.

PR #1151, opened 2023-10-24
deprecation schema & import/render support

grokspawn:
[parent tracker](https://github.com/operator-framework/operator-registry/issues/1155)

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/declcfg.go
+++ b/alpha/declcfg/declcfg.go
@@ -16,16 +16,18 @@ import (
 )
 
 const (
-	SchemaPackage = "olm.package"
-	SchemaChannel = "olm.channel"
-	SchemaBundle  = "olm.bundle"
+	SchemaPackage     = "olm.package"
+	SchemaChannel     = "olm.channel"
+	SchemaBundle      = "olm.bundle"
+	SchemaDeprecation = "olm.catalog.deprecation"
 )
 
 type DeclarativeConfig struct {
-	Packages []Package
-	Channels []Channel
-	Bundles  []Bundle
-	Others   []Meta
+	Packages     []Package
+	Channels     []Channel
+	Bundles      []Bundle
+	Deprecations []Deprecation
+	Others       []Meta
 }
 
 type Package struct {
@@ -88,6 +90,19 @@ type Bundle struct {
 type RelatedImage struct {
 	Name  string `json:"name"`
 	Image string `json:"image"`
+}
+
+type Deprecation struct {
+	Schema       string             `json:"schema"`
+	Package      string             `json:"package"`
+	Name         string             `json:"name,omitempty"`
+	Deprecations []DeprecationEntry `json:"deprecations"`
+}
+
+type DeprecationEntry struct {
+	Schema  string          `json:"schema"`
+	Name    string          `json:"name,omitempty"`
+	Message json.RawMessage `json:"message"`
 }
 
 type Meta struct {
@@ -180,4 +195,12 @@ func extractUniqueMetaKeys(blobMap map[string]any, m *Meta) error {
 		*ptr = v
 	}
 	return nil
+}
+
+func (destination *DeclarativeConfig) Merge(src *DeclarativeConfig) {
+	destination.Packages = append(destination.Packages, src.Packages...)
+	destination.Channels = append(destination.Channels, src.Channels...)
+	destination.Bundles = append(destination.Bundles, src.Bundles...)
+	destination.Others = append(destination.Others, src.Others...)
+	destination.Deprecations = append(destination.Deprecations, src.Deprecations...)
 }
--- a/alpha/declcfg/load.go
+++ b/alpha/declcfg/load.go
@@ -227,12 +227,8 @@ func mergeCfgs(ctx context.Context, cfgChan <-chan *DeclarativeConfig, fcfg *Dec
 			if !ok {
 				return nil
 			}
-			fcfg.Packages = append(fcfg.Packages, cfg.Packages...)
-			fcfg.Channels = append(fcfg.Channels, cfg.Channels...)
-			fcfg.Bundles = append(fcfg.Bundles, cfg.Bundles...)
-			fcfg.Others = append(fcfg.Others, cfg.Others...)
+			fcfg.Merge(cfg)
 		}
-
 	}
 }
 
@@ -297,6 +293,12 @@ func LoadReader(r io.Reader) (*DeclarativeConfig, error) {
 				return fmt.Errorf("parse bundle: %v", err)
 			}
 			cfg.Bundles = append(cfg.Bundles, b)
+		case SchemaDeprecation:
+			var d Deprecation
+			if err := json.Unmarshal(in.Blob, &d); err != nil {
+				return fmt.Errorf("parse deprecation: %w", err)
+			}
+			cfg.Deprecations = append(cfg.Deprecations, d)
 		case "":
 			return fmt.Errorf("object '%s' is missing root schema field", string(in.Blob))
 		default:
--- a/alpha/declcfg/load_test.go
+++ b/alpha/declcfg/load_test.go
@@ -97,13 +97,14 @@ func TestLoadReader(t *testing.T) {
 
 func TestWalkMetasFS(t *testing.T) {
 	type spec struct {
-		name              string
-		fsys              fs.FS
-		assertion         require.ErrorAssertionFunc
-		expectNumPackages int
-		expectNumChannels int
-		expectNumBundles  int
-		expectNumOthers   int
+		name                  string
+		fsys                  fs.FS
+		assertion             require.ErrorAssertionFunc
+		expectNumPackages     int
+		expectNumChannels     int
+		expectNumBundles      int
+		expectNumDeprecations int
+		expectNumOthers       int
 	}
 	specs := []spec{
 		{
@@ -122,19 +123,20 @@ func TestWalkMetasFS(t *testing.T) {
 			assertion: require.Error,
 		},
 		{
-			name:              "Success/ValidDir",
-			fsys:              validFS,
-			assertion:         require.NoError,
-			expectNumPackages: 3,
-			expectNumChannels: 0,
-			expectNumBundles:  12,
-			expectNumOthers:   1,
+			name:                  "Success/ValidDir",
+			fsys:                  validFS,
+			assertion:             require.NoError,
+			expectNumPackages:     3,
+			expectNumChannels:     0,
+			expectNumBundles:      12,
+			expectNumDeprecations: 1,
+			expectNumOthers:       1,
 		},
 	}
 
 	for _, s := range specs {
 		t.Run(s.name, func(t *testing.T) {
-			numPackages, numChannels, numBundles, numOthers := 0, 0, 0, 0
+			numPackages, numChannels, numBundles, numDeprecations, numOthers := 0, 0, 0, 0, 0
 			err := WalkMetasFS(s.fsys, func(path string, meta *Meta, err error) error {
 				if err != nil {
 					return err
@@ -146,6 +148,8 @@ func TestWalkMetasFS(t *testing.T) {
 					numChannels++
 				case SchemaBundle:
 					numBundles++
+				case SchemaDeprecation:
+					numDeprecations++
 				default:
 					numOthers++
 				}
@@ -330,6 +334,18 @@ func TestLoadFS(t *testing.T) {
 					},
 					{
 						Schema: "olm.bundle",
+					},
+				},
+				Deprecations: []Deprecation{
+					{
+						Schema:  "olm.catalog.deprecation",
+						Package: "kiali",
+						Name:    "bobs-discount-name",
+						Deprecations: []DeprecationEntry{
+							{Schema: "olm.bundle", Name: "kiali-operator.v1.68.0", Message: json.RawMessage(`"kiali-operator.v1.68.0 is deprecated. Uninstall and install kiali-operator.v1.72.0 for support.\n"`)},
+							{Schema: "olm.package", Name: "kiali", Message: json.RawMessage(`"package kiali is end of life.  Please use 'kiali-new' package for support.\n"`)},
+							{Schema: "olm.channel", Name: "alpha", Message: json.RawMessage(`"channel alpha is no longer supported.  Please switch to channel 'stable'.\n"`)},
+						},
 					},
 				},
 				Others: []Meta{
@@ -881,6 +897,25 @@ present in the .indexignore file.`),
 	unrecognizedSchema = &fstest.MapFile{
 		Data: []byte(`{"schema":"olm.package"}{"schema":"unexpected"}{"schema":"olm.bundle"}`),
 	}
+	deprecations = &fstest.MapFile{
+		Data: []byte(`---
+schema: olm.catalog.deprecation
+package: kiali
+name: bobs-discount-name
+deprecations:
+- schema: olm.bundle
+  name: kiali-operator.v1.68.0
+  message: |
+     kiali-operator.v1.68.0 is deprecated. Uninstall and install kiali-operator.v1.72.0 for support.
+- schema: olm.package
+  name: kiali
+  message: |
+     package kiali is end of life.  Please use 'kiali-new' package for support.
+- schema: olm.channel
+  name: alpha
+  message: |
+     channel alpha is no longer supported.  Please switch to channel 'stable'.`),
+	}
 
 	validFS = fstest.MapFS{
 		".indexignore":     indexIgnore,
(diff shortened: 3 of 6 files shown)


alpha/declcfg/declcfg.go
```diff
@@ -16,16 +16,18 @@ import (
...
 
 type DeclarativeConfig struct {
-	Packages []Package
-	Channels []Channel
-	Bundles  []Bundle
-	Others   []Meta
+	Packages     []Package
+	Channels     []Channel
+	Bundles      []Bundle
+	Deprecations map[string]Deprecation // mapping package name to deprecation
```
stevekuznetsov, line 29: Comment is superfluous if you name it `DeprecationsByPackage`?

Alternatively, why is this not a field on `Package`?

alpha/declcfg/declcfg.go
```diff
@@ -181,3 +196,27 @@ func extractUniqueMetaKeys(blobMap map[string]any, m *Meta) error {
...
+	if err := destination.mergeDeprecations(src); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (destination *DeclarativeConfig) mergeDeprecations(src *DeclarativeConfig) error {
+	for p, d := range src.Deprecations {
+		if _, ok := destination.Deprecations[p]; ok {
+			return fmt.Errorf("deprecation for package %q already exists", d.Package)
```
stevekuznetsov, line 214: This is a surprising semantic for a method called `Merge` - why is it necessary?

alpha/declcfg/load.go
```diff
@@ -272,7 +268,7 @@ func extractCSV(objs []string) string {
 
 // LoadReader reads yaml or json from the passed in io.Reader and unmarshals it into a DeclarativeConfig struct.
 func LoadReader(r io.Reader) (*DeclarativeConfig, error) {
-	cfg := &DeclarativeConfig{}
+	cfg := &DeclarativeConfig{Deprecations: make(map[string]Deprecation)}
```
stevekuznetsov, line 271: I'm surprised you need to initialize the map here - why?

alpha/declcfg/load.go
```diff
@@ -297,6 +293,15 @@ func LoadReader(r io.Reader) (*DeclarativeConfig, error) {
 				return fmt.Errorf("parse bundle: %v", err)
 			}
 			cfg.Bundles = append(cfg.Bundles, b)
+		case SchemaDeprecation:
+			var d Deprecation
+			if err := json.Unmarshal(in.Blob, &d); err != nil {
+				return fmt.Errorf("parse deprecation: %v", err)
```
stevekuznetsov, line 299:
```suggestion
				return fmt.Errorf("parse deprecation: %w", err)
```

alpha/declcfg/load.go
```diff
@@ -297,6 +293,15 @@ func LoadReader(r io.Reader) (*DeclarativeConfig, error) {
...
+			}
+			if _, ok := cfg.Deprecations[d.Package]; ok {
+				return fmt.Errorf("deprecation for package %q already exists", d.Package)
```
stevekuznetsov, line 302: Packages, channels and bundles above don't seem to have this check. Why are deprecations different?

alpha/declcfg/load_test.go
```diff
@@ -337,6 +337,182 @@ func TestLoadFS(t *testing.T) {
...
... 0x2e, 0x35, 0x20, 0x31, 0x35, 0x2e, 0x35, 0x20, 0x30, 0x20, 0x30, 0x20, 0x31, 0x20, 0x34, 0x2e, 0x32, 0x35, 0x2d, 0x31, 0x30, 0x2e, 0x36, 0x39, 0x20, 0x31, 0x35, 0x2e, 0x36, 0x32, 0x20, 0x31, 0x35, 0x2e, 0x36, 0x32, 0x20, 0x30, 0x20, 0x30, 0x20, 0x31, 0x20, 0x2e, 0x37, 0x32, 0x20, 0x34, 0x2e, 0x36, 0x38, 0x22, 0x20, 0x66, 0x69, 0x6c, 0x6c, 0x3d, 0x22, 0x23, 0x33, 0x34, 0x38, 0x35, 0x34, 0x30, 0x22, 0x2f, 0x3e, 0x3c, 0x70, 0x61, 0x74, 0x68, 0x20, 0x64, 0x3d, 0x22, 0x4d, 0x31, 0x35, 0x20, 0x32, 0x33, 0x2e, 0x34, 0x38, 0x61, 0x31, 0x35, 0x2e, 0x35, 0x35, 0x20, 0x31, 0x35, 0x2e, 0x35, 0x35, 0x20, 0x30, 0x20, 0x30, 0x20, 0x31, 0x2d, 0x2e, 0x37, 0x32, 0x20, 0x34, 0x2e, 0x36, 0x38, 0x20, 0x31, 0x35, 0x2e, 0x35, 0x34, 0x20, 0x31, 0x35, 0x2e, 0x35, 0x34, 0x20, 0x30, 0x20, 0x30, 0x20, 0x31, 0x2d, 0x33, 0x2e, 0x35, 0x33, 0x2d, 0x31, 0x35, 0x2e, 0x33, 0x37, 0x41, 0x31, 0x35, 0x2e, 0x35, 0x20, 0x31, 0x35, 0x2e, 0x35, 0x20, 0x30, 0x20, 0x30, 0x20, 0x31, 0x20, 0x31, 0x35, 0x20, 0x32, 0x33, 0x2e, 0x34, 0x38, 0x22, 0x20, 0x66, 0x69, 0x6c, 0x6c, 0x3d, 0x22, 0x23, 0x37, 0x64, 0x62, 0x63, 0x34, 0x32, 0x22, 0x2f, 0x3e, 0x3c, 0x2f, 0x73, 0x76, 0x67, 0x3e}, MediaType: "image/svg+xml"}, Description: ""},
```
stevekuznetsov, line 346: Is the icon data important here?

alpha/declcfg/load_test.go
```diff
@@ -337,6 +337,182 @@ func TestLoadFS(t *testing.T) {
...
+							{Type: "olm.gvk", Value: json.RawMessage(`{"group":"etcd.database.coreos.com","kind":"EtcdBackup","version":"v1beta2"}`)},
+							{Type: "olm.channel", Value: json.RawMessage(`{"name":"clusterwide-alpha","replaces":"etcdoperator.v0.9.2-clusterwide"}`)},
+						},
+						RelatedImages: []RelatedImage{{Name: "etcdv0.9.2", Image: "quay.io/coreos/etcd-operator@sha256:66a37fd61a06a43969854ee6d3e21087a98b93838e284a6086b13917f96b0d9b"}},
+					},
+					{
+						Schema: "olm.bundle",
+					},
+				},
+				Deprecations: map[string]Deprecation{
```
stevekuznetsov, line 487: Why does this test load anything other than deprecations?

alpha/declcfg/load_test.go
```diff
@@ -961,6 +1166,15 @@ func TestLoadFile(t *testing.T) {
 			expectNumBundles:  6,
 			expectNumOthers:   0,
 		},
+		{
+			name:              "Success/ValidFile/Deprecations",
+			fsys:              validFSD,
+			path:              "deprecations.yaml",
+			assertion:         require.NoError,
+			expectNumPackages: 0,
+			expectNumBundles:  0,
```
stevekuznetsov, line 1175: Wouldn't it be useful to expect deprecations as opposed to not (only) expecting anything else?

alpha/declcfg/declcfg.go
grokspawn: It's not a field on Package because they are separate schema, and I wanted to avoid deprecations being written out under it when we encode it as json/yaml. 

I was arbitrarily not wanting a long variable name here, but you're right that it could be a self-documenting replacement for the comment.

alpha/declcfg/declcfg.go
grokspawn: That a merge can fail?  That we allow 0..1 of a Deprecations per Package?

alpha/declcfg/load.go
grokspawn: If you don't, then the merge generates a panic assigning to nil map. 
We could catch it when we sense our first deprecations schema instance, but it seemed cleaner to init here.

alpha/declcfg/load.go
grokspawn: Each of those types can have an arbitrary quantity but Deprecations are singletons per package.  One of the complexities here is keeping these closely-coupled objects disjoint.

alpha/declcfg/load_test.go
grokspawn: It's not mandatory, and it's noisy, so we don't always include it.

alpha/declcfg/load_test.go
grokspawn: Right now it's not so significant, but we will build on these to validate other characteristics where we validate the deprecations against the other declcfg content (for e.g. to ensure that deprecations only refer to objects present in the catalog).

alpha/declcfg/load.go
stevekuznetsov: Your merge code already has:

```go
		if destination.Deprecations == nil {
			destination.Deprecations = make(map[string]Deprecation)
		}
```

Are you certain it will fail? :)

I think it's nicer to always handle the `nil`-map there instead of hoping that everyone initializes it correctly.

alpha/declcfg/load.go
stevekuznetsov: I think it's true that some set of properties on a package, channel and bundle would be uniquely identifiable, yes? Is it allowed for a catalog to contain two bundles that are identical? If yes, why? If no, why is the validation for that not here? Can the validation for deprecations live where the validation for bundles does?

alpha/declcfg/declcfg.go
stevekuznetsov: Surprising because of https://github.com/operator-framework/operator-registry/pull/1151/files/2b2e3547018b958f3223a7c6a8b54f72aec3f5c2#r1374666692

alpha/declcfg/load.go
grokspawn: Son of a ... I thought I took that out in favor of the init policy.  That's why this is a draft, man.  It's all belt-and-suspenders up in here. :)

alpha/declcfg/load.go
grokspawn: Can-o-worms territory here. Historically the sole validation of declcfg here was "can we unmarshal it".  
For `opm validate` command, it's /slightly/ more, as [can we unmarshal and create a model.Model from it](https://github.com/operator-framework/operator-registry/blob/master/pkg/lib/config/validate.go). 

I wanted a loud/early/lightweight failure if we violated singleton-per-package.

grokspawn: There's no reason we couldn't implement a validation phase after completely processing the FS, but it doesn't currently exist and I didn't feel that this work rose to the level of demanding it.

alpha/declcfg/declcfg.go
grokspawn: Because of the per-path concurrency, it seemed possible that we could conceivably end up with two files which each possessed an instance of deprecations for the same package (so they would pass the unmarshaling check) but when we merged them as part of the concurrency coalesce they would collide.

alpha/declcfg/load.go
stevekuznetsov: Perhaps sketch out the warpath in the future for the other parts here so it's clear why we have one validation for one segment but not the rest?

alpha/declcfg/load_test.go
stevekuznetsov: Perhaps scope this to what is needed today?

alpha/declcfg/load.go
grokspawn: Have to init that map here or we can't create a struct with a non-nil map.  Merge would propagate it to the destination safely if that path hadn't previously encountered a deprecation instance.

alpha/declcfg/load.go
grokspawn: Sorry, I thought that I understood the validation approach here, but it's only validated when it's converted to a Model type in the [Validate](https://github.com/grokspawn/operator-registry/blob/3432b12a869e37e0456567b0c93edb679f372a58/alpha/model/model.go#L49-L8) method.

I haven't gotten that far with this work yet, but that is also where we'll move this check.

alpha/declcfg/declcfg.go
grokspawn: Learning that the data is generally validated in the conversion to the `Model` type, I'll implement that conversion/validation in the next PR.

alpha/declcfg/load_test.go
grokspawn: Eh, my goal with this "stuff"+deprecations test was just that: that adding the new schema did not impact existing parsing.

alpha/declcfg/load_test.go
grokspawn: Had some fun playing with options pattern, reflection and generics here.  In the end, the simplest approach was to inline a function for the comparisons.

alpha/declcfg/load_test.go
stevekuznetsov: The point of this test is to test loading deprecations. Perhaps remove unnecessary test fixures that bloat the review and are not precise to what is being loaded. If I commit something that breaks icon loading, I'd love to only see the icon-related tests fail.

alpha/declcfg/load_test.go
```diff
@@ -890,17 +1073,27 @@ present in the .indexignore file.`),
...
+		"cockroachdb.json": cockroachdb,
+		"etcd.yaml":        etcd,
+		"etcdoperator.v0.6.1.clusterserviceversion.yaml": etcdCSV,
+		"README.md":                readme,
+		"deprecations.yaml":        deprecations,
+		"unrecognized-schema.json": unrecognizedSchema,
+	}
 )
 
+type EvaluationFunc func(*testing.T, *DeclarativeConfig)
```
stevekuznetsov, line 1088: Why don't the tests simply expose the `*DeclarativeConfig` they expect and use a `cmp.Diff()` on the output of the test?

alpha/declcfg/load_test.go
```diff
@@ -337,6 +337,170 @@ func TestLoadFS(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:      "Success/ValidDir with deprecations",
```
everettraven, line 341: Is there a particular reason to make this a separate test case completely as opposed to adding the deprecations to the existing test case named `Success/ValidDir`?

alpha/declcfg/load_test.go
grokspawn: Attempting to leave a minimal footprint and just change the source of friction.  TestLoadFS already does validation over the generated `DeclarativeConfig`. In fact, TestLoadFile is probably superfluous, but I didn't think it was harmful to leave it.

alpha/declcfg/load_test.go
grokspawn: [e9afb84b36b44e76b20defac5b73747afd7fdd46](https://github.com/operator-framework/operator-registry/pull/1151/commits/e9afb84b36b44e76b20defac5b73747afd7fdd46)

everettraven (approved):
Changes look good to me

/lgtm

ncdc (approved):

alpha/declcfg/declcfg.go
```diff
@@ -90,6 +92,19 @@ type RelatedImage struct {
...
+	Schema       string             `json:"schema"`
+	Package      string             `json:"package"`
+	Name         string             `json:"name,omitempty"`
+	Deprecations []DeprecationEntry `json:"deprecations"`
+}
+
+type DeprecationEntry struct {
+	Schema  string          `json:"schema"`
+	Name    string          `json:"name,omitempty"`
+	Message json.RawMessage `json:"message"`
```
joelanford, line 105: I know this is already merged, but:
```suggestion
	Message string `json:"message"`
```

alpha/declcfg/declcfg.go
```diff
@@ -16,16 +16,18 @@ import (
 )
 
 const (
-	SchemaPackage = "olm.package"
-	SchemaChannel = "olm.channel"
-	SchemaBundle  = "olm.bundle"
+	SchemaPackage     = "olm.package"
+	SchemaChannel     = "olm.channel"
+	SchemaBundle      = "olm.bundle"
+	SchemaDeprecation = "olm.catalog.deprecation"
```
joelanford, line 22: Did we ever debate what the name of this should be? The most intuitive name that comes to mind for me is is `olm.deprecations`, but I can't remember if we had `"catalog"` in there for a reason.

PR #1157, opened 2023-11-03
Update api dependency to 0.19.0

kevinrizza:
Bump api dependency.

Since that api change is related to validation types, rather than ingest the validation data this change does the first step of https://github.com/operator-framework/operator-registry/issues/1158 and marks `opm alpha bundle validate` as deprecated.

### DIFF
--- a/cmd/opm/alpha/bundle/validate.go
+++ b/cmd/opm/alpha/bundle/validate.go
@@ -35,10 +35,13 @@ Optional validators. These validators are disabled by default and can be enabled
  * Operatorhub validator - performs operatorhub.io validation. To validate a bundle using custom categories use with the OPERATOR_BUNDLE_CATEGORIES environmental variable to point to a json-encoded categories file.
  * Bundle objects validator - performs validation on resources like PodDisruptionBudgets and PriorityClasses. 
 
-See https://olm.operatorframework.io/docs/tasks/validate-package/#validation for more info.`,
-		Example: `$ opm alpha bundle validate --tag quay.io/test/test-operator:latest --image-builder docker`,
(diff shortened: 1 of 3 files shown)


joelanford:
Are we also going to add the new validators for `opm alpha bundle validate`?
- standardcategories
- standardcapabilities
- operatorhubv2

Also, is there a reason that both operator-sdk and opm have entrypoints to run the `api`-provided validators? Also, why is opm's set of available validators much smaller that SDK's?

joelanford:
To me, it kinda seems like we should remove `opm alpha bundle validate` entirely and point everyone to the SDK validation.

kevinrizza:
@joelanford totally agree, I actually took a quick look at the validation bits and decided against changing anything there because it just doesn't seem valuable. I'm bumping this dependency not to update anything there but just because it's got a transitive dependency to the api that needs to be updated before the sdk can consume the latest version.

If you want, I can just get rid of the alpha bundle validate command as part of this pull request

joelanford:
I don't know if anyone is using the `opm alpha bundle validate` command, so I'm hesitant to just drop it out of nowhere (and yeah, I realize it's alpha, so we have every right to). I'll make an issue to capture our desire to remove it.

Maybe for now, we update that command to output a clear deprecation message? Something along the lines of "this subcommand is deprecated and will be removed. Migrate to `operator-sdk bundle validate`"?

joelanford:
https://github.com/operator-framework/operator-registry/issues/1158

kevinrizza:
@joelanford makes sense to me. I added a deprecation warning to the command.

joelanford (approved):

Issue #1154, opened 2023-10-30
Implement GRPC API changes to expose deprecations


grokspawn:
As per #1146 

GRPC API (pkg/api/*; pkg/registry/types.go; utests)

PR #1161, opened 2023-11-06
grpc changes, model conversion, validation

grokspawn:
Contributes to https://github.com/operator-framework/operator-registry/issues/1154

NB:  this is declarativeconfig --> grpc path only, right now.  We need future issues for 
- [x] backwards conversions to support round trip: #1162 
- [x] some unit tests specific to deprecation representation: #1163 

**Reviewer Checklist**


pkg/api/registry.proto
```diff
@@ -22,6 +22,7 @@ service Registry {
 message Channel{
 	string name = 1;
 	string csvName = 2;
+	DeprecationEntry deprecation = 3;
```
joelanford, line 25: Here and elsewhere:
```suggestion
	optional DeprecationEntry deprecation = 3;
```

pkg/api/grpc_health_v1/health.pb.go
joelanford, line 1: How did these health files change?

pkg/api/registry.proto
grokspawn: As we discussed, `required` and `optional` were discontinued circa 2015, and these are already represented as pointers.  
We need to capture a test somewhere against an 'older' olm version to verify that it doesn't die when meeting the 'newer' opm serving catalogs that hold deprecation metadata.

pkg/api/grpc_health_v1/health.pb.go
grokspawn: new protoc makes a readability-only change to use consts instead of literal strings (see https://github.com/operator-framework/operator-registry/pull/1161/files#diff-088170ff8b63e91b9321b60879eb555a1bc8e2192c198806f9cf0531400fdd8dR21-R24)

alpha/declcfg/declcfg_to_model.go
```diff
@@ -168,6 +168,70 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
 		}
 	}
 
+	// deprecationsDuplicateCheck tracks the set of package names
```
joelanford, line 171: We missed this comment yesterday when we changed the variable name.
```suggestion
	// deprecationsByPackage tracks the set of package names
```

alpha/declcfg/declcfg_to_model.go
```diff
@@ -168,6 +168,70 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
...
+		if deprecationsByPackage.Has(deprecation.Package) {
+			return nil, fmt.Errorf("expected a maximum of one deprecation per package: %q", deprecation.Package)
+		}
+		deprecationsByPackage.Insert(deprecation.Package)
+
+		references := sets.New[PackageScopedReference]()
+
+		for j, entry := range deprecation.Entries {
+			if entry.Reference.Schema == "" {
+				return nil, fmt.Errorf("schema must be set for deprecation entry reference %q[%v]", deprecation.Package, j)
```
joelanford, line 199:
```suggestion
				return nil, fmt.Errorf("schema must be set for deprecation entry[%v] for package %q", deprecation.Package, j)
```

alpha/declcfg/declcfg_to_model.go
```diff
@@ -168,6 +168,70 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
...
+			}
+
+			if references.Has(entry.Reference) {
+				return nil, fmt.Errorf("duplicate deprecation entry %#v for package %q", entry.Reference, deprecation.Package)
+			}
+
+			switch entry.Reference.Schema {
+			case SchemaBundle:
+				if !packageBundles[deprecation.Package].Has(entry.Reference.Name) {
+					return nil, fmt.Errorf("cannot deprecate bundle %q for package %q: bundle not found", entry.Reference, deprecation.Package)
```
joelanford, line 209:
```suggestion
					return nil, fmt.Errorf("cannot deprecate bundle %q for package %q: bundle not found", entry.Reference.Name, deprecation.Package)
```

alpha/declcfg/declcfg_to_model.go
```diff
@@ -168,6 +168,70 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
...
+				}
+				for _, mch := range mpkg.Channels {
+					if mb, ok := mch.Bundles[entry.Reference.Name]; ok {
+						mb.Deprecation.Message = entry.Message
+					}
+				}
+			case SchemaChannel:
+				ch, ok := mpkg.Channels[entry.Reference.Name]
+				if !ok {
+					return nil, fmt.Errorf("cannot deprecate channel %q for package %q: channel not found", entry.Reference, deprecation.Package)
```
joelanford, line 219:
```suggestion
					return nil, fmt.Errorf("cannot deprecate channel %q for package %q: channel not found", entry.Reference.Name, deprecation.Package)
```

pkg/api/registry.proto
```diff
@@ -124,3 +127,7 @@ message GetDefaultProviderRequest{
 	string kind = 3;
 	string plural = 4;
 }
+
+message DeprecationEntry{
```
joelanford, line 131:
```suggestion
message Deprecation{
```

grokspawn: ??

alpha/declcfg/write.go
```diff
@@ -426,8 +426,8 @@ func writeToEncoder(cfg DeclarativeConfig, enc encoder) error {
 		}
 
 		deprecations := deprecationsByPackage[pName]
-		sort.SliceStable(deprecations, func(i, j int) bool {
-			return deprecations[i].Name < deprecations[j].Name
+		sort.Slice(deprecations, func(i, j int) bool {
+			return deprecations[i].Package < deprecations[j].Package
```
joelanford, line 430: I realized my suggestion to change to sort by package name results in this being a no-op because `deprecations` is all of the `Deprecation` objects with the same `pName`.

Perhaps we just _don't_ sort? Because we expect this `deprecations` slice to be length 0 or 1.

joelanford:
I think this looks good to me implementation-wise. What are you thinking re: tests?

alpha/declcfg/write.go
grokspawn: That wasn't you.  I did that 9 days ago, and you're right ... with ordinality [0,1] there's no need to sort.

grokspawn:
> I think this looks good to me implementation-wise. What are you thinking re: tests?

Tracked in #1163

alpha/declcfg/declcfg.go
```diff
@@ -93,16 +93,19 @@ type RelatedImage struct {
...
+	Schema  string             `json:"schema"`
+	Package string             `json:"package"`
+	Entries []DeprecationEntry `json:"entries"`
 }
 
 type DeprecationEntry struct {
-	Schema  string          `json:"schema"`
-	Name    string          `json:"name,omitempty"`
-	Message json.RawMessage `json:"message"`
+	Reference PackageScopedReference `json:"reference"`
```
everettraven, line 102: What is the motivation for making this a separate type as opposed to keeping the flat structure?

Before this change, the YAML to represent a `Deprecation` is:
```yaml
schema: olm.deprecations
package: somepackage
entries:
  - schema: olm.package
    name: somepackage
    message: |
       somepackage is deprecated as it has been archived and is no longer maintained.
```

After the change it is now:
```yaml
schema: olm.deprecations
package: somepackage
entries:
  - reference:
     schema: olm.package
     name: somepackage
    message: |
       somepackage is deprecated as it has been archived and is no longer maintained.
```

In this context it feels unnecessary to have this nested and requires extra typing (maybe I'm just lazy?) to get the same thing conveyed.

alpha/declcfg/declcfg_to_model.go
```diff
@@ -168,6 +168,70 @@ func ConvertToModel(cfg DeclarativeConfig) (model.Model, error) {
...
+			return nil, fmt.Errorf("cannot apply deprecations to an unknown package %q", deprecation.Package)
+		}
+
+		// must be unique per package
+		if deprecationsByPackage.Has(deprecation.Package) {
+			return nil, fmt.Errorf("expected a maximum of one deprecation per package: %q", deprecation.Package)
+		}
+		deprecationsByPackage.Insert(deprecation.Package)
+
+		references := sets.New[PackageScopedReference]()
```
everettraven, line 195: I see now. Is using the `PackageScopedReference` type required here? Could we have the same functionality using the flat `DeprecationEntry` type by using `sets.New[string]()` and doing some string concatenation of the `DeprecationEntry.Schema` and `DeprecationEntry.Name` fields?

alpha/declcfg/declcfg.go
grokspawn: It felt like having the flat structure was doing some "this is really a reference" hand-waving, using a nested, repeated fieldname (schema).  Really prone to confusion, and the new structure makes it clear that this is a reference to a schema+name in a manner that loosely follows general format for a kube GVK-like thing.

alpha/declcfg/declcfg_to_model.go
grokspawn: We totally could, but it's cleaner if we can acknowledge the composite type instead of doing some behind-the-scenes string concatenation.

alpha/declcfg/declcfg.go
joelanford: Jordan and I collab'd on this one. First it was:
```
- schema: olm.package
  name: foo
  message: foo is deprecated
```

Then Jordan had:
```
- scope: olm.package
  reference: foo
  message: foo is deprecated
```

And we settled on this because I thought it was important that `schema` and `name` show up here, but I also saw how Jordan thought that an inlined schema/name might cause confusion because it isn't an obvious reference to another object.

everettraven: I'm not opinionated enough to make this block this PR, but tried to approach it from a user's perspective. I can see the confusion but in the context of FBC objects that contain an `entries` field there isn't an explicit `reference` field because it is a reference to another object. I also think if this is well documented it mitigates _most_ of that confusion. As a user I would also prefer to type less if it accomplishes the same thing, but again I could just be a bit lazy.

Again, shouldn't block this PR but was something I thought was worth at least bringing up. Thanks for the insights!

alpha/declcfg/declcfg_to_model.go
joelanford: Also string concatenation is a place for bugs to creep in if you aren't extremely careful and have some validation to make sure that you don't end up with apparent, but not actual duplicates:
- `foo`, `bar` => `foobar`
- `foob`, `ar` => `foobar`

everettraven: Fair enough.

everettraven (approved):

Issue #1162, opened 2023-11-09
grpc api-->declarativeconfig types conversions


grokspawn:
`opm` typically functions to make FBC catalogs available over GRPC by converting the FBC to declcfg.DeclarativeConfig, storing/validating as model.Model, and expressing via api.Bundle/Package/Channel and #1161 handled that forward-conversion. 

However, the library also supports backwards conversions which should be updated to support the modified types.

PR #1164, opened 2023-11-13
WIP: Deprecation utests

grokspawn:
Currently off branch for #1161.  Will rebase once that merges. 

**Reviewer Checklist**


alpha/declcfg/declcfg_to_model_test.go
```diff
@@ -291,7 +440,7 @@ func TestConvertToModel(t *testing.T) {
 }
 
 func TestConvertToModelRoundtrip(t *testing.T) {
-	expected := buildValidDeclarativeConfig(true)
+	expected := buildValidDeclarativeConfig(validDeclarativeConfigSpec{IncludeUnrecognized: true, IncludeDeprecations: false}) // TODO: turn on deprecation when we have model-->declcfg conversion
```
grokspawn, line 443: This function goes declcfg --> model --> declcfg and compares the start and finish products.  We don't yet have backwards conversion, so this can't be enabled until #1162 lands.

alpha/declcfg/model_to_declcfg_test.go
```diff
@@ -19,7 +19,7 @@ func TestConvertFromModel(t *testing.T) {
 		{
 			name:      "Success",
 			m:         buildTestModel(),
-			expectCfg: buildValidDeclarativeConfig(false),
+			expectCfg: buildValidDeclarativeConfig(validDeclarativeConfigSpec{IncludeUnrecognized: false, IncludeDeprecations: false}),
```
grokspawn, line 22: #1162 will make this `false, true`

alpha/declcfg/declcfg_to_model_test.go
```diff
@@ -280,6 +280,155 @@ func TestConvertToModel(t *testing.T) {
 				Bundles:  []Bundle{newTestBundle("foo", "0.1.0")},
 			},
 		},
+		{
+			name:      "Error/Deprecation/UnspecifiedPackage",
```
stevekuznetsov, line 284: nit: names are for humans - machines handle whatever we throw at them. I know you're following the pattern here and you don't need to change it, but in the future we should consider using names that are English sentences

alpha/declcfg/declcfg_to_model_test.go
```diff
@@ -280,6 +280,155 @@ func TestConvertToModel(t *testing.T) {
...
+			assertion: hasError(`package name must be set for deprecation item 0`),
```
stevekuznetsov, line 285: If the content of the error is valuable to callers, it should be exposed by a mechanism that they can query without resorting to stringly-typing - exported error types, `errors.Is()`, etc. Again, I know you're following the pattern here but this adds brittleness to the tests here

pkg/server/server_test.go
```diff
@@ -75,6 +80,17 @@ func fbcJsonCache(catalogDir, cacheDir string) (cache2.Cache, error) {
 	return store, nil
 }
 
+func fbcJsonCacheFromFs(catalogFS fs.FS, cacheDir string) (cache2.Cache, error) {
+	store := cache2.NewJSON(cacheDir)
```
stevekuznetsov, line 84: nit: I know it's also pre-existing, but prefer a more human-friendly import alias to `cache2`

pkg/server/server_test.go
```diff
@@ -113,6 +129,13 @@ func TestMain(m *testing.M) {
 		logrus.Fatalf("failed to create json cache: %v", err)
 	}
 	s2 := server(fbcJsonStore)
+
+	fbcJsonDeprecationStore, err := fbcJsonCacheFromFs(validFS, filepath.Join(tmpDir, "json-deprecation-cache"))
+	if err != nil {
+		logrus.Fatalf("failed to create json deprecation cache: %v", err)
+	}
+	s3 := server(fbcJsonDeprecationStore)
```
stevekuznetsov, line 137: Similarly, this can be `deprecationServer` ... when I saw `s3.Serve()` I for a moment wondered if you were serving an S3-compatible object storage ... ?

pkg/server/server_test.go
```diff
@@ -147,12 +179,81 @@ func client(t *testing.T, address string) (api.RegistryClient, *grpc.ClientConn)
 	return api.NewRegistryClient(conn), conn
 }
 
+var (
```
stevekuznetsov, line 182: Test data should be declared within the test using it.

pkg/server/server_test.go
```diff
@@ -787,3 +854,206 @@ func etcdoperator_v0_9_2(channel string, addSkipsReplaces, addExtraProperties bo
 	}
 	return b
 }
+
+func getCockroachBundle() *api.Bundle {
```
stevekuznetsov, line 858: If this is static data, why is it behind a function?

pkg/server/server_test.go
```diff
@@ -787,3 +854,206 @@ func etcdoperator_v0_9_2(channel string, addSkipsReplaces, addExtraProperties bo
...
...cgMCAwIDAgMCAuMDkgMTcuMzIgMTcuMzIgMCAwIDAgLjgzIDEuNTcuMDcuMDcgMCAwIDAgLjA4IDAgMTYuMzkgMTYuMzkgMCAwIDEgMS44MS0uNTQgMTUuNjUgMTUuNjUgMCAwIDEgMTEuNTkgMS44OCAxNy41MiAxNy41MiAwIDAgMC0zLjc4IDQuNDhjLS4yLjMyLS4zNy42NS0uNTUgMXMtLjIyLjQ1LS4zMy42OS0uMzEuNzItLjQ0IDEuMDhhMTcuNDYgMTcuNDYgMCAwIDAgNC4yOSAxOC43Yy4yNi4yNS41My40OS44MS43M3MuNDQuMzcuNjcuNTQuNTkuNDQuODkuNjRhLjA3LjA3IDAgMCAwIC4wOCAwYy4zLS4yMS42LS40Mi44OS0uNjRzLjQ1LS4zNS42Ny0uNTQuNTUtLjQ4LjgxLS43M2ExNy40NSAxNy40NSAwIDAgMCA1LjM4LTEyLjYxIDE3LjM5IDE3LjM5IDAgMCAwLTEuMDktNi4wOWMtLjE0LS4zNy0uMjktLjczLS40NS0xLjA5cy0uMjItLjQ3LS4zMy0uNjktLjM1LS42Ni0uNTUtMWExNy42MSAxNy42MSAwIDAgMC0zLjc4LTQuNDggMTUuNjUgMTUuNjUgMCAwIDEgMTEuNi0xLjg0IDE2LjEzIDE2LjEzIDAgMCAxIDEuODEuNTQuMDcuMDcgMCAwIDAgLjA4IDBxLjQ0LS43Ni44Mi0xLjU2YS4wNy4wNyAwIDAgMCAwLS4wOUExNi44OSAxNi44OSAwIDAgMCAzMCAuNTd6IiBmaWxsPSIjMTUxZjM0Ii8+PHBhdGggZD0iTTIxLjgyIDE3LjQ3YTE1LjUxIDE1LjUxIDAgMCAxLTQuMjUgMTAuNjkgMTUuNjYgMTUuNjYgMCAwIDEtLjcyLTQuNjggMTUuNSAxNS41IDAgMCAxIDQuMjUtMTAuNjkgMTUuNjIgMTUuNjIgMCAwIDEgLjcyIDQuNjgiIGZpbGw9IiMzNDg1NDAiLz48cGF0aCBkPSJNMTUgMjMuNDhhMTUuNTUgMTUuNTUgMCAwIDEtLjcyIDQuNjggMTUuNTQgMTUuNTQgMCAwIDEtMy41My0xNS4zN0ExNS41IDE1LjUgMCAwIDEgMTUgMjMuNDgiIGZpbGw9IiM3ZGJjNDIiLz48L3N2Zz4=",
```
stevekuznetsov, line 891: Can we remove these binary data blobs from tests that do not need them?

pkg/server/server_test.go
grokspawn: It is mimicking an existing test pattern here, c.f. https://github.com/operator-framework/operator-registry/pull/1164/files#diff-fe7a244e18e92d8f347af17363209172e1d79224f086b0585a1bbeef3b626c4eR777

pkg/server/server_test.go
grokspawn: Short answer: 
I'll check for this case. 

Long answer:
A lot of these are seeded from existing catalog data.  What purpose are we serving by skipping it?  These tests are literally "what if we have something like before, but now add deprecation metadata".

pkg/server/server_test.go
grokspawn: I pondered this one, but whatever logic drove the name is lost in time.  The best correlation I had was that "this was the second server cache created in the test".  
This is easily enough changed to `fbccache`

pkg/server/server_test.go
grokspawn: The predecessor objects are `s` and `s2`.  This holds to that pattern.

pkg/server/server_test.go
stevekuznetsov: Sorry, but I'm not sure that is sufficient justification - is there a reason that is independent of that?

pkg/server/server_test.go
stevekuznetsov: The more "stuff" is here, the harder it is for someone to see the diff and to grok which parts are important and which are not - I'll flip the question, since you are proposing to commit this - what value does having this icon bring to this test?

grokspawn: I skipped it in this case, but I didn't strip it out of the bundle manifests already in the tests.

pkg/server/server_test.go
stevekuznetsov: I know. I am saying it's a bad pattern.

pkg/server/server_test.go
grokspawn: I'm anticipating that if there was a need to pivot the original, we might want to do that as well in the future.  I'm not doing it right now, but trying to match the idiom of the existing tests in a manner that lets them expand in a similar manner.

PR #1169, opened 2023-11-17
Semver lastedge error

grokspawn:
During an onboarding session, we discovered that an author was missing an expected replaces between their last bundle version and its semver-ordered predecessor.  

On investigation, it was because the channel linking step always seeks forward for a y-version change and runs off the end of the version list, and thus fails to detect the change. 

When investigating, it became clear that the method was over-engineered as well, so this PR substantially simplifies the implementation.

### DIFF
--- a/alpha/template/semver/semver.go
+++ b/alpha/template/semver/semver.go
@@ -6,14 +6,13 @@ import (
 	"io"
 	"sort"
 
-	"github.com/blang/semver/v4"
-	"k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
-	"sigs.k8s.io/yaml"
-
 	"github.com/operator-framework/operator-registry/alpha/action"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/alpha/property"
+
+	"github.com/blang/semver/v4"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"sigs.k8s.io/yaml"
 )
 
 func (t Template) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error) {
@@ -224,7 +223,6 @@ func (sv *semverTemplate) generateChannels(semverChannels *bundleVersions) []dec
 	hwc := highwaterChannel{archetype: archetypesByPriority[0], version: semver.Version{Major: 0, Minor: 0}}
 
 	unlinkedChannels := make(map[string]*declcfg.Channel)
-	unassociatedEdges := []entryTuple{}
 
 	for _, archetype := range archetypesByPriority {
 		bundles := (*semverChannels)[archetype]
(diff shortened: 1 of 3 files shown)


everettraven (commented):
From a code standpoint, this look ok to me but I am pretty unfamiliar with this section of the code. In order to give what I feel would be a proper review I might need a more synchronous walkthrough of the changes.

grokspawn:
> From a code standpoint, this look ok to me but I am pretty unfamiliar with this section of the code. In order to give what I feel would be a proper review I might need a more synchronous walkthrough of the changes.

Happy to schedule something or grab an impromptu huddle.  The short version of what happened is
1. the previous code used to keep a complete accounting of every edge, and had complex logic to identify and handle correctly changes in channel archetype, major/minor channel type, and the relevant bundle version for each edge.  I basically went "this is already in unlinked channels, except for a bundle-name --> version lookup"; and
2. the previous code would miss a minor-stream increment last-1 .. last because 
   1. the iterators were bound/contained by the array
   2. the yProbe looking for such iterations would have to be outside the array to detect it, which isn't allowed  

  so I just added a quick "did it happen" check for those edges.

alpha/template/semver/semver.go
```diff
@@ -272,88 +270,84 @@ func (sv *semverTemplate) generateChannels(semverChannels *bundleVersions) []dec
...
+				zmaxQueue = (*entries)[preChangeIndex].Name
+			}
+			for curEdge < preChangeIndex {
+				// add skips edges to y-1 from z < y
+				(*entries)[preChangeIndex].Skips = append((*entries)[preChangeIndex].Skips, (*entries)[curEdge].Name)
+				curEdge += 1
+			}
+			curEdge += 1
+			yProbe = curEdge + 1
+		}
```
everettraven, line 339: So after spending more time dissecting this for loop to really understand what each step is doing (took ~30min) I think it is good to go logically. Maybe it's just me coming back from being sick and still being a bit foggy (still recovering) but I felt that it was a bit difficult to understand this change in what I felt was a reasonable time frame (~20 min) as a reviewer, which led me to feeling rather unconfident in my review and thus withholding an "approval".

Feel free to take this recommendation or leave it as it, but I think a slight refactor for readability could be useful here. Maybe something along the lines of:
```suggestion
			yStreamHead := getYStreamHead(curEntry, *entries) // This function would perform the same logic as already exists in the yProbe for loop 

			if curEdge != yStreamHead {
				if zmaxQueue != "" {
					// add skips edge to allow skipping over y iterations within an x stream
					(*entries)[yStreamHead].Skips = append((*entries)[yStreamHead].Skips, zmaxQueue)
					(*entries)[yStreamHead].Replaces = zmaxQueue
				}
				zmaxQueue = (*entries)[yStreamHead].Name
			}
			
			// could also be made into a separate function if desired
			for _, edge := range (*entries)[curEdge:yStreamHead] {
				// add skips edges to y-1 from z < y
				(*entries)[yStreamHead].Skips = append((*entries)[yStreamHead].Skips, edge.Name)
			}
			curEdge = yStreamHead + 1
		}
```

everettraven: Overall, what threw me for a loop (pun intended) the most was wrapping my head around the multiple curEdge increments (and why) as we progressed through the loop and the initially unclear discovery of the y stream head (although looking back at it it totally makes sense as is, so feel free to keep the loop instead of breaking out to another function if you want - I just ate lunch so my mind is feeling a bit more clear now 😆 ).

everettraven (approved):
Have the one nit, but I don't think my concerns of readability should block this PR since functionally it is 👍 .

PR #1172, opened 2023-12-07
graph deprecation mermaid visualization update

rashmigottipati:
**Description of the change:**
Add colorblind-friendly color indicators for deprecation of each scope (bundle, package and channel)

**Motivation for the change:**
Make it easy to visually identify the deprecated packages, bundles and channels from the mermaid graphs

**Reviewer Checklist**
- [ ] Implementation matches the proposed design, or proposal is updated to match implementation
- [ ] Sufficient unit test coverage 
- [ ] Sufficient end-to-end test coverage
- [ ] Docs updated or added to `/docs` 
- [ ] Commit messages sensible and descriptive

Closes [1160](https://github.com/operator-framework/operator-registry/issues/1160)

Previous mermaid graph output without the color indications for deprecations of each scope: 

original-mermaid-graph

With the changes in this PR, example mermaid graph output showcasing a deprecated channel `kiali-alpha`, bundle `kiali-operator.v1.68.0`, package `kiali`:

new-mermaid-graph

PR #1173, opened 2023-12-12
(bugfix) Update to handle mermaid visualization for deprecation of multiple packages

rashmigottipati:
[PR](https://github.com/operator-framework/operator-registry/pull/1172) to add colorblind-friendly color indicators for deprecation of each scope (bundle, package and channel) was merged last week. Based on this [review comment](https://github.com/operator-framework/operator-registry/pull/1172#discussion_r1423996999), previously merged PR presumes that we will only ever have at most one deprecated package.  This PR fixes that by handling the case of a declarative config includes multiple deprecated packages.

### DIFF
--- a/alpha/declcfg/write.go
+++ b/alpha/declcfg/write.go
@@ -202,16 +202,12 @@ func (writer *MermaidWriter) WriteChannels(cfg DeclarativeConfig, out io.Writer)
 		out.Write([]byte("  end\n"))
 	}
 
-	if len(deprecatedPackages) > 0 {
-		for _, deprecatedPackage := range deprecatedPackages {
-			out.Write([]byte(fmt.Sprintf("style %s fill:#989695\n", deprecatedPackage)))
-		}
+	for _, deprecatedPackage := range deprecatedPackages {
+		out.Write([]byte(fmt.Sprintf("style %s fill:#989695\n", deprecatedPackage)))
 	}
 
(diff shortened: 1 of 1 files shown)


alpha/declcfg/write.go
```diff
@@ -202,8 +202,10 @@ func (writer *MermaidWriter) WriteChannels(cfg DeclarativeConfig, out io.Writer)
 		out.Write([]byte("  end\n"))
 	}
 
-	if deprecatedPackage != "" {
-		out.Write([]byte(fmt.Sprintf("style %s fill:#989695\n", deprecatedPackage)))
+	if len(deprecatedPackages) > 0 {
```
grokspawn, line 205: question... is there any reason that we need to check the size of the slice?  If we've encountered no matching deprecations the loop would become a NOOP, right?

(No need to take out, because we're doing the same for 211, but I wanted to grab an opportunity to ask.)

grokspawn (approved):

rashmigottipati:
Closing this as this is not a bug and we're indeed handling the deprecation of multiple packages in this PR https://github.com/operator-framework/operator-registry/pull/1172 as https://github.com/operator-framework/operator-registry/blob/e5dca1a649fae543cb1caa12ae15dd5163442321/alpha/declcfg/write.go#L142 is within the channel loop and we're bound to hit every single package that is being deprecated and capture it as deprecated.

PR #1221, opened 2024-02-07
use composite template's contributions path as fallback search for template inputs

grokspawn:

since using the composite template means the union of requirements from two parties, it can be confusing sometimes to understand the context of resource references.  

For example, in the contribution schema `olm.composite` there are references to input filesystem elements in `Components[].Strategy.Template.Config.Input` ([ref](https://github.com/operator-framework/operator-registry/blob/4c69b4af4437039212dd83fbe049fa30efb0b2cc/alpha/template/composite/types.go#L22-L23)) but these may be relative or absolute references.  And if relative... then relative to what?  

This PR attempts to address this by fallback searching for input references relative to the cwd of the contribution file (if driven by CLI) or the `Components[].Strategy.Template.Config.ContributionPath` (if provided). 

For example, given a contribution of the format
```yaml
schema: olm.composite
components:
  - name: catalog-of-wonders
    destination:
      path: my-operator
    strategy:
      name: semver
      template:
        schema: olm.builder.semver
        config:
          input: semver-catalog-template.yaml
          output: catalog.yaml
```

the template expansion will look for `semver-catalog-template.yaml` in the CWD of the executed command.  But if the file is actually in a `catalog-data` subdirectory but the command is executed in the parent, then it won't be solved, for e.g. 
```tree
catalog-data
├── basic-catalog-template.yaml
└── contributions.yaml
```

The catalogs.yaml file which fulfills the `olm.composite.catalogs` spec may be local or remote, so it is unsuitable as a search location, but the `olm.composite` source is always local, which allows us to infer a possible anchor point for relative references. 

In case a naiive access attempt for the specified contribution input file fails, we use the resulting anchor in a fallback attempt to resolve input references.

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/template/composite.go
+++ b/cmd/opm/alpha/template/composite.go
@@ -4,6 +4,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 
@@ -49,6 +50,11 @@ and a 'composite template' file`,
 			}
 			defer compositeReader.Close()
 
+			compositePath, err := filepath.Abs(filepath.Dir(compositeFile))
+			if err != nil {
+				log.Fatalf("getting absolute path of composite config file %q: %v", compositeFile, err)
+			}
+
 			// catalog maintainer's 'catalogs.yaml' file
 			tempCatalog, err := composite.FetchCatalogConfig(catalogFile, http.DefaultClient)
 			if err != nil {
@@ -58,7 +64,7 @@ and a 'composite template' file`,
 
 			template := composite.NewTemplate(
 				composite.WithCatalogFile(tempCatalog),
-				composite.WithContributionFile(compositeReader),
+				composite.WithContributionFile(compositeReader, compositePath),
 				composite.WithOutputType(output),
 				composite.WithRegistry(reg),
 			)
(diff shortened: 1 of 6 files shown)


cmd/opm/alpha/template/composite.go
```diff
@@ -49,6 +50,8 @@ and a 'composite template' file`,
 			}
 			defer compositeReader.Close()
 
+			compositePath, _ := filepath.Abs(filepath.Dir(compositeFile))
```
everettraven, line 53: Looks like we are eating an error here. Is there a particular reason for this?

Looking into the `filepath.Abs()` source code this returns an error if `os.Getwd()` returns an error (at least on unix based systems).

grokspawn: My first thoughts were that any access error would be caught in line 49 where the same file is opened.  However, it's not really a big deal if we catch the error.  I just can't imagine a case where `Abs` generates an error that `Open` would not.

everettraven: Yeah, that's fair. I just personally prefer to not eat errors because it can cause a lot of issues if we assume something that has the potential to error will never error

everettraven (approved):
/lgtm

PR #1231, opened 2024-03-14
render: add alpha support for filtering

joelanford:

This commit adds two alpha flags to `opm render` and updates FBC library code to support efficient filtering of FBC:
- `--alpha-filter-config=./path/to/config`
- `--alpha-keep-packages=pkg1,pkg2`

To support the filter config option, this commits defines a new configuration file format that supports:
- filtering by package
- within a package, filtering by channel
- within a package, selecting a new default channel (required if the original default channel is filtered out)
- within each channel, filtering by a version range

In order to support correct version range filtering, some bundles that fall outside the version range may remain in the channel to ensure that the filtered channel is still valid and includes all bundles that did fall in the range

Included in this commit is an update to declcfg.WalkMetasFS to enable concurrency, similar to recent changes in declcfg.LoadFS. declcfg.LoadFS is also refactored internally to use the new declcfg.WalkMetasFS implementation.

There are two motivations:
1. Filtering by individual versions is an often requested feature, especially in workflows that requiring mirroring catalogs to disconnected environments. In OLM, filtering versions requires semantic knowledge of the upgrade graph rules, so it is not as simple as keeping or removing arbitrary bundles and channel entries. This PR includes that semantic knowledge.
2. When rendering an existing FBC, `render` is significantly faster than rendering an entire catalog and then using a post-render tool like jq or yq to filter after the fact.

I ran the existing benchmark to confirm that this update does not regress performance, even in the case of no filtering being applied. I will post those results in a comment.

**Reviewer Checklist**


alpha/declcfg/filter/filter.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+)
+
+type Filterer interface {
+	FilterCatalog(ctx context.Context, fbc *declcfg.DeclarativeConfig) (*declcfg.DeclarativeConfig, error)
+}
+
+type MetaFilterer interface {
+	KeepMeta(meta *declcfg.Meta) bool
+}
```
everettraven, line 15: Nit: Andy did a great job of getting this stuck in my head over the course of a few PRs: https://go.dev/wiki/CodeReviewComments#interfaces

The gist is that the interfaces should be defined where the values of the interface are used and not by the package that implements them. I'm not strongly opinionated on this (yet) but is something that crossed my mind while reviewing so figured I would point it out.

alpha/action/render.go
```diff
@@ -237,6 +247,19 @@ func (r Render) imageToDeclcfg(ctx context.Context, imageRef string) (*declcfg.D
...
+func (r *Render) renderFBCDirectory(ctx context.Context, dir string) (*declcfg.DeclarativeConfig, error) {
+	// If a filterer is provided that can tell us which meta objects to keep, we
+	// can optimize FBC loading and post-filtering by ignoring meta objects that
+	// we know we won't need to keep.
+	keepMeta := func(meta *declcfg.Meta) bool { return true }
+	if r.Filterer != nil {
+		if metaFilterer, ok := r.Filterer.(filter.MetaFilterer); ok {
+			keepMeta = metaFilterer.KeepMeta
+		}
+	}
```
everettraven, line 259: Nit: Why not have an explicit field for the `MetaFilterer`?

alpha/declcfg/filter/config/v1alpha1/config.go
```diff
@@ -0,0 +1,77 @@
...
+	Name string `json:"name"`
+
+	// VersionRange is a semver range to filter the versions of the channel.
+	// If not set, all versions will be included.
+	VersionRange string `json:"versionRange,omitempty"`
+}
+
+func LoadFilterConfiguration(data []byte) (*FilterConfiguration, error) {
+	cfg := &FilterConfiguration{}
+	if err := yaml.Unmarshal(data, cfg); err != nil {
```
everettraven, line 46: Can this `yaml.Unmarshal` handle both YAML and JSON formatted data? If not, a go doc that the input data _has_ to be valid YAML would be nice.

alpha/declcfg/filter/packages.go
```diff
@@ -0,0 +1,50 @@
...
+func NewPackageFilter(keepPackages ...string) Filterer {
+	packageSet := sets.New[string](keepPackages...)
+	return &packageFilter{keepPackages: packageSet}
+}
+
+type packageFilter struct {
+	keepPackages sets.Set[string]
+}
+
+func (f *packageFilter) FilterCatalog(ctx context.Context, fbc *declcfg.DeclarativeConfig) (*declcfg.DeclarativeConfig, error) {
```
everettraven, line 21: My eyes could be deceiving me, but looks like this function doesn't use the `ctx` parameter.

alpha/declcfg/load.go
```diff
@@ -196,7 +183,17 @@ func sendPaths(ctx context.Context, root fs.FS, pathChan chan<- string) error {
...
-func parsePaths(ctx context.Context, root fs.FS, pathChan <-chan string, cfgChan chan<- *DeclarativeConfig) error {
+type pathMetas struct {
+	path  string
+	metas []pathMeta
+}
+
+type pathMeta struct {
+	meta *Meta
+	err  error
+}
```
everettraven, line 194: What are these for? I can't seem to see where they are being used, am I just missing it?

cmd/opm/render/cmd.go
```diff
@@ -100,3 +127,36 @@ func nullLogger() *logrus.Entry {
...
+	switch obj.APIVersion {
+	case "olm.operatorframework.io/v1alpha1":
+		cfg, err := configv1alpha1.LoadFilterConfiguration(data)
+		if err != nil {
+			return nil, err
+		}
+		return configv1alpha1.NewFilterer(cfg, configv1alpha1.WithLogger(log)), nil
+	default:
+		return nil, fmt.Errorf("unexpected API version %q", obj.APIVersion)
+	}
```
everettraven, line 161: It looks like the `configv1alpha1.LoadFilterConfiguration` function already performs validation. Could this whole thing just be simplified to something like:
```go
	if filterFile == "" {
		return nil, nil
	}

	data, err := os.ReadFile(filterFile)
	if err != nil {
		return nil, err
	}
	
	cfg, err := configv1alpha1.LoadFilterConfiguration(data)
	if err != nil {
		return nil, err
	}
	return configv1alpha1.NewFilterer(cfg, configv1alpha1.WithLogger(log)), nil
```
?

everettraven: I do realize that maybe this isn't as extensible because it only validates against the specific config version but this could be updated when/if we add another version

alpha/declcfg/load.go
joelanford: Oh good catch. PR went through some optimization iteration and these are now unused.

cmd/opm/render/cmd.go
joelanford: +1, this is premature optimization. I'll reduce and add some comments about what we might do if we add new config file versions.

alpha/declcfg/filter/config/v1alpha1/filter.go
```diff
@@ -0,0 +1,280 @@
...
+		keeps[pkg.Name] = channels
+	}
+	return &filterer{
+		pkgConfigs: pkgConfigs,
+		keeps:      keeps,
+		opts:       opts,
+	}
+}
+
+func (f *filterer) FilterCatalog(_ context.Context, fbc *declcfg.DeclarativeConfig) (*declcfg.DeclarativeConfig, error) {
```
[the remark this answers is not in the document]
joelanford: I've seen `context.Context` get added to so many functions at this point that I pretty much add it defensively for functions I expect:
- could be long running
- could need to call other functions that accept a context

alpha/declcfg/filter/filter.go
joelanford: Yep, this is an outcome of the way this code evolved before I submitted it. I'll move it to the `declcfg` package.

alpha/action/render.go
joelanford: At the time, I think it was because I was actually implementing inlined functions with that same signature, so it was lighter weight than implementing a full struct.

I'll add a function type that implements the interface (sorta like http.Handler and http.HandlerFunc).

joelanford: I'm also thinking about names... the age old problem.

alpha/declcfg/filter/config/v1alpha1/filter.go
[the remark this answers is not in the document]
joelanford: Ok, I'm going to leave it in. One possibility I thought of is an interaction with an OCI registry to get the labels of bundle images and filter based on them. I'm considering adding a context to the meta filter signature as well for the same reason. WDYT?

alpha/declcfg/filter/config/v1alpha1/config.go
joelanford: Yes, this supports both.

alpha/declcfg/filter/config/v1alpha1/filter.go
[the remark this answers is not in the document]
grokspawn: IMHO, we should have it for consistency reasons (most of the declcfg code follows this convention), but it's also proven really helpful when we refactored for parallelism previously, and the downside ... ?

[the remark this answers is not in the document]
everettraven: I think it's fine to include if we are anticipating the addition of a filtering method that will use it in the near future. Just calling out that it may be a premature optimization that could be added in the future when needed.

I also am not strongly opinionated here so if folks feel strongly we should keep it - fine by me

joelanford:
Here's the promised `benchstat` output comparing `master` and this PR.

```
goos: darwin
goarch: arm64
                      │ master.out  │           filter-fbc.out           │
                      │   sec/op    │   sec/op     vs base               │
LoadFS/1_routines-10    718.5m ± 0%   717.7m ± 0%       ~ (p=0.353 n=10)
LoadFS/10_routines-10   110.7m ± 2%   110.5m ± 1%       ~ (p=0.684 n=10)
LoadFS/20_routines-10   110.9m ± 0%   110.9m ± 1%       ~ (p=0.971 n=10)
geomean                 206.6m        206.4m       -0.09%

                      │  master.out  │           filter-fbc.out            │
                      │     B/op     │     B/op      vs base               │
LoadFS/1_routines-10    234.8Mi ± 0%   233.6Mi ± 0%  -0.49% (p=0.000 n=10)
LoadFS/10_routines-10   237.7Mi ± 0%   236.6Mi ± 0%  -0.43% (p=0.000 n=10)
LoadFS/20_routines-10   237.1Mi ± 0%   236.0Mi ± 0%  -0.47% (p=0.000 n=10)
geomean                 236.5Mi        235.4Mi       -0.47%

                      │ master.out  │           filter-fbc.out           │
                      │  allocs/op  │  allocs/op   vs base               │
LoadFS/1_routines-10    2.561M ± 0%   2.559M ± 0%  -0.09% (p=0.000 n=10)
LoadFS/10_routines-10   2.562M ± 0%   2.559M ± 0%  -0.09% (p=0.000 n=10)
LoadFS/20_routines-10   2.561M ± 0%   2.559M ± 0%  -0.09% (p=0.000 n=10)
geomean                 2.561M        2.559M       -0.09%
```

everettraven (approved):

perdasilva:
@joelanford I tried to resolve conflicts using the GitHub UI - feel free to force push over it =D

PR #1278, opened 2024-04-18
pkg/cache: add preferred pogreb database cache impl

joelanford:
This PR adds a new caching implementation that has some improvements over our existing JSON cache implementation. It also maintains backward compatibility to gracefully handle serving existing JSON-based caches.

I ran a few tests for compatibility:
1. With an `opm` from this PR, the following happens:
   - Success when the cache contains JSON files
   - Success when the cache contains a Pogreb DB
   - Success when the cache contains both the JSON and Pogreb files side-by-side (and it uses Pogreb)

2. With an `opm` from the current master branch, the following happens:
   - Success when the cache contains JSON files
   - Failure when the cache contains a Pogreb DB
   - Success when the cache contains both the JSON and Pogreb files side-by-side (and it uses JSON). Therefore _adding_ a Pogreb DB to the cache is forward compatible.

- It stores items in an embedded database (see https://github.com/akrylysov/pogreb), which seems to fit the mold of our use case (infrequent bulk load, read heavy clients)
- It serializes/deserializes data as protobuf rather than JSON (so smaller and faster)
- It uses ~50% less CPU to serve similar request loads
- It uses ~66% less peak memory to serve similar request loads
- Its steady state memory is about the same.
- It is much faster to serve (ran some local benchmarks with `ghz` with 20 concurrent connections)
    - JSON
        - 10% latency 1.63s
        - 99% latency 2.53s
    - Pogreb
        - 10% latency 681ms
        - 99% latency 841ms
- The digest calculation is based purely on the FBC and cache data (the current JSON digest uses the tar format and is pretty finicky)
    - For FBC, walk meta blobs, writing them each to the digester
    - For the cache, iterate all the items in the k/v store, writing each k+v to the digester

Cache size improvements
- I've been testing with a migrated operatorhub catalog (that is one that has `olm.csv.metadata`)
- Pogreb: 70M
- Json: 89M

**Reviewer Checklist**


joelanford:
I had an epiphany overnight!

The last commit in the PR changes the way the cache is built. It writes meta objects to a temporary file and records the locations of each meta in the file, grouped by package. That way we can later read _just_ the metas for a particular package into memory.

Then we go package by package building a model, converting to the package index, and writing api bundles to the cache. The beauty is that only a single package's model is loaded in memory at any given time.

**This may mean that we can stop storing caches in the catalog image and can go back to building them on-the-fly when the container starts!** 

I noticed that when using an FBC with `olm.csv.metadata`, startup peak memory and time was basically inconsequential when building a cache on the fly.

pkg/server/server_test.go
```diff
@@ -124,15 +130,15 @@ func TestMain(m *testing.M) {
 
 	grpcServer := server(dbStore)
 
-	fbcJsonStore, err := fbcJsonCache(fbcDir, filepath.Join(tmpDir, "json-cache"))
+	fbcJsonStore, err := fbcCache(fbcDir, filepath.Join(tmpDir, "cache"))
```
grokspawn, line 133:
```suggestion
	fbcStore, err := fbcCache(fbcDir, filepath.Join(tmpDir, "cache"))
```
?

Same for `fbcJsonDeprecationStore` --> `fbcDeprecationStore` ?

grokspawn: Basically nits, but it seems like most variable names have had the `json` removed from them since now they can have protobuf backing instead. Given the variability, it would be nice to only have json in vars which are definitively of that type.

joelanford: Good catch! Done.

PR #1281, opened 2024-04-22
pkg/cache: optimize peak memory usage during cache build

joelanford:
I extracted a commit from #1278 , which can be implemented on its own with our existing caching algorithm. As I mentioned there:

> [This commit] changes the way the cache is built. It writes meta objects to a temporary file and records the locations of each meta in the file, grouped by package. That way we can later read just the metas for a particular package into memory.
>
> Then we go package by package building a model, converting to the package index, and writing api bundles to the cache. The beauty is that only a single package's model is loaded in memory at any given time.
>
> **This may mean that we can stop storing caches in the catalog image and can go back to building them on-the-fly when the container starts!**
>
> I noticed that when using an FBC with `olm.csv.metadata`, startup peak memory and time was basically inconsequential when building a cache on the fly.

In order to maintain the performance of cache building, we need to ensure that WalkMetasFS can make use of concurrency in the same way that LoadFS can (which is what the cache builder currently uses). Therefore, the first commit in this PR includes those changes.

There have been numerous issues reported about how finicky pre-built caches are. There are cases where a catalog image with a pre-built cache works correctly on one node, but not another. There are other cases where caches built outside the image and then injected in are mangled enough to throw off the digest calculation. While these cases are likely problems with the specific digest algorithm we use, this could all be avoided if we were able to build the cache on-the-fly.

**Reviewer Checklist**

### DIFF
--- a/pkg/cache/json.go
+++ b/pkg/cache/json.go
@@ -10,13 +10,17 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
+	"sync"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/pkg/api"
 	"github.com/operator-framework/operator-registry/pkg/registry"
-	"github.com/sirupsen/logrus"
 )
 
 var _ Cache = &JSON{}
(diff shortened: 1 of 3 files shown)


joelanford:
/hold
I'm going to make this faster by adding conccurency support to WalkMetasFS.

pkg/cache/json.go
```diff
@@ -230,56 +234,141 @@ func (q *JSON) Build(ctx context.Context, fbcFsys fs.FS) error {
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case pkgChan <- pkgName:
+			}
+		}
+		return nil
+	})
+
+	var mapMu sync.Mutex
+	for i := 0; i < runtime.NumCPU(); i++ {
```
joelanford, line 291: This means that data for up to `runtime.NumCPU` packages will be in memory at once. I noticed the following when building a cache for the operatorhub catalog on my 10-core Mac M1 Pro:

On `master`:
- With unmigrated catalog
	- With `GOMAXPROCS=1`: Peak memory=778Mb, Duration=15s
	- With unset GOMAXPROCS, on master branch: Peak memory=731Mb, Duration=10s
- With migrated catalog
	- With `GOMAXPROCS=1` on master branch: Peak memory=166Mb, Duration=2.7s
	- With unset GOMAXPROCS, on master branch: Peak memory=141Mb, Duration=2.0s

On PR branch:
- With unmigrated catalog
	- With `GOMAXPROCS=1` on PR branch: Peak memory=234Mb, Duration=16s
	- With unset GOMAXPROCS, on PR branch: Peak memory=290Mb, Duration=6.6s
- With migrated catalog
	- With `GOMAXPROCS=1` on PR branch: Peak memory=115Mb, Duration=3s
	- With unset GOMAXPROCS, on PR branch: Peak memory=117Mb, Duration=1.39s

joelanford:
I'm going to close this one. I think we should focus on #1278.

Issue #1331, opened 2024-05-29
ensure existing templates enforce explicit schema


grokspawn:

grokspawn:
Fixed by #1335

Issue #1330, opened 2024-05-29
[epic] eliminate the need to specify the catalog template type in render-template


grokspawn:
We can reduce the level of knowledge that opm users need to have about our supported catalog templates if we require an explicit schema for each.  

Steps to accomplish this:
- [x] #1331
- [x] #1336
- [x] #1332
 (for e.g. [verb] [file(s)] [options]) right now basic/semver templates already follow this format, but composite template with its two required imports does not presently
- [x] #1333
- [x] #1334

PR #1335, opened 2024-05-29
provide required schema for basic template

grokspawn:
This standardizes all templates on formal schema (#1331), and lays groundwork for standardization of input (#1332) to support #1334 following a roadmap listed in #1330.

IMPORTANT NOTE!

Working on this, it made a lot of sense for `opm` to own the conversion to a template from FBC, since it controls/supports the templates' definitions.  So this PR introduces a new `opm alpha convert-template` command which explicitly only provides support for the basic template. 

Solves #1332 

Eliminating the need for users (pipelines, authors, etc.) to be aware of which supported template type is being processed. 

**Reviewer Checklist**

### DIFF
--- a/alpha/template/basic/basic.go
+++ b/alpha/template/basic/basic.go
@@ -2,25 +2,58 @@ package basic
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 
 	"github.com/operator-framework/operator-registry/alpha/action"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/pkg/image"
+	"k8s.io/apimachinery/pkg/util/yaml"
 )
+
+const schema string = "olm.template.basic"
 
 type Template struct {
 	Registry image.Registry
 }
 
+type BasicTemplate struct {
+	Schema  string          `json:"schema"`
+	Entries []*declcfg.Meta `json:"entries"`
+}
+
+func parseSpec(reader io.Reader) (*BasicTemplate, error) {
+	bt := &BasicTemplate{}
+	btDoc := json.RawMessage{}
+	btDecoder := yaml.NewYAMLOrJSONDecoder(reader, 4096)
+	err := btDecoder.Decode(&btDoc)
+	if err != nil {
+		return nil, fmt.Errorf("decoding template schema: %v", err)
+	}
+	err = json.Unmarshal(btDoc, bt)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshalling template: %v", err)
+	}
+
+	if bt.Schema != schema {
+		return nil, fmt.Errorf("template has unknown schema (%q), should be %q", bt.Schema, schema)
+	}
+
+	return bt, nil
+}
+
 func (t Template) Render(ctx context.Context, reader io.Reader) (*declcfg.DeclarativeConfig, error) {
-	cfg, err := declcfg.LoadReader(reader)
+	bt, err := parseSpec(reader)
+	if err != nil {
+		return nil, err
+	}
+	cfg, err := declcfg.LoadSlice(bt.Entries)
 	if err != nil {
 		return cfg, err
 	}
 
-	outb := cfg.Bundles[:0] // allocate based on max size of input, but empty slice
+	outb := cfg.Bundles[:0]
 	// populate registry, incl any flags from CLI, and enforce only rendering bundle images
 	r := action.Render{
 		Registry:       t.Registry,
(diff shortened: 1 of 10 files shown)


alpha/template/basic/basic.go
```diff
@@ -48,3 +81,38 @@ func (t Template) Render(ctx context.Context, reader io.Reader) (*declcfg.Declar
 func isBundleTemplate(b *declcfg.Bundle) bool {
 	return b.Schema != "" && b.Image != "" && b.Package == "" && len(b.Properties) == 0 && len(b.RelatedImages) == 0
 }
+
+// FromReader reads FBC from a reader and generates a BasicTemplate from it
+func FromReader(r io.Reader) (*BasicTemplate, error) {
```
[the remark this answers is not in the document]
grokspawn: Once we get some more use/feedback/maturity in the interface, the plan is to eventually promote the catalog templates out of alpha and into pkg.

cmd/opm/alpha/convert-template/convert.go
```diff
@@ -0,0 +1,66 @@
...
+This command outputs a basic catalog template to STDOUT from input FBC.
+If no argument is specified or is '-' input is assumed from STDIN.
+`,
+		RunE: func(c *cobra.Command, args []string) error {
+
+			switch output {
+			case "yaml":
+				fallthrough
+			case "json":
+				converter.OutputFormat = output
```
everettraven, line 44:
```suggestion
			case "yaml", "json":
				converter.OutputFormat = output
```

cmd/opm/alpha/template/cmd.go
```diff
@@ -1,29 +1,27 @@
...
 	runCmd := &cobra.Command{
 		Use:   "render-template",
 		Short: "Render a catalog template type",
 		Args:  cobra.NoArgs,
 	}
 
-	runCmd.AddCommand(newBasicTemplateCmd())
-	runCmd.AddCommand(newSemverTemplateCmd())
+	bc := newBasicTemplateCmd()
+	// bc.Hidden = true
```
everettraven, line 17: Remove?

cmd/opm/alpha/template/cmd.go
```diff
@@ -1,29 +1,27 @@
...
+	runCmd.AddCommand(bc)
 
-	return runCmd
-}
+	sc := newSemverTemplateCmd()
+	// sc.Hidden = true
```
everettraven, line 21: Remove?

cmd/opm/internal/util/util.go
```diff
@@ -66,3 +67,11 @@ func CreateCLIRegistry(cmd *cobra.Command) (*containerdregistry.Registry, error)
 	}
 	return reg, nil
 }
+
+func OpenFileOrStdin(cmd *cobra.Command, args []string) (io.ReadCloser, string, error) {
```
everettraven, line 71: Nit: This is an internal function, and I know it was mostly a move, but the implementation only ever looks at the first argument. It seems reasonable to just make this a single string argument instead of a slice of strings
```suggestion
func OpenFileOrStdin(cmd *cobra.Command, path string) (io.ReadCloser, string, error) {
```

everettraven (approved):
Saw a few small things, but nothing that I think is worth holding this up over if you feel they aren't worth addressing

cmd/opm/alpha/convert-template/convert.go
grokspawn: Reasoning?  When I wrote it I found (for e.g. [here](https://go.dev/wiki/Switch)) no idiomatic argument for either approach.

cmd/opm/internal/util/util.go
grokspawn: This has bugged me too, but it does actually check for [empty args](https://github.com/operator-framework/operator-registry/pull/1335/files#diff-9373b2d6ffc7be864852e489772ff2062a0a2359ff9f3be9f574ba26d69b8fedR72) as a condition for reading from stdin instead.
We could make it pass just the input path and have each caller test for empty args (and pass in an empty string) but then the functions utility is significantly reduced.

cmd/opm/alpha/template/cmd.go
grokspawn: A planned follow-up in https://github.com/operator-framework/operator-registry/issues/1330 will use these.  I was initially doing it as part of this PR, but realized that it would balloon this PR and likely impact the speed of its review.  If you don't feel strongly, I'd just leave these since for some time the commands would be hidden-but-callable.

cmd/opm/internal/util/util.go
everettraven: I don't feel super strongly one way or the other, and saw the empty args check but figured an empty string would work equivalently. This shouldn't block this PR if you'd rather keep it as is.

cmd/opm/alpha/convert-template/convert.go
perdasilva: I was gonna say, if this is possible, that's nice =D I can never remember fallthrough - well not exactly true, diarrhea in German is durchfall. So, I have that mnemonic XD

everettraven: IMO using the `fallthrough` directive:
- increases mental complexity for a reader. They have to first understand what the fallthrough directive does and then build the mental model of "If the yaml case is hit it will fallthrough and do whatever the case underneath of it is, which is the json case and therefore will set the OutputFormat"
- `fallthrough` means execute the logic of the case underneath and could result in an easier introduction of bugs. What happens if someone adds an additional case below the `yaml` case that slips through and completely changes the behavior when `yaml` is specified as the output format?

Instead, using the OR syntax for a switch case is:
- Easier to reason about and build a mental model. Simply looking at the case statement tells me that either the value `yaml` or `json` executes the exact same code.
- Less prone to introducing bugs inadvertently as there is a well defined execution path that is not dependent on the ordering of case statements

perdasilva (approved):
lgtm - though I'd also suggest a someone more across the subject to ensure edge cases, etc. don't fall through the cracks.

cmd/opm/alpha/convert-template/convert.go
grokspawn: I guess it's a question of subjective contexts here.  I infinitely prefer the explicit fallthrough idiom over for e.g. C's implicit (and problematic) fallthrough.  
And similarly, I prefer to have a single evaluated condition examined per case, based on experience with other languages.  
Like I said... context. 

Not a significant factor for me so I made the changes to unblock the review.

(Also, eww Per. 😉)

everettraven (approved):

PR #1338, opened 2024-05-31
generate binary-less dockerfiles

grokspawn:
provide the capability for `opm generate dockerfiles` to generate Dockerfiles which build a binary-less catalog image based on scratch.  

We expose a new `builder-image` flag, allowing users to specify a new image to be used as a builder in the multi-stage dockerfile generated.  This may not be `scratch`. 
`builder-image` and `binary-image` flags both default to `quay.io/operator-framework/opm:latest`
In order to generate a binary-less catalog image dockerfile, users just need to specify `binary-image` of `scratch`. 

***Users who previously set `binary-image` may need to check if they need to also specify `builder-image` to preserve existing behavior***

binary-less catalogs can be built on scratch to make tiny, platform-less catalog sources which can leverage the `.spec.grpcPodConfig.extractContent` fields to allow for on-cluster unpacking.

**Reviewer Checklist**

### DIFF
--- a/alpha/action/generate_dockerfile.go
+++ b/alpha/action/generate_dockerfile.go
@@ -9,11 +9,11 @@ import (
 )
 
 type GenerateDockerfile struct {
-	BaseImage   string
-	IndexDir    string
-	ExtraLabels map[string]string
-	Writer      io.Writer
-	Lite        bool
+	BaseImage    string
+	BuilderImage string
+	IndexDir     string
+	ExtraLabels  map[string]string
+	Writer       io.Writer
 }
 
 func (i GenerateDockerfile) Run() error {
@@ -21,14 +21,7 @@ func (i GenerateDockerfile) Run() error {
 		return err
 	}
 
-	var dockerfileTemplate string
-	if i.Lite {
-		dockerfileTemplate = binlessDockerfileTmpl
-	} else {
-		dockerfileTemplate = dockerfileTmpl
-	}
-
-	t, err := template.New("dockerfile").Parse(dockerfileTemplate)
+	t, err := template.New("dockerfile").Parse(dockerfileTmpl)
 	if err != nil {
 		// The template is hardcoded in the binary, so if
 		// there is a parse error, it was a programmer error.
@@ -47,44 +40,36 @@ func (i GenerateDockerfile) validate() error {
 	return nil
 }
 
-const binlessDockerfileTmpl = `# The builder image is expected to contain
+const dockerfileTmpl = `# The builder image is expected to contain
 # /bin/opm (with serve subcommand)
-FROM {{.BaseImage}} as builder
+FROM {{.BuilderImage}} as builder
 
 # Copy FBC root into image at /configs and pre-populate serve cache
 ADD {{.IndexDir}} /configs
 RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
 
-FROM scratch
+FROM {{.BaseImage}}
+
+{{- if ne .BaseImage "scratch" }}
+# The base image is expected to contain
+# /bin/opm (with serve subcommand) and /bin/grpc_health_probe
+
+# Configure the entrypoint and command
+ENTRYPOINT ["/bin/opm"]
+CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
+{{- else }}
+# OLMv0 CatalogSources that use binary-less images must set:
+# spec:
+#   grpcPodConfig:
+#     extractContent:
+#       catalogDir: /configs
+#       cacheDir: /tmp/cache
+{{- end }}
 
 COPY --from=builder /configs /configs
 COPY --from=builder /tmp/cache /tmp/cache
 
 # Set FBC-specific label for the location of the FBC root directory
-# in the image
-LABEL ` + containertools.ConfigsLocationLabel + `=/configs
-{{- if .ExtraLabels }}
-
-# Set other custom labels
-{{- range $key, $value := .ExtraLabels }}
-LABEL "{{ $key }}"="{{ $value }}"
-{{- end }}
-{{- end }}
-`
-
-const dockerfileTmpl = `# The base image is expected to contain
-# /bin/opm (with a serve subcommand) and /bin/grpc_health_probe
-FROM {{.BaseImage}}
-
-# Configure the entrypoint and command
-ENTRYPOINT ["/bin/opm"]
-CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
-
-# Copy declarative config root into image at /configs and pre-populate serve cache
-ADD {{.IndexDir}} /configs
-RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
-
-# Set DC-specific label for the location of the DC root directory
 # in the image
 LABEL ` + containertools.ConfigsLocationLabel + `=/configs
 {{- if .ExtraLabels }}
--- a/alpha/action/generate_dockerfile.go
+++ b/alpha/action/generate_dockerfile.go
@@ -13,6 +13,7 @@ type GenerateDockerfile struct {
 	IndexDir    string
 	ExtraLabels map[string]string
 	Writer      io.Writer
+	Lite        bool
 }
 
 func (i GenerateDockerfile) Run() error {
@@ -20,7 +21,14 @@ func (i GenerateDockerfile) Run() error {
 		return err
 	}
 
-	t, err := template.New("dockerfile").Parse(dockerfileTmpl)
+	var dockerfileTemplate string
+	if i.Lite {
+		dockerfileTemplate = binlessDockerfileTmpl
+	} else {
+		dockerfileTemplate = dockerfileTmpl
+	}
+
+	t, err := template.New("dockerfile").Parse(dockerfileTemplate)
 	if err != nil {
 		// The template is hardcoded in the binary, so if
 		// there is a parse error, it was a programmer error.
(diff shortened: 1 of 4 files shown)


grokspawn:
Generated dockerfile is in the format:
```dockerfile
# The builder image is expected to contain
# /bin/opm (with serve subcommand)
FROM quay.io/operator-framework/opm:latest as builder

# Copy FBC root into image at /configs and pre-populate serve cache
ADD configs /configs
RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]

FROM scratch

COPY --from=builder /configs /configs
COPY --from=builder /tmp/cache /tmp/cache

# Set FBC-specific label for the location of the FBC root directory
# in the image
LABEL operators.operatorframework.io.index.configs.v1=/configs
```

grokspawn:
binless-catalog-demo

grokspawn:
Solves #1349

cmd/opm/generate/cmd.go
```diff
@@ -83,6 +85,7 @@ value of each duplicate key will be added to the generated Dockerfile.
 		},
 	}
 	cmd.Flags().StringVarP(&baseImage, "binary-image", "i", containertools.DefaultBinarySourceImage, "Image in which to build catalog.")
+	cmd.Flags().BoolVarP(&lite, "lite", "t", false, "Generate a smaller, binary-less Dockerfile.")
```
everettraven, line 88: What do you think of something like `--include-binary=false` to generate the binary-less Dockerfile? To maintain existing functionality it could default to `--include-binary=true`

grokspawn: It's analogous to what I modeled, so I'm fine with it.  I expect there will be a few opinions on this one, so I'm waiting until we get a few before we let them battle royale.
😁

everettraven: I'm not strongly opinionated but just thought I'd make a suggestion for another name since you mentioned folks might not like the `--lite` flag :) - I actually don't mind it so whatever floats your boat here is fine by me

grokspawn: Yeah, `lite` is living on borrowed time. 😄  
We have to effectively communicate that we're both
1. deciding whether or not to use a builder stage; AND 
2. specifying a base image

joelanford: I'm still interested in the option of `--binary-image=scratch` (and we could beef up the CLI help text slightly to explain the pairing with CatalogSource's `spec.grpcPodConfig.extractContent`).

AND I want to again discuss the validity of changing the Dockerfile template to just _always_ be multi-stage.  If you squint, you could call it a bug that `--binary-image=scratch` results in a broken Dockerfile and it should have been multi-stage from the beginning.

For users out there that generate the Dockerfile and then manipulate it afterward, my suggestion in retrospect would be: don't. If you want to modify the Dockerfile after generation, you are better off just maintaining your own Dockerfile.

Long-term I want to eliminate this command and build a better tool that can build and push images without a container runtime, so IMO this is living on borrowed time.

grokspawn: > AND I want to again discuss the validity of changing the Dockerfile template to just always be multi-stage. If you squint, you could call it a bug that --binary-image=scratch results in a broken Dockerfile and it should have been multi-stage from the beginning.

If --binary-image=scratch is the way to indicate that we are building a binless image, how does a user ask for a different builder image?

I'm pushing a new commit to switch to a two-flag approach to reduce ambiguity.  I think maybe we can still get to single-template-land though.

grokspawn:
I think this is ready for another round of review

grokspawn:
Ooh.  Found a cleaner way to handle baseImage conditionality. 😄

cmd/opm/generate/cmd.go
joelanford: > If --binary-image=scratch is the way to indicate that we are building a binless image, how does a user ask for a different builder image?

Great point!

alpha/action/generate_dockerfile.go
```diff
@@ -39,19 +40,28 @@ func (i GenerateDockerfile) validate() error {
...
+
 FROM {{.BaseImage}}
 
+{{- if ne .BaseImage "scratch" }}
+# The base image is expected to contain
+# /bin/opm (with serve subcommand)
 # Configure the entrypoint and command
 ENTRYPOINT ["/bin/opm"]
 CMD ["serve", "/configs", "--cache-dir=/tmp/cache"]
+{{- end }}
```
joelanford, line 59: Maybe add an `else` here (for binless images) to add a comment like:

```
OLMv0 Catalog Sources that use binary-less images must set:

spec:
  grpcPodConfig:
    extractContent:
      catalogDir: /configs
      cacheDir: /tmp/cache

```

alpha/action/generate_dockerfile_test.go
```diff
@@ -41,49 +41,160 @@ func TestGenerateDockerfile(t *testing.T) {
...
+COPY --from=builder /configs /configs
+COPY --from=builder /tmp/cache /tmp/cache
+
+# Set FBC-specific label for the location of the FBC root directory
+# in the image
+LABEL operators.operatorframework.io.index.configs.v1=/configs
+`,
+		},
+		{
+			name: "Lite/Success/WithExtraLabels",
```
joelanford, line 168:
```suggestion
			name: "Scratch/Success/WithExtraLabels",
```

cmd/opm/generate/cmd.go
```diff
@@ -27,7 +27,8 @@ func NewCmd() *cobra.Command {
 
 func newDockerfileCmd() *cobra.Command {
 	var (
-		baseImage      string
+		binaryImage    string
```
joelanford, line 30: You know, I actually prefer `baseImage` because it makes more sense in the context of using "scratch", but the flag name is `binary-image`, so I understand this change.

Another option that crossed my mind (which is more complex code, but better end-user UX):
- Deprecate/hide `--binary-image`
- Add `--base-image` (and let it take over the shortened `-i`)
- Make `--base-image` and `--binary-image` mutually exclusive.

But then I also thought to myself: "But long-term I still want better tooling that doesn't require users to have to do their own docker builds, so maybe we minimize work and complexity here because \[the above suggested changes] don't actually matter all that much."

pkg/containertools/dockerfilegenerator.go
```diff
@@ -8,10 +8,11 @@ import (
 )
 
 const (
-	DefaultBinarySourceImage = "quay.io/operator-framework/opm:latest"
-	DefaultDbLocation        = "/database/index.db"
-	DbLocationLabel          = "operators.operatorframework.io.index.database.v1"
-	ConfigsLocationLabel     = "operators.operatorframework.io.index.configs.v1"
+	DefaultScratchSourceImage = "scratch"
```
joelanford, line 11: I don't think there's any such this as "a different scratch image", so the naming of this feels weird.

I think I'd suggest one of:
- no constant at all
- an unexported constant in `cmd/opm/generate/cmd.go`, but named simply `scratchImage` or similar.

cmd/opm/generate/cmd.go
grokspawn: I mean ... me too on the `base-image` thing, because that's what it is.  In this case, it was exposed as `binary-image` so I just aligned things. 
And I had already played with the same kind of thinking w.r.t. mutual-exclusivity and renaming to `base-image` and it's not much code, so let me push up what I think that looks like.

pkg/containertools/dockerfilegenerator.go
grokspawn: I was originally using this in cmd as well as actions, so a constant where existing constants were made sense.  However, the template changes necessitate a bare string comparison, and evidently the utest uses got lost too, leaving only the one use in cmd... and this is not my hill to die on. ;)

cmd/opm/generate/cmd.go
grokspawn:
```shell
./bin/opm generate dockerfile --binary-image scratch --base-image scratch foo
Flag --binary-image has been deprecated, use --base-image instead
Error: if any flags in the group [binary-image base-image] are set none of the others can be; [base-image binary-image] were all set
Usage:
  opm generate dockerfile <dcRootDir> [flags]
```

grokspawn:
Updated in last so that old calling patterns are preserved, for e.g. if caller only uses deprecated `binary-image` flag, then both `base-image` and `builder-image` are inferred from it.  The resulting Dockerfile is still multi-stage, but using the same image for both stages.

cmd/opm/generate/cmd.go
```diff
@@ -71,19 +83,24 @@ value of each duplicate key will be added to the generated Dockerfile.
...
 			return nil
 		},
 	}
-	cmd.Flags().StringVarP(&baseImage, "binary-image", "i", containertools.DefaultBinarySourceImage, "Image in which to build catalog.")
+	cmd.Flags().StringVar(&baseImage, "binary-image", containertools.DefaultBinarySourceImage, "Image in which to build catalog.")
+	cmd.Flags().StringVarP(&baseImage, "base-image", "i", containertools.DefaultBinarySourceImage, "Image base to use to build catalog.")
+	cmd.Flags().StringVarP(&builderImage, "builder-image", "b", containertools.DefaultBinarySourceImage, "Image to use as a build stage.")
 	cmd.Flags().StringSliceVarP(&extraLabelStrs, "extra-labels", "l", []string{}, "Extra labels to include in the generated Dockerfile. Labels should be of the form 'key=value'.")
+	cmd.Flags().MarkDeprecated("binary-image", "use --base-image instead")
+	cmd.MarkFlagsMutuallyExclusive("binary-image", "base-image")
```
joelanford, line 103: Where did we land on adding this?
```
cmd.MarkFlagsMutuallyExclusive("binary-image", "builder-image")
```

cmd/opm/generate/cmd.go
```diff
@@ -42,10 +43,21 @@ Dockerfile with the same name already exists, this command will fail.
 
 When specifying extra labels, note that if duplicate keys exist, only the last
 value of each duplicate key will be added to the generated Dockerfile.
+
+A separate builder and binary image can be specified. The builder image may not be "scratch".
```
joelanford, line 47:
```suggestion
A separate builder and base image can be specified. The builder image may not be "scratch".
```

joelanford: While we're at it, we could also change the `dc` and `declarative config` references over to `FBC` and `file-based catalog`.

cmd/opm/generate/cmd.go
grokspawn: I thought we said that we'd allow mixing them, while complaining. 
So someone gives us binary-file and we go "that's deprecated", but we also use it appropriately. 
The only special handling is if they use `binary-image` and nothing else, where we use that value for builder AND base.
But I'm not strongly opinionated on this, so let me know if you are.

PR #1384, opened 2024-07-23
add optional schema migrations; default to olm.bundle.object instead of olm.csv.metadata

grokspawn:
Introduces the concept of optional schema migrations to `opm` available to the render and migrate actions, and template expansions (basic and semver).  This includes a sequencing and extension mechanism since we presume that successive schema migrations might be dependent on the successful execution of prior ones. 

Schema migrations represent a level, are explicitly requested by name, and the sequence of all migrations up to and including the named one will be executed in sequence.

If any schema migration fails, it is immediately terminal to the sequence.

The existing `--migrate` flag for `opm render` is equivalent to specifying the last possible `--migrate-level`:
```sh
./bin/opm render --help
Generate a stream of file-based catalog objects to stdout from the provided
catalog images, file-based catalog directories, bundle images, and sqlite
database files.

...

      --migrate                Perform all available schema migrations on the rendered FBC

...
```

Information on available migrations can be obtained from the help for associated migrate/render/semver/basic commands, for e.g.:
```sh
./bin/opm render --help
Generate a stream of file-based catalog objects to stdout from the provided
catalog images, file-based catalog directories, bundle images, and sqlite
database files.

...
      --migrate-level string   Name of the last migration to run (default: none)

                               The migrator will run all migrations up to and including the selected level.

                               Available migrators:
                                 - none                          : do nothing
                                 - bundle-object-to-csv-metadata : migrates bundles' "olm.bundle.object" to "olm.csv.metadata"
...
```

When we introduced the new [bundle metadata type](https://github.com/operator-framework/operator-registry/pull/1094) `olm.csv.metadata` and made it the default, we started hearing of authors having to select non-latest versions of `opm` in order to preserve their pipelines' functionality.  This is an anti-pattern, since we want to ensure that authors can safely use versions of `opm` that have the most fixes. 

**Reviewer Checklist**


joelanford:
> The existing --migrate flag for opm render is deprecated, and will exit with an error if used

That sounds like a breaking change. We _could_ make `--migrate` and `--migrate-level` mutually exclusive. If we did that, I think we would need to decide what `--migrate` means. I can think of two options:
1. It means the exact same thing as `--migrate-level=bundle-object-to-csv-metadata` (because that's what it means right now).
2. It means "do all of the available migrations". That's what I had imagined it would mean when we introduced it.

alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,91 @@
...
+
+func HelpText() string {
+	var help strings.Builder
+	help.WriteString("  The migrator will run all migrations up to and including the selected level.\n\n")
+	help.WriteString("  Available migrators:\n")
+	if len(allMigrations) == 0 {
+		help.WriteString("   (no migrations available in this version)\n")
+	}
+	for _, migration := range allMigrations {
+		help.WriteString(fmt.Sprintf("  - %s\n", migration.Name()))
```
joelanford, line 79: Include `migration.Help()` here?

alpha/action/migrations/000_bundle_object_to_csv_metadata.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"encoding/json"
+
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
+)
+
+var BundleObjectToCSVMetadata = newMigration(
```
joelanford, line 12: This should probably be unexported now that I'm looking closer at it.

alpha/action/render.go
```diff
@@ -416,50 +414,20 @@ func moveBundleObjectsToEndOfPropertySlices(cfg *declcfg.DeclarativeConfig) {
...
+		return err
 	}
 
-	for _, m := range migrations {
-		if err := m(cfg); err != nil {
+	for _, m := range (*mobj).Migrations {
+		if err := m.Migrate(cfg); err != nil {
 			return err
 		}
 	}
```
joelanford, line 427: Call `mobj.Migrate()` here (it encapsulates this loop).

alpha/property/property.go
```diff
@@ -257,6 +257,9 @@ func MustBuildGVKRequired(group, version, kind string) Property {
 func MustBuildBundleObject(data []byte) Property {
 	return MustBuild(&BundleObject{Data: data})
 }
+func MustBuildBundleObjectData(data []byte) Property {
+	return MustBuild(&BundleObject{Data: data})
+}
```
joelanford, line 262: This looks identical to `MustBuildBundleObject` above. Do we need both?

alpha/template/semver/types.go
```diff
@@ -4,13 +4,15 @@ import (
...
 	"github.com/operator-framework/operator-registry/pkg/image"
 )
 
 // data passed into this module externally
 type Template struct {
-	Data     io.Reader
-	Registry image.Registry
+	Data         io.Reader
+	Registry     image.Registry
+	MigrateLevel string
```
joelanford, line 15: Nit: I see some fields called `MigrateLevel` and others called `MigrationLevel`. Should we make those consistent?

Only exception to this rule I could see if the `Migrate.Level` field, which makes sense to me to be named differently to avoid stuttering.

cmd/opm/alpha/render-graph/cmd.go
```diff
@@ -61,6 +62,9 @@ $ opm alpha render-graph quay.io/operatorhubio/catalog:latest | \
 			render.AllowedRefMask = action.RefDCImage | action.RefDCDir | action.RefSqliteImage | action.RefSqliteFile
 			render.Registry = registry
 
+			// Run all migrations
+			render.MigrationLevel = ""
```
joelanford, line 66: Doesn't this mean "run no migrations"?

cmd/opm/migrate/cmd.go
```diff
@@ -51,5 +52,10 @@ parsers that assume that a file contains exactly one valid JSON object.
 		},
 	}
 	cmd.Flags().StringVarP(&output, "output", "o", "json", "Output format (json|yaml)")
+	cmd.Flags().StringVar(&migrate.Level, "migrate-level", "", "Name of the last migration to run (default: none)\n"+migrations.HelpText())
+	_ = cmd.RegisterFlagCompletionFunc("migrate-level", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{migrations.HelpText()}, cobra.ShellCompDirectiveDefault
```
joelanford, line 57: What does this do? I understand at a high level that it helps with shell completion when a user is tabbing in their CLI after typing `--migrate <TAB>`.

But what happens with that? Will a user be able to tab through the names of each level?

alpha/property/property.go
grokspawn: Vestigial.  Cleaned up.

cmd/opm/alpha/render-graph/cmd.go
grokspawn: Good catch.  I didn't write that, and `opm render-graph` does not care about bundle metadata, so we shouldn't set anything.

cmd/opm/migrate/cmd.go
grokspawn: This was a false start.  I'd hoped that it would allow us to provide a list of available migrators as a tab-completion, but it doesn't work except for bash, and I'm generally disenchanted with it.  
I left in, in case someone was going to point out my error, since I exhausted search results. 
I'll kill it, at least for now.

grokspawn:
> > The existing --migrate flag for opm render is deprecated, and will exit with an error if used
> 
> That sounds like a breaking change. We _could_ make `--migrate` and `--migrate-level` mutually exclusive. If we did that, I think we would need to decide what `--migrate` means. I can think of two options:
> 
>     1. It means the exact same thing as `--migrate-level=bundle-object-to-csv-metadata` (because that's what it means right now).
> 
>     2. It means "do all of the available migrations". That's what I had imagined it would mean when we introduced it.

This is a breaking change as it is, and I tried to be really up-front to it.  I wanted something more elegant for the default-runs-all case than [this](https://github.com/operator-framework/operator-registry/pull/1384/commits/9e68d1c63e2303bcc99b16fadae9f87b33beb9bb#diff-e8b54ecee4543731d84f335c27814412b3419c50c057696d104012ede0b981f8R44).  There's also the argument to be made that if using the deprecation flag is an easy way to 'select all' then it would be the most common interaction users will have, and why would we want to deprecate it (versus providing overlapping options).

Separately, I'm uncomfortable with the reliance on bare strings instead of untyped data for the migration levels, and I was going to spend some more time with it, but in all honesty that can be follow-on work after this merges.

grokspawn:
NB: I haven't updated the PR description based on the recent decision to reverse what the migrate flag means to render, since I think we're on the same page here but not 100% certain.  I'll update that with the ultimate result of the review.

alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,122 @@
...
+
+type Migration interface {
+	Token() MigrationToken
+	Name() string
+	Help() string
+	Migrate(*declcfg.DeclarativeConfig) error
+}
+
+func newMigration(name string, help string, fn func(config *declcfg.DeclarativeConfig) error) Migration {
+	s := strings.ReplaceAll(help, "`", "\"")
```
grokspawn, line 24: This is because of [this](https://github.com/spf13/pflag/issues/200), where expansion of the template for the usage string imparts specific meaning to the first set of backticks it encounters.

alpha/action/migrate_test.go
```diff
@@ -16,6 +17,13 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
 )
 
+type migrationLevel string
+
+const (
+	bundleObjectType migrationLevel = "bundle-object"
```
joelanford, line 23: This is basically "no migrations", right?

I think we talked about this, but I feel like these tests could get very cumbersome once we have more than 2 migrations.

What if, instead, these tests ran all the migrations, and then we have a separate set of unit tests for the `migrations` package?

alpha/action/migrate.go
```diff
@@ -5,13 +5,15 @@ import (
...
 
+	"github.com/operator-framework/operator-registry/alpha/action/migrations"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/pkg/image"
 )
 
 type Migrate struct {
 	CatalogRef string
 	OutputDir  string
+	Level      migrations.MigrationToken
```
joelanford, line 16: Thinking about this a little more, what if we put `*migrations.Migrations` here (and on the other render-based structs) instead of the level? That way:
- tests of the render-based structs could use a simple test-only `Migrations` struct to verify that the migrations are executed (it doesn't really matter _what_ those migrations are)
- Other users of the library can use other migrations if they want.

alpha/action/migrations/migrations.go
joelanford: Instead of doing this "magic" replacement, should we just have the original `help` string use `"` directly?

alpha/action/migrations/000_bundle_object_to_csv_metadata.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
+
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
+	"github.com/operator-framework/operator-registry/alpha/property"
+)
+
+var bundleObjectToCSVMetadata = newMigration(
+	"bundle-object-to-csv-metadata",
+	"migrates bundles' `olm.bundle.object` to `olm.csv.metadata`",
+	func(cfg *declcfg.DeclarativeConfig) error {
```
joelanford, line 15: Nit: If we wanted the names and descriptions of all the migrations in one place, we could do something like the following instead.

Here: get rid of the `newMigration` wrapping and directly declare the migration function:
```go
func migrateBundleObjectsToCSVMetadata(cfg *declcfg.DeclarativeConfig) error { ... }
```

In `migrations.go` in line 74 where we setup `allMigrations`, add the `newMigration` wrapping:
```go
newMigration("bundle-object-to-csv-metadata", `migrates bundles' "olm.bundle.object" to "olm.csv.metadata"`, migrateBundleObjectsToCSVMetadata)
```

alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,122 @@
...
+	return s.help
+}
+
+type Migrations struct {
+	Migrations []Migration
+}
+
+func GetLastMigration() MigrationToken {
+	if len(allMigrations) == 0 {
+		return ""
```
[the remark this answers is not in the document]
joelanford: Or even something like 
```go
newMigrations("none", "no migration", func(_ *declcfg.DeclarativeConfig) error { return nil}))
```

alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,122 @@
...
+	var help strings.Builder
+	help.WriteString("\nThe migrator will run all migrations up to and including the selected level.\n\n")
+	help.WriteString("Available migrators:\n")
+	if len(allMigrations) == 0 {
+		help.WriteString("   (no migrations available in this version)\n")
+	}
+
+	tabber := tabwriter.NewWriter(&help, 0, 0, 1, ' ', 0)
+	for _, migration := range allMigrations {
+		fmt.Fprintf(tabber, "  - %s\t: %s\n", migration.Name(), migration.Help())
```
joelanford, line 109: We can use `migration.Token()` without other changes:
```suggestion
		fmt.Fprintf(tabber, "  - %s\t: %s\n", migration.Token(), migration.Help())
```

Go is able to figure out that `MigrationToken` is a string, so when it interprets it with `%s` formatting, it just uses the raw underlying string. If, in the future we need a special "token" to "name" function for some reason, I think we can define `String()` of the `MigrationToken` type and put whatever translation logic we want.

joelanford: If we do this, I think we can remove the `name` field and `Name()` method related to a migration.

alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func ValidateName(name string) (MigrationToken, error) {
+	for _, migration := range allMigrations {
+		if migration.Name() == name {
+			return migration.Token(), nil
+		}
+	}
+	return MigrationTokenInvalid, fmt.Errorf("unknown migration level %q", name)
+}
```
joelanford, line 69: I feel like we should be able to incorporate this function into `NewMigrations` (or make it unexported and call it from there), especially if you agree that it makes sense for the `Render` struct (and others) to accept a *migrations.Migrations` directly, as I suggested elsewhere.

alpha/action/migrate_test.go
joelanford: (actually see my other comment about putting `*migrations.Migrations` on these structs and then just defining a dummy test `Migrations` instead)

alpha/action/migrations/000_bundle_object_to_csv_metadata.go
grokspawn: LOL!  I /almost/ did that already, but I was making a concerted effort to not make 'nice to have' changes in my update, in order to get this in sooner.  
But now I think there's no excuse not to make the changes. 😁

alpha/action/migrate_test.go
grokspawn: There's a nuance here which is not the same as elsewhere.  I agree that we ought to be clear that what we're testing here is all/nothing and not incremental.  But in this case these are just coincidentally using identifiers that align with the current [0...1] options. 
We're using it in testdata construction to align with the selected all/none migrations. 
I've renamed/retyped this area to make it hopefully more clear.  WDYT?

grokspawn:
/hold 
still working on the migrations_test in alpha/action/migrations.  Just pushed up to share a reference.

alpha/action/migrate_test.go
grokspawn: Adjusted this as we discussed to 
1. test only the none/arbitrary migrations cases, to verify that migrations-as-a-mechanism works;
2. introduce unit tests in actions.migrations which allow us to evaluate our sequence of migrators to protect against surprises along the processing chain

alpha/action/migrate_test.go
```diff
@@ -126,6 +182,10 @@ func TestMigrate(t *testing.T) {
 				require.Equal(t, expectedData, string(actualData))
 				return nil
 			})
+
+			if s.migrationCount != 0 {
+				require.Equal(t, s.migrationCount, migrationPre+s.migrationCount)
```
joelanford, line 187: I very well could be misunderstanding this code, but this will fail if `migrationPre` is ever non-zero, right?

And the only reason it doesn't fail now is because there's only one spec that increments the counter?

I'm having trouble conjuring an approach that remains simple and overcomes this problem, but I'll clone and tinker later.

alpha/action/migrations/migrations.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, migration := range migrations {
+		keep = append(keep, migration)
+		if migration.Token() == token {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return nil, fmt.Errorf("unknown migration level %q", name)
+	}
```
joelanford, line 94: Now that we `validateName`, it seems like this is unreachable code? Should we drop this (or validateName)?

alpha/action/migrations/migrations_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func TestMigrations(t *testing.T) {
+	noneMigration, err := NewMigrations(NoMigrations)
+	require.NoError(t, err)
+	csvMigration, err := NewMigrations("bundle-object-to-csv-metadata")
+	require.NoError(t, err)
+	allMigrations, err := NewMigrations(AllMigrations)
+	require.NoError(t, err)
+
+	var evaluators map[MigrationToken]func(*declcfg.DeclarativeConfig) error = map[MigrationToken]func(*declcfg.DeclarativeConfig) error{
```
joelanford, line 25:
```suggestion
	evaluators := map[MigrationToken]func(*declcfg.DeclarativeConfig) error{
```

alpha/action/migrations/migrations_test.go
```diff
@@ -0,0 +1,132 @@
...
+			if diff := cmp.Diff(*d, csvMetadataCatalogFBC); diff == "" {
+				return fmt.Errorf("unexpected result of migration\n%s", diff)
+			}
+			return nil
+		},
+	}
+
+	tests := []struct {
+		name      string
+		migrators *Migrations
```
joelanford, line 42: WDYT about adding `input` and `expectedOutput` struct fields here instead of the `evaluators` indirection?

If we did that, the test code would turn into something like:
```
assert.NoError(t, migators.Migrate(t.input))
assert.Empty(t, cmp.Diff(t.input, t.expectedOutput))
```

alpha/action/migrations/migrations_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	var evaluators map[MigrationToken]func(*declcfg.DeclarativeConfig) error = map[MigrationToken]func(*declcfg.DeclarativeConfig) error{
+		MigrationToken(NoMigrations): func(d *declcfg.DeclarativeConfig) error {
+			if diff := cmp.Diff(*d, unmigratedCatalogFBC); diff != "" {
+				return fmt.Errorf("'none' migrator is not expected to change the config\n%s", diff)
+			}
+			return nil
+		},
+		MigrationToken("bundle-object-to-csv-metadata"): func(d *declcfg.DeclarativeConfig) error {
+			if diff := cmp.Diff(*d, csvMetadataCatalogFBC); diff == "" {
```
joelanford, line 33: ?
```suggestion
			if diff := cmp.Diff(*d, csvMetadataCatalogFBC); diff != "" {
```

alpha/action/migrations/migrations.go
grokspawn: Yeah, this is dead, but my intention was to throw in validateName temporarily to validate the tokenizing, and evidently forgot to replace.  I don't think we gain much efficiency from skipping one loop, but we'll have it now. 😉

alpha/action/migrations/migrations_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+		{
+			name:      "MigrationSequence",
+			migrators: allMigrations,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var config declcfg.DeclarativeConfig = unmigratedCatalogFBC
```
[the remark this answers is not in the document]
grokspawn: I was thinking to eval how golang performed a shallow copy by doing the value assignment, but... it seemed to work.  I definitely don't want to pollute the test sources, so I'll still func-ionalize it even though it appeared to be doing OK.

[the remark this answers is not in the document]
joelanford: yeah, you'll get a different `declcfg.DeclarativeConfig` object, but the copy and the original would have the same underlying slices.

alpha/action/migrate_test.go
grokspawn: Yes, this does an explicit compare against the counter value in the test spec... which will only work when the count of TCs attempting this feat is one.  
It should be verifying that the count went up by the amount in the test spec, instead:
```golang
require.Equal(t, migrationCounter, migrationPre+s.migrationCount)
```

alpha/action/migrations/migrations_test.go
grokspawn: This revealed that I was missing bundle.Image in my testdata, which was an oversight that makes the whole scenario collapse... so well spotted!

alpha/action/migrations/migrations_test.go
grokspawn: evaluators solves the problem of validating the sequence of migrators indicated by a migration label, whereas a test case may invoke multiple migrators (since the migrators label indicates a sequence [1..n] rather than a single value).  It's one-to-one with the available migrators so we can attest to no side effects as we progress through a migrator chain, not one-to-spec. 
i.e. 
 input --> none(unmutated) --> csvMetadata(mutated) --> futureMigration(mutatedagain).

pkg/registry/registry_to_model_test.go
```diff
@@ -57,8 +57,6 @@ func testExpectedProperties(t *testing.T) []property.Property {
 			Type:  "olm.constraint",
 			Value: json.RawMessage(`{"cel":{"rule":"properties.exists(p, p.type == \"certified\")"},"failureMessage":"require to have \"certified\""}`),
 		},
-		// property.MustBuildCSVMetadata(csv),
-
```
joelanford: Delete this entirely?

joelanford (approved):

PR #1417, opened 2024-08-19
model: validate unique bundle versions

joelanford:
Validate that bundles within a package use unique semver versions.

This has always been assumed. Now we are explicitly checking.

I verified that the operatorhubio catalog and all of Red Hat's catalogs for OCP 4.15 and 4.16 continue to pass with this new validation.

Are there any other catalogs that should be checked?

**Reviewer Checklist**


grokspawn:
solves #1572

PR #1423, opened 2024-08-22
template/*: allow passing custom bundle renderer

joelanford:
Make the template rendering libraries accept a bundle renderer function.

This allows importers of operator-registry to provide custom rendering logic for bundle references that appear in template configuration files.

This would enable third-parties to build experimental bundle transport formats and continue using the standard operator-registry template tooling with them.

**Reviewer Checklist**

### DIFF
--- a/alpha/template/semver/semver.go
+++ b/alpha/template/semver/semver.go
@@ -10,7 +10,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 	"sigs.k8s.io/yaml"
 
-	"github.com/operator-framework/operator-registry/alpha/action"
 	"github.com/operator-framework/operator-registry/alpha/declcfg"
 	"github.com/operator-framework/operator-registry/alpha/property"
 )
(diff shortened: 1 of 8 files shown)


joelanford:
Seems like the semver template explicitly expects the references in the template file to show up in the rendered FBC. Looking to see if that can be relaxed somehow because that breaks the use case I'm trying to make work (render a bundle from a local OCI archive file)

alpha/template/semver/semver.go
```diff
@@ -25,22 +24,16 @@ func (t Template) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error
 
 	var cfgs []declcfg.DeclarativeConfig
 
-	bundleDict := make(map[string]struct{})
-	buildBundleList(&sv.Candidate.Bundles, &bundleDict)
-	buildBundleList(&sv.Fast.Bundles, &bundleDict)
-	buildBundleList(&sv.Stable.Bundles, &bundleDict)
-
+	bundleDict := buildBundleList(*sv)
```
perdasilva, line 27: nit: would it be worth renaming sv to make the code a bit more readable?

alpha/template/semver/semver.go
```diff
@@ -61,12 +54,16 @@ func (t Template) Render(ctx context.Context) (*declcfg.DeclarativeConfig, error
 	return &out, nil
 }
 
-func buildBundleList(bundles *[]semverTemplateBundleEntry, dict *map[string]struct{}) {
-	for _, b := range *bundles {
-		if _, ok := (*dict)[b.Image]; !ok {
-			(*dict)[b.Image] = struct{}{}
+func buildBundleList(t semverTemplate) map[string]string {
+	dict := make(map[string]string)
+	for _, bl := range []semverTemplateChannelBundles{t.Candidate, t.Fast, t.Stable} {
```
perdasilva, line 59: nit: wdyt about making the making the semverTemplateChannelBundles list a parameter? Before we could see which channel templates were being used to create the bundle list at the top level.  Might also make the method more flexible? (though given that it isn't exported - maybe not...XD).

alpha/template/semver/types.go
```diff
@@ -1,19 +1,18 @@
...
 
 	"github.com/blang/semver/v4"
 
-	"github.com/operator-framework/operator-registry/alpha/action/migrations"
-	"github.com/operator-framework/operator-registry/pkg/image"
+	"github.com/operator-framework/operator-registry/alpha/declcfg"
 )
 
 // data passed into this module externally
 type Template struct {
```
perdasilva, line 13: suggestion: should we just make Template (as defined in basic with just the RenderBundle method) a shared type that basic and semver sub-type and implement? The basic and semver packages could even export a factory functions to get new instances of templates?

perdasilva (commented):
If I'm understanding it right, aside from updating some error messages to be more readable (thank you!), we're moving the Template type(s) to be more flexible, allowing users to implement their own rendering behavior. In general it lgtm. I've left a few comments and suggestions, but nothing that I think is blocking or that we couldn't address later. I don't know enough about the downstream (_not_ as in to _the_ downstream) side-effects of this change to approve. I'd think @grokspawn would be the best person here to reason about the big picture effects of this change (though I doubt there will be anything too serious).

perdasilva:
@joelanford Oh shit! ocp-bot merged it coz I put /lgtm :face_with_diagonal_mouth: 
should we disable it?

PR #1448, opened 2024-09-10
explicitly handle SIGINT and SIGTERM

joelanford:

This PR adds explicit handling for SIGINT and SIGTERM, so that we can catch those signals and propagate them via a Go context to long running processes in order to perform a graceful shutdown.

Temporary files are being left behind because `defer` statements are not executed when these signals are not caught and handled by the program. This PR ensures that we run `defer` statements prior to exiting.

**Reviewer Checklist**


perdasilva (approved):
If I understand the changes correctly, we are now using the `signal` package to create a top context that will cancel on SIGINT or SIGTERM. This context is set as the cmd context, and each  (long running) opm command creates a sub-context (with the top level context as parent) for its work. These commands also spawn a go routine to monitor the context and shutdown gracefully if it gets cancelled. Therefore, if the command's context gets cancelled -> graceful shutdown, if the process gets a SIGINT or SIGTERM, the top level process gets cancelled and propagates down to the sub-contexts -> graceful shutdown.

We're getting a breaking api change by deleting shutdown.go - but that _should_ be ok as this package is only really meaningful for the command structure (as opposed to manipulating catalog content). So, lgtm ^^

PR #1491, opened 2024-11-12
add macos codesigning of opm binary for apple silicon

grokspawn:
Due to apple's requirement that all apple silicon apps must be signed, the app make targets here will not be able to be executed unless they are signed in macos sequoia and later.

For example, if this generates the `opm` binary at $HOME/devel/operator-registry/bin/opm and we invoke it, without being signed the result will be `Killed: 9` messages for each attempted invocation of the compiled unsigned app, for e.g.:

```sh
╰$ ./bin/opm --help
[1]    14243 killed     ./bin/opm --help
```

This is a completely naive approach to rectifying this, and we probably need some more thought on the subject, but this approach at least lets me execute locally-built binaries to examine catalog content.

**Reviewer Checklist**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,10 @@ $(CMDS):
 $(OPM): opm_version_flags=-ldflags "-X '$(PKG)/cmd/opm/version.gitCommit=$(GIT_COMMIT)' -X '$(PKG)/cmd/opm/version.opmVersion=$(OPM_VERSION)' -X '$(PKG)/cmd/opm/version.buildDate=$(BUILD_DATE)'"
 $(OPM):
 	$(extra_env) $(GO) build $(opm_version_flags) $(extra_flags) $(TAGS) -o $@ ./cmd/$(notdir $@)
+ifeq ($(shell go env GOARCH),arm64)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -55,6 +55,10 @@ $(CMDS):
 $(OPM): opm_version_flags=-ldflags "-X '$(PKG)/cmd/opm/version.gitCommit=$(GIT_COMMIT)' -X '$(PKG)/cmd/opm/version.opmVersion=$(OPM_VERSION)' -X '$(PKG)/cmd/opm/version.buildDate=$(BUILD_DATE)'"
 $(OPM):
 	$(extra_env) $(GO) build $(opm_version_flags) $(extra_flags) $(TAGS) -o $@ ./cmd/$(notdir $@)
+ifeq ($(shell go env GOARCH),arm64)
```
tmshort, line 58: This will also hit Raspberry Pi, and miss those older Intel models that can run Sequoia. Did you want to check explicitly for macOS?
```suggestion
ifeq ($(shell go env GOOS),darwin)
```

grokspawn: I didn't want to test GOOS, because intel-based darwin doesn't have this requirement.  I couldn't figure out how to do a double conditional well to select only the darwin+arm64 set of targets.

PR #1517, opened 2024-12-12
Update operator.yaml - Just remove kube-rbac-proxy from RelatedImages at fbc-dir

camilamacedo86:
The motivation here is to remove from the mock gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0 since it is deprecated and, at any point, can no longer be available.


camilamacedo86:
I will open a PR with removing the reference for all places

PR #1524, opened 2024-12-20
enable linting and bingo management of the lint tool

### LINKED ISSUE #1525: lint make target... doesn't lint
Discovered reviewing #1523 that the lint target [only formats imports](https://github.com/operator-framework/operator-registry/pull/1523#issuecomment-2557445207).

Solved by #1524

grokspawn:
Closes #1525 

Discovered that the lint make target doesn't lint anything.  It only checks whether go imports are in expected grouping/ordering. 

This PR introduces gci support (replacing goimports) as a targeted linter, and implements golang-linter support via a bingo-managed version. 

**There are discrete commits to topically collect:**
1. adding lint infrastructure
2. automated lint fixes
3. manual test lint fixes
4. manual functional lint fixes
5. bingo update of golangci-lint version
6. review updates

**The parent issue has a link to a google sheet which includes a list of every single update made here, categorized and broken down by reporting linter.**

The ONLY FOCUS of this PR is to establish a "bright line" beyond which no new reports will slip through the cracks, but due to the age (and the "feature completeness") of much of the reported code, there are a LOT of issues which are **specifically and locally disabled** to remove them from lint reports without making updates. 

I'd initially felt sanguine about updating style changes which modified (read: broke) APIs, but I backed away from this after seeing how prevalent those changes would need to be.

The only included case of a stylistic breaking API change was the JSONUnmarshaller, since this is a recent addition that was added for internal use, so it seemed a safe bet to not break anyone's access. 

I was initially much more willing to make updates to test code to bring them up to date, but 
1. again, the scope of this was soul-destroying; and
2. I started encountering some unit test failures because the tests weren't seamlessly constructed.  These were of the nature of "called twice" or "doesn't need to be called" failure scenarios, and I just restored the original flows and disabled the specific linter in that specific scenario

**Reviewer Checklist**


.golangci.yaml
```diff
@@ -0,0 +1,73 @@
...
+#  of the above repositories!
+#
+#  TODO: Find a way to have a shared golangci config.
+########
+
+run:
+  # Default timeout is 1m, up to give more room
+  timeout: 4m
+
+linters:
```
perdasilva, line 20: you don't have to, but if it's easy enough, we could add govet here as well (then maybe drop it from the Makefile if it's there)

.golangci.yaml
```diff
@@ -68,6 +68,15 @@ linters-settings:
     - pkg: github.com/blang/semver/v4
       alias: bsemver
 
+issues:
+  # exclusion rules, mostly to avoid functionally complete areas and reduce unit test noise
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - unused
+        - govet
```
perdasilva, line 77: Just because I left the comment in the other commit XD

alpha/action/migrate_test.go
```diff
@@ -125,7 +125,7 @@ func TestMigrate(t *testing.T) {
 				return
 			}
 			actualFS := os.DirFS(s.migrate.OutputDir)
-			fs.WalkDir(actualFS, ".", func(path string, d fs.DirEntry, err error) error {
+			_ = fs.WalkDir(actualFS, ".", func(path string, d fs.DirEntry, err error) error {
```
perdasilva, line 128: we could also maybe `require.NoError`

alpha/declcfg/write_test.go
```diff
@@ -500,13 +500,13 @@ func removeJSONWhitespace(cfg *DeclarativeConfig) {
 	for ib := range cfg.Bundles {
 		for ip := range cfg.Bundles[ib].Properties {
 			var buf bytes.Buffer
-			json.Compact(&buf, cfg.Bundles[ib].Properties[ip].Value)
+			_ = json.Compact(&buf, cfg.Bundles[ib].Properties[ip].Value)
```
perdasilva, line 503: maybe also `require.NoError` ?

alpha/declcfg/write_test.go
```diff
@@ -500,13 +500,13 @@ func removeJSONWhitespace(cfg *DeclarativeConfig) {
...
 			cfg.Bundles[ib].Properties[ip].Value = buf.Bytes()
 		}
 	}
 	for io := range cfg.Others {
 		var buf bytes.Buffer
-		json.Compact(&buf, cfg.Others[io].Blob)
+		_ = json.Compact(&buf, cfg.Others[io].Blob)
```
perdasilva, line 509: ditto

test/e2e/bundle_image_test.go
```diff
@@ -276,7 +276,7 @@ func pushLoadImages(client *kubernetes.Clientset, w io.Writer, images ...string)
 		}
 	} else {
 		for _, image := range images {
-			pushWith("docker", image)
+			err = pushWith("docker", image)
```
perdasilva, line 279: nit:

```
if err := pushWith("docker", image); err != nil {
	return err
}
```

.golangci.yaml
grokspawn: the .golangci.yaml file in this project is a copy of that in op-con.  Since that one has just updated, I can take that change, but IMO we can update the Makefile in a separate PR.  This one is already way beyond any reasonable complexity.

alpha/declcfg/write_test.go
grokspawn: No test object available in this scope, or I would've. 🤷

alpha/declcfg/write_test.go
grokspawn: Same problem as pre-ditto.

test/e2e/bundle_image_test.go
grokspawn: err is declared in line 262.  Is there some benefit to re-defining it in this scope? (I'd missed this)

.golangci.yaml
camilamacedo86: Hi @grokspawn 

You are with an outdated version :-P 
See that we added govet there and we cleanup the staff: https://github.com/operator-framework/operator-controller/commit/099a6cf8ce5f38f35a109f27b49f8ae2f0129290

camilamacedo86: But I am ok with we move forward as it is and add govet in a follow up. 
It is fine for me too. @perdasilva wdyt?

grokspawn: I did add govet in [ac872deec7a3f69452cdf7d1cdcc38b7dec70911](https://github.com/operator-framework/operator-registry/pull/1524/commits/ac872deec7a3f69452cdf7d1cdcc38b7dec70911)

azych (approved):
That's great and I completely agree with your decision to back out of actually fixing those lint issues in favor of silencing most of them just to establish a baseline given the already significant scope of work - I'm sure people can be good scouts and gradually make fixes as they touch upon different areas of the code they work on.

camilamacedo86 (approved):
/lgtm

🚀

Issue #1591, opened 2025-02-24
Add flag for auth file


jotak:
Hi,
I've seen that [opm supports auth file](https://github.com/operator-framework/operator-registry/issues/919) through `REGISTRY_AUTH_FILE` env
For better discoverability, would it be possible to add that as a global flag, and document it in `opm --help` ?

whatsacomputertho:
Agreed with OP.  To add to this, is there any chance we could even default `REGISTRY_AUTH_FILE` to `$HOME/.config/containers/auth.json` (mac / win) / `${XDG_RUNTIME_DIR}/containers/auth.json` (linux)?  This way we could just `podman login` / `skopeo login` and inherit that auth file in `opm`

joelanford:
> To add to this, is there any chance we could even default REGISTRY_AUTH_FILE to $HOME/.config/containers/auth.json (mac / win) / ${XDG_RUNTIME_DIR}/containers/auth.json (linux)? This way we could just podman login / skopeo login and inherit that auth file in opm

This feature was implemented already and landed in 1.33.0. For backward compatibility reasons, it is a bit tricky though.

First we derive an auth file. Precedence for selecting an authfile is:
- `REGISTRY_AUTH_FILE`, if set.
- `$DOCKER_CONFIG/config.json`, if set.
- `$HOME/.docker/config`

Once we have chosen an auth file path we pick a credential, with the following precedence:
- A credential in the auth file, if the file is present and readable and it contains a matching credential.
- A credential based on the `containers/image` default file locations and lookup logic. (This is what will provide access to the `.config/containers/auth.json` entries).

Given the existing complexity of this logic, I'm hesitant to add yet another possible option. More likely, we are looking to revamp and cut out large portions of `opm`'s legacy code, and either make a `v2` or release an entire new tool (or tools). At that point, I have hope that the CLI will have a near identical UX as the rest of the `containers/image` ecosystem.

(Only concern there is `containers/image` lack of support for arbitrary OCI artifact hierarchies, last time I checked at least).

PR #1626, opened 2025-03-25
Containers image registry implementation with opt-in caching

joelanford:
This change adds a new `containers/image`-based registry client and updates various (non-deprecated) commands to use it. Specifically `render` (and other commands that use `action.Render`) and `alpha bundle validate`. Deprecated sqlite based commands are not updated.

The intent of the PR is to keep essentially full compatibility with the `containerd` implementation, and to additionally gain the features of containers/image-based clients.

This PR also includes the ability to opt into caching (and reusing) downloaded blobs (when `OLM_CACHE_DIR` is set). The behavior around the cache is essentially `--pull-policy=Always`, where existing layers do not need to be redownloaded. While this PR does garbage collect unused blobs in the cache after each image pull, specifically cleaning up the cache is (for now at least) left as an exercise for the user.

Improves alignment of `opm` with the ecosystem of tools it is often used with. The main benefit here is that `opm` calls that use this client will now honor user's existing `containers/image`-based configuration, including `registries.conf` CAs and image mappings.

**Reviewer Checklist**


Makefile
```diff
@@ -102,8 +100,7 @@ image-upstream:
 	docker build -f upstream-example.Dockerfile .
 
 .PHONY: lint
-#lint:
-#	find . -type f -name '*.go' ! -name '*.pb.go' -print0 | xargs -0 goimports -w
+GOLANGCI_LINT_ARGS := --build-tags=$(TAGS)
```
grokspawn, line 103: When I was adding the linter to this repo I was making use of an externally-declared GOLANGCI_LINT_ARGS.  Can we either
```suggestion
GOLANGCI_LINT_ARGS += --build-tags=$(TAGS)
```
or
```suggestion
GOLANGCI_LINT_ARGS := "$(GOLANGCI_LINT_ARGS) --build-tags=$(TAGS)"
```

joelanford: Ah, I was wondering why I wasn't seeing that anywhere. Yeah, I can make this change.

pkg/image/containersimageregistry/registry.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func newCacheConfig(dir string, preserve bool) *cacheConfig {
+	return &cacheConfig{
+		baseDir:  dir,
+		preserve: preserve,
+	}
+}
+
+func ForceTemporaryImageCache() Option {
```
grokspawn, line 104: Why not the idiomatic `With$OPTION` syntax?

pkg/image/containersimageregistry/registry.go
```diff
@@ -0,0 +1,290 @@
...
+			return err
+		}
+		return nil
+	}
+}
+
+func InsecureSkipTLSVerify(insecureSkipTLSVerify bool) Option {
+	return func(r *Registry) error {
+		if r.sourceCtx == nil {
+			r.sourceCtx = &types.SystemContext{}
```
grokspawn, line 118: We wouldn't use the `DefaultSystemContext` here?

pkg/image/containersimageregistry/registry.go
```diff
@@ -0,0 +1,290 @@
...
+	policyContext, err := signature.NewPolicyContext(policy)
+	if err != nil {
+		return err
+	}
+
+	sourceCtx := r.sourceCtx
+	authFile := getAuthFile(r.sourceCtx, namedRef.String())
+	if authFile != "" {
+		if sourceCtx == nil {
+			sourceCtx = &types.SystemContext{}
```
grokspawn, line 158: use default?

pkg/image/containersimageregistry/registry.go
```diff
@@ -0,0 +1,290 @@
...
+	cache     *cacheConfig
+}
+
+var DefaultSystemContext = &types.SystemContext{OSChoice: "linux"}
+
+func New(sourceCtx *types.SystemContext, opts ...Option) (orimage.Registry, error) {
+	reg := &Registry{
+		sourceCtx: sourceCtx,
+	}
+
```
grokspawn, line 40:
```suggestion
if sourceCtx == nil {
   reg.sourceCtx = &types.SystemContext{}
```
would avoid checks at 117 & 157.

pkg/image/containersimageregistry/registry.go
```diff
@@ -0,0 +1,290 @@
...
+)
+
+var _ orimage.Registry = (*Registry)(nil)
+
+type Registry struct {
+	sourceCtx *types.SystemContext
+	cache     *cacheConfig
+}
+
+var DefaultSystemContext = &types.SystemContext{OSChoice: "linux"}
```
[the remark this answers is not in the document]
joelanford: In our case, we are pulling catalog and bundle images, which are never meant to be OS or arch-specific. Also, with the exception of a some niche windows containers use cases, linux is the only OS that ever shows up in image indexes or manifest lists because linux is the only OS that actually supports containers.

This defaulting essentially guarantees that MacOS and windows users can natively interact with the current incarnation of OLM's OCI formats (which use standard images to carry the FBC and bundle filesystems) without having to always override the OS from darwin to linux.

pkg/image/containersimageregistry/registry.go
joelanford: I don't think so, no. Because `InsecureSkipTLSVerify` will be called by `New`, which accepts an explicit `*types.SystemContext` from the caller. So if the user explicitly passes `nil` as the `*types.SystemContext`, we should not override that to the `DefaultSystemContext`.

pkg/image/containersimageregistry/registry.go
joelanford: No, with pretty much the same reasoning as above. At this point, the caller has already explicitly provided us with a `nil` SystemContext and has opted _not_ to use the `DefaultSystemContext`.

pkg/image/containersimageregistry/registry.go
joelanford: So technically, it is valid to send a `nil` SystemContext all the way into the `container/image` library calls. So the reason I have things setup this way is to only actually make a `types.SystemContext` when we actually _need_ to set a field. Otherwise, we roll with `nil` all the way through.

But I don't think there is a semantic difference between `&types.SystemContext{}` and `nil`, so I can get on board with this suggestion!

pkg/image/containersimageregistry/registry.go
```diff
@@ -0,0 +1,290 @@
...
+	namedRef, err := reference.ParseNamed(ref.String())
+	if err != nil {
+		return err
+	}
+	dockerRef, err := docker.NewReference(namedRef)
+	if err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(r.cache.ociLayoutDir(), 0700); err != nil {
```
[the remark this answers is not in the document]
joelanford: No complaints from the linter. I think `0700` is standard for directories because the execute bit for directories is what gives a user permission to traverse the directory.

internal/testutil/image/registry.go
joelanford, line 1: This used to be in `pkg/lib/image`, but it is only use by tests. Given that its implicit intention is as a test utility, I think it is reasonable to move this into `internal/testutil` to make that intention explicit.

This is technically a breaking change, but one that I am comfortable with because a reasonable person would not expect this to actually be part of the public API surface of operator-registry.

pkg/image/containersimageregistry/registry.go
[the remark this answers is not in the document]
grokspawn: LOL.  I was going to point to https://github.com/operator-framework/operator-registry/issues/1591 and say that I was reviewing to evaluate similar goals ... but then I see you replied to that issue in a way I feel resolves the questions.

PR #1630, opened 2025-04-04
pkg/configmap: sort configmap data by key when loading bundle

joelanford:
Sort configmap data by key when loading bundle (which is derived from filenames in the bundle directories)

This functionality is used by OLMv0 to construct an `api.Bundle` object that is needed to create install plan steps, which are eventually put into the installplan.status.plan field. It is important to have deterministic ordering of slice fields in Kubernetes APIs to avoid constant re-reconciliation and/or false positive change detection.

Specifically in the case of InstallPlan steps, the order of the steps does have _some_ semantic requirements (CSV, then CRDs, then everything else). The problem is that if "everything else" is in a random order, the order of the resulting steps will still have non-determinism.

If we have deterministic order of steps, OLM may be able to resolve a class of bugs related to unnecessary creation of multiple install plans that would otherwise be identical.

**Reviewer Checklist**

### DIFF
--- a/pkg/configmap/configmap.go
+++ b/pkg/configmap/configmap.go
@@ -3,6 +3,8 @@ package configmap
 import (
 	"errors"
 	"fmt"
+	"maps"
+	"slices"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


dinhxuanvu:
While I don't see any harms on sorting items in configmap data using their names, ultimately I don't see the point of doing so. The order that actually matters is the order of steps in InstallPlan. While steps are being installed, at each step, OLM simply looks up the correlative item in the configmap. So changing the order in configmap has no impact on anything. If there is, you need to add a test case to show the correlation between InstallPlan steps and Configmap items.
Secondly, new InstallPlan is being generated when the resolution state changes AKA new version, new operator and etc. I haven't seen the case there are multiple identical InstallPlans. Each IP should be different. InstallPlans or Configmaps are not a part of resolution. I fail to see how they matter in that area.

PR #1657, opened 2025-04-28
🌱 Add go-version check to avoid introducing a version that we cannot support yet

camilamacedo86:
…

**Reviewer Checklist**

### DIFF
--- a/hack/tools/check-go-version.sh
+++ b/hack/tools/check-go-version.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+###########################################
+# Check Go version in go.mod files
+# and ensure it is not greater than the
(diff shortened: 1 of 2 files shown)


tmshort (approved):

hack/tools/check-go-version.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
```
grokspawn, line 2: I know this is just a copy, but can you add a comment in here that the original source of this is o-f/op-con and to keep them in sync if there is a change?
We have the same kind of standard [here](https://github.com/operator-framework/operator-registry/blob/master/.golangci.yaml#L1-L14).

PR #1664, opened 2025-05-05
containers/image: remove signatures when copying in to OCI layout

joelanford:
This does not affect signature validation, and we do not need to preserve signatures _after_ validation because we will never need to propagate those signatures to another image transport/destination.

When using the containers/image registry client implementation, remove signatures when copying in to OCI layout

The OCI layout destination does not support copying signatures. This causes a bug (render/migrate fail) when using a policy that requires signature validation.

This change does not affect signature validation, and we do not need to preserve signatures _after_ validation because we will never need to propagate those signatures to another image transport/destination.

**Reviewer Checklist**

### DIFF
--- a/pkg/image/containersimageregistry/registry.go
+++ b/pkg/image/containersimageregistry/registry.go
@@ -161,6 +161,13 @@ func (r *Registry) Pull(ctx context.Context, ref orimage.Reference) error {
 		SourceCtx:                             sourceCtx,
 		DestinationCtx:                        r.cache.getSystemContext(),
 		OptimizeDestinationImageAlreadyExists: true,
+
+		// We use the OCI layout as a temporary storage and
+		// pushing signatures for OCI images is not supported
+		// so we remove the source signatures when copying.
+		// Signature validation will still be performed
+		// accordingly to a provided policy context.
+		RemoveSignatures: true,
 	}); err != nil {
 		return err
 	}


pkg/image/containersimageregistry/registry.go
```diff
@@ -161,6 +161,13 @@ func (r *Registry) Pull(ctx context.Context, ref orimage.Reference) error {
 		SourceCtx:                             sourceCtx,
 		DestinationCtx:                        r.cache.getSystemContext(),
 		OptimizeDestinationImageAlreadyExists: true,
+
+		// We use the OCI layout as a temporary storage and
+		// pushing signatures for OCI images is not supported
+		// so we remove the source signatures when copying.
+		// Signature validation will still be performed
+		// accordingly to a provided policy context.
+		RemoveSignatures: true,
```
grokspawn, line 170: Maybe we also want 
```golang
PreserveDigests: true,
```
It doesn't impact this particular operation, but it also seems a good option which doesn't default on.  
I verified that this works with upstream (unsigned) and downstream (signed) example catalogs.

joelanford: GoDoc for that field is:
```
// Preserve digests, and fail if we cannot.
```

I'm hesitant about the "fail if we cannot" part. We have [similar code in operator-controller](https://github.com/operator-framework/operator-controller/blob/061da6a62527598c0727e032f50465af80e5b1c3/internal/shared/util/image/pull.go#L150-L156) that does not set this value, so I don't think it is necessary, and likely not for this bug fix specifically.

grokspawn: I agree that it's not necessary for resolving this bug.  But the bug arose because we didn't consider other fields that might've been pertinent when we introduced the new containers/image registry support, so it felt like we should at least discuss any other possible missed configuration. 

I think we can agree that mutating digests across our local copy would be an undesirable side effect.

grokspawn: On second thought, even skopeo has this option as a non-main-path flag, which suggests that there are conditions under which this is desirable, but not in general. 
If we wanted this, we should plumb through a controlling flag.  Which I don't think we need to do, at this time.

PR #1707, opened 2025-06-27
🌱 remove bingo upgrade target

grokspawn:
Removes the bingo-upgrade target

This well-intentioned target is incapable of discerning infra tooling from public API tooling (which is similarly pinned, but has very different support expectations). 

**Reviewer Checklist**


camilamacedo86:
NIT:
It does not change anything to end users so the right emoji is 🌱 
I updated that

PR #1744, opened 2025-08-15
initial working viz of ignored v0 edges

grokspawn:

This is a first pass in making OLMv0 upgrade graph mechanics visible in rendered graphs, adding to existing `olm.deprecations` indication rendering.

Bundle versions which are skipped in a graph are outlined in red and have all their contributory edges changed to red, dashed arrows to indicate that they are ignored when calculating upgrades, for e.g. 
Screenshot 2025-08-18 at 9 58 21 AM

in this graph, cert-manager-operator.v1.15.1 is skipped, so any graph updates by this version are similarly ignored.  This means that users cannot upgrade from v1.15.0 to v1.15.1. 

These features are controlled by a new `draw-v0-semantics` flag (defaults off).  
By default, render-graph displays simple upgrade graph mechanics, which aligns to OLMv1 functionality.

Answering user questions recently, it became apparent that OLMv0 interpretation of an upgrade graph includes some interpretation mechanics which aren't well-communicated.  These mechanics can result in some user stranding on operator versions because of ignored graph contributions. 

**Reviewer Checklist**

### DIFF
--- a/cmd/opm/alpha/render-graph/cmd.go
+++ b/cmd/opm/alpha/render-graph/cmd.go
@@ -18,7 +18,7 @@ func NewCmd() *cobra.Command {
 		render               action.Render
 		minEdge              string
 		specifiedPackageName string
-		drawV0Semantics   bool
+		drawV0Semantics      bool
 	)
 	cmd := &cobra.Command{
 		Use:   "render-graph [index-image | fbc-dir]",
--- a/cmd/opm/alpha/render-graph/cmd.go
+++ b/cmd/opm/alpha/render-graph/cmd.go
@@ -18,6 +18,7 @@ func NewCmd() *cobra.Command {
 		render               action.Render
 		minEdge              string
 		specifiedPackageName string
+		drawV0Semantics   bool
 	)
 	cmd := &cobra.Command{
 		Use:   "render-graph [index-image | fbc-dir]",
@@ -67,7 +68,10 @@ $ opm alpha render-graph quay.io/operatorhubio/catalog:latest | \
 				log.Fatal(err)
 			}
 
-			writer := declcfg.NewMermaidWriter(declcfg.WithMinEdgeName(minEdge), declcfg.WithSpecifiedPackageName(specifiedPackageName))
+			writer := declcfg.NewMermaidWriter(
+				declcfg.WithMinEdgeName(minEdge),
+				declcfg.WithSpecifiedPackageName(specifiedPackageName),
+				declcfg.WithV0Semantics(drawV0Semantics))
 			if err := writer.WriteChannels(*cfg, os.Stdout); err != nil {
 				log.Fatal(err)
 			}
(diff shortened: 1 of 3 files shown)


bentito:
~When I run this, even if I try to send 2 to /dev/null, I can't get the warnings out of my way to send the graph output to, for instance, mermaid:~

```bash
bin/opm alpha render-graph quay.io/operatorhubio/catalog:latest|head 2>/dev/null
```

edit: the above command is wrong, stderr is correctly used, but maybe we should give an example in comments of redirecting as needed:

```
bin/opm alpha render-graph quay.io/operatorhubio/catalog:latest 2>/dev/null | 
```

grokspawn:
@bentito I think trying to graph an entire catalog will not function properly (though I'd love to see the warnings readout).  You would want to limit to a single package, and possibly a lower version range via flags. 

I generated customized FBC here, based on a semver catalog template expansion.  (Original report [here](https://github.com/openshift/cert-manager-operator-release/blob/504b6dad6f92cb033a5d180ee025193013019f1d/catalogs/v4.19/catalog/openshift-cert-manager-operator/channel.yaml#L15-L16) )
I have all the artifacts along the way at https://gist.github.com/grokspawn/e822d025f6718428f08efc2bd02a0d9d

The only missing detail is that, since OPM handles FBC as a hierarchical filesystem, I had to drop the FBC into a directory to manipulate it.  

For e.g.:
```bash
mkdir /tmp/c4
./bin/opm alpha render-template semver reproducer-semver.yaml > /tmp/c4/catalog.json

./bin/opm alpha render-graph /tmp/c4
```

bentito:
lgtm

cmd/opm/alpha/render-graph/cmd.go
```diff
@@ -67,13 +68,17 @@ $ opm alpha render-graph quay.io/operatorhubio/catalog:latest | \
...
+				declcfg.WithSpecifiedPackageName(specifiedPackageName),
+				declcfg.WithV0Semantics(includeV0Semantics))
 			if err := writer.WriteChannels(*cfg, os.Stdout); err != nil {
 				log.Fatal(err)
 			}
 		},
 	}
 	cmd.Flags().StringVar(&minEdge, "minimum-edge", "", "the channel edge to be used as the lower bound of the set of edges composing the upgrade graph; default is to include all edges")
 	cmd.Flags().StringVarP(&specifiedPackageName, "package-name", "p", "", "a specific package name to filter output; default is to include all packages in reference")
+	cmd.Flags().BoolVar(&includeV0Semantics, "include-v0-semantics", false, "whether to indicate OLMv0 semantics in the output; default is to simply represent the upgrade graph")
```
joelanford, line 82: My nit is `s/include/use/` for the name of the flag/variable, etc.

"Include" implies multiple $things could be included, all contributing. Whereas "use" has more of a mutual exclusive feel, which I think is more of what we're going for? I also notice that the actual mermaid renderer uses "use" terminology, so this would have better overall alignment as well.

PR #1750, opened 2025-08-19
validate fail for stranded channel entries

grokspawn:
`opm validate` fails when an edge is stranded because the replaces chain from the head is broken by skipped edges.  

Due to OLMv0 graph mechanics, any skips edge will cause OLMv0 to ignore the bundle version when considering upgrades (since v0 discards graph contribution from skipped bundle versions). 
Since the purpose of a replaces edge is to enable upgrade mobility across a graph, allowing the bundle version to be ignored (due to the skips entry) is an error, and potentially results in stranding. 

For example, take input `olm.channel`:
```yaml
schema: olm.channel
name: stable-v1
package: test-operator
entries:
  - name: test-operator-v1.0.0 # stranded due to skip of test-operator-v1.1.2
  - name: test-operator-v1.1.0 # stranded due to skip of test-operator-v1.1.2
    replaces: test-operator-v1.0.0
  - name: test-operator-v1.1.2
    replaces: test-operator-v1.1.0
  - name: test-operator-v1.1.4
    replaces: test-operator-v1.1.2
    skips:
      - test-operator-v1.1.2
  - name: test-operator-v1.2.0
  - name: test-operator-v1.2.1
    replaces: test-operator-v1.1.4
    skips:
      - test-operator-v1.1.4
      - test-operator-v1.2.0
  - name: test-operator-v1.3.0
    replaces: test-operator-v1.2.1
    skips:
      - test-operator-v1.2.1
  - name: test-operator-v1.4.0
    replaces: test-operator-v1.3.0
    skips:
      - test-operator-v1.3.0

```

Using a [new version of opm](https://github.com/operator-framework/operator-registry/pull/1744) which can optionally display OLMv0 graph semantics, skipped objects are limned in red and ignored edges are red dashed arrows to help visualize the stranded edges. 

image

**Reviewer Checklist**

### DIFF
--- a/alpha/model/model.go
+++ b/alpha/model/model.go
@@ -288,6 +288,10 @@ func (c *Channel) validateReplacesChain() error {
 		if _, ok := chainFrom[cur.Name]; !ok {
 			chainFrom[cur.Name] = []string{cur.Name}
 		}
+		// if the replaces edge is known to be skipped, disregard it
+		if skippedBundles.Has(cur.Replaces) {
+			break
+		}
 		for k := range chainFrom {
 			chainFrom[k] = append(chainFrom[k], cur.Replaces)
 		}
--- a/alpha/model/model_test.go
+++ b/alpha/model/model_test.go
@@ -164,7 +164,15 @@ func TestValidReplacesChain(t *testing.T) {
 			}},
 			assertion: hasError(`channel contains one or more stranded bundles: anakin.v0.0.1`),
 		},
-	}
+		{
+			name: "Error/SkippedReplacesStranded",
+			ch: Channel{Bundles: map[string]*Bundle{
+				"anakin.v0.0.1": {Name: "anakin.v0.0.1"},
+				"anakin.v0.0.2": {Name: "anakin.v0.0.2", Replaces: "anakin.v0.0.1"},
+				"anakin.v0.0.3": {Name: "anakin.v0.0.3", Replaces: "anakin.v0.0.2", Skips: []string{"anakin.v0.0.2"}},
+			}},
+			assertion: hasError(`channel contains one or more stranded bundles: anakin.v0.0.1`),
+		}}
 	for _, s := range specs {
 		t.Run(s.name, func(t *testing.T) {
 			err := s.ch.validateReplacesChain()


joelanford:
This validation check seems to be very narrowly tailored to "can't both skip and replace the same thing in one entry", which is good!

However, I think it very slightly misses the point and the broader problem.
1. It is actually okay to both `skip` and `replace` a bundle that is already a leaf node in the graph.
2. When a node is `skip`-ed and causes other entries to no longer have a path to the channel head, _that_ is the real problem that we need to check for.

grokspawn:
> This validation check seems to be very narrowly tailored to "can't both skip and replace the same thing in one entry", which is good!
> 
> However, I think it very slightly misses the point and the broader problem.
> 
>     1. It is actually okay to both `skip` and `replace` a bundle that is already a leaf node in the graph.

This is totally fine in any OLMv1 context, but I'd argue that since it comes with migration side-effects for OLMv0 that it's never OK.  In general, we should not have these kind of surprises, and I think it's reasonable to enforce the most-restrictive case here (because it's easier to grow-permissive than -restrictive). 

> 
>     2. When a node is `skip`-ed and causes other entries to no longer have a path to the channel head, _that_ is the real problem that we need to check for.

That's a specific flavor of this more general issue.  But I'd argue that it is also resolved by preventing the more general issue.

pkg/cache/cache_test.go
```diff
@@ -402,7 +402,7 @@ var validFS = fstest.MapFS{
 	"name": "clusterwide-alpha",
 	"entries": [
 		{"name": "etcdoperator.v0.9.0"},
-		{"name": "etcdoperator.v0.9.2-clusterwide", "replaces": "etcdoperator.v0.9.0", "skips": ["etcdoperator.v0.6.1","etcdoperator.v0.9.0"], "skipRange": ">=0.9.0 <=0.9.1"},
```
joelanford, line 405: Is this change related to the model validation change somehow? It seems unrelated to me at first glance.

grokspawn: Checked the catalogs impact and it looks the same as before.  HOWEVER, we no longer need this test change, because somehow it's OK for 0.9.2-clusterwide to skip AND replace v0.9.0 ...?

bandrade:
I've tested this PR using a synthetic catalog with the following structure to explicitly trigger the `replaces` + `skips` conflict detection introduced by this PR.

---

### 📦 Catalog Structure

All files were placed under a directory named `minimal-test`, and used `.yaml` extension to ensure `opm validate` would parse them.

#### `test.package.yaml`
```yaml
schema: olm.package
name: test-operator
defaultChannel: stable-v1
```

stable-v1.channel.yaml
```yaml
schema: olm.channel
name: stable-v1
package: test-operator
entries:
  - name: test-operator-v1.0.0
  - name: test-operator-v1.1.0
  - name: test-operator-v1.1.2
  - name: test-operator-v1.1.4
    replaces: test-operator-v1.0.0
    skips:
      - test-operator-v1.0.0
      - test-operator-v1.1.0
      - test-operator-v1.1.2
  - name: test-operator-v1.2.0
  - name: test-operator-v1.2.1
    replaces: test-operator-v1.1.4
    skips:
      - test-operator-v1.1.4     # <- conflict with replaces
      - test-operator-v1.2.0
  - name: test-operator-v1.3.0
    replaces: test-operator-v1.2.1
    skips:
      - test-operator-v1.2.1     # <- conflict with replaces
  - name: test-operator-v1.4.0
    replaces: test-operator-v1.3.0
    skips:
      - test-operator-v1.3.0     # <- conflict with replaces
```

test-operator-<version>.bundle.yaml (for each version)

Each file includes the required olm.package property, e.g.:
```
schema: olm.bundle
name: test-operator-v1.2.1
package: test-operator
image: quay.io/example/test-operator:v1.2.1
properties:
  - type: olm.package
    value:
      packageName: test-operator
      version: 1.2.1
````

opm validate completed without any errors.

I was expecting results like this
```
invalid package "test-operator", channel "stable-v1": entry "test-operator-v1.2.1" has identical replaces and skips: "test-operator-v1.1.4"
invalid package "test-operator", channel "stable-v1": entry "test-operator-v1.3.0" has identical replaces and skips: "test-operator-v1.2.1"
invalid package "test-operator", channel "stable-v1": entry "test-operator-v1.4.0" has identical replaces and skips: "test-operator-v1.3.0"
```
Could you validate if is there something wrong on my test? Thanks

**Comment by grokspawn** [comment]
> I've tested this PR using a synthetic catalog with the following structure to explicitly trigger the `replaces` + `skips` conflict detection introduced by this PR.

> Could you validate if is there something wrong on my test? Thanks

Hey @bandrade I'll need to update the PR description, because the new commit changed the functionality to not merely refuse skipped-replaces, but to really consider if a skipped-replace strands bundles across the replaces chain.  

The original example essentially ignores ALL lower bundle versions, so the new check does not identify it as a failure.  
In order for it to be identified, there have to be non-skipped bundles earlier in the replaces chain which are stranded because intermediary links are ignored (belong to a skipped edge).  

For e.g., this modification to your channel.yaml results in a failure: 
```yaml
schema: olm.channel
name: stable-v1
package: test-operator
entries:
  - name: test-operator-v1.0.0  # stranded because of skip on v1.1.4
  - name: test-operator-v1.1.0  # stranded because of skip on v1.1.4
    replaces: test-operator-v1.0.0
  - name: test-operator-v1.1.2
    replaces: test-operator-v1.1.0
  - name: test-operator-v1.1.4
    replaces: test-operator-v1.1.2
    skips:
      - test-operator-v1.1.2
  - name: test-operator-v1.2.0
  - name: test-operator-v1.2.1
    replaces: test-operator-v1.1.4
    skips:
      - test-operator-v1.1.4
      - test-operator-v1.2.0
  - name: test-operator-v1.3.0
    replaces: test-operator-v1.2.1
    skips:
      - test-operator-v1.2.1 
  - name: test-operator-v1.4.0
    replaces: test-operator-v1.3.0
    skips:
      - test-operator-v1.3.0
```

results in the message
```tree
FATA[0000] invalid index:
└── invalid package "test-operator":
    └── invalid channel "stable-v1":
        └── channel contains one or more stranded bundles: test-operator-v1.0.0, test-operator-v1.1.0
```

grokspawn:
/hold cancel
Removing the hold after consensus with stakeholders of existing, known catalogs.

grokspawn:
There is a coverage gap in this PR, in that if there are no non-skipped edges below a skipped edge, then it cannot identify stranded edges.  
This is because it is edge availability (they are not skipped) in the replacement chain is the key to identifying this problem.  
This is an edge case which already exists if folks use `startingCsv` to pin an initial-installation version, and we can add new validation to catch it later, like https://github.com/operator-framework/operator-registry/pull/1762 which attempts to adjust the underlying heuristics to cover all stranding cases.

bandrade:
Thanks for the clarification and updated logic — I just reproduced the new stranded bundle detection using the modified `channel.yaml` you suggested.

I created a synthetic catalog with the following `replaces` chain:
```
 ./bin/opm validate /Users/bandrade/Downloads/stranded-replaces-chain
FATA[0000] invalid index:
└── invalid package "test-operator":
    └── invalid channel "stable-v1":
        └── channel contains one or more stranded bundles: test-operator-v1.0.0, test-operator-v1.1.0 

```

/lgtm

mantomas:
It looks like thanks to this change, the `k8s-operatorhub/community-operators` are unable to release anything new, as there is `grafana-operator` present and is failing the new validation (eg. [here](https://github.com/k8s-operatorhub/community-operators/actions/runs/17729485391/job/50378500309)). We will pin the previous opm version for now.

grokspawn:
Assessment of existing, known catalogs was tabulated here: https://docs.google.com/spreadsheets/d/1ngHlFDOflLkpzf7Fd3_AAet_wqx65vjV8PkpHPonO2w/ ("[old] Summary" tab). 

For community-operators, @mantomas it appears that the infinispan-operator has unhealthy graph edges which are identified with this change. 

operatorhubio catalog has failures with grafana-operator contributions.  I'd suggest touching base with the maintainers there to get them to make updates.

PR #1762, opened 2025-08-21
opm validate: make sure skipped bundles are not included in replaces chain

joelanford:

When verifying channel validity, halt replaces chain traversal when we encounter a bundle that is skipped by any other entry anywhere in the channel.

`opm validate` currently says that invalid graphs are actually valid. In certain circumstances when a bundle is both skipped and is also somewhere in the replaces chain, the current `opm validate` logic has a preference for the bundle being in the replaces chain despite being skipped. In reality, skipped bundles are _NOT_ in the replaces chain and their upgrade edges are NOT honored.

**Reviewer Checklist**


grokspawn:
This is an aspirational, holistic attempt to resolve the family of problems characterized by inadvertent invalid channels, as well as to buttress the lax off-cluster validations (to enable earlier-detected errors and avoid on-cluster surprise).  

Versus #1750 which seeks to narrowly-define a single validation failure case -- and more importantly, to navigate a publishing path through existing catalog content to ameliorate impacts.  

#1762 also runs the risks of introducing v0-isms into validation as default, by introducing the [semver-noncompliant build metadata processing](https://github.com/operator-framework/operator-registry/pull/1762/files#diff-d8b78496e2c2574d1cb809798e7a1d7eb5f315de4940964cf0ac4199fda6b08bR328-R347) v0 uses.

I didn't see an impact assessment of applying the new rules to existing catalog content, and I think we absolutely need to understand any impacts there to understand how viable this PR is, and/or the nature/scope of follow-up actions we need to take to enable it. 

At this point, I'd favor working to get #1750 to consensus and giving ourselves some more time to work through the complexities this is trying to tackle.

PR #1979, opened 2026-05-11
fix(declcfg): WriteFS now preserves Others and Deprecations

tmshort:
WriteFS was iterating over Packages, Channels, and Bundles, but never consulted cfg.Others or cfg.Deprecations when building the per-package DeclarativeConfig written to disk.  Any custom FBC blobs (e.g. olm.lifecycle metadata used by the Operator Update Planner) and olm.deprecations entries were silently dropped.

This matches the existing behaviour of writeToEncoder, which already handles all five schema types correctly.  Others with an empty Package field are now written to a root-level catalog file, consistent with how writeToEncoder emits them at the end of its output.

Fixes: OCPBUGS-85347

**Reviewer Checklist**

### DIFF
--- a/alpha/declcfg/write.go
+++ b/alpha/declcfg/write.go
@@ -467,52 +467,70 @@ type encoder interface {
 	Encode(interface{}) error
 }
 
-func writeToEncoder(cfg DeclarativeConfig, enc encoder) error {
-	pkgNames := sets.NewString()
+func configsByPackage(cfg DeclarativeConfig) (sets.Set[string], map[string]DeclarativeConfig, []Meta) {
+	pkgNames := sets.New[string]()
+	byCfg := map[string]DeclarativeConfig{}
+	var rootOthers []Meta
 
-	packagesByName := map[string][]Package{}
+	add := func(name string, fn func(DeclarativeConfig) DeclarativeConfig) {
+		if name == "" {
+			return
+		}
+		pkgNames.Insert(name)
+		byCfg[name] = fn(byCfg[name])
+	}
+
 	for _, p := range cfg.Packages {
-		pkgName := p.Name
-		pkgNames.Insert(pkgName)
-		packagesByName[pkgName] = append(packagesByName[pkgName], p)
+		add(p.Name, func(c DeclarativeConfig) DeclarativeConfig {
+			c.Packages = append(c.Packages, p)
+			return c
+		})
 	}
-	channelsByPackage := map[string][]Channel{}
 	for _, c := range cfg.Channels {
-		pkgName := c.Package
-		pkgNames.Insert(pkgName)
-		channelsByPackage[pkgName] = append(channelsByPackage[pkgName], c)
+		add(c.Package, func(dc DeclarativeConfig) DeclarativeConfig {
+			dc.Channels = append(dc.Channels, c)
+			return dc
+		})
 	}
-	bundlesByPackage := map[string][]Bundle{}
 	for _, b := range cfg.Bundles {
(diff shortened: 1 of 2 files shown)


grokspawn:
Can we converge implementation with [writeToEncoder](https://github.com/operator-framework/operator-registry/blob/b058f57535c4e174956d050cb359420c1e26bc68/alpha/declcfg/write.go#L470C2-L470C2)?
Existing WriteJSON & WriteYAML methods there use a unified output approach which includes all currently-known schemas, plus a single touchpoint to make future schemas available easily.

alpha/declcfg/write_test.go
```diff
@@ -621,3 +624,39 @@ class anakin-dark-anakin.v0.1.0 skipped
...
+	// Others with no package name must appear in a root-level catalog file.
+	rootData, err := os.ReadFile(filepath.Join(dir, "catalog.json"))
+	require.NoError(t, err)
+	rootStr := string(rootData)
+	require.Contains(t, rootStr, `"schema": "custom.1"`, "root catalog must contain package-less Others")
+	require.Contains(t, rootStr, `"schema": "custom.2"`, "root catalog must contain package-less Others")
+
+	// Round-trip: loading the written files must reproduce the Others and Deprecations.
+	// Normalize JSON whitespace in Blob fields before comparing because WriteJSON
+	// reserializes blobs through the JSON encoder, which compacts whitespace.
```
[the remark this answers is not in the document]
tmshort: Addressed.

alpha/declcfg/write.go
```diff
@@ -572,34 +592,35 @@ func writeToEncoder(cfg DeclarativeConfig, enc encoder) error {
...
 
-	for _, p := range cfg.Packages {
-		fcfg := DeclarativeConfig{
-			Packages: []Package{p},
-			Channels: channelsByPackage[p.Name],
-			Bundles:  bundlesByPackage[p.Name],
+	for _, pName := range pkgNames.List() {
+		if pName == "" {
+			continue
 		}
```
joelanford: In theory `pName == ""` should never happen because anything without a package name would be funneled over to `rootOthers`?

alpha/declcfg/write.go
```diff
@@ -467,52 +467,74 @@ type encoder interface {
 	Encode(interface{}) error
 }
 
-func writeToEncoder(cfg DeclarativeConfig, enc encoder) error {
-	pkgNames := sets.NewString()
+// configsByPackage groups a DeclarativeConfig into per-package sub-configs plus a
+// package-less remainder (Others with no Package field).  It is the single touchpoint
```
grokspawn: Please remodel/remove the comment here.  I'm not comfortable with a claude-based rewording of my review comment.

tmshort: I did have to clarify what you wanted, and I hope I was correct.

alpha/declcfg/write.go
```diff
@@ -467,52 +467,70 @@ type encoder interface {
 	Encode(interface{}) error
 }
 
-func writeToEncoder(cfg DeclarativeConfig, enc encoder) error {
-	pkgNames := sets.NewString()
+func configsByPackage(cfg DeclarativeConfig) (sets.Set[string], map[string]DeclarativeConfig, []Meta) {
```
grokspawn, line 470: nit: This triple-return really feels like a wanna-be struct.

alpha/declcfg/write.go
grokspawn: I think you absolutely were.  Appreciate it.

PR #1986, opened 2026-05-15
BREAKING CHANGE: drop all sqlite support

grokspawn:
A view of how opm would look if we drop all SQLite support (which has been deprecated for 4+ years).  
An earlier iteration of this PR also dropped the model intermetiate package, but findings of current bugs as well as the added complexity suggest we'd be better off doing a separate PR for it. 

**Reviewer Checklist**


test/e2e/opm_test.go
```diff
@@ -21,10 +21,7 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/image/containerdregistry"
 	"github.com/operator-framework/operator-registry/pkg/image/execregistry"
 	"github.com/operator-framework/operator-registry/pkg/lib/bundle"
-	"github.com/operator-framework/operator-registry/pkg/lib/indexer"
-	lregistry "github.com/operator-framework/operator-registry/pkg/lib/registry"
 	"github.com/operator-framework/operator-registry/pkg/registry"
-	"github.com/operator-framework/operator-registry/pkg/sqlite"
 )
```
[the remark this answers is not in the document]
grokspawn: stale

alpha/action/render.go
```diff
@@ -213,93 +171,6 @@ func (r Render) imageToDeclcfg(ctx context.Context, imageRef string) (*declcfg.D
...
-			if ris.Has(ri.Image) {
-				ris.Delete(ri.Image)
-			}
-		}
-		for ri := range ris {
-			cfg.Bundles[i].RelatedImages = append(cfg.Bundles[i].RelatedImages, declcfg.RelatedImage{Image: ri})
-		}
-	}
-	return nil
-}
```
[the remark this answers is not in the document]
grokspawn: stale

test/e2e/opm_test.go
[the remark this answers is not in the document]
grokspawn: stale

go.mod
```diff
@@ -136,18 +131,16 @@ require (
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.18.4 // indirect
 	github.com/klauspost/pgzip v1.2.6 // indirect
-	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.19 // indirect
+	github.com/mattn/go-sqlite3 v1.14.32 // indirect
 	github.com/miekg/pkcs11 v1.1.2 // indirect
```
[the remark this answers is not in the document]
grokspawn: stale

go.mod
```diff
@@ -136,18 +131,16 @@ require (
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.18.4 // indirect
 	github.com/klauspost/pgzip v1.2.6 // indirect
-	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.19 // indirect
+	github.com/mattn/go-sqlite3 v1.14.32 // indirect
```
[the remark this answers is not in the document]
grokspawn: podman still brings it in and we cannot control transitive dependencies

grokspawn:
Applied the api-diff override label as this would be part of a proposed opm v2.0 with breaking changes.

grokspawn:
I ran a functional and output-format-based gap analysis on the second commit (which eliminates packages/layers).  The results are [here](https://hackmd.io/@of-olm/H1EHKZjMMe).   Aside from a restriction where the old code assumed that bundles had a 1:1: mapping with channels, the FBC-native validation recognizes that a bundle may belong to an arbitrary set of channels.  

[EDIT: rescoped this PR to ONLY the elimination of SQLite.  Additional layer elimination is in a follow-up: https://github.com/operator-framework/operator-registry/pull/2025]

alpha/declcfg/api_conversions.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	csv := buildCSV(b, pkg, props)
+	csvData, err := json.Marshal(csv)
+	if err != nil {
+		return b.CsvJSON
+	}
+	csvJSON := string(csvData)
+	if len(b.Objects) == 0 {
+		b.Objects = []string{csvJSON}
```
perdasilva: Since we're receiving b by value here, will b actually be mutated on the caller's side? Should we pass a pointer?

alpha/declcfg/api_conversions.go
```diff
@@ -0,0 +1,398 @@
...
+			}
+		}
+	}
+
+	apiDeps, err := convertPropertiesToAPIDependencies(b.Properties)
+	if err != nil {
+		return nil, fmt.Errorf("convert properties to api dependencies: %v", err)
+	}
+
+	return &api.Bundle{
```
perdasilva: Do we also need to set `Deprecations` here?

alpha/declcfg/validate.go
```diff
@@ -0,0 +1,447 @@
...
+		cde := sets.Set[string]{}
+		seenEntries := sets.New[string]()
+		for _, entry := range c.Entries {
+			if seenEntries.Has(entry.Name) {
+				pkgResult(c.Package).subErrors = append(pkgResult(c.Package).subErrors, fmt.Errorf("channel %q: duplicate entry %q", c.Name, entry.Name))
+			}
+			seenEntries.Insert(entry.Name)
+			cde = cde.Insert(entry.Name)
+		}
+		channelDefinedEntries[c.Package] = cde
```
perdasilva: should we be keying by channel? Are we rewriting the previous channels entries?

alpha/declcfg/api_conversions.go
```diff
@@ -0,0 +1,398 @@
...
+	props, err := parseProperties(b.Properties)
+	if err != nil {
+		return nil, fmt.Errorf("parse properties: %v", err)
+	}
+
+	csvJSON := generateCSVJSON(b, pkg, props)
+
+	// Find which channel this bundle belongs to
+	channelName := ""
+	for _, ch := range channels {
```
perdasilva: what does it looks like if the package belongs to multiple channels?

alpha/declcfg/api_conversions.go
```diff
@@ -0,0 +1,398 @@
...
+		if ch.Package != b.Package {
+			continue
+		}
+		for _, entry := range ch.Entries {
+			if entry.Name == b.Name {
+				replaces = entry.Replaces
+				skips = entry.Skips
+				skipRange = entry.SkipRange
+				break
```
perdasilva: should we also break out from the outer loop? Could we be overwriting previous results?

pkg/cache/cache.go
```diff
@@ -397,26 +395,50 @@ func (c *cache) processPackage(ctx context.Context, reader io.Reader) (packageIn
...
+			}
+			for _, entry := range ch.Entries {
+				if entry.Name == b.Name {
+					relevantChannels = append(relevantChannels, ch)
+					break
 				}
 			}
 		}
+
+		apiBundle, err := declcfg.ConvertBundleToAPIBundle(b, *pkg, relevantChannels)
```
perdasilva: api.Bundle has only a single channel in its struct - could this cause problems if a bundle is present in multiple channels?

pkg/cache/pkgs.go
```diff
@@ -190,19 +182,19 @@ func (pkgs packageIndex) GetBundleThatProvides(ctx context.Context, c Cache, gro
...
 
 type cChannel struct {
 	Name        string
 	Head        string
 	Bundles     map[string]cBundle
-	Deprecation *model.Deprecation `json:"deprecation,omitempty"`
+	Deprecation *string `json:"deprecation,omitempty"`
 }
 
 type cBundle struct {
```
perdasilva, line 200: do we need `Deprecation`?

pkg/server/server_test.go
```diff
@@ -369,14 +261,6 @@ func testGetBundle(addr string, expected *api.Bundle) func(*testing.T) {
 }
 
 func TestGetBundleForChannel(t *testing.T) {
```
perdasilva, line 309: We're leaving the test method and just nuking its implementation

pkg/server/server_test.go
```diff
@@ -481,8 +342,6 @@ func testGetChannelEntriesThatReplace(addr string, expected []*api.ChannelEntry)
 }
 
 func TestGetBundleThatReplaces(t *testing.T) {
-	t.Run("Sqlite", testGetBundleThatReplaces(dbAddress, etcdoperatorV0_9_2("alpha", false, false, includeManifestsAll)))
```
perdasilva, line 484: Same, empty impl

pkg/server/server_test.go
```diff
@@ -497,8 +356,6 @@ func testGetBundleThatReplaces(addr string, expected *api.Bundle) func(*testing.
 }
 
 func TestGetBundleThatReplacesSynthetic(t *testing.T) {
-	t.Run("Sqlite", testGetBundleThatReplacesSynthetic(dbAddress, etcdoperatorV0_9_2("alpha", false, false, includeManifestsAll)))
```
perdasilva, line 500: Here too! There are a few more below this as well

alpha/action/list.go
```diff
@@ -48,64 +46,119 @@ func (r *ListPackagesResult) WriteColumns(w io.Writer) error {
...
 	csv := v1alpha1.ClusterServiceVersion{}
-	if err := json.Unmarshal([]byte(head.CsvJSON), &csv); err != nil {
+	if err := json.Unmarshal([]byte(headBundle.CsvJSON), &csv); err != nil {
 		return ""
 	}
 	return csv.Spec.DisplayName
 }
 
+// findChannelHead finds the head bundle of a channel by analyzing the replaces chain.
+func findChannelHead(entries []declcfg.ChannelEntry) (string, error) {
```
perdasilva: duplicated logic also found in pkg/cache/pkgs.go:288-320 and possibly also alpha/declcfg/validate.go:369-398

AGENTS.md
perdasilva: Was this function deleted?

AGENTS.md
```diff
@@ -117,19 +114,7 @@ model, err := declcfg.ConvertToModel(cfg)
 err = declcfg.Write(cfg, "output.yaml")
```
perdasilva: Moved to WriteJSON/WriteYAML/WriteFS possibly?

AGENTS.md
perdasilva: Call needs ctx param

AGENTS.md
```diff
@@ -33,7 +30,7 @@ The Operator Registry is a Kubernetes/OpenShift component that provides operator
 - **Go 1.24.4**: Minimum Go version required
```
perdasilva: should we bump this to the current 1.26.3?

AGENTS.md
```diff
@@ -9,18 +9,15 @@ The Operator Registry is a Kubernetes/OpenShift component that provides operator
...
 ### Binaries
-- **`opm`**: Main CLI tool for generating and updating registry databases and index images
-- **`initializer`**: **(Deprecated)** Converts operator manifests to SQLite database format
-- **`registry-server`**: **(Deprecated)** Exposes gRPC interface to SQLite databases
-- **`configmap-server`**: Parses ConfigMaps into SQLite databases and exposes gRPC interface
+- **`opm`**: Main CLI tool for rendering and serving file-based catalogs
 
 ### Libraries
 - **`pkg/client`**: High-level client interface for gRPC API
 - **`pkg/api`**: Low-level client libraries for gRPC interface
```
perdasilva, line 16: only the proto stubs remain - should we reword?

AGENTS.md
```diff
@@ -9,18 +9,15 @@ The Operator Registry is a Kubernetes/OpenShift component that provides operator
...
 - **`pkg/registry`**: Core registry types (Packages, Channels, Bundles)
```
perdasilva, line 17: only `Bundle` remains

test/e2e/opm_test.go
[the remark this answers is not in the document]
grokspawn: Not there anymore.  Either stale or hallucinated.

pkg/registry/populator.go
```diff
@@ -28,6 +28,9 @@ type DirectoryPopulator struct {
 	overwrittenImages map[string][]string
 }
 
+// TODO(follow-up): Unit tests for DirectoryPopulator, SemverPackageManifest, and ValidateEdgeBundlePackage
+// were removed along with SQLite support (they depended on SQLite-backed Load/GraphLoader/Query implementations).
+// These functions are still used in production and need test coverage with non-SQLite implementations.
```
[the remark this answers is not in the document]
grokspawn: Tests for these were entirely reliant on SQLite-based functions, which haven't been rewired here.  This is captured in a TODO and will either be resolved here with later effort (if the codecov threshold for CI is not met) or in a follow-up PR.

cmd/opm/render/cmd.go
```diff
@@ -26,11 +25,10 @@ func NewCmd(showAlphaHelp bool) *cobra.Command {
...
 	)
 	cmd := &cobra.Command{
-		Use:   "render [catalog-image | catalog-directory | bundle-image | bundle-directory | sqlite-file]...",
+		Use:   "render [catalog-image | catalog-directory | bundle-image | bundle-directory]...",
 		Short: "Generate a stream of file-based catalog objects from catalogs and bundles",
 		Long: `Generate a stream of file-based catalog objects to stdout from the provided
-catalog images, file-based catalog directories, bundle images, and sqlite
-database files.
+catalog images, file-based catalog directories, and bundle images.
 `,
```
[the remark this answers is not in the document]
grokspawn: Fixed — added "bundle directories" to the Long description to match the Use string.

README.md
```diff
@@ -73,26 +67,36 @@ Of course, this build step can be done with any other OCI spec container tools l
...
+
+To make your bundle available, you build a file-based catalog (FBC) that references your bundle images. You can render bundle images into FBC format and then serve the catalog via a container image.
+
+First, render your bundle into a file-based catalog:
 
 ```sh
-opm index add --bundles quay.io/my-container-registry-namespace/my-manifest-bundle:0.0.1 --tag quay.io/my-container-registry-namespace/my-index:1.0.0
-podman push quay.io/my-container-registry-namespace/my-index:1.0.0
+mkdir my-catalog
+opm render quay.io/my-container-registry-namespace/my-manifest-bundle:0.0.1 -o yaml >> my-catalog/catalog.yaml
```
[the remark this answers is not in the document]
grokspawn: Fixed — the README section has been rewritten to use FBC workflows (opm render/validate/generate dockerfile). The render example now uses > not >>.

pkg/server/server_test.go
grokspawn: This is what happens when you let LLMs take the first pass.  Resolving in a follow up PR where we do the model layer collapse which is removed from this PR.

pkg/server/server_test.go
grokspawn: This is what happens when you let LLMs take the first pass.  Resolving in a follow up PR where we do the model layer collapse which is removed from this PR.

pkg/server/server_test.go
grokspawn: This is what happens when you let LLMs take the first pass. Resolving in a follow up PR where we do the model layer collapse which is removed from this PR.

pkg/cache/pkgs.go
grokspawn: Either unnecessary or missing in the initial feature info: [original branch](https://github.com/operator-framework/operator-registry/blob/e79d272673a2be29b8270ce2f3323554acde20be/pkg/cache/pkgs.go#L208) did not have it.

alpha/action/list.go
grokspawn: Resolved.

AGENTS.md
grokspawn: It was, but now this PR is only the SQLite changes.  This is resolved in https://github.com/operator-framework/operator-registry/pull/2025

AGENTS.md
grokspawn: Resolved in https://github.com/operator-framework/operator-registry/pull/2025

AGENTS.md
grokspawn: Resolved in https://github.com/operator-framework/operator-registry/pull/2025

AGENTS.md
grokspawn: Resolved in https://github.com/operator-framework/operator-registry/pull/2025

AGENTS.md
grokspawn: Yes, these shouldn't have been removed.  Resolved in https://github.com/operator-framework/operator-registry/pull/2025

AGENTS.md
grokspawn: Resolved in https://github.com/operator-framework/operator-registry/pull/2025

alpha/declcfg/api_conversions.go
grokspawn: Yes, this was an error.  Restored it in https://github.com/operator-framework/operator-registry/pull/2025

alpha/declcfg/api_conversions.go
grokspawn: The cache for FBC would mimic the SQLite bundle "has-a" channel by caching an instance of the bundle for each channel in which the bundle appeared.  We need to do the same in https://github.com/operator-framework/operator-registry/pull/2025

pkg/cache/cache.go
grokspawn: Existing code would slam the bundle for each channel, losing all but the last channel in which it appeared.  Instead, it needs to fulfill the GRPC API needs by (as original code was doing) generating a bundle in the cache for each relevant channel. 
Resolved in https://github.com/operator-framework/operator-registry/pull/2025

alpha/declcfg/api_conversions.go
grokspawn: Resolved in https://github.com/operator-framework/operator-registry/pull/2025

grokspawn:
codecov/project results are misleading, because there are at least a couple of areas where we went 75% --> 100%, but the delta is captured as 0%.

pkg/server/server_test.go
```diff
@@ -24,59 +24,19 @@ import (
...
 )
 
 const (
-	dbPort    = ":50052"
-	dbAddress = "localhost" + dbPort
-	dbName    = "test.db"
-
-	cachePort    = ":50053"
-	cacheAddress = "localhost" + cachePort
-
```
joelanford, line 42: The origin of this test setup was just expectations from the sqlite-based GRPC responses.

When we added FBC, we updated this test to migrate/render the DB to FBC and then do the exact same request/response/assertion for both DB and FBC to make sure the GRPC response was the same for the equivalent DB and FBC.

Now that sqlite is not a thing, we obviously can't translate sqlite to FBC for the cachePort server, so I assume that is why it is removed here.

However, I'd suggest that we keep the cachePort tests by committing the outputDir of the migration in lines 130-138.  That way we don't lose any of the original base test cases.

joelanford:
We may be able to drastically simplify our github release workflows now that the CGO dependency is no more!

Totally do-able in a follow-up, but don't want that to be overlooked!

go.mod
```diff
@@ -134,19 +129,17 @@ require (
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/klauspost/compress v1.18.6 // indirect
 	github.com/klauspost/pgzip v1.2.6 // indirect
-	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.23 // indirect
+	github.com/mattn/go-sqlite3 v1.14.44 // indirect
```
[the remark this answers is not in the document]
grokspawn: no, copilot.  This would just be litter in the go.mod file.

pkg/server/server_test.go
grokspawn: Done.

grokspawn:
> We may be able to drastically simplify our github release workflows now that the CGO dependency is no more!
> 
> Totally do-able in a follow-up, but don't want that to be overlooked!

Agreed.  There are a few things we can iterate on after this lands.

rashmigottipati (commented):
changes look good to me overall, will wait for others to review as well

README.md
```diff
@@ -73,26 +67,36 @@ Of course, this build step can be done with any other OCI spec container tools l
...
+podman build -t quay.io/my-container-registry-namespace/my-catalog:1.0.0 -f my-catalog.Dockerfile .
+podman push quay.io/my-container-registry-namespace/my-catalog:1.0.0
 ```
 
-For more detail on using `opm` to generate index images, take a look at the [documentation](docs/design/opm-tooling.md).
+The resulting image is a catalog image containing file-based catalog content that is served via a gRPC API when the container image is run.
+
+For more detail on using `opm`, see the [file-based catalog documentation](https://olm.operatorframework.io/docs/tasks/creating-a-catalog/).
 
 # Using the index with Operator Lifecycle Manager
```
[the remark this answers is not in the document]
grokspawn: Defer.  Disputable.  
Even your proposal was not comprehensive:  There is another use immediately after.

Makefile
```diff
@@ -62,8 +62,8 @@ build: clean $(CMDS) $(OPM)
 cross: opm_version_flags=-ldflags "-X '$(PKG)/cmd/opm/version.gitCommit=$(GIT_COMMIT)' -X '$(PKG)/cmd/opm/version.opmVersion=$(OPM_VERSION)' -X '$(PKG)/cmd/opm/version.buildDate=$(BUILD_DATE)'"
 cross:
 ifeq ($(shell go env GOARCH),amd64)
-	GOOS=darwin CC=o64-clang CXX=o64-clang++ CGO_ENABLED=1 CGO_LDFLAGS='-Wl,-undefined,dynamic_lookup' $(GO) build $(opm_version_flags) -tags=$(TAGS) -o "bin/darwin-amd64-opm" --ldflags "-extld=o64-clang -extldflags=-Wl,-undefined,dynamic_lookup" ./cmd/opm
-	GOOS=windows CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ CGO_ENABLED=1 $(GO) build $(opm_version_flags) -tags=$(TAGS)  -o "bin/windows-amd64-opm" --ldflags "-extld=x86_64-w64-mingw32-gcc" -buildmode=exe ./cmd/opm
+	GOOS=darwin CGO_ENABLED=0 $(GO) build $(opm_version_flags) -tags=$(TAGS) -o "bin/darwin-amd64-opm" ./cmd/opm
+	GOOS=windows CGO_ENABLED=0 $(GO) build $(opm_version_flags) -tags=$(TAGS) -o "bin/windows-amd64-opm" -buildmode=exe ./cmd/opm
```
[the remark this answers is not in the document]
grokspawn: This PR is one of several we've conceptualized, and that work can be done in a follow-on effort.

cmd/opm/root/cmd.go
```diff
@@ -44,8 +42,7 @@ To view help related to alpha features, set HELP_ALPHA=true in the environment.`
 		logrus.Panic(err.Error())
 	}
 
-	cmd.AddCommand(registry.NewOpmRegistryCmd(showAlphaHelp), alpha.NewCmd(showAlphaHelp), initcmd.NewCmd(), migrate.NewCmd(), serve.NewCmd(), render.NewCmd(showAlphaHelp), validate.NewCmd(), generate.NewCmd())
-	index.AddCommand(cmd, showAlphaHelp)
+	cmd.AddCommand(alpha.NewCmd(showAlphaHelp), initcmd.NewCmd(), migrate.NewCmd(), serve.NewCmd(), render.NewCmd(showAlphaHelp), validate.NewCmd(), generate.NewCmd())
```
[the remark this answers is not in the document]
grokspawn: Already captured in a TODO for a follow-up.

alpha/action/render.go
```diff
@@ -125,21 +110,8 @@ func (r Render) renderReference(ctx context.Context, ref string) (*declcfg.Decla
...
-	}
-
-	db, err := sqlite.Open(ref)
-	if err != nil {
-		return nil, err
-	}
-	defer db.Close()
-	return sqliteToDeclcfg(ctx, db)
+	// Only directories are supported for file-based catalogs and bundles.
+	return nil, fmt.Errorf("ref %q is not a directory", ref)
```
perdasilva: **Bug: file-path ref returns an error that doesn't wrap `ErrNotAllowed`**

When a caller passes a plain file path (e.g. a `.db` or `.json` file that exists on disk), `renderReference` hits this branch and returns a raw error. `list.go`'s `indexRefToModel` checks `errors.Is(err, ErrNotAllowed)` to produce a friendly message; because this error doesn't wrap `ErrNotAllowed`, that check is false and the user sees the internal string instead:

> `ref "./catalog.json" is not a directory`

…instead of the intended:

> `cannot list non-index "./catalog.json"`

**Fix:** wrap `ErrNotAllowed` just like the directory branches above do:
```suggestion
	return nil, fmt.Errorf("ref %q is not a directory: %w", ref, ErrNotAllowed)
```

alpha/action/render.go
```diff
@@ -160,145 +132,61 @@ func (r Render) imageToDeclcfg(ctx context.Context, imageRef string) (*declcfg.D
...
 	}
 	return cfg, nil
 }
 
-// checkDBFile returns an error if ref is not an sqlite3 database.
-func checkDBFile(ref string) error {
-	typ, err := filetype.MatchFile(ref)
-	if err != nil {
-		return err
+func (r Render) renderImageConfig(ctx context.Context, ref image.SimpleReference, tmpDir string, labels map[string]string) (*declcfg.DeclarativeConfig, error) {
```
perdasilva, line 142: **Bug: sqlite index images get a cryptic generic error with no actionable guidance**

`renderImageConfig` checks `ConfigsLocationLabel` then `PackageLabel` but has no branch for `containertools.DbLocationLabel` (`operators.operatorframework.io.index.database.v1`). A user who passes an existing sqlite-backed index image will have it fully pulled and unpacked before landing in `imageTypeError`, which prints:

> `render "...": image type could not be determined, found labels`
> `  operators.operatorframework.io.index.database.v1=/database/index.db`

There is no indication that sqlite support was removed or that they should use `opm migrate` first.

**Fix:** add a fast-fail branch before the fallthrough:
```suggestion
func (r Render) renderImageConfig(ctx context.Context, ref image.SimpleReference, tmpDir string, labels map[string]string) (*declcfg.DeclarativeConfig, error) {
	if _, ok := labels[containertools.DbLocationLabel]; ok {
		return nil, fmt.Errorf("render %q: sqlite-based index images are no longer supported; migrate the catalog to file-based format first using 'opm migrate'", ref)
	}
	if configsDir, ok := labels[containertools.ConfigsLocationLabel]; ok {
		return r.renderDeclcfgImage(ctx, tmpDir, configsDir)
	}
	if _, ok := labels[bundle.PackageLabel]; ok {
		return r.renderBundleImage(ref, tmpDir)
	}
	return nil, r.imageTypeError(ref.String(), labels)
}
```

pkg/server/catalog_build_test.go
```diff
@@ -0,0 +1,394 @@
...
+
+		entry := declcfg.ChannelEntry{Name: name}
+		if csv.Spec.Replaces != "" {
+			entry.Replaces = csv.Spec.Replaces
+			if !visited[csv.Spec.Replaces] {
+				queue = append(queue, csv.Spec.Replaces)
+			}
+		}
+		if len(csv.Spec.Skips) > 0 {
+			entry.Skips = csv.Spec.Skips
```
perdasilva, line 345: **Bug: skip-only bundles are never enqueued — the dead `!ok` branch and its comment are misleading**

The function doc says _"Bundles listed in skips that do not exist in csvFields are included as skip-only entries (no replaces, not traversed further)"_ and the inline comment at the `!ok` branch says _"Bundle referenced via skips but not present in the manifests"_. Neither is true: skip targets are copied into `entry.Skips` here but **never appended to the queue**. The `!ok` branch is only reachable for bundles that entered the queue via a `replaces` chain.

Concrete impact: a channel where a bundle appears solely in another bundle's `skips` list (and not in any `replaces` chain) will produce no `ChannelEntry` for that bundle, diverging from what a real FBC loader would emit.

**Fix:** append skip targets to the queue the same way replaces targets are:
```suggestion
		if len(csv.Spec.Skips) > 0 {
			entry.Skips = csv.Spec.Skips
			for _, s := range csv.Spec.Skips {
				if !visited[s] {
					queue = append(queue, s)
				}
			}
		}
```

pkg/server/catalog_build_test.go
```diff
@@ -0,0 +1,394 @@
...
+		}
+		if pm.PackageName != "" {
+			return pm, nil
+		}
+	}
+	return nil, fmt.Errorf("no package manifest found in %s", pkgDir)
+}
+
+func isYAML(name string) bool {
+	return len(name) > 5 && (name[len(name)-5:] == ".yaml" || name[len(name)-4:] == ".yml")
```
perdasilva: **Bug: `isYAML` off-by-one silently drops valid `.yml` files**

The single guard `len(name) > 5` is correct for `.yaml` (min length 6 for `"a.yaml"`) but too strict for `.yml`, which only needs length ≥ 5. `isYAML("a.yml")` evaluates `5 > 5 = false` and returns false, silently skipping any `.yml` file whose base name is one character long.

The simplest correct fix uses `filepath.Ext`:
```suggestion
func isYAML(name string) bool {
	ext := filepath.Ext(name)
	return ext == ".yaml" || ext == ".yml"
}
```

pkg/server/catalog_build_test.go
```diff
@@ -0,0 +1,394 @@
...
+			if err := json.Unmarshal(jsonData, &obj); err != nil || obj.Kind == "" {
+				continue
+			}
+
+			info := &manifestInfo{jsonData: jsonData, kind: obj.Kind}
+
+			switch obj.Kind {
+			case "ClusterServiceVersion":
+				var csv rawCSVFields
+				if err := yaml.Unmarshal(raw, &csv); err != nil {
```
perdasilva: **Efficiency: CSV bytes decoded from YAML to JSON twice**

`jsonData` (already JSON-converted YAML) is computed on the line above, but `yaml.Unmarshal(raw, &csv)` re-parses the original `raw` bytes — `sigs.k8s.io/yaml.Unmarshal` internally calls `yamlToJSONTarget` before invoking `json.Unmarshal`. Every CSV file therefore pays two full YAML-to-JSON passes.

Since `jsonData` is already available, replace this line with:
```suggestion
			if err := json.Unmarshal(jsonData, &csv); err != nil {
```
(also add `"encoding/json"` to the import block if not already present — it already is in this file).

pkg/server/catalog_build_test.go
```diff
@@ -0,0 +1,394 @@
...
+				}
+				return string(props[i].Value) < string(props[j].Value)
+			})
+
+			image := ""
+			if name == "etcdoperator.v0.9.2" {
+				image = "fake/etcd-operator:v0.9.2"
+			}
+
+			bundleMap[name] = declcfg.Bundle{
```
perdasilva, line 252: **Bug (latent): same CSV name in both flat root and versioned subdir silently overwrites the bundle**

`buildPackageFBC` collects from the package root dir (for flat packages) and all subdirectories (for versioned packages) in a single pass, then merges into the same `bundleMap` keyed by CSV name. If the same `csv.Metadata.Name` appears in both the root and a subdir — a layout the function header explicitly says it intends to support — the second write silently wins with a different CRD set, producing a bundle whose properties are internally inconsistent.

Current test data doesn't trigger this (versioned packages have only a `*.package.yaml` at the root), but a guard would make the intent explicit and catch future test-data additions:
```go
if _, exists := bundleMap[name]; exists {
    return nil, fmt.Errorf("duplicate bundle name %q found in multiple directories", name)
}
```

pkg/server/catalog_build_test.go
```diff
@@ -0,0 +1,394 @@
...
+	entries, err := os.ReadDir(pkgDir)
+	if err != nil {
+		return nil, err
+	}
+	for _, e := range entries {
+		if e.IsDir() || !isYAML(e.Name()) {
+			continue
+		}
+		path := filepath.Join(pkgDir, e.Name())
+		f, err := os.Open(path)
```
perdasilva, line 371: **Nit: `os.Open` failure is fatal while `DecodePackageManifest` failure is skipped — inconsistent**

Two lines below, a `DecodePackageManifest` error causes the loop to `continue` (skip to the next file) because not every YAML file is a package manifest. But an `os.Open` error immediately propagates and aborts the whole search.

If a YAML file is enumerable by `os.ReadDir` but not openable (e.g. permissions tightened between the two calls), the caller receives an OS error rather than the package manifest it would have found in the next iteration.

Consider making `os.Open` failures skip the same way:
```suggestion
		f, err := os.Open(path)
		if err != nil {
			continue
		}
```

perdasilva: Idk that I agree necessarily with Claude here. Seems like we'd want to bail if there are issues opening the FBC/yaml.

grokspawn:
@perdasilva I resolved all your comments above.  It would be good if your LLM reviewer agent could classify the bugs by severity, and would also be a bit more relaxed for test-only versus product classification.  For e.g., the catalog_built_test findings are really relevant for devs only, adding new testdata or scenarios.

grokspawn:
IMHO, https://github.com/operator-framework/operator-registry/pull/2034 should be resolved (merged or responded-to with rationale to delay until re-versioning) before this merges.
/hold until some terminal state is reached on #2034

grokspawn:
/hold cancel
#2034 cannot be merged until we have access to go builder images supporting the required version.

pedjak (approved):
Thorough removal of 4+ years of deprecated SQLite support — solid cleanup.

A few things I noticed beyond what Copilot and @perdasilva already covered:

**Good catches in this PR:**
- The `FBCCacheNoResults` test case fix (`cacheAddress` → `customSchemaCacheAddress`) is a legitimate pre-existing bug fix: the old test was querying for `custom.operator.io` schema against a server without custom schemas, so the zero-result assertion was trivially true.
- The `render.go` refactoring into `renderImageConfig`/`renderDeclcfgImage`/`renderBundleImage`/`imageTypeError` is a big readability win over the nested if/else.
- Nice that `CGO_ENABLED=0` for cross-compilation falls out naturally from this.

**Noted for follow-up (not blocking):**
- `Makefile` `static` target still has `-tags "json1"` (Copilot raised this, agree it's fine as follow-up).
- The `Channel.Priority` field removal from `alpha/property/property.go` is a separate API change from the SQLite removal — see inline comment.

alpha/property/property.go
pedjak, line 1: **suggestion (non-blocking):** The removal of the `Priority` field from `Channel` and the `MustBuildChannelPriority` function is a separate API change from dropping SQLite. While the struct is marked "internal use only" and `MustBuildChannelPriority` has no external callers, removing the `json:"priority"` field means any existing serialized `olm.channel` properties containing `"priority"` will silently drop that data on deserialization.

This is probably fine given the "internal use only" contract, but worth calling out since the PR description only mentions SQLite removal. If this is intentional, no action needed — just noting it for the record.

grokspawn: It looks like when I split this PR between the "no sqlite" and "deprecate old layers" this change made it over.  I don't think it really matters since the immediate follow-up PR would remove it anyway, and it is an internal-only API that is also deprecated.  
Still .... darnit, I thought I had completely divided the conjoined twins here.

rashmigottipati (approved):
/lgtm

PR #2038, opened 2026-07-23
Reject cyclic `substitutesFor` chains instead of looping until OOM

haklein:
A bundle whose `olm.substitutesFor` annotation references its own CSV name (or a set of bundles forming an indirect cycle) causes an endless loop when walking the chain in `addSubstitutesFor`. This finally leads to a OOM condition: `fatal error: runtime: out of memory`. ([OCPBUGS-37284](https://issues.redhat.com/browse/OCPBUGS-37284)).

Guard every `substitutesFor` chain walk with a visited set seeded with the walk's starting node, so any cycle - direct or indirect - fails fast with an error naming the offending bundle instead of exhausting memory.

  Add a visited-set guard (`markSeen`) to all five substitutesFor chain
  walks in `pkg/sqlite/addSubstitutesFor`. A cyclic chain now fails with
  `cyclic substitutesFor chain detected involving "<name>"` instead of
  looping forever.

  Regression tests: self-reference (`olm.substitutesFor` == own CSV name)
  and indirect cycle (A→B→A). Both hang on master, "green" with the fix.

  OCPBUGS-37284 — a bundle whose `olm.substitutesFor` references its own
  name sends the walk at `load.go:293` into an infinite loop, growing
  `skips` each iteration until `fatal error: runtime: out of memory`
  (broke IIB builds for web-terminal-operator 1.7.x).

  Same guard pattern as the existing `replaces` cycle checks:
  `pkg/sqlite/load.go` (`replaceCycle`) and
  `alpha/model/model.go` (`validateReplacesChain`).

  **Reviewer Checklist**

### DIFF
--- a/pkg/sqlite/load.go
+++ b/pkg/sqlite/load.go
@@ -162,6 +162,18 @@ func (s *sqlLoader) addOperatorBundle(tx *sql.Tx, bundle *registry.Bundle) error
 	return s.addAPIs(tx, bundle)
 }
 
+// markSeen records name in seen, erroring if it was already there. It guards
+// the substitutesFor chain walks below against cyclic references (e.g. a
+// bundle substituting for itself), which would otherwise loop until the
+// process runs out of memory (OCPBUGS-37284).
+func markSeen(seen map[string]struct{}, name string) error {
+	if _, ok := seen[name]; ok {
+		return fmt.Errorf("cyclic substitutesFor chain detected involving %q", name)
+	}
+	seen[name] = struct{}{}
+	return nil
+}
+
 func (s *sqlLoader) addSubstitutesFor(tx *sql.Tx, bundle *registry.Bundle) error {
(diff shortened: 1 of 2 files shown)


grokspawn:
Hey @haklein thank you for your PR.  
While this PR tackles a legitimate issue, it is for a feature which was deprecated 4+ years ago (SQLite catalogs) and which will shortly [be deleted](https://github.com/operator-framework/operator-registry/pull/1986).

IMO we should merge this PR but we should not allow further PRs against deprecated SQLite code.

haklein:
> IMO we should merge this PR but we should not allow further PRs against deprecated SQLite code.

fair enough, thanks for your feedback. Will keep that in mind when picking further backlog candidates.

grokspawn:
> fair enough, thanks for your feedback. Will keep that in mind when picking further backlog candidates.

No fault of yours.  I just want to make sure we're careful about communicating that this is going away soon, so there will be extremely limited opportunity for anyone to make use of this fix.

PR #2078, opened 2026-09-03
trigger release workflow either by tag push or GUI

grokspawn:
Allows the release and all payload to be generated either by
1. (existing) pushing a new tag to the main branch
2. (new) generating a new release through the GH GUI

Other o-f repos do things differently, and some are flexible enough to do it correctly either way.  op-reg has always required a pushed tag, and not handled the case well where the release was first created in the GUI instead.

**Reviewer Checklist**

## Summary by CodeRabbit

- **Release Management**
  - Release automation now creates releases for new tags and updates existing releases without creating duplicates.
  - Release notes are generated or refreshed automatically during release processing.
  - Release assets are uploaded reliably to the associated release, including when a release already exists.
  - Tag-based releases continue to support automatic note generation and asset publishing.


joelanford:
It is my understanding that creating a release via the UI ends up creating a tag, which then triggers our existing release workflows via that tag push.

And the one thing we've needed to do in the past to make sure that works is allow the GH goreleaser workflow to overwrite/replace the GH release that already exists (since it was pre-created in the UI).

grokspawn:
Updated flow to overwrite existing release, with same flow for tag-push releases, and the same results w.r.t. release notes and artifacts.

.github/workflows/goreleaser.yaml
```diff
@@ -44,7 +46,30 @@ jobs:
...
+              --title "$(jq -r '.name' "${notes_file}")" \
+              --notes "$(jq -r '.body' "${notes_file}")"
+            gh release upload "${RELEASE_TAG}" --clobber ./assets/*
+          else
+            gh release create "${RELEASE_TAG}" \
+              --verify-tag \
+              --generate-notes \
+              ./assets/* || {
+                # A GUI-created release may race this tag-triggered run.
+                gh release upload "${RELEASE_TAG}" --clobber ./assets/*
```
joelanford: If the create fails, it seems like we'd want to basically do the "already exists" path of generating notes and then editing.

WDYT about unconditionally calling `create`, and if that fails, the the path of generating notes, editing the release, and uploading the assets.

PR #2088, opened 2026-09-15
Add CSV metadata minification migration

joelanford:
## Summary

Minify CRD and APIService descriptions in `olm.csv.metadata` by retaining only the fields consumed by OLM's package server:

- `name`
- `group`, where applicable
- `version`
- `kind`
- `displayName`
- `description`

Minification occurs whenever `CSVMetadata` is marshaled, so newly constructed metadata is always minimized without requiring a separate opt-in migration.

The existing `bundle-object-to-csv-metadata` migration now also rewrites pre-existing `olm.csv.metadata` properties. This check occurs before the `Image` and `CsvJSON` early exits, allowing already-migrated bundles without `CsvJSON` to be minimized while leaving their other properties unchanged.

This matches the field filtering performed by OLM's package server:

https://github.com/operator-framework/operator-lifecycle-manager/blob/cf27af72356d95b33c2fcb38dcff4d9ca2e7d475/pkg/package-server/apis/operators/packagemanifest.go#L28-L35

For the Red Hat 4.22 catalog, this reduces the catalog size from 118,396 KiB to 47,200 KiB, approximately 60.1%.

## Testing

```text
go test ./alpha/property ./alpha/action/migrations ./alpha/action/...
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bundle migration when CSV metadata already exists, preserving the existing metadata while avoiding unnecessary changes.
  - CSV metadata is now minimized during serialization, retaining key identifying and descriptive information while omitting detailed resource, descriptor, deployment, and container data.
  - Improved handling of bundles containing existing CSV metadata without requiring additional CSV data.


grokspawn:
Comments [here](https://kubernetes.slack.com/archives/C0181L6JYQ2/p1789502928748849) but I'll reproduce here and start using this as primary coordination to stay in-sync with other reviewers:

> I'm reviewing it, but I think I expected the "new minified version" to become the new de facto structure instead of retaining the "bloated original" status.
>
> (and relying on skipUnknownFields serialization capabilities to successfully ingest legacy instances)

and

>If you think that we cannot safely adopt the "new reality" of the minified structure ... can we establish a parallel to the migrations (maybe called finalizers) which are represented as mandatory in-order mutations?
>
>I say that because the current migration used to be side-effect-free for existing csv.metadata, but now the flow isn't so clear and really breaks the logic into two seemingly-unrelated chunks/locations.
```

grokspawn:
From offline conversation: 
Complication to establishing a canonical type lies in the re-use of v1alpha1 types from o-f/api in the csv.metadata type.  Options include
1. introducing new types in csv.metadata and dealing with sync/drift issues
2. dissembling the current flow into helpers to aid composition in both old & new contexts
3. establishing a "finalizer/normalizer" toolflow to enforce mandatorily formatted output

joelanford:
> establishing a "finalizer/normalizer" toolflow to enforce mandatorily formatted output

I think this is essentially what is implemented now, at this from a serialization perspective, which is nice because it doesn't require callers to remember to use any explicit handlers. I'm definitely open to more ideas though. Do you see any fatal flaws with the JSON marshaller approach?

